PR #81, opened 2017-12-12
Qinchen/GPU allocation do local GPU label filtering

qinchen123:
Currently, only support single label.


yqwang-ms:
>     LOGGER.logInfo(

Since we enabled Relax and GPU request convert NodeLabel and Gputype to specific nodes request, so how can you ensure allocated container is what your requested NodeLabel and Gputype?

---
Refers to: frameworklauncher/src/main/java/com/microsoft/frameworklauncher/applicationmaster/GpuAllocationManager.java:62 in 78c9e9d. [](commit_id = 78c9e9d180bc2c34b424487cf392bb28ddb73c32, deletion_comment = False)

qinchen123:
Solution change, close this com.

PR #100, opened 2017-12-18
Add gpu exporter docker 

YanjieGao:
Add gpu exporter docker which will collect node and gpu metrics
Add prometheus root folder


service-deployment/src/exporter/start.sh
```diff
@@ -0,0 +1,23 @@
...
+# to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+# The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING
+# BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+# DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+go build -v -o /usr/local/app /usr/local/app.go
```
fanyangCS: it looks that this line can be moved into dockerfile

fanyangCS: can we move this line into dockerfile, between ln31 and 32? this way we can bring the go build process ahead to docker build time, instead of docker run time

YanjieGao: No problem, i will refine this

YanjieGao: #Resolved

PR #103, opened 2017-12-18
Make datafolder as a template parameter

ydye:
Change the default /mnt as a template parameter  dataPath in clusterconfig.

### DIFF
--- a/service-deployment/bootstrap/frameworklauncher/frameworklauncher.yaml.template
+++ b/service-deployment/bootstrap/frameworklauncher/frameworklauncher.yaml.template
@@ -62,7 +62,7 @@ spec:
       volumes:
       - name: launcher-log
         hostPath:
-          path: /mnt/sdb1/logs
+          path: {{ clusterinfo[ 'dataPath' ] }}/launcherlogs
       - name: hadoop-config-volume
         configMap:
           name: {{ clusterinfo[ 'hadoopinfo' ][ 'configmapname' ] }}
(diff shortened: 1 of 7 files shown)


service-deployment/src/cleaning-image/start.sh
```diff
@@ -24,9 +24,9 @@ if [ -d "/mnt/hdfs" ]; then
 
 fi
 
-if [ -d "/mnt/sdb1" ]; then
+if [ -d "/mnt/launcherlogs" ]; then
```
fanyangCS: >mnt [](start = 10, length = 3)

just to confirm: we don't need to change /mnt to a template because it is defined in cleaning.yaml.template (volumnMounts), right?

service-deployment/bootstrap/frameworklauncher/frameworklauncher.yaml.template
```diff
@@ -62,7 +62,7 @@ spec:
       volumes:
       - name: launcher-log
         hostPath:
-          path: /mnt/sdb1/logs
+          path: {{ clusterinfo[ 'dataPath' ] }}/launcherlogs
```
fanyangCS: >auncherlogs [](start = 49, length = 11)

will the change from logs to launcherlogs affect frameworklauncher? please double check with @Yuqi

service-deployment/src/cleaning-image/start.sh
ydye: The folder in container won't be affected.

service-deployment/bootstrap/frameworklauncher/frameworklauncher.yaml.template
ydye: The log folder in launcher's logic is container's file system. The host's path change won't affect it.

PR #109, opened 2017-12-20
Add version in RESTful API

abuccts:
Add version in RESTful API. Update webportal ajax link accordingly.

### DIFF
--- a/rest-server/src/config/express.js
+++ b/rest-server/src/config/express.js
@@ -40,7 +40,7 @@ app.use(cookieParser());
 app.use(morgan('dev', { 'stream': logger.stream }));
 
 // mount all routes to /api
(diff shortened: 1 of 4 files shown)


rest-server/src/config/express.js
```diff
@@ -40,7 +40,7 @@ app.use(cookieParser());
 app.use(morgan('dev', { 'stream': logger.stream }));
 
 // mount all routes to /api
-app.use('/api', router);
+app.use('/api/v1', router);
```
fanyangCS: >/api/v1' [](start = 9, length = 8)

do we need to fix web portal?

abuccts: will fix in next PR

PR #111, opened 2017-12-20
Save authentication database to host file system

abuccts:
Save authentication database to host file system.

### DIFF
--- a/service-deployment/bootstrap/rest-server/rest-server.yaml.template
+++ b/service-deployment/bootstrap/rest-server/rest-server.yaml.template
@@ -37,8 +37,8 @@ spec:
         securityContext:
           privileged: true
         volumeMounts:
-        - mountPath: /var/databases
-          name: database-path
+        - mountPath: /var/pai
+          name: pai-path
         env:
         - name: LAUNCHER_WEBSERVICE_URI
           value: {{ clusterinfo['restserverinfo']['webservice_uri'] }}
@@ -47,7 +47,7 @@ spec:
         - name: JWT_SECRET
           value: {{ clusterinfo['restserverinfo']['jwt_secret'] }}
         - name: LOWDB_FILE
-          value: /var/databases/{{ clusterinfo['restserverinfo']['lowdb_file'] }}
+          value: /var/pai/{{ clusterinfo['restserverinfo']['lowdb_path'] }}
         - name: LOWDB_ADMIN
           value: {{ clusterinfo['restserverinfo']['lowdb_admin'] }}
         - name: LOWDB_PASSWD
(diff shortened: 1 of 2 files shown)


service-deployment/bootstrap/rest-server/rest-server.yaml.template
```diff
@@ -53,4 +57,8 @@ spec:
           containerPort: 8080
           hostPort: {{ clusterinfo['restserverinfo']['server_port'] }}
       imagePullSecrets:
-      - name: {{ clusterinfo['dockerregistryinfo']['secretname'] }}
\ No newline at end of file
+      - name: {{ clusterinfo['dockerregistryinfo']['secretname'] }}
+      volumes:
+      - name: database-path
+        hostPath:
+          path: /var/databases
```
fanyangCS: >/var/databases [](start = 16, length = 14)

make this configurable in clusterconfig-example.yaml

sterowang: I would suggest to use a common parent folder to store the all data on host for PAI, for example, "/var/pai/webportal" so that we store the all PAI data under "/var/pai".

service-deployment/bootstrap/rest-server/rest-server.yaml.template
```diff
@@ -43,7 +47,7 @@ spec:
         - name: JWT_SECRET
           value: {{ clusterinfo['restserverinfo']['jwt_secret'] }}
         - name: LOWDB_FILE
-          value: {{ clusterinfo['restserverinfo']['lowdb_file'] }}
+          value: /var/databases/{{ clusterinfo['restserverinfo']['lowdb_file'] }}
```
hwuu: Configurable?

service-deployment/bootstrap/rest-server/rest-server.yaml.template
```diff
@@ -35,6 +36,9 @@ spec:
         imagePullPolicy: Always
         securityContext:
           privileged: true
+        volumeMounts:
```
hwuu: volumesToMount?

service-deployment/bootstrap/rest-server/rest-server.yaml.template
abuccts: We can't strict the data on one host, so `/var/pai` path on each host?
For auth database, only need to store on the host where runs the rest-server pod.

service-deployment/bootstrap/rest-server/rest-server.yaml.template
abuccts: It's `volumnMounts`

service-deployment/bootstrap/rest-server/rest-server.yaml.template
abuccts: Do not need to make it configurable inside container.

service-deployment/bootstrap/rest-server/rest-server.yaml.template
hwuu: I don't think it's correct English.

PR #118, opened 2017-12-21
Add a separate 'k8s dashboard' page and a 'cluster view' page.

hwuu:
The previous 'cluster status' is rename to 'k8s dashboard'. And now a new section, 'cluster view' is added. It will link to the grafana portal, when the grafana is ready.


webportal/config/webpack.common.js
```diff
@@ -182,6 +183,14 @@ const config = {
       cache: true,
       chunks: ['layout', 'docs']
     }),
+    new HtmlWebpackPlugin({
+      title: 'Platform for AI',
+      filename: 'k8s.html',
```
fanyangCS: >k8s [](start = 17, length = 3)

only for dashboard or k8s in general?

service-deployment/clusterconfig-example.yaml
```diff
@@ -81,8 +81,10 @@ clusterinfo:
   webportalinfo:
     # uri for rest server
     rest_server_uri: http://10.0.3.9:9186
-    # uri for cluster status dashboard
-    cluster_status_uri: http://10.0.3.9:9090
+    # uri of k8s dashboard
+    k8s_dashboard_uri: http://10.0.3.9:9090
+    # uri of grafana portal
+    grafana_uri: http://10.0.3.9:9090
```
fanyangCS: > http://10.0.3.9:9090 [](start = 16, length = 21)

grafana and k8s dashboard use the same port?

PR #123, opened 2017-12-21
Hadoop service dependency

ydye:
When deploying hadoop service， wait until deployment finished.

### DIFF
--- a/service-deployment/bootstrap/hadoop-service/hadoop-data-node.yaml.template
+++ b/service-deployment/bootstrap/hadoop-service/hadoop-data-node.yaml.template
@@ -40,6 +40,13 @@ spec:
           name: hadoop-config-volume
         - mountPath: /host-configuration
           name: host-confg-volume
+        readinessProbe:
+          exec:
+            command:
+            - cat
+            - /jobstatus/jobok
+          initialDelaySeconds: 5
+          periodSeconds: 3
         env:
         - name: HDFS_ADDRESS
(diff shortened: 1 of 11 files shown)


service-deployment/bootstrap/hadoop-service/hadoop-data-node.yaml.template
```diff
@@ -40,6 +40,13 @@ spec:
           name: hadoop-config-volume
         - mountPath: /host-configuration
           name: host-confg-volume
+        readinessProbe:
+          exec:
+            command:
+            - cat
+            - /jobstatus/jobok
```
fanyangCS: >/jobstatus/jobok [](start = 14, length = 16)

is this file in host filesystem? if yes, what if datanode node manager collocated together?

ydye:
@fanyangCS  Add dependency solve and fix some issue in hadoop-service.

service-deployment/src/hadoop-run/start.sh
```diff
@@ -27,5 +27,8 @@ chmod u+x generate_config.sh
 cp /hadoop-configuration/${START_SERVICE}  start_service.sh
 chmod u+x start_service.sh
 
+mkdir -p /jobstatus
+touch /jobstatus/jobok
```
fanyangCS: > /jobstatus/jobok [](start = 5, length = 17)

usually we start the service first and touch jobok.

PR #124, opened 2017-12-21
Fix bugs in webportal deployment

abuccts:
Fix bug in webportal deployment:
- Always pull docker image.
- Use `yarn install` instead of `npm install`.

### DIFF
--- a/service-deployment/bootstrap/webportal/webportal-ds.yaml.template
+++ b/service-deployment/bootstrap/webportal/webportal-ds.yaml.template
@@ -34,8 +34,6 @@ spec:
       - name: webportal
         image: {{ clusterinfo['dockerregistryinfo']['prefix'] }}webportal
         imagePullPolicy: Always
-        securityContext:
-          privileged: true
         env:
         - name: REST_SERVER_URI
           value: {{ clusterinfo['webportalinfo']['rest_server_uri'] }}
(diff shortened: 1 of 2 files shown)


service-deployment/bootstrap/webportal/webportal-ds.yaml.template
```diff
@@ -33,6 +33,9 @@ spec:
       containers:
       - name: webportal
         image: {{ clusterinfo['dockerregistryinfo']['prefix'] }}webportal
+        imagePullPolicy: Always
+        securityContext:
+          privileged: true
```
hwuu: Does this mean the docker command will use the `--privileged` parameter? I heard from @fanyangCS that this is not secure.

hwuu: Does this mean the docker command will use the `--privileged` parameter? I heard from @fanyangCS that this is not secure.

service-deployment/src/webportal/dockerfile
```diff
@@ -24,7 +24,7 @@ ENV NODE_ENV=production \
 
 COPY copied_file/ /usr/src/
 COPY copied_file/webportal/ /usr/src/app/
-RUN npm install
+RUN npm run yarn install
```
hwuu: Is it really necessary to use yarn? I tried `npm install` and it worked fine.

abuccts: of course, we use yarn as package manager.
you can find the doc [here](https://yarnpkg.com/en/docs/migrating-from-npm)

fanyangCS: [https://www.sitepoint.com/yarn-vs-npm/](https://www.sitepoint.com/yarn-vs-npm/) here are more info about why yarn is better than npm

---
In reply to: [158422658](https://github.com/Microsoft/pai/pull/124#discussion_r158422658) [](ancestors = 158422658)

PR #125, opened 2017-12-21
Solve service dependencies for rest server

### LINKED ISSUE #96: rest server potential deployment issue
rest server deployment assumes the deployment of hadoop and frameworklauncher has been completed successfully. need to detect the service status of hadoop and frameworklauncher.
check https://github.com/Microsoft/pai/tree/master/service-deployment/kubernetesTool for details

abuccts:
Add framework launcher health check. Solve service dependencies for rest
server.

Fix #96.

### DIFF
--- a/rest-server/src/config/launcher.js
+++ b/rest-server/src/config/launcher.js
@@ -20,6 +20,8 @@
 const path = require('path');
 const fse = require('fs-extra');
 const Joi = require('joi');
+const async = require('async');
+const unirest = require('unirest');
 const childProcess = require('child_process');
 const config = require('./index');
 const logger = require('./logger');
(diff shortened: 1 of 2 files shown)


rest-server/src/config/launcher.js
```diff
@@ -92,46 +98,70 @@ if (error) {
...
+unirest.get(launcherConfig.healthCheckPath())
+    .headers(launcherConfig.webserviceRequestHeaders)
+    .timeout(2000)
+    .end((res) => {
+      if (res.status === 200) {
+        logger.info('connected to framework launcher successfully');
+        prepareHdfsPath();
+        prepareLocalPath();
+      } else {
+        throw new Error('cannot connect to framework launcher');
```
fanyangCS: what now? I suggest we retry. also, you can detect whether launcher server is ready in deployment code.https://github.com/Microsoft/pai/blob/master/service-deployment/kubernetesTool/servicestatus.py

abuccts: the pod will restart after the error thrown. If we retry, since the health check request is async, rest server still can be visited.
launcher server may running for a while then crash and restart, there may exist a race condition for service status detection.

PR #126, opened 2017-12-21
Removing private SSH key

oudeismetis:
I noticed you had commited a private ssh key to your repo. 

**"I know. It was intentional."** 
Sorry to bother you then. Feel free to ignore this PR. 
You know you can also [secure private data while keeping it under source control](https://keybase.io/blog/encrypted-git-for-everyone).

**"I didn't notice. What is this file? Is it a problem?"**
[This can be a very bad thing](http://www.securityweek.com/github-search-makes-easy-discovery-encryption-keys-passwords-source-code)

**"I keep accidentally typing `git add .`, any way I can prevent that from being an issue?"**
There are a couple of steps you should take immediately:
1. Avoid commands like `git add .` and `git add *`. It's better to be more explicit when you can.
1. Assume your key is compromised. Unassociate it from any accounts you are using it for and [generate a new key](https://help.github.com/articles/reviewing-your-ssh-keys/).
1. Store your keys in a location outside of your project.
1. Use a [.gitignore file](https://git-scm.com/docs/gitignore) to prevent accidental commits of certain files.

Github also has a page about how to [remove sensitive data from a repo](https://help.github.com/articles/removing-sensitive-data-from-a-repository/).

Hope this PR finds you well


fanyangCS:
thanks for the kind reminder. the checkin was intentional.

PR #129, opened 2017-12-22
Refine rest server storage on hdfs

abuccts:
Refine rest server storage on hdfs:
- Use `path/username/jobname` structure.
- Store job config file on hdfs.

### DIFF
--- a/rest-server/src/models/job.js
+++ b/rest-server/src/models/job.js
@@ -118,7 +118,7 @@ class Job {
 
   putJob(name, data, next) {
     if (!data.outputDir.trim()) {
-      data.outputDir = `${launcherConfig.hdfsUri}/Output/${name}`;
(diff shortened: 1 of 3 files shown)


rest-server/src/models/job.js
```diff
@@ -118,7 +118,7 @@ class Job {
 
   putJob(name, data, next) {
     if (!data.outputDir.trim()) {
-      data.outputDir = `${launcherConfig.hdfsUri}/Output/${name}`;
+      data.outputDir = `${launcherConfig.hdfsUri}/Output/${data.username}/${name}`;
```
fanyangCS: >{data.username [](start = 58, length = 14)

to clarify, even we fix the name here, we still need to fix launcher, otherwise, different user still cannot use same job name, right?

abuccts: right

PR #134, opened 2017-12-25
hwuu/standardize rest api

hwuu:
Changes:

- Authentication API: changed from `/api/v1/auth` to `/api/v1/token`.
- User add / remove APIs: changed from `/api/v1/auth` to `/api/v1/user`.
- Job APIs: changed from `/api/v1/job...` to `/api/v1/jobs...`.
- Return codes of APIs.


fanyangCS:
> // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.

conflict with master branch, pls resolve

---
Refers to: rest-server/src/config/launcher.js:16 in f299948. [](commit_id = f299948bfa791cc1456d93a0178c723ee2b3ad78, deletion_comment = False)

rest-server/src/config/launcher.js
```diff
@@ -92,6 +92,7 @@ if (error) {
 }
 launcherConfig = value;
 
+// TODO: Replace hdfs commands with webhdfs APIs?
```
abuccts: why not use native hdfs?

rest-server/src/controllers/auth.js
```diff
@@ -24,16 +24,16 @@ const logger = require('../config/logger');
...
-const login = (req, res) => {
+const getToken = (req, res) => {
   const username = req.body.username;
   const password = req.body.password;
   const expiration = req.body.expiration;
   authModel.check(username, password, (err, state, admin) => {
     if (err || !state) {
       logger.warn('user %s authentication failed', username);
-      return res.status(401).json({
+      return res.status(500).json({
```
abuccts: UNAUTHORIZED instead of INTERNAL SERVER ERROR

rest-server/src/controllers/auth.js
```diff
@@ -44,12 +44,12 @@ const login = (req, res) => {
       }, authConfig.secret, { expiresIn: expiration }, (signError, token) => {
         if (signError) {
           logger.warn('sign token error\n%s', signError.stack);
-          return res.status(500).json({
+          return res.status(401).json({
```
abuccts: INTERNAL SERVER ERROR instead of UNAUTHORIZED

rest-server/src/controllers/job.js
```diff
@@ -26,15 +26,15 @@ const logger = require('../config/logger');
  */
 const load = (req, res, next, jobName) => {
   job = new Job(jobName, () => {
-    if (job.state === 'JOB_NOT_FOUND' && req.method !== 'PUT') {
+    if ((req.method === 'GET' || req.method === 'DELETE') && job.state === 'JOB_NOT_FOUND') {
```
abuccts: check job.state first

rest-server/src/routes/index.js
```diff
@@ -19,17 +19,18 @@
 // module dependencies
 const express = require('express');
 const controller = require('../controllers/index');
-const authRoute = require('./auth');
-const jobRoute = require('./job');
-
+const tokenRouter = require('./token');
+const userRouter = require('./user');
+const jobsRouter = require('./jobs');
```
abuccts: do not need to rename filename

rest-server/src/controllers/auth.js
```diff
@@ -44,12 +44,12 @@ const login = (req, res) => {
...
             error: 'SignTokenFailed',
             message: 'sign token failed'
           });
         }
-        return res.json({
+        return res.status(200).json({
```
abuccts: default status is 200

rest-server/src/config/auth.js
```diff
@@ -47,5 +47,5 @@ const authSchema = Joi.object().keys({
 module.exports = {
   secret: config.jwtSecret,
   check: jwtCheck,
-  schema: authSchema
+  schema: authConfigSchema
```
abuccts: this schema is used for parameter validation instead of server config

rest-server/src/config/launcher.js
fanyangCS: in the future, to improve modularity, we need to make this file system agnostic. so users can replace hdfs with whatever storage system they want. 

---
In reply to: [158621888](https://github.com/Microsoft/pai/pull/134#discussion_r158621888) [](ancestors = 158621888)

rest-server/src/controllers/job.js
```diff
@@ -56,14 +56,8 @@ const list = (req, res) => {
...
       });
-    } else if (jobList === undefined || jobList.length === 0) {
-      logger.warn('list jobs error, no job found');
-      return res.status(500).json({
-        error: 'JobListNotFound',
-        message: 'could not find job list'
-      });
     } else {
-      return res.json(jobList);
+      return res.status(200).json(jobList);
```
abuccts: what if error occurs in launcher webservice?

rest-server/src/config/launcher.js
abuccts: changing from native hdfs to web hdfs api doesn't achieve this

rest-server/src/routes/user.js
```diff
@@ -19,21 +19,18 @@
...
+    /** PUT /api/v1/user - Create or update a user */
+    .put(authConfig.check, param.validate(authConfig.schema), authController.updateUser)
 
-    /** PUT /api/auth - Update user */
-    .put(authConfig.check, param.validate(authConfig.schema), authCtrl.update)
-
-    /** DELETE /api/auth - Delete user */
-    .delete(authConfig.check, authCtrl.remove);
+    /** DELETE /api/v1/user - Remove a user */
+    .delete(authConfig.check, authController.removeUser);
```
abuccts: if you change `auth` route, `auth` controller and model should be changed correspondingly.

rest-server/src/config/launcher.js
hwuu: Rest-server is rest-server. Using the native hdfs binary introduces too many dependencies between modules -- you have to prepare the hdfs binary in the image at least. In addition, we just want to create directories here, it's not efficient to use such a heavy thing.

rest-server/src/config/auth.js
hwuu: ok. Will change.

rest-server/src/controllers/auth.js
hwuu: ok. Will change.

rest-server/src/controllers/auth.js
hwuu: ok. But is the error message 'sign token error' accurate to indicate an internal server error?

rest-server/src/routes/index.js
hwuu: Any principle to name the module? `jobs.js` here is just a router, not a class, why do we need to use `job.js`?

abuccts: the principle is to use singular form and be consistent.

rest-server/src/config/launcher.js
abuccts: rest-server is part of the infrastructure and it is inside the hadoop cluster. webhdfs uses http and needs to access all nodes which will limit the performance, so why should we use the http api instead of the shell command? the cluster prepared hadoop binary during deployment and other pods like frameworklauncher also needs it.

rest-server/src/controllers/auth.js
abuccts: right, it means 'error occurs during server signing token'.

rest-server/src/controllers/job.js
hwuu: I thought all errors have been caught by your above code. Then what status code we return do you suggest?

rest-server/src/config/launcher.js
hwuu: hdfs does not need to access data node? what is its mechanism? in addition, in micro-service architecture, http apis are the recommended way to interact with other micro service components (see: https://en.m.wikipedia.org/wiki/Microservices). correct me if my understanding is wrong.

rest-server/src/controllers/auth.js
hwuu: sounds good. will change.

rest-server/src/routes/user.js
hwuu: ok. but ‘user’ and ‘token’ share the same underlying model to access lowdb and to encrypt passwords. do you suggest we should split the model into two also?

rest-server/src/routes/index.js
hwuu: do you mean all .js files should be organized by module names? modules include: token, user, job. my understanding correct? Then how do you feel about the inconsistency of the api endpoint name (plural form) and the module name (singular form as you suggested)?

rest-server/src/config/job.js
```diff
@@ -18,7 +18,12 @@
 
 // module dependencies
 const Joi = require('joi');
+const jwt = require('express-jwt');
+const config = require('./index');
 
+const jwtCheck = jwt({
```
abuccts: why do we need it here?

rest-server/src/config/token.js
```diff
@@ -15,19 +15,17 @@
...
 
-
 const jwtCheck = jwt({
   secret: config.jwtSecret
 });
 
-// define auth schema
-const authSchema = Joi.object().keys({
+// define input schema
+const inputSchema = Joi.object().keys({
```
abuccts: this is not the only input schema, use the model name instead of `input`

rest-server/src/config/user.js
```diff
@@ -0,0 +1,36 @@
...
+// BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+// module dependencies
+const Joi = require('joi');
+
+// define auth schema
+const inputSchema = Joi.object().keys({
```
abuccts: same

rest-server/src/config/user.js
```diff
@@ -0,0 +1,36 @@
...
+  username: Joi.string()
```
abuccts: need to duplicate with token here?

rest-server/src/controllers/job.js
abuccts: you deleted an `if else` branch so the last `else` branch includes errors

abuccts:
by the way, have you tested the code, especially the authentication part?

rest-server/src/config/token.js
hwuu: The context is clear. No need to mention the module name again.

rest-server/src/config/user.js
hwuu: same

rest-server/src/config/user.js
hwuu: The two schema share the same part but they also have differences. Is there a better way to do so?

rest-server/src/controllers/job.js
hwuu: Thanks for pointing it out. Must be a mistake.

rest-server/src/config/job.js
hwuu: VS code's bug...

rest-server/src/controllers/user.js
```diff
@@ -15,67 +15,28 @@
...
+    userModel.update(username, password, admin, modify, (err, state) => {
       if (err || !state) {
         logger.warn('update user %s failed', username);
         return res.status(500).json({
           error: 'UpdateFailed',
           message: 'update failed'
         });
       } else {
-        return res.json({
+        return res.status(201).json({
```
abuccts: why not 202?

rest-server/src/controllers/job.js
```diff
@@ -85,10 +83,10 @@ const update = (req, res) => {
       logger.warn('update job %s error\n%s', req.job.name, err.stack);
       return res.status(500).json({
         error: 'JobUpdateError',
-        message: 'job updated error'
+        message: 'job update error'
       });
     } else {
-      return res.json({
+      return res.status(201).json({
```
abuccts: 202?

rest-server/src/config/user.js
```diff
@@ -0,0 +1,44 @@
...
+// BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+// module dependencies
+const Joi = require('joi');
+
+// define the input schema for the 'update user' api
+const userInputSchema_PUT = Joi.object().keys({
```
abuccts: do not use this kind of variable name, camel case only.

rest-server/src/controllers/job.js
```diff
@@ -15,26 +15,24 @@
...
  */
 const load = (req, res, next, jobName) => {
   job = new Job(jobName, () => {
     if (job.jobStatus.state === 'JOB_NOT_FOUND' && req.method !== 'PUT') {
       logger.warn('load job %s error, could not find job', jobName);
-      return res.status(500).json({
-        error: job.state,
+      return res.status(404).json({
+        error: "JobNotFound",
         message: `could not find job ${jobName}`
```
sterowang: Suggest capitalize the first letter of each message.

rest-server/src/controllers/job.js
hwuu: Correct. Will change.

rest-server/src/controllers/user.js
hwuu: According to http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html, 201 means you get the resource immediately while 202 means the request is successfully queued. So I think it should be 201 here.

rest-server/src/controllers/job.js
hwuu: I didn't change your code here. Would you like me to change?

hwuu: And I suggest we do it in another PR. This PR is sufficiently large.

abuccts: yes, we can standardize error message as Bin suggested in another pr.

PR #135, opened 2017-12-25
Make ssh port configurable in k8s deployment

abuccts:
Make ssh port configurable in k8s deployment.

### DIFF
--- a/kubernetes-deployment/bootstrap.py
+++ b/kubernetes-deployment/bootstrap.py
@@ -96,6 +96,8 @@ def sftp_paramiko(src, dst, filename, host_config):
     username = host_config['username']
     password = host_config['password']
     port = 22
+    if (host_config['hostport']):
+        port = host_config['hostport']
 
     # First make sure the folder exist.
     ssh = paramiko.SSHClient()
@@ -128,6 +130,8 @@ def ssh_shell_paramiko(host_config, commandline):
     username = host_config['username']
     password = host_config['password']
     port = 22
+    if (host_config['hostport']):
+        port = host_config['hostport']
 
     ssh = paramiko.SSHClient()
     ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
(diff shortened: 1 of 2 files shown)


kubernetes-deployment/cluster-config-example.yaml
```diff
@@ -123,9 +123,11 @@ mastermachinelist:
     # The nodename in k8s. To avoid complication, it is suggested to fill in the hostip for the nodename.
     # If you are confident that DNS can resolve your named nodename correctly, you can specify a UNIQUE_HOST_NAME
     # hostip: the static IP of this node. We assume the IP of a host will not change.
+    # hostport: the ssh port of this node. Default port is 22.
```
fanyangCS: sshport seems to be a better name?

kubernetes-deployment/bootstrap.py
```diff
@@ -96,6 +96,8 @@ def sftp_paramiko(src, dst, filename, host_config):
     username = host_config['username']
     password = host_config['password']
     port = 22
+    if (host_config['hostport']):
```
fanyangCS: why do we need to make ssh port in a host env configuration?

kubernetes-deployment/cluster-config-example.yaml
abuccts: updated

kubernetes-deployment/bootstrap.py
abuccts: some env may not allow ssh to use 22 port

kubernetes-deployment/cluster-config-example.yaml
```diff
@@ -123,9 +123,11 @@ mastermachinelist:
     # The nodename in k8s. To avoid complication, it is suggested to fill in the hostip for the nodename.
     # If you are confident that DNS can resolve your named nodename correctly, you can specify a UNIQUE_HOST_NAME
     # hostip: the static IP of this node. We assume the IP of a host will not change.
+    # sshport: the ssh port of this node. Default port is 22.
```
fanyangCS: >Default [](start = 42, length = 7)

if not specified, the port is 22

PR #139, opened 2017-12-27
Add parameter in job view page

abuccts:
Add jobName parameter in job view page.

### DIFF
--- a/webportal/src/app/job/job-view/job-view.component.js
+++ b/webportal/src/app/job/job-view/job-view.component.js
@@ -17,6 +17,7 @@
 
 
 // module dependencies
+const url = require('url');
 const breadcrumbComponent = require('../breadcrumb/breadcrumb.component.ejs');
 const loadingComponent = require('../loading/loading.component.ejs');
 const jobTableComponent = require('./job-table.component.ejs');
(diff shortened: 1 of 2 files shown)


webportal/src/app/job/job-view/job-view.component.js
```diff
@@ -176,7 +177,12 @@ window.loadJobDetail = loadJobDetail;
 
 $('#content-wrapper').html(jobViewHtml);
 $(document).ready(() => {
-  loadJobs();
+  const query = url.parse(window.location.href, true).query;
+  if (query['jobName']) {
+    loadJobDetail(query['jobName']);
+  } else {
+    loadJobs();
```
fanyangCS: >loadJobs(); [](start = 4, length = 11)

should we show "job not found?"

abuccts: show the empty page if no job returned? what if error occurs in the back-end, should we split those two cases?

fanyangCS: good idea. show "job not found" if the backend returns no result, and show whatever error code if backend reports error.

---
In reply to: [158775305](https://github.com/Microsoft/pai/pull/139#discussion_r158775305) [](ancestors = 158775305)

abuccts: it's the current implementation.

fanyangCS: really? I thought if the query['jobname'] return failure you show all the jobs, instead of jobNotFound?

---
In reply to: [158776110](https://github.com/Microsoft/pai/pull/139#discussion_r158776110) [](ancestors = 158776110)

abuccts: if `query[jobName]` is empty, it will show all jobs. otherwise show the specify job.

PR #152, opened 2018-01-02
Single build and push

ydye:
1: refactor build and push script.
2: enable Single image build and push.

### DIFF
--- a/service-deployment/docker_build.py
+++ b/service-deployment/docker_build.py
@@ -246,12 +246,31 @@ def dependency_solve(service_config, image_name, created_image, prefix):
 
 
 
-def build_docker_images(cluster_config, service_config):
+def build_docker_images(cluster_config, service_config, target):
 
     image_list = service_config['imagelist']
     created_image = {}
 
-    for image in image_list:
+    if target == 'all':
+
+        for image in image_list:
+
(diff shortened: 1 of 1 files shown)


fanyangCS:
do you need to update readme?

service-deployment/docker_build.py
```diff
@@ -283,6 +312,7 @@ def hadoop_binary_prepare(custom_hadoop_path, hadoop_version):
 def main():
     parser = argparse.ArgumentParser()
     parser.add_argument('-p', '--path', required=True, help="path to cluster configuration file")
+    parser.add_argument('-n', '--name', default='all', help="Build and push an image to the registry")
```
fanyangCS: >('-n', '--name', d [](start = 23, length = 18)

need to update readme?

ydye:
@fanyangCS  I will continue to develop single service deploy script. And then will update the readme together. And this PR's update is compatible with our previous build process.

service-deployment/docker_build.py
ydye: Expose it in the readme.md in the future.

fanyangCS: >name [](start = 33, length = 4)

imagename?

ydye: yes

PR #156, opened 2018-01-02
[Launcher]: Fix issues on Windows platform

yqwang-ms:


frameworklauncher/src/main/java/com/microsoft/frameworklauncher/service/Service.java
```diff
@@ -251,20 +251,17 @@ private ContainerLaunchContext setupContainerLaunchContextForDefaultAM(
     // SetupLocalEnvironment
     Map<String, String> localEnvs = new HashMap<>();
 
-    StringBuilder classpath = new StringBuilder(ApplicationConstants.Environment.CLASSPATH.$$())
-        .append(ApplicationConstants.CLASS_PATH_SEPARATOR).append("./*");
+    // Internal class is preferred over external class
+    localEnvs.put(ApplicationConstants.Environment.CLASSPATH_PREPEND_DISTCACHE.name(), "true");
```
fanyangCS: >localEnvs [](start = 4, length = 9)

Do you think we can create a unit test to make sure the Windows/Linux OS env. won't break this?

frameworklauncher/src/main/java/com/microsoft/frameworklauncher/utils/HadoopUtils.java
```diff
@@ -127,7 +127,7 @@ private static FileStatus getFileStatusInHdfsInternal(String hdfsPath) throws Ex
       LOGGER.logInfo("[hadoop fs -stat %%Y %s]", hdfsPath);
       FileStatus fileStatus = fs.getFileStatus(new Path(hdfsPath));
       return fileStatus;
-    } catch (PathNotFoundException e) {
+    } catch (PathNotFoundException | FileNotFoundException e) {
       throw new NonTransientException("Path does not exist", e);
```
qinchen123: >Path does not exist [](start = 39, length = 19)

"hdfsPath"  Path or File Node Found in the message ?

yqwang-ms: Internal Exception itself contains it

---
In reply to: [159354625](https://github.com/Microsoft/pai/pull/156#discussion_r159354625) [](ancestors = 159354625)

frameworklauncher/src/main/java/com/microsoft/frameworklauncher/service/Service.java
yqwang-ms: Good point. But it is hard to UT in Launcher, since it at least need to mock a NM, and do the same UT as NM did. Since it is a feature in NM, we can just believe the NM UT is passed.

---
In reply to: [159354155](https://github.com/Microsoft/pai/pull/156#discussion_r159354155) [](ancestors = 159354155)

fanyangCS: let's create an issue to track this then.

---
In reply to: [159355911](https://github.com/Microsoft/pai/pull/156#discussion_r159355911) [](ancestors = 159355911,159354155)

PR #159, opened 2018-01-03
Single deploy and stop

ydye:
Deploy a single service.

### DIFF
--- a/service-deployment/deploy.py
+++ b/service-deployment/deploy.py
@@ -194,15 +194,7 @@ def generate_template_file(cluster_config, service_config):
 
     for serv in service_list:
 
-        template_list = service_list[serv]['templatelist']
-        if 'None' in template_list:
-            continue
-
-        for template in template_list:
-
-            template_data = read_template("bootstrap/{0}/{1}.template".format(serv, template))
-            generate_data = generate_from_template(template_data, cluster_config)
-            write_generated_file("bootstrap/{0}/{1}".format(serv, template), generate_data)
+        generate_template_file_service(serv, cluster_config, service_config)
 
 
 
@@ -293,21 +285,7 @@ def copy_arrangement(service_config):
 
     for srv in service_list:
 
-        if 'copy' not in service_list[srv]:
-            continue
-
-        for target in service_list[srv]['copy']:
-            dst = "bootstrap/{0}/{1}".format(srv, target['dst'])
-            src = target['src']
-
-            if os.path.exists(dst) == False :
-                shell_cmd = "mkdir -p {0}".format(dst)
-                error_msg = "failed to mkdir -p {0}".format(dst)
-                execute_shell(shell_cmd, error_msg)
-
-            shell_cmd = "cp -r {0} {1}".format(src, dst)
-            error_msg = "failed to copy {0}".format(src)
-            execute_shell(shell_cmd, error_msg)
+        copy_arrangement_service(srv, service_config)
 
 
 
--- a/service-deployment/deploy.py
+++ b/service-deployment/deploy.py
@@ -199,17 +199,21 @@ def single_service_bootstrap(serv, service_config):
     if 'stopscript' not in service_config['servicelist'][serv]:
         return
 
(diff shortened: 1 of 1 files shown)


fanyangCS:
@sterowang is added to the review. #Closed

fanyangCS:
@mhwu is added to the review. #Closed

fanyangCS:
>         clean_up_generated_file(service_config)

maybe you can add single service cleanup here in the next PR

---
Refers to: service-deployment/deploy.py:361 in 5f48ae4. [](commit_id = 5f48ae49c0c22fe5e589546b1189612aead81a36, deletion_comment = False)

ydye:
@fanyangCS Sure, I will do it, in the next PR.

service-deployment/deploy.py
```diff
@@ -223,6 +265,28 @@ def bootstrap_service(service_config):
...
+            error_msg = "failed to mkdir -p {0}".format(dst)
+            execute_shell(shell_cmd, error_msg)
+
+        shell_cmd = "cp -r {0} {1}".format(src, dst)
+        error_msg = "failed to copy {0}".format(src)
+        execute_shell(shell_cmd, error_msg)
+
+
+
 def copy_arrangement(service_config):
```
fanyangCS: >copy_arrangement [](start = 4, length = 16)

maybe to reuse copy_arrangement_service here.

service-deployment/deploy.py
```diff
@@ -173,6 +173,21 @@ def clean_up_generated_file(service_config):
...
+        return
+
+    for template in template_list:
+        template_data = read_template("bootstrap/{0}/{1}.template".format(serv, template))
+        generate_data = generate_from_template(template_data, cluster_config)
+        write_generated_file("bootstrap/{0}/{1}".format(serv, template), generate_data)
+
+
+
 def generate_template_file(cluster_config, service_config):
```
fanyangCS: >generate_template_file [](start = 4, length = 22)

reuse generate_template_file_service?

service-deployment/deploy.py
```diff
@@ -191,6 +206,33 @@ def generate_template_file(cluster_config, service_config):
 
 
 
+def single_service_bootstrap(serv, service_config):
```
fanyangCS: >single_service_bootstrap [](start = 4, length = 24)

what is the difference from dependency_bootstrap? do you need to consider service dependency here?

service-deployment/deploy.py
ydye: Yes, I will change it.

service-deployment/deploy.py
ydye: Ok.

service-deployment/deploy.py
ydye: Single service startup will ignore the dependency. User should ensure their operation is in a correctly order.

ydye: dependency_bootstrap will auto-start all service in a correctly order.

abuccts: `single_service_bootstrap` can also be reused in `bootstrap_service`?

fanyangCS: got it. please add a comment here.

---
In reply to: [159383031](https://github.com/Microsoft/pai/pull/159#discussion_r159383031) [](ancestors = 159383031)

ydye: @abuccts @fanyangCS  should we run a stop script in a function?

ydye: If we reuse single startup in bootstrap_service. If user mis-use our command. Then all service will be redeploy. In current process, if service already exists in the cluster, re-executing the startscript won't have any influence. And that's the one reason why i will make the function single_service_bootstrap  ignore dependency.

fanyangCS: I understand you introduce this function for debug purpose. I prefer everything done manually in this case.

---
In reply to: [159384788](https://github.com/Microsoft/pai/pull/159#discussion_r159384788) [](ancestors = 159384788)

PR #162, opened 2018-01-03
[Webportal] Check uploaded json in job submission

### LINKED ISSUE #161: No response in job submission view when submit an invalid Json file
There is no response, no error message,  nothing . when submit invalid Json file to the PAI system.

abuccts:
Check uploaded json in job submission.

Fixed #161.

### DIFF
--- a/webportal/src/app/job/job-submit/job-submit.component.js
+++ b/webportal/src/app/job/job-submit/job-submit.component.js
@@ -30,7 +30,7 @@ const jobSubmitHtml = jobSubmitComponent({
   loading: loadingComponent
 });
 
-const isValidJSON = (str) => {
+const isValidJson = (str) => {
   try {
     JSON.parse(str);
     return true;
@@ -75,7 +75,7 @@ $(document).ready(() => {
     const reader = new FileReader();
     reader.onload = (event) => {
       const jobConfig = event.target.result;
-      if (isValidJSON(jobConfig)) {
+      if (isValidJson(jobConfig)) {
         submitJob(JSON.parse(jobConfig));
       } else {
         alert('Please upload a valid json file.');
--- a/webportal/src/app/job/job-submit/job-submit.component.js
+++ b/webportal/src/app/job/job-submit/job-submit.component.js
@@ -30,6 +30,15 @@ const jobSubmitHtml = jobSubmitComponent({
   loading: loadingComponent
 });
 
+const isValidJSON = (str) => {
+  try {
+    JSON.parse(str);
+    return true;
+  } catch (e) {
+    return false;
+  }
+};
+
 const submitJob = (jobConfig) => {
   userAuth.checkToken((token) => {
     loading.showLoading();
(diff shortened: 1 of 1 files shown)


webportal/src/app/job/job-submit/job-submit.component.js
```diff
@@ -65,8 +74,12 @@ $(document).ready(() => {
   $(document).on('change', '#file', (event) => {
     const reader = new FileReader();
     reader.onload = (event) => {
-      const jobConfig = JSON.parse(event.target.result);
-      submitJob(jobConfig);
+      const jobConfig = event.target.result;
+      if (isValidJson(jobConfig)) {
+        submitJob(JSON.parse(jobConfig));
+      } else {
+        alert('Please upload a valid json file.');
```
qinchen123: >alert('Please upload a valid json file.'); [](start = 8, length = 42)

Alert the exception information will more helpful for user to detect the Json format issue

webportal/src/app/job/job-submit/job-submit.component.js
```diff
@@ -30,6 +30,15 @@ const jobSubmitHtml = jobSubmitComponent({
...
 });
 
+const isValidJson = (str) => {
+  try {
+    JSON.parse(str);
+    return true;
+  } catch (e) {
+    return false;
+  }
+};
```
fanyangCS: shall we implement it in rest server?

webportal/src/app/job/job-submit/job-submit.component.js
abuccts: The js SyntaxError doesn't provide much information and I think print error stack in production env is not a good idea.

webportal/src/app/job/job-submit/job-submit.component.js
abuccts: no. If the uploaded file isn't a valid json, ajax can not send it as a json object to rest api.

fanyangCS: agree we should check the json format here. but it looks like we need to do it on rest-server as well, since in the future users can submit jobs through restAPIs

---
In reply to: [159381738](https://github.com/Microsoft/pai/pull/162#discussion_r159381738) [](ancestors = 159381738)

webportal/src/app/job/job-submit/job-submit.component.js
fanyangCS: You can show the alert and attach original exception somewhere for interested users (like us)

---
In reply to: [159381548](https://github.com/Microsoft/pai/pull/162#discussion_r159381548) [](ancestors = 159381548)

webportal/src/app/job/job-submit/job-submit.component.js
sterowang: Agree. One thing is not print the full stack, but just show the exception type and message.

abuccts: @fanyangCS rest-server's several middlewares have checked json format:
1. if user specified `Content-Type: application/json` header, [`Joi`](https://github.com/Microsoft/pai/blob/master/rest-server/src/config/job.js#L22) will make sure it's a valid json.
2. if `Content-Type: application/json` is not set in header, [`bodyParse`](https://github.com/Microsoft/pai/blob/master/rest-server/src/config/express.js#L36) will try to parse the request body to json and return error message if it's not a valid json.

what #161 met is error occurs in browser's `JSON.parse()` so the rest code wasn't executed.

webportal/src/app/job/job-submit/job-submit.component.js
abuccts: updated

PR #164, opened 2018-01-03
Issue fix

ydye:
When run this command, user is root. So just change the path to Relative path will delete the file correctly


abuccts:
I think I just met this problem yesterday on my two-node cluster.

ydye:
@abuccts 
I suggest to cleanup the old script with command ```sudo ./deploy -p cluster-config -c```
It will be centralized to clean the data and then run ``` sudo ./deploy -p cluster-config ``` to deploy kubernetes.

PR #166, opened 2018-01-03
Bug Fix

ydye:
It's a bug. Fix it. Then cleanup vm should be done centralized

### DIFF
--- a/kubernetes-deployment/start.sh
+++ b/kubernetes-deployment/start.sh
@@ -73,9 +73,9 @@ fi
 
 #tar -xvf kubernetes.tar
 
-chmod u+x src/cleanup.sh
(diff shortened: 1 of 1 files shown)


kubernetes-deployment/start.sh
```diff
@@ -73,9 +73,9 @@ fi
 
 #tar -xvf kubernetes.tar
 
-chmod u+x src/cleanup.sh
+#chmod u+x src/cleanup.sh
 
-./src/cleanup.sh
+#./src/cleanup.sh
```
fanyangCS: we should just remove it?

ydye: Removed.

PR #168, opened 2018-01-04
Pylon: Universal entry to inner APIs

hwuu:
Based on nginx. Need to refine the interface in the next version.


service-deployment/clusterconfig-example.yaml
```diff
@@ -160,6 +165,7 @@ machinelist:
     webportal: "true"
     prometheus: "true"
     grafana: "true"
+    pylon: "true"
```
ydye: Can this proxy run on other node？

hwuu: Yes, as long as the node have a public IP address.

pylon/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+URL pattern:
+
+```
+http://<pylon_ip>:<pylon_port>/r/http/<api_server_ip>/<api_server_port>/...
+```
+
+For example:
+
+- Example 1 – Kubernetes api: http://10.151.40.179:8086/r/http/10.151.40.179/8080/api/v1/nodes
```
sterowang: It would be wonderful if this "/http/ip/port/" can be replaced with "/kubernetes/" so that caller can get a fixed URL whatever where we deployed kubernetes.

sterowang: test reply

sterowang: The same to other urls.

hwuu: Yes Bin, I totally agree. I plan to do this in the next step -- In this PR, we implement the basic functionality; and then in the next PR, we refine the interface. This is because the nginx-based reverse proxy is not the highest priority. How about this idea?

sterowang: Yes, that's what we discussed on the scrum meeting. Putting comments here is for reminding.

pylon/src/nginx.conf
```diff
@@ -0,0 +1,58 @@
+# Copyright (c) Microsoft Corporation
+# All rights reserved.
+#
+# MIT License
```
sterowang: Test1

sterowang: Test 2

sterowang: Test 3

---
In reply to: [159610875](https://github.com/Microsoft/pai/pull/168#discussion_r159610875) [](ancestors = 159610875)

sterowang: test 4

pylon/README.md
hwuu: Will do.

service-deployment/clusterconfig-example.yaml
ydye: Can this proxy run in the ha mode?

hwuu: It's stateless. So theoretically it can run in HA mode.

pylon/README.md
```diff
@@ -0,0 +1,58 @@
+# Pylon: Universal Entry to API Servers
```
fanyangCS: >Pylon: Universal Entry to API Servers [](start = 2, length = 37)

need to explain why we need this

hwuu: Good suggestion. Will add.

PR #169, opened 2018-01-04
[Tests] Setup end to end test

abuccts:
Setup end to end test:
- Check pods status
- Check hdfs root path
- Check launcher webservice
- Check rest api

### DIFF
--- a/tests/end_to_end/install.sh
+++ b/tests/end_to_end/install.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+# Copyright (c) Microsoft Corporation
+# All rights reserved.
+#
+# MIT License
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
+# documentation files (the "Software"), to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and
+# to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+# The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING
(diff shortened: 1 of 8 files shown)


fanyangCS:
@ydye is added to the review. #Closed

tests/end_to_end/install.sh
```diff
@@ -0,0 +1,29 @@
...
+# The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING
+# BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+# DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+
+install_bats() {
```
fanyangCS: >install_bats [](start = 0, length = 12)

who runs this script? should we add a readme?

abuccts: Updated.

hwuu:
Do we have a document that lists all the aspects we should test? There are the following items in my mind:

- Build-time:
    - Check whether the configurations are correct. For example, no conflict IPs, etc.
- Run-time:
    - Check whether all services are online (already in Yifan's code).
    - Check whether each component works properly (maybe this is the job of unit test?).
    - Define some user scenarios, such as submitting a certain job, adding / removing users, etc., and check whether the system works in all of these scenarios.

abuccts:
This is only for end-to-end test. Unit test + CI + CD may be added later.

PR #174, opened 2018-01-05
fix user specified GPU atrribute not compatible with GPU type

qinchen123:
Also do GPU type filtering when user specified the GPU attribute


qinchen123:
@fanyangCS is added to the review. #Closed

PR #179, opened 2018-01-05
add node label for node-exporter[for issue 178]

YanjieGao:
add node label for node-exporter[for issue 178]


ydye:
#179

service-deployment/clusterconfig-example.yaml
```diff
@@ -168,11 +168,12 @@ machinelist:
     ip: 10.0.3.11
     hdfsrole: worker
     yarnrole: worker
-  
+    node-exporter: true
```
ydye: Add quotation. Like "true"

PR #193, opened 2018-01-08
[Webportal] Fix AppExitDiagnostics issue

abuccts:
Fix AppExitDiagnostics issue.


webportal/src/app/job/job-view/job-detail-table.component.ejs
```diff
@@ -47,7 +47,11 @@
 
 <div class="collapse" id="appExitDiagnostics">
   <div class="well">
-    <%= jobStatus.appExitDiagnostics.replace(/\n/g, "<br>") %>
+    <% if (jobStatus.appExitDiagnostics) { %>
```
fanyangCS: >jobStatus.appExitDiagnostics [](start = 11, length = 28)

we need to confirm this is the right flag to determine the job status.

abuccts: updated. change to `not available yet`

PR #195, opened 2018-01-08
[Tests] Update end to end test

abuccts:
Update end to end test:
- Install pai-fs
- Prepare CNTK job code and data
- Check hdfs root path through pai-fs
- Make Test dir on hdfs through pai-fs
- Upload job data to hdfs through pai-fs
- Submit a test job through framework launcher
- Submit a CNTK test job through rest server


tests/end_to_end/install.sh
```diff
@@ -25,5 +25,23 @@ install_bats() {
...
 apt-get install -y dos2unix
-install_bats
\ No newline at end of file
+
+mkdir -p local
+cd local
+install_bats
+install_paifs
+prepare_cntk_job
+cd -
```
sterowang: How about we put the testing code and dependencies into a Docker image and execute there in the next step?

sterowang:
> cluster_config=$cluster_config bats test_rest_server.sh

We can consider to run tests regular and send out an alert email and testing result if some errors occurred.

---
Refers to: tests/end_to_end/start.sh:56 in 0e51aa8. [](commit_id = 0e51aa89806e9089824e386f7a9897ab6ce6f236, deletion_comment = False)

tests/end_to_end/install.sh
abuccts: make it as an option?

abuccts:
We need to submit jobs to rest server which requires an auth token, so we also need to update the token regularly.

tests/end_to_end/test_hdfs.sh
```diff
@@ -21,11 +21,37 @@
 . utils.sh
 
 eval $(parse_yaml $cluster_config "pai_")
-hdfs_uri=$pai_clusterinfo_restserverinfo_hdfs_uri
-webhdfs_uri="$(echo $hdfs_uri | sed -e "s/^hdfs/http/g" | sed -e "s/9000/50070/g")"
+hdfs_host=$pai_clusterinfo_hadoopinfo_hadoop_vip
+paifs_arg="--host $hdfs_host --port 50070 --user root"
```
fanyangCS: >50070 [](start = 36, length = 5)

configurable?

tests/end_to_end/test_hdfs.sh
```diff
@@ -21,11 +21,37 @@
...
+
+@test "make hdfs test sub dir" {
+  result="$(python local/pai-fs/pai-fs.py $paifs_arg -mkdir hdfs://Test/launcher)"
+  [[ ! $result == *Error* ]]
+  result="$(python local/pai-fs/pai-fs.py $paifs_arg -mkdir hdfs://Test/cntk)"
+  [[ ! $result == *Error* ]]
+}
+
+@test "upload cntk data to hdfs" {
+  result="$(python local/pai-fs/pai-fs.py $paifs_arg -cp -r local/CNTK/Examples/SequenceToSequence/CMUDict/Data hdfs://Test/cntk/)"
```
fanyangCS: >local/CNTK/Examples/SequenceToSequence/CMUDict/Data [](start = 60, length = 51)

assume the data is available locally? where do we download the data?

tests/end_to_end/etc/cntk.json
```diff
@@ -0,0 +1,17 @@
+{
+  "jobName": "CNTK_TEST",
+  "image": "aiplatform/pai.run.cntk",
```
fanyangCS: >aiplatform/pai.run.cntk [](start = 12, length = 23)

where do we config the docker registry?
maybe we should put everythying in a docker image, regarding to environment configuration: docker registery, hdfs lib, cntk/tf/mxnet dependency

tests/end_to_end/test_hdfs.sh
abuccts: - hdfs host is read from clusterconfig.yaml
- webhdfs port is the default port and also be read from clusterconfig.yaml, but our hadoop config uses the default port and can't config it

tests/end_to_end/test_hdfs.sh
abuccts: already done in `./install.sh`, it will download requirements in `local` dir

tests/end_to_end/etc/cntk.json
abuccts: read the docker registry prefix from clusterconfig.yaml?

fanyangCS: the docker registry in clusterconfig.yaml is for deployment. the DL framework image can be from different docker registry?

---
In reply to: [161502752](https://github.com/Microsoft/pai/pull/195#discussion_r161502752) [](ancestors = 161502752)

tests/end_to_end/test_hdfs.sh
fanyangCS: user can manually change hadoop config after the prepare-hadoop step. but yeah we can do this later

---
In reply to: [161502491](https://github.com/Microsoft/pai/pull/195#discussion_r161502491) [](ancestors = 161502491)

tests/end_to_end/test_hdfs.sh
fanyangCS: in prepare_cntk_job()?

---
In reply to: [161502651](https://github.com/Microsoft/pai/pull/195#discussion_r161502651) [](ancestors = 161502651)

tests/end_to_end/test_hdfs.sh
abuccts: ok

tests/end_to_end/test_hdfs.sh
abuccts: right

tests/end_to_end/etc/cntk.json
abuccts: yes. we can add a separate config file for end to end test later.

PR #208, opened 2018-01-10
Improve loading performance for the hardware page

hwuu:


hwuu:
@fanyangCS : According to https://datatables.net/reference/api/draw(), this function redraws the whole table to let data changes take effect. It document said 'most' apis do not automatically call it so I called it when I updated each cell. However, my yesterday's experiment told me that this function was not necessary.

PR #210, opened 2018-01-10
Refine job status returned by API

hwuu:
Changes:
1) Correct retry count.
2) Convert framework states to 3 job states.

### DIFF
--- a/rest-server/src/models/job.js
+++ b/rest-server/src/models/job.js
@@ -97,33 +97,38 @@ class Job {
             if (framework.frameworkStatus) {
               let jobState = "";
               const frameworkState = framework.frameworkStatus.frameworkState;
+              const applicationExitType = framework.frameworkStatus.applicationExitType;
               if (
-                  frameworkState == "FRAMEWORK_WAITING" ||
-                  frameworkState == "APPLICATION_CREATED" ||
-                  frameworkState == "APPLICATION_LAUNCHED" ||
-                  frameworkState == "APPLICATION_WAITING"
+                  frameworkState == 'FRAMEWORK_WAITING' ||
+                  frameworkState == 'APPLICATION_CREATED' ||
+                  frameworkState == 'APPLICATION_LAUNCHED' ||
+                  frameworkState == 'APPLICATION_WAITING'
                 ) {
-                jobState = "WAITING";
+                jobState = 'WAITING';
               } else if (
-                  frameworkState == "APPLICATION_RUNNING" ||
-                  frameworkState == "APPLICATION_RETRIEVING_DIAGNOSTICS" ||
-                  frameworkState == "APPLICATION_COMPLETED"
+                  frameworkState == 'APPLICATION_RUNNING' ||
+                  frameworkState == 'APPLICATION_RETRIEVING_DIAGNOSTICS' ||
+                  frameworkState == 'APPLICATION_COMPLETED'
                 ) {
-                jobState = "RUNNING";
+                jobState = 'RUNNING';
               } else if (
-                  frameworkState == "FRAMEWORK_COMPLETED"
+                  frameworkState == 'FRAMEWORK_COMPLETED' &&
+                  applicationExitType == 'SUCCEEDED'
                 ) {
-                jobState = "COMPLETED";
+                jobState = 'SUCCEEDED';
+              } else if (
+                  frameworkState == 'FRAMEWORK_COMPLETED' &&
+                  applicationExitType != 'SUCCEEDED'
+                ) {
+                jobState = 'FAILED';
               }
               let jobRetryCount = 0;
               const jobRetryCountInfo = framework.frameworkStatus.frameworkRetryPolicyState;
-              if (jobRetryCountInfo) {
-                jobRetryCount =
-                  jobRetryCountInfo.transientNormalRetriedCount +
-                  jobRetryCountInfo.transientConflictRetriedCount +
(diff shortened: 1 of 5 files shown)


rest-server/src/models/job.js
```diff
@@ -94,11 +95,41 @@ class Job {
             next(job, new Error('job not found'));
           } else {
             if (framework.frameworkStatus) {
+              let jobState = "";
+              const frameworkState = framework.frameworkStatus.frameworkState;
+              if (
+                  frameworkState == "FRAMEWORK_WAITING" ||
```
abuccts: pls use single quotes rather than double quotes

rest-server/src/models/job.js
```diff
@@ -94,11 +95,41 @@ class Job {
...
+                jobState = "COMPLETED";
+              }
+              let jobRetryCount = 0;
+              const jobRetryCountInfo = framework.frameworkStatus.frameworkRetryPolicyState;
+              if (jobRetryCountInfo) {
+                jobRetryCount =
+                  jobRetryCountInfo.transientNormalRetriedCount +
+                  jobRetryCountInfo.transientConflictRetriedCount +
+                  jobRetryCountInfo.nonTransientRetriedCount +
+                  jobRetryCountInfo.unKnownRetriedCount;
```
abuccts: check before add

hwuu: @abuccts : I asked Yuqi. These fields will always be there. Then do we still need to check them?

rest-server/src/models/job.js
hwuu: @abuccts : What policy shall we follow to choose from single quotes and double quotes?

rest-server/src/models/job.js
```diff
@@ -94,11 +95,41 @@ class Job {
...
+              } else if (
+                  frameworkState == "APPLICATION_RUNNING" ||
+                  frameworkState == "APPLICATION_RETRIEVING_DIAGNOSTICS" ||
+                  frameworkState == "APPLICATION_COMPLETED"
+                ) {
+                jobState = "RUNNING";
+              } else if (
+                  frameworkState == "FRAMEWORK_COMPLETED"
+                ) {
+                jobState = "COMPLETED";
```
fanyangCS: >COMPLETED [](start = 28, length = 9)

@github@xiongyf.com and @micw.mm@gmail.com, I assume if jobState == complete, we will further tell if the job is "succeeded" or "failed", right?

rest-server/src/models/job.js
```diff
@@ -94,11 +95,41 @@ class Job {
...
+                  frameworkState == "FRAMEWORK_WAITING" ||
+                  frameworkState == "APPLICATION_CREATED" ||
+                  frameworkState == "APPLICATION_LAUNCHED" ||
+                  frameworkState == "APPLICATION_WAITING"
+                ) {
+                jobState = "WAITING";
+              } else if (
+                  frameworkState == "APPLICATION_RUNNING" ||
+                  frameworkState == "APPLICATION_RETRIEVING_DIAGNOSTICS" ||
+                  frameworkState == "APPLICATION_COMPLETED"
```
yqwang-ms: It is better to move them to be WAITING

rest-server/src/models/job.js
hwuu: Currently the state of succeeded or failed is further distinguished in web portal using the appExitStaues field returned by Launcher. Shall we do it here?

fanyangCS: I don't think we should do it here. but please check with Yuqi if appExitStatues is the legit field to detect the success of a job completion.

---
In reply to: [160622710](https://github.com/Microsoft/pai/pull/210#discussion_r160622710) [](ancestors = 160622710)

yqwang-ms: You can use applicationExitCode to judge succeed/fail when the state = APPLICATION_COMPLETED or FRAMEWORK_COMPLETED

---
In reply to: [160626847](https://github.com/Microsoft/pai/pull/210#discussion_r160626847) [](ancestors = 160626847,160622710)

sterowang: As FRAMEWORK_COMPLETED will arrived finally, so I suggest set jobState to SUCCEEDED or FAILED here depends on applicationExitCode, this will be more user friendly on web portal.

rest-server/src/models/job.js
abuccts: 1. Current js files use single quotes, so please be **consistent**. Otherwise you have to change all other quotes.
2. You can check [node](https://github.com/jscs-dev/node-jscs/blob/master/presets/node-style-guide.json#L47), [google](https://github.com/jscs-dev/node-jscs/blob/master/presets/google.json#L21) and [fb](https://github.com/facebook/fbjs/blob/c69904a511b900266935168223063dd8772dfc40/packages/eslint-config-fbjs-opensource/index.js#L247) for other style guides.

rest-server/src/models/job.js
abuccts: Isn't this the current solution?
@hwuu you have to update code in webportal accordingly if changed states here.

hwuu: Thanks guys. I will make the change according to your suggestions.

rest-server/src/models/job.js
hwuu: Good suggestion. I will make the change.

PR #211, opened 2018-01-10
Enable tag of docker image

ydye:
1) Easy to manage image of different version
2) For future rolling-update and rolling-back.

### DIFF
--- a/service-deployment/clusterconfig-example.yaml
+++ b/service-deployment/clusterconfig-example.yaml
@@ -44,6 +44,8 @@ clusterinfo:
     docker_username: your_registry_username
     docker_password: your_registry_password
 
+    docker_tag: your_image_tag
+
     # The name of the secret in kubernetes will be created in your cluster
     # Must be lower case, e.g., regsecret.
     secretname: your_secret_name
--- a/service-deployment/bootstrap/cleaning/cleaning.yaml.template
+++ b/service-deployment/bootstrap/cleaning/cleaning.yaml.template
@@ -29,7 +29,7 @@ spec:
       hostPID: true
       containers:
       - name: cleaning-one-shot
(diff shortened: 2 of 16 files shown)


service-deployment/clusterconfig-example.yaml
```diff
@@ -44,6 +44,8 @@ clusterinfo:
     docker_username: your_registry_username
     docker_password: your_registry_password
 
+    docker_tag: your_image_tag
```
hwuu: Is it mandatory? What will happen if it's not specified?

sterowang: After we finished initial deployment and want to upgrade cluster images to a newer tag, what's the procedures to do it?

ydye: @sterowang 
1) Rolling-update: https://kubernetes.io/docs/tasks/manage-daemon/update-daemon-set/
2) Rolling-back: https://kubernetes.io/docs/tasks/manage-daemon/rollback-daemon-set/

Currently we can't directly enable the upgrade feature. Before it, we should add the option ``` .spec.updateStrategy.type = RollingUpdate.```. Then we have two methods to do it.

a: Directly use kubectl
b: Develop a tool to do upgrading and package the function.

ydye: It's mandatory now. If it's not specified, the program will exist. Should we set thie value as an option?

abuccts: @ydye you can set the tag to `latest` in `docker_build.py` if it's not specified.

ydye: I think keeping mandatory is a good way. Because cluster administrator should know what they are doing. If administrator want to deploy ```latest``` tag, they can also fill ```latest``` here.

ydye: @hwuu @abuccts , I set tag as an option. and default value is latest.

PR #217, opened 2018-01-12
Filter labels & Float table head

YitongFeng:
image


webportal/package.json
```diff
@@ -37,6 +37,7 @@
     "express": "4.16.2",
     "extract-text-webpack-plugin": "3.0.2",
     "file-loader": "1.1.5",
+    "floatthead": "2.0.3",
```
hwuu: Please update `yarn.lock` file accordingly. [Install yarn](https://yarnpkg.com/en/docs/install#windows-tab) and then run: `npm run yarn install`.

webportal/src/app/cluster-view/services/services.component.ejs
```diff
@@ -4,8 +4,8 @@
   <div class="row">
     <div class="col-xs-12">
       <div class="box">
-        <div id="service-table" class="box-body table-responsive no-padding">
-          <%= serviceTable({ data: []  }) %>
+        <div id="service-table" class="box-body table-responsive no-padding" style="max-height:<%= $(window).height() * 0.85 %>px">
```
hwuu: Did you try using an absolute value other than a percentage?

webportal/package.json
YitongFeng: Solved

PR #218, opened 2018-01-12
Web portal: Add a new page to show hardware utilization detail for each node.

hwuu:
image

### DIFF
--- a/webportal/src/app/cluster-view/hardware/hardware-detail.component.js
+++ b/webportal/src/app/cluster-view/hardware/hardware-detail.component.js
@@ -1,0 +1,54 @@
+// Copyright (c) Microsoft Corporation
+// All rights reserved.
+//
+// MIT License
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
+// documentation files (the "Software"), to deal in the Software without restriction, including without limitation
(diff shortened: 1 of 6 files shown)


webportal/src/app/cluster-view/hardware/hardware-detail.component.js
```diff
@@ -0,0 +1,54 @@
...
+const hardwareDetailComponent = require('./hardware-detail.component.ejs');
+const breadcrumbComponent = require('../../job/breadcrumb/breadcrumb.component.ejs');
+const webportalConfig = require('../../config/webportal.config.json');
+const url = require('url');
+
+//
+
+const resizeContentWrapper = () => {
+  $('#content-wrapper').css({'height': $(window).height() + 'px'});
+  $('#content-iframe').css('height', (($(window).height() - 120)) + 'px');
```
sterowang: What does the magic "120" means?

hwuu: It is the number of pixels from the top of the page to the top of the iframe -- I haven't found a solution to make the iframe automatically fit into the space.

sterowang: Not looks perfect, but let's move on first.

hwuu: Agree. Will find a way eventually.

PR #219, opened 2018-01-12
Web portal: apply the style of table in hardware page to job-view page

hwuu:
New features:

- Better look-and-feel.
- Paging.
- Searching.
- Sorting.
- And the 'delete' button is also removed.

image

### DIFF
--- a/webportal/src/app/job/job-view/job-table.component.ejs
+++ b/webportal/src/app/job/job-view/job-table.component.ejs
@@ -1,4 +1,4 @@
-<table class="table table-hover">
+<table id="job-table" class="table table-bordered table-hover">
   <thead>
     <tr>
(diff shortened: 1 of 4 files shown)


webportal/src/app/job/job-view/job-table.component.ejs
```diff
@@ -7,7 +7,7 @@
       <th>ElapsedTime</th>
       <th>Retries</th>
       <th>Status</th>
-      <th>Delete</th>
+      <!-- <th>Delete</th> -->
```
sterowang: Later, we should change it to "Cancel" to stop job.

hwuu: Agree. I've put this into my own backlog.

PR #220, opened 2018-01-12
Add the driver's version option in cluster-configration

ydye:


service-deployment/clusterconfig-example.yaml
```diff
@@ -26,6 +26,9 @@ clusterinfo:
   # PAI data root path
   paiPath: "/var/pai"
 
+  # Choose proper nvidia driver version from this url http://www.nvidia.com/object/linux-amd64-display-archive.html
+  nvidia_drivers_version: "384.111"
```
sterowang: Next time, should we update the deployment in PaiDeployment git instead?

ydye: Yes, I will update it.

PR #221, opened 2018-01-14
[Webportal] Add user logout and registration

abuccts:
Add user logout and registration.


webportal/src/app/user/user-login/user-login-nav.component.ejs
```diff
@@ -1,9 +1,13 @@
 <% if (cookies.get("token")) { %>
-  <li class="dropdown user user-menu">
-    <a href="#" data-toggle="control-sidebar">
+  <li class="dropdown">
+    <a href="#" class="dropdown-toggle" data-toggle="dropdown">
       <!-- img src="" class="user-image" alt="User Image" -->
       <span class="hidden-xs">Welcome, <%= cookies.get("user") %></span>
     </a>
+    <ul class="dropdown-menu">
+      <li><a href="#" onclick="window.location.replace('/register.html')">New user</a></li>
```
hwuu: Will the 'new user' link still be available even if the current user is not an admin? Adding/removing users is a separate functionality (management operations of admins only) other than login/logout and registration (self-serving operations available for all users). In addition, the top-right menu usually contains the operations closely related to the current user only (see github's menu). I suggest we put this user management feature in a separate place.

webportal/src/app/user/user-register/user-register.component.ejs
```diff
@@ -0,0 +1,21 @@
+<section class="content-header"><%= breadcrumb({ breadcrumbTitle: "User Register" }) %></section>
```
hwuu: Register --> Registration

abuccts: only change this place?

webportal/src/app/user/user-login/user-login-nav.component.ejs
abuccts: need to support in authentication first

webportal/src/app/user/user-register/user-register.component.ejs
hwuu: Yes please.

webportal/src/app/user/user-login/user-login-nav.component.ejs
hwuu: Let's do it in this way for now and then improve in the future.

webportal/src/app/user/user-register/user-register.component.ejs
abuccts: updated

PR #231, opened 2018-01-15
[Webportal] Protect user management

abuccts:
Protect user management.

### DIFF
--- a/webportal/config/webpack.common.js
+++ b/webportal/config/webpack.common.js
@@ -163,6 +163,10 @@ const config = {
       jQuery: 'jquery',
       'window.jQuery': 'jquery'
     }),
+    new webpack.ProvidePlugin({
+      cookies: 'js-cookie',
+      'window.cookies': 'js-cookie'
+    }),
     new HtmlWebpackPlugin({
       title: 'Platform for AI',
       filename: 'index.html',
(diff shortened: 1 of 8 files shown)


webportal/src/app/user/user-auth/user-auth.component.js
```diff
@@ -26,4 +26,8 @@ const checkToken = (callback) => {
   }
 };
 
-module.exports = { checkToken };
\ No newline at end of file
+const checkAdmin = () => {
+  return cookies.get('admin') === 'true';
```
hwuu: I’m wondering what if the response is hijacked and modified, or the webpage code is changed dynamically in Chrome’s F12 console?

webportal/config/webpack.common.js
```diff
@@ -163,6 +163,10 @@ const config = {
       jQuery: 'jquery',
       'window.jQuery': 'jquery'
     }),
+    new webpack.ProvidePlugin({
```
hwuu: Haven’t we used this cookies plugin before?

abuccts: not global

webportal/src/app/user/user-auth/user-auth.component.js
abuccts: anyone can change anything in his browser, it doesn't have any effect to the back end.

PR #233, opened 2018-01-16
Node manager start script update to retry nvidia-smi

ydye:


service-deployment/bootstrap/hadoop-service/hadoop-configuration/nodemanager-start-service.sh
```diff
@@ -33,7 +33,26 @@ then
   $HADOOP_YARN_HOME/bin/yarn nodemanager
 else
   echo gpu machine
-  nvidia-smi
+
+  # The loop is designed for the node restart or kubelet restart.
+  # Because when kubelet crushed, all service in the node will be started at the same time.
+  # Usually drivers' startup process is much slower than node-manager.
+  # So set the try times to 10. If after 10 time retris, the nm service still can't find gpu.
```
fanyangCS: > 10 time retris, the nm service still can't find gpu. [](start = 40, length = 53)

shall we generate a warning in this case?

ydye: No. If we can't successfully run nvidia-smi, I will assume the node's drivers is error. Hadoop service shouldn‘t handle too much drivers‘ logic

service-deployment/bootstrap/hadoop-service/hadoop-configuration/nodemanager-start-service.sh
```diff
@@ -33,7 +33,26 @@ then
...
+  # Please check the node status.
+  for (( i=1; i <= 10; i++ ))
+  do
+
+    if nvidia-smi
+    then
+      echo GPUs are found.
+      break
+    fi
```
fanyangCS: >fi [](start = 4, length = 2)

I suggest to add an else here: echo waiting for GPU driver to be ready

ydye: But actually, the script will continue executing. And the NM will be started in a states without GPU dirvers. And additionally i have considered to write exit here. But if the drivers is broken and hard to be fix, the nm will never ready (In an endless loop of exit and restart).

fanyangCS: yeah. the else seems redundant

---
In reply to: [161729505](https://github.com/Microsoft/pai/pull/233#discussion_r161729505) [](ancestors = 161729505)

PR #234, opened 2018-01-16
[Rest Server] Refine job constructor

abuccts:
Refine job constructor.

### DIFF
--- a/rest-server/src/models/job.js
+++ b/rest-server/src/models/job.js
@@ -37,7 +37,7 @@ class Job {
       for (let key of Object.keys(job)) {
         this[key] = job[key];
       }
-      next();
+      next(this);
     });
   }
 
--- a/rest-server/src/controllers/job.js
+++ b/rest-server/src/controllers/job.js
@@ -23,7 +23,7 @@ const logger = require('../config/logger');
  * Load job and append to req.
  */
 const load = (req, res, next, jobName) => {
-  job = new Job(jobName, () => {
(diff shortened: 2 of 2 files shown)


rest-server/src/models/job.js
```diff
@@ -37,7 +37,7 @@ class Job {
       for (let key of Object.keys(job)) {
         this[key] = job[key];
       }
-      next();
+      next(this);
```
hwuu: Will the constructor function return AFTER this next(this) function call returns? If not, the ‘undefined’ issue may still be there.

hwuu: OK. I found the 'load' function calls 'next()' when Job's constructor function is done: https://github.com/Microsoft/pai/blob/def8a2a7b8cf2a22aa7e65dfec72f805dc88f20e/rest-server/src/controllers/job.js#L41. So this method works correctly.

abuccts: it uses aync callback instead of sync return

sterowang:
>     this.name = name;

If a object needs to take sometime to initialize itself, suggest we don't provides a public constructor, instead change the public constructor to a static method without return value. This can avoid someone use the partially initialized object returned by constructor.

---
Refers to: rest-server/src/models/job.js:35 in 0901061. [](commit_id = 090106198ed7ef0fb970764e76b7f2b9cacfc5b7, deletion_comment = False)

PR #241, opened 2018-01-17
Change the threshold of image gc.

ydye:
--image-gc-high-threshold=98. 
--image-gc-low-threshold=95

```
image-gc-high-threshold, the percent of disk usage which triggers image garbage collection. Default is 90%.
image-gc-low-threshold, the percent of disk usage to which image garbage collection attempts to free. Default is 80%.
```

### DIFF
--- a/kubernetes-deployment/template/kubelet.sh.template
+++ b/kubernetes-deployment/template/kubelet.sh.template
@@ -43,4 +43,6 @@ docker run \
   --pod-manifest-path=/etc/kubernetes/manifests \
   --allow-privileged=true \
   --logtostderr=true \
+  --image-gc-high-threshold=98 \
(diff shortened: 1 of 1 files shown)


kubernetes-deployment/template/kubelet.sh.template
```diff
@@ -43,4 +43,6 @@ docker run \
   --pod-manifest-path=/etc/kubernetes/manifests \
   --allow-privileged=true \
   --logtostderr=true \
+  --image-gc-high-threshold=98 \
```
fanyangCS: >image-gc-high-threshold=98 [](start = 4, length = 26)

[https://github.com/kubernetes/kubernetes/pull/51423](https://github.com/kubernetes/kubernetes/pull/51423) in the future we may use this.

PR #242, opened 2018-01-18
Add nvidia_uvm check in function nvidiaPresent

ydye:


ydye:
@fanyangCS @sterowang  After this fix, when the nvidia-uvm is missing, just restart drivers.

PR #243, opened 2018-01-19
Fix Bug in cleanup-service.py

ydye:
In the situation of private registry. When the cluster-configuration is cleaned up, then the last step to pull cleaning-image in the registry will fail. So I move the cluster-configuration cleaning up to the last.

### DIFF
--- a/service-deployment/cleanup-service.py
+++ b/service-deployment/cleanup-service.py
@@ -66,6 +66,11 @@ def main():
         if 'stopscript' not in service_config['servicelist'][serv]:
             continue
 
+        # Cleanup cluster-configuration will lose secret of private registry.
+        # Then the cleaning-image will be failed to pull.
+        if serv == 'cluster-configuration':
+            continue
+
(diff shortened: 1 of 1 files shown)


service-deployment/cleanup-service.py
```diff
@@ -96,6 +101,15 @@ def main():
...
 
+    serv = 'cluster-configuration'
+    shell_cmd = 'chmod a+x ./bootstrap/{0}/{1}'.format(serv, service_config['servicelist'][serv]['stopscript'])
+    error_msg = 'Failed make the {0} stop script executable'.format(serv)
+    execute_shell(shell_cmd, error_msg)
+
+    shell_cmd = './bootstrap/{0}/{1}'.format(serv, service_config['servicelist'][serv]['stopscript'])
+    error_msg = 'Failed stop the service {0}'.format(serv)
+    execute_shell(shell_cmd, error_msg)
+
```
fanyangCS: this could become a function: cleanOneService

ydye: Resolved.

Issue #247, opened 2018-01-19
pai-fs cp bug


abuccts:
If files already exist on hdfs, `pai-fs -cp` will append those files to the same file instead of overriding it.

fanyangCS:
https://github.com/Microsoft/pai/pull/249 fixed this bug

PR #249, opened 2018-01-20
Fix bug of copy the same file with pai-fs

xiaoxinla:
delete the file if we find it has already exist in the path, and then continue the copy process.

### DIFF
--- a/pai-fs/fsimpl/BaseFs.py
+++ b/pai-fs/fsimpl/BaseFs.py
@@ -203,8 +203,10 @@ class FileSystem:
             if self.verbose and is_normal_stdout():
                 sys.stdout.write(progressString)
                 sys.stdout.flush()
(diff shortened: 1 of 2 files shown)


fanyangCS:
@sterowang is added to the review. #Closed

fanyangCS:
@abuccts is added to the review. #Closed

abuccts:
#247 is caused by hdfs instead of pai-fs and it can not be reproduce now.
It seems that `pai-fs.py -cp -r` called `cp_file` instead of `remote_cp_file`.

xiaoxinla:
@abuccts I have tested the code of cp_file, it called remote_cp actually. How do you test the case?

xiaoxinla:
@fanyangCS @abuccts @sterowang Can you merge it into master?

Issue #252, opened 2018-01-23
service view, some information not searchable


fanyangCS:
for information listed after page 3

PR #255, opened 2018-01-23
fix table sorting bug

YitongFeng:


fanyangCS:
fix issue https://github.com/Microsoft/pai/issues/252

PR #259, opened 2018-01-24
[Rest Server] Handle Docker image pulling failure

### LINKED ISSUE #257: handle docker image pulling failure
should handle the failure here.
https://github.com/Microsoft/pai/blob/66cc43b57f34ebc74313036e2e7a8e71e063e1b5/rest-server/src/templates/yarnContainerScript.mustache#L54

The job should fail due to user (since docker registry is provided by user) after trying a few times.

abuccts:
Handle Docker image pulling failure.
Fixes #257.

### DIFF
--- a/rest-server/src/templates/yarnContainerScript.mustache
+++ b/rest-server/src/templates/yarnContainerScript.mustache
@@ -46,12 +46,13 @@ printf "%s %s\n%s\n\n" "[INFO]" "NVIDIA_DEVICES" "$nvidia_devices"
 {{# jobData.authFile }}
 IFS=$'\r\n' GLOBIGNORE='*' \
   eval 'cred=($(hdfs dfs -cat {{{ jobData.authFile }}}))' \
-  || { echo 'Read authFile Failed' ; exit 1; }
+  || { echo 'Read authFile failed' ; exit 1; }
 docker login --username ${cred[1]} --password ${cred[2]} ${cred[0]} \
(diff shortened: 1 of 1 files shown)


rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -46,12 +46,13 @@ printf "%s %s\n%s\n\n" "[INFO]" "NVIDIA_DEVICES" "$nvidia_devices"
...
   eval 'cred=($(hdfs dfs -cat {{{ jobData.authFile }}}))' \
-  || { echo 'Read authFile Failed' ; exit 1; }
+  || { echo 'Read authFile failed' ; exit 1; }
 docker login --username ${cred[1]} --password ${cred[2]} ${cred[0]} \
-  || { echo 'Authorized Failed' ; exit 1; }
+  || { echo 'Authorized failed' ; exit 1; }
 {{/ jobData.authFile }}
 
-docker pull {{{ jobData.image }}}
+docker pull {{{ jobData.image }}} \
```
fanyangCS: >docker pull {{{ jobData.image }}} [](start = 0, length = 33)

should we try more than once?

abuccts: Not sure about the retry policy of `docker pull`, it seems that `docker pull` will wait if some errors occur. Does @ydye have any idea?

PR #264, opened 2018-01-25
rename the log path [fix]

YanjieGao:


fanyangCS:
here is another mis-spelling https://github.com/Microsoft/pai/blob/97358995e611dbb889d3bdaedc7e7d3327d1afa0/prometheus/exporter/gpu_exporter.go#L154

PR #272, opened 2018-01-25
[Rest Server] Add job name validation

abuccts:
Add job name validation, only allow `^[a-zA-Z0-9._\-()]+$` format.


abuccts:
Fixes #262

PR #273, opened 2018-01-25
[Rest Server] Make codeDir optional in job config file

abuccts:
Make codeDir optional in job config file, allow empty code directory.


abuccts:
Fixes #265

fanyangCS:
In readme, please explain: if code path unspecified, PAI assumes user’s has built the code into the docker image and has its own startup script. Also, please test this type of docker image before merging this PR.

rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -67,20 +67,24 @@ if [ "$(cat ContainerList | wc -l)" -ne $PAI_TASKS_NUM ]; then
...
 
-cd $HOME
-hdfs dfs -get $PAI_CODE_DIR code || exit 1
-cd code
+if [[ -z $PAI_CODE_DIR ]]; then
+  cd $HOME
+  hdfs dfs -get $PAI_CODE_DIR code || exit 1
+  cd code
+fi
+
```
sterowang: One thing we need to discuss here:

If user specifies code, then we goto "$WORKDIR/code", else we just stay at "$WORKDIR". This is inconsistent. Also the subfolder "code" is created by us implicitly, this might conflict with user container folder structure.

A possible solution is let user determine their code folder name entirely. For example, suppose user specified "hdfs://xxx/nni", we will download nni to "$WORKDIR/nni". Also we don't cd into that folder, just keep $WORKDIR unchanged, so it's user's duty to input command line like "nni/start.sh". By this way we can keep consistency if user specified codeDir or not.

abuccts: updated

PR #278, opened 2018-01-26
Move graf‘s json.template to bootstrap 

ydye:
Refactor grafana image to make it reuseable


ydye:
Test passed on the local cluster.

PR #281, opened 2018-01-29
Kubernetes Deploy Tool

ydye:
1) Support repair option in bootstrap，but now， we can only repair worker-node ( Due to Etcd dynamical configuration issue) 
2) Implement plugin module in maintainlib for future. So that we can easily expend it if we want add more main job to the script
3) All plugin class defined in maintainlib, and the script package should be listed in maintain.yaml.  Now admin could specify an job （repair）with the name in maintain.yaml. And bootstrap will call the class instance to package the script and run the job. 
4) Add dockerfile to make it possible for user to run the deploy tool in a docker container where all the prerequisite are prepared.


ydye:
Why plugin?  Different environment has different maintain method. So I think make a job as plugin will be easily to expend.  
And the repair function has been tested on local.

kubernetes-deployment/cluster-config-example.yaml
```diff
@@ -184,3 +184,5 @@ workermachinelist:
 
 
 
+
+
```
fanyangCS: no change in this yaml, excluded from this PR?

fanyangCS:
> sudo ./bootstrap.py -p yourclusterconfig.yaml -f yournodelist.yaml -a

we need an example nodelist.yaml in readme. 

---
Refers to: kubernetes-deployment/README.md:154 in ea24bf5. [](commit_id = ea24bf5bdd472547b94576ec10373ecd59f47c8a, deletion_comment = False)

kubernetes-deployment/maintainlib/common.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def write_generated_file(generated_file, file_path):
+
+    with open(file_path, "w+") as fout:
+        fout.write(generated_file)
+
+
+
+def sftp_paramiko(src, dst, filename, host_config):
```
sterowang: >sftp_paramiko [](start = 4, length = 13)

What major difference between sftp and scp so that we need to use sftp?

kubernetes-deployment/maintainlib/common.py
```diff
@@ -0,0 +1,176 @@
...
+    hostip = host_config['hostip']
+    username = host_config['username']
+    password = host_config['password']
+    port = 22
+    if 'sshport' in host_config:
+        port = host_config['sshport']
+
+    ssh = paramiko.SSHClient()
+    ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+    ssh.connect(hostname=hostip, port=port, username=username, password=password)
```
sterowang: >, password=password [](start = 61, length = 19)

Password write into stdin, does it means "password=password" not workable?

kubernetes-deployment/bootstrap.py
```diff
@@ -395,9 +397,33 @@ def remove_nodes(cluster_config, node_list_config):
...
+    # Todo: load maintain from a DB such as etcd instead of a yaml file.
+    maintain_config = load_yaml_file("maintain.yaml")
+
+    for host in node_list_config['machinelist']:
+
+        maintain_expert(cluster_config, maintain_config, node_list_config['machinelist'][host], job_name)
+
+
+
 def option_validation(args):
```
fanyangCS: >option_validation [](start = 4, length = 17)

given that we have CI in place, can you add unit test for this function?

kubernetes-deployment/maintainlib/common.py
```diff
@@ -0,0 +1,176 @@
...
+    stdin, stdout, stderr = ssh.exec_command(commandline, get_pty=True)
+    stdin.write(password + '\n')
+    stdin.flush()
+
+    for response_msg in stdout:
+        print response_msg.strip('\n')
+
```
sterowang: As you printed out the ssh response, suggest you print out the ssh command line also.

kubernetes-deployment/bootstrap.py
```diff
@@ -395,9 +397,33 @@ def remove_nodes(cluster_config, node_list_config):
...
+
+
+def maintain_nodes(cluster_config, node_list_config, job_name):
+
+    # Todo: load maintain from a DB such as etcd instead of a yaml file.
+    maintain_config = load_yaml_file("maintain.yaml")
+
+    for host in node_list_config['machinelist']:
+
+        maintain_expert(cluster_config, maintain_config, node_list_config['machinelist'][host], job_name)
```
fanyangCS: >maintain_expert [](start = 8, length = 15)

renamed to maintain_one_node?

kubernetes-deployment/maintainlib/common.py
```diff
@@ -0,0 +1,176 @@
...
+    sftp = paramiko.SFTPClient.from_transport(transport)
+    sftp.put('{0}/{1}'.format(src, filename), '{0}/{1}'.format(dst, filename))
+    sftp.close()
+
+    transport.close()
+
+
+
+def ssh_shell_paramiko(host_config, commandline):
```
fanyangCS: >ssh_shell_paramiko [](start = 4, length = 18)

all functions before this one have been used elsewhere (in service-deployment. can you refactor the code so k8s deployment and service deployment can reuse them?
Not necessary done in this PR though

kubernetes-deployment/bootstrap.py
```diff
@@ -407,9 +433,27 @@ def option_validation(args):
...
 
+    if args.maintain != None:
+
+        if args.file == None:
+            print "Please specify the nodelist.yaml's path"
+            return False
+        if args.deploy or args.clean:
+            print "You could not specify the option (-a or -r) with the option (-d or -c)"
+            return False
+        if args.maintain == "repair":
```
sterowang: This is common to the all commands, suggest you put it out in a common place to avoid code duplication.

kubernetes-deployment/bootstrap.py
```diff
@@ -436,6 +480,7 @@ def main():
     parser.add_argument('-f', '--file', default=None, help="An yamlfile with the nodelist to maintain")
     parser.add_argument('-a', '--add', action="store_true", help="Add the node from nodelist.yaml")
     parser.add_argument('-r', '--remove', action="store_true", help="Remove the node from nodelist.yaml")
+    parser.add_argument('-m', '--maintain', default=None, help="maintain the cluster")
```
fanyangCS: >add_argument [](start = 11, length = 12)

this is getting more and more complicated. we need a printUsage to help user.

kubernetes-deployment/bootstrap.py
```diff
@@ -407,9 +433,27 @@ def option_validation(args):
         if args.deploy or args.clean:
             print "You could not specify the option (-a or -r) with the option (-d or -c)"
             return False
+        if args.maintain != None:
+            print "You could not specify the option (-a or -r) with the option -m"
         # Add or remove node-list
         return True
 
+    if args.maintain != None:
+
```
sterowang: If add, remove and maintain are mutual exclusives, suggest put them as the values of a common parameter. For example, replace "-a", "-r", "-m" with a single acton command "-a" with values "add", "remove", "repair".

So internally you can use switch statement to handle them easily.

And internally, you can make "repair" and all the other maintenance command goes into a same branch like:

if args.action="repair" or args.action="xxx".

So generally you could make the mutual exclusive commands to be values of one parameter. This will makes users to do single choice naturally.

kubernetes-deployment/maintain-tool/repair-worker-node.sh
```diff
@@ -0,0 +1,37 @@
...
+# check etc/ exist or not.
+staticpod="src/etc"
+if [ -d "$staticpod" ]; then
+
+    cp -r src/etc /
+
+fi
+
+chmod u+x src/kubelet.sh
+./src/kubelet.sh
```
fanyangCS: >kubelet.sh [](start = 6, length = 10)

does it always work? for example, what if some service on this node requires to clean up the corresponding folder, otherwise the restart will fail?

kubernetes-deployment/cluster-config-example.yaml
ydye: Resolved.

kubernetes-deployment/bootstrap.py
ydye: Add unitest for validation?

kubernetes-deployment/maintainlib/common.py
ydye: Better not. This will introduce dependency. Then we have to prepare another script for user to prepare environment,

kubernetes-deployment/bootstrap.py
ydye: The logic of ```if args.maintain == "repair":```?

kubernetes-deployment/bootstrap.py
ydye: For example?

kubernetes-deployment/maintainlib/common.py
ydye: Resolved.

kubernetes-deployment/bootstrap.py
ydye: For easy understand?

kubernetes-deployment/maintainlib/common.py
ydye: Maybe this question could solve the problem. https://superuser.com/questions/134901/whats-the-difference-between-scp-and-sftp
But in our script, I don't want to import another package "scp" to handle file transmission, because paramiko has a sftp solution.

kubernetes-deployment/bootstrap.py
sterowang: I mean we use one single parameter as "-a" for args.action, it can be "add/remove/repair/deploy/clean" instead of -a/-r/-m/-d/-c. Think about what if user specifies -a and -d, user might expect add a machine and deploy, however what we will do is only adding machines, to me looks like a little confusing.

---
In reply to: [164983160](https://github.com/Microsoft/pai/pull/281#discussion_r164983160) [](ancestors = 164983160)

kubernetes-deployment/bootstrap.py
fanyangCS: yes, add unit test to see if validation can catch it when user submit with wrong flag 

---
In reply to: [164982731](https://github.com/Microsoft/pai/pull/281#discussion_r164982731) [](ancestors = 164982731)

kubernetes-deployment/bootstrap.py
fanyangCS: yes. maintain_expert is hard to understand. maintain_one_node is clear semantically

---
In reply to: [164986808](https://github.com/Microsoft/pai/pull/281#discussion_r164986808) [](ancestors = 164986808)

kubernetes-deployment/bootstrap.py
ydye: I see, in next pr do it.

kubernetes-deployment/bootstrap.py
ydye: Next pr will merge them into one. Do ut in the future.

kubernetes-deployment/bootstrap.py
ydye: Resolved.

kubernetes-deployment/maintain-tool/repair-worker-node.sh
fanyangCS: can you add some code to detect whether this is a worker node?

---
In reply to: [164973547](https://github.com/Microsoft/pai/pull/281#discussion_r164973547) [](ancestors = 164973547)

kubernetes-deployment/maintainlib/common.py
fanyangCS: I suggest to submit a PR for maintainlib only. add unit tests for functions in commom.py and repair.py

---
In reply to: [164973939](https://github.com/Microsoft/pai/pull/281#discussion_r164973939) [](ancestors = 164973939)

ydye: This is the PR for maintainlib. Only one instance ```repair```

sterowang: I saw we have already published so many comments here, maybe we can continue from here in order not to start from the beginning?

---
In reply to: [165922550](https://github.com/Microsoft/pai/pull/281#discussion_r165922550) [](ancestors = 165922550,164973939)

ydye:
@sterowang @fanyangCS 
Cause all commits have been merged in small PR. I will close this PR.

PR #295, opened 2018-02-02
[Rest Server] Allow user to change his own password

abuccts:
Allow user to change his own password.


abuccts:
should we allow user to delete himself in the platform?

fanyangCS:
no

---
In reply to: [362515127](https://github.com/Microsoft/pai/pull/295#issuecomment-362515127) [](ancestors = 362515127)

PR #311, opened 2018-02-07
add cadvisor metrics as prometheus exporter

YanjieGao:
1 It could expose container level metrics. Later we could leverage it to aggregate job level or hardware metrics.
2 use its metrics to more convenient support disk space usage metrics


service-deployment/clusterconfig-example.yaml
```diff
@@ -115,6 +115,8 @@ clusterinfo:
     prometheus_port: 9090
     # port for node exporter
     node_exporter_port: 9100
+    # cadvisor port
+    cadvisor_port: 4194
```
ydye: Why config this as a value.

service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
```diff
@@ -30,4 +30,11 @@ data:
         {% for host in machinelist %}
         - {{ machinelist[host]['ip']}}:{{ clusterinfo['prometheusinfo']['node_exporter_port'] }}
         {% endfor %}
+    - job_name: "cadvisor"
+      scrape_interval: "15s"
+      static_configs:
+      - targets: 
+        {% for host in machinelist %}
+        - {{ machinelist[host]['ip']}}:{{ clusterinfo['prometheusinfo']['cadvisor_port'] }}
```
ydye: Aggregate all host infomation through ca in one node?

service-deployment/clusterconfig-example.yaml
YanjieGao: as kubernetes cadvisor default expose prometheus format metrics through this port

service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
YanjieGao: This will aggregate all node's cadvisor's metrics, as cadvisor expose prometheus format metrics through this port. current version use Kube default cadvisor, if deploy by ourself will add node label.

service-deployment/clusterconfig-example.yaml
ydye: Because we will use our own ca, can you modify this port?

PR #309, opened 2018-02-06
[Launcher]: Refine GET APIs

yqwang-ms:
1. Make Status consistent with Request's life cycle from APIs' view
2. Provide more info for GET Frameworks API
3. Remove GET TaskStatuses and TaskRoleStatus API

PR #314, opened 2018-02-07
[Rest Server] Update GET jobs API

abuccts:
Update GET jobs API, corresponding to #309.

### DIFF
--- a/rest-server/src/models/job.js
+++ b/rest-server/src/models/job.js
@@ -47,34 +47,20 @@ class Job {
         .end((res) => {
           const resJson = typeof res.body === 'object' ?
               res.body : JSON.parse(res.body);
-          const jobList = [];
-          async.each(resJson.frameworkNames, (frameworkName, callback) => {
-            this.getJobStatus(frameworkName, (jobStatus) => {
-              const jobOverview = {
-                name: frameworkName,
-                username: jobStatus.username,
-                state: jobStatus.state,
-                subState: jobStatus.subState,
-                retries: jobStatus.retries,
-                createdTime: jobStatus.createdTime,
-                completedTime: jobStatus.completedTime,
-                appTrackingUrl: jobStatus.appTrackingUrl,
-                appExitType: jobStatus.appExitType
-              };
-              jobList.push(jobOverview);
-              callback();
-            });
-          }, (err) => {
-            jobList.sort((a, b) => b.createdTime - a.createdTime);
-            next(jobList, err);
+          const jobList = res.summarizedFrameworkInfos.map((frameworkInfo) => {
+            return {
+              name: frameworkInfo.frameworkName,
+              username: frameworkInfo.userName,
+              state: frameworkInfo.frameworkState,
+              retries: frameworkInfo.frameworkRetryPolicyState.retriedCount,
+              createdTime: frameworkInfo.firstRequestTimestamp,
+              completedTime: frameworkInfo.frameworkCompletedTimestamp,
+              appExitCode: frameworkInfo.applicationExitCode
+            };
           });
+          jobList.sort((a, b) => b.createdTime - a.createdTime);
+          next(jobList, err);
         });
-  }
-
-  getJobStatus(name, next) {
-    this.getJob(name, (job, err) => {
-      next(job.jobStatus, err);
-    });
   }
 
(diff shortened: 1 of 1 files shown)


rest-server/src/models/job.js
```diff
@@ -47,36 +47,22 @@ class Job {
...
   }
 
-  getJobStatus(name, next) {
-    this.getJob(name, (job, err) => {
-      next(job.jobStatus, err);
-    });
-  }
-
   getJob(name, next) {
     unirest.get(launcherConfig.frameworkPath(name))
```
yqwang-ms: >frameworkPath [](start = 31, length = 13)

I also changed this API.
launcherConfig.frameworkPath = (frameworkName) => {
  return `${launcherConfig.webserviceUri}/v1/Frameworks/${frameworkName}`;
};

Do you need to adjust here?
Or use another API:
launcherConfig.frameworkStatusPath.

rest-server/src/models/job.js
```diff
@@ -47,36 +47,22 @@ class Job {
...
-            });
-          }, (err) => {
-            jobList.sort((a, b) => b.createdTime - a.createdTime);
-            next(jobList, err);
+          const jobList = res.summarizedFrameworkInfos.map((frameworkInfo) => {
+            return {
+              name: frameworkInfo.frameworkName,
+              username: frameworkInfo.userName,
+              state: frameworkInfo.frameworkState,
+              retries: frameworkInfo.frameworkRetryPolicyState.retriedCount,
```
yqwang-ms: >retriedCount [](start = 63, length = 12)

Seems we need to use this?

transientNormalRetriedCount +
transientConflictRetriedCount +
nonTransientRetriedCount +
unKnownRetriedCount

rest-server/src/models/job.js
```diff
@@ -47,36 +47,22 @@ class Job {
...
+              createdTime: frameworkInfo.firstRequestTimestamp,
+              completedTime: frameworkInfo.frameworkCompletedTimestamp,
+              appExitCode: frameworkInfo.applicationExitCode
+            };
           });
+          jobList.sort((a, b) => b.createdTime - a.createdTime);
```
yqwang-ms: >createdTime [](start = 35, length = 11)

Have you handled null case?

rest-server/src/models/job.js
abuccts: updated.
what's the relationship between `retriedCount` and those four retriedCounts? I thought you added a sum count

rest-server/src/models/job.js
abuccts: updated.

rest-server/src/models/job.js
abuccts: Updated.

rest-server/src/models/job.js
```diff
@@ -41,42 +41,61 @@ class Job {
...
-          }, (err) => {
-            jobList.sort((a, b) => b.createdTime - a.createdTime);
-            next(jobList, err);
+            return {
+              name: frameworkInfo.frameworkName,
+              username: frameworkInfo.userName,
+              state: this.convertJobState(frameworkInfo.frameworkState, frameworkInfo.applicationExitCode),
+              subState: frameworkInfo.frameworkState,
+              retries: retries,
+              createdTime: frameworkInfo.firstRequestTimestamp || new Date(2018, 1, 1).getTime(),
```
hwuu: In what circumstance will this field be empty? And I think we shall not use a default value.

yqwang-ms: The old frameworks do not have this field, we ensure new frameworks will always have.

---
In reply to: [167141691](https://github.com/Microsoft/pai/pull/314#discussion_r167141691) [](ancestors = 167141691)

rest-server/src/models/job.js
yqwang-ms: I have not added the counter yet. retriedCount is the number against the retrymaxnumber user set. It may be or may not be all the retries depends on whether fancy RetryPolicy is enabled

---
In reply to: [167132637](https://github.com/Microsoft/pai/pull/314#discussion_r167132637) [](ancestors = 167132637)

yqwang-ms:
LGTM, anyway, pls end to end test before check in.

PR #319, opened 2018-02-09
[Tests] Setup unit test for rest server

abuccts:
Setup unit test for rest server, use [Mocha](https://github.com/mochajs/mocha) for testing and [Chai](https://github.com/chaijs/chai) for Assertion.


fanyangCS:
do you plan to replace npm with yarn in rest server?

abuccts:
It's not urgent to use `yarn`, we can replace `npm` later in another PR.

PR #324, opened 2018-02-13
add cadvisor as exporter

YanjieGao:
deploy cadvisor to collect container level metrics.
This deployment support get container label and container gpu metrics.

### DIFF
--- a/service-deployment/bootstrap/prometheus/node-exporter-ds.yaml.template
+++ b/service-deployment/bootstrap/prometheus/node-exporter-ds.yaml.template
@@ -35,6 +35,18 @@ spec:
         volumeMounts:
         - mountPath: {{ clusterinfo[ 'dataPath' ] }}/prometheus
           name: collector-mount
+        - name: rootfs
+          mountPath: /rootfs
+          readOnly: true
+        - name: var-run
+          mountPath: /var/run
+          readOnly: false
+        - name: sys
+          mountPath: /sys
+          readOnly: true
+        - name: docker
+          mountPath: /var/lib/docker
+          readOnly: true
         name: node-exporter
         args:
           - '--collector.textfile.directory={{ clusterinfo[ 'dataPath' ] }}/prometheus'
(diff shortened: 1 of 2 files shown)


ydye:
After this pr, should we still need aggregating kube c-advisor from prom?

ydye:
And we should make the user know that the ca port shouldn't be the default port because of the ca in kubelet.

YanjieGao:
Kubelet use 4194 as its default ca port. ca docker use 8089 as default port. We could add a comment at the above the config

YanjieGao:
After this PR, prometheus will aggregate ca metrics from PAI deployed ca. Do not use kubelet's ca.

ydye:
So can we close another PR?

YanjieGao:
This PR is export ca metrics. Another PR is prom fetch ca metrics. Not duplicate.

service-deployment/bootstrap/prometheus/node-exporter-ds.yaml.template
```diff
@@ -55,6 +67,36 @@ spec:
         - mountPath: {{ clusterinfo[ 'dataPath' ] }}/prometheus
           name: collector-mount
         name: gpu-exporter
+      - image: google/cadvisor:latest
```
ydye: Do not use latest.

YanjieGao: ok, fix

service-deployment/bootstrap/prometheus/node-exporter-ds.yaml.template
```diff
@@ -55,6 +67,36 @@ spec:
         - mountPath: {{ clusterinfo[ 'dataPath' ] }}/prometheus
           name: collector-mount
         name: gpu-exporter
+      - image: google/cadvisor:v0.29.0
```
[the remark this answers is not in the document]
YanjieGao: yes, hostnet config is at the bottom of the file

ydye:
Merge it, after passing test.

service-deployment/clusterconfig-example.yaml
```diff
@@ -117,7 +115,9 @@ clusterinfo:
     prometheus_port: 9090
     # port for node exporter
     node_exporter_port: 9100
-  
+    # cadvisor port
```
ydye: Why put this configuration in grafana?

YanjieGao: Put it in promethues. This config currently is used by prometheus. The web shows the config is folded.

ydye: Sorry, I didn't expand all the code. 😆

YanjieGao:
pass test

YanjieGao:
pass test

PR #334, opened 2018-02-24
Move ssh function to common lib. And add validation and ut.

ydye:


kubernetes-deployment/test/test_maintainlib_common.py
```diff
@@ -119,6 +119,40 @@ def test_package(self):
...
+
+        port1 = 22
+        self.assertTrue(common.port_validation(port1))
+
+        port2 = "232"
+        self.assertTrue(common.port_validation(port2))
+
+
+
+    def test_port_validation_wrong(self):
```
fanyangCS: >test_port_validation_wrong [](start = 8, length = 26)

no need to use two tests for valid and invalid parameters. please merge the two into one.

kubernetes-deployment/test/test_maintainlib_common.py
```diff
@@ -119,6 +119,40 @@ def test_package(self):
...
+        self.assertFalse(common.ipv4_address_validation(addr2))
+
+
+
+    def test_port_validation_correct(self):
+
+        port1 = 22
+        self.assertTrue(common.port_validation(port1))
+
+        port2 = "232"
```
fanyangCS: please also try 65536, -22. also should we consider port 0 as invalid port (anyport)?

kubernetes-deployment/test/test_maintainlib_common.py
```diff
@@ -119,6 +119,40 @@ def test_package(self):
...
+    def test_ipv4_address_validation_correct(self):
+
+        addr = "128.0.0.1"
+        self.assertTrue(common.ipv4_address_validation(addr))
+
+
+
+    def test_ipv4_address_validation_wrong(self):
+
+        addr1 = "128.0.0.x"
```
fanyangCS: >128.0.0.x [](start = 17, length = 9)

please merge the valid and invalid tests into one.
also, please test 
1) ipv6 address, e.g., 0x:ef:ff:....
2) localhost
3) 127.0.0.1 (we should consider this as invalid, right?)
4) string, e.g., "mydefaultip"
5) invalid value in each ., e.g., 0.-1.0.0, 0.0.358.0

ydye: 127.0.0.1 is valid. other are all invalid.

ydye: 256.0.0.0 will cover the scenario of 0.0.358. (0  -  255)

kubernetes-deployment/test/test_maintainlib_common.py
ydye: 0 is valid.

ydye: Both port and ipv4 address checking should be treated as format validation, not definition validation.

kubernetes-deployment/test/test_maintainlib_common.py
ydye: Done

PR #351, opened 2018-03-01
[REST Server] Show err.stack when job config is not valid.

hwuu:


rest-server/src/middlewares/parameter.js
```diff
@@ -20,18 +20,19 @@
...
 /**
  * Validate parameters.
  */
 const validate = (schema) => {
   return (req, res, next) => {
     Joi.validate(req.body, schema, (err, value) => {
       if (err) {
-        logger.warn('parameters validation error\n%s', err.stack);
+        errorType = 'ParameterValidationError';
+        errorMessage = 'Could not validate request data.\n' + err.stack;
```
abuccts: variable not defined

abuccts:
Do we really need to show error stack to user?
If so, set `NODE_ENV` to `development` is enough.

abuccts:
Please refer to [rest-server/src/config/express.js#L52-L59](https://github.com/Microsoft/pai/blob/master/rest-server/src/config/express.js#L52-L59) for error handler.

rest-server/src/middlewares/parameter.js
```diff
@@ -20,18 +20,19 @@
...
+        logger.warn('[%s] %s', errorType, errorMessage);
         return res.status(500).json({
-          error: 'ParameterValidationError',
-          message: 'could not validate request data'
+          error: errorType,
+          message: errorMessage
```
abuccts: please add trailing comma to pass lint check

hwuu: Will do.

hwuu:
@abuccts : I think we should provide users, rather than us, sufficient hint to help them spot the causes of errors. If we rely on the 'env' config, we should restart the service to make it take effect.

PR #352, opened 2018-03-01
Fix A BUG in drivers script.

ydye:


service-deployment/src/drivers/dockerfile.template
```diff
@@ -61,6 +62,7 @@ COPY install-* enable-nvidia-persistenced-mode.sh $STAGE_DIR/
 RUN chmod u+x install-nvidia-drivers
 RUN chmod u+x install-all-drivers
 RUN chmod u+x enable-nvidia-persistenced-mode.sh
+RUN dos2unix *
```
fanyangCS: >dos2unix [](start = 4, length = 8)

do you assume all input files are dos style?

PR #364, opened 2018-03-06
Can add ssh

wangcan0329:
docker container support ssh


rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -52,7 +52,25 @@ export PAI_JOB_TASK_ROLE_COUNT={{{ taskRolesNumber }}}
 export PAI_JOB_TASK_ROLE_LIST={{{ taskRoleList }}}
 export PAI_KILL_ALL_ON_COMPLETED_TASK_NUM={{{ jobData.killAllOnCompletedTaskNumber }}}
 
-export PAI_CONTAINER_HOST_PORT=$(( (RANDOM % 55535) + 10001 ))
+# generate container host port
+PAI_CONTAINER_HOST_PORT=$(( (RANDOM % 55535) + 10001 ))
```
hwuu: Can you write `PAI_CONTAINER_HOST_PORT=$(( (RANDOM % 55535) + 10001 ))` once? For example:
```
while true
  CONTAINER_HOST_PORT=...
  if not lsof -i:${PAI_CONTAINER_SSH_PORT}
    break
  else
    sleep 1
```
In this way, we don't have to change two lines if we need to modify the port range in the future.

rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -105,6 +123,73 @@ export PAI_TASK_ROLE_LIST=$PAI_JOB_TASK_ROLE_LIST
...
+      "[INFO]" "begin to download ssh key pair from hdfs ..."
+  hdfs dfs -get "${hdfs_ssh_folder}/.ssh/" "/root/" || exit 1
+fi
+# Start ssh service
+start_ssh_service
+# Generate ssh connect info file in "PAI_CONTAINER_ID-PAI_CURRENT_CONTAINER_IP-PAI_CONTAINER_SSH_PORT" format on hdfs
+hdfs dfs -touchz ${hdfs_ssh_folder}/$PAI_CONTAINER_ID-$PAI_CONTAINER_HOST_IP-$PAI_CONTAINER_SSH_PORT || exit 1
+
+# Write env to bashrc
+env | sed "s/^/export /" >> ~/.bashrc
```
abuccts: `grep "^PAI"` before `sed`?

rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -52,7 +52,25 @@ export PAI_JOB_TASK_ROLE_COUNT={{{ taskRolesNumber }}}
 export PAI_JOB_TASK_ROLE_LIST={{{ taskRoleList }}}
 export PAI_KILL_ALL_ON_COMPLETED_TASK_NUM={{{ jobData.killAllOnCompletedTaskNumber }}}
 
-export PAI_CONTAINER_HOST_PORT=$(( (RANDOM % 55535) + 10001 ))
+# generate container host port
```
hwuu: `generate` --> `Determine the`

rest-server/src/templates/dockerContainerScript.mustache
wangcan0329: I discussed with Bin, if we only keep environment started with "PAI", some configurations like hdfs will be incorrect when SSH connect into container.

abuccts: ok

PR #371, opened 2018-03-07
[fix] fix the prom tag

YanjieGao:


service-deployment/bootstrap/prometheus/node-label.sh.template
```diff
@@ -18,7 +18,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 {% for host in machinelist %}
-    {% if 'prometheus' in machinelist[ host ] and machinelist[ host ][ 'prometheus' ] == 'true' %}
+    {% if 'prometheus' in machinelist[ host ] and machinelist[ host ][ 'prom' ] == 'true' %}
```
ydye: Please use prometheus as a tag

service-deployment/bootstrap/prometheus/prometheus-deployment.yaml.template
```diff
@@ -30,7 +30,7 @@ spec:
       hostNetwork: true
       hostPID: true
       nodeSelector:
-        prom: "true"
+        prometheus: "true"
```
ydye: The label in the create-label.sh should be also changed.

service-deployment/bootstrap/prometheus/node-label.sh.template
```diff
@@ -18,7 +18,7 @@
...
 kubectl label nodes {{ machinelist[ host ][ 'nodename' ] }} prom=true
```
ydye: Here

ydye:
Before check in this code, please test it first.

PR #377, opened 2018-03-10
[Pylon] Support Yarn web portal

hwuu:
And other improvements included.


pylon/src/nginx.conf.template
```diff
@@ -33,20 +33,35 @@ http {
...
     # General-purpose reverse proxy.
-    location ~ ^/r/(http|https)/([^/]+)/(\d*)(.*)$ {
-      set $target_scheme $1;
-      set $target_host   $2;
-      set $target_port   :$3;
-      set $target_path   $4$is_args$args;
+    # The proxy to WebHDFS API depends on it.
+    #
+
+    location ~ ^/a/([^/]+)/(\d*)(.*)$ {
```
sterowang: >/a [](start = 16, length = 2)

Why change from "/r" to "/a"?

pylon/src/nginx.conf.template
```diff
@@ -33,20 +33,35 @@ http {
...
+      set $target_host   $1;
+      set $target_port   :$2;
+      set $target_path   $3$is_args$args;
       if ($target_port = ":0") {
         set $target_port "";
       }
-      proxy_pass $target_scheme://$target_host$target_port$target_path;
+      proxy_pass http://$target_host$target_port$target_path;
```
sterowang: >http [](start = 17, length = 4)

Will not support https by now?

service-deployment/src/pylon/dockerfile
```diff
@@ -44,6 +44,9 @@ RUN wget http://www.zlib.net/zlib-1.2.11.tar.gz && \
 RUN wget https://www.openssl.org/source/openssl-1.1.0f.tar.gz && \
     tar -zxf openssl-1.1.0f.tar.gz
 
+# subs_filter
+RUN git clone git://github.com/yaoweibin/ngx_http_substitutions_filter_module.git
```
sterowang: >yaoweibin [](start = 31, length = 9)

Does subs_filter has an official git? Or this is the official git?

hwuu: This is the official git. The module is not a standard nginx module listed [here](http://nginx.org/en/docs/), but it is included in the list of officially supported modules from NGINX, Inc. and NGINX Plus Certified Modules from NGINX's partners (see here). In addition, the moudle has a BSD license (see its doc [here](https://github.com/yaoweibin/ngx_http_substitutions_filter_module/tree/master/doc#Copyright__License)) so we have sufficient freedom to control it.

pylon/src/nginx.conf.template
hwuu: If the communication inside the cluster does not have to enable SSL, then we don't have to support https.

pylon/src/nginx.conf.template
hwuu: I think `/a` looks more like a temporary functionality that will eventually be retired.

pylon/src/nginx.conf.template
```diff
@@ -74,26 +97,99 @@ http {
...
     #
-    # Web portals
+    # YARN web portal
+    #
+
+    location ~ ^/yarn$ {
+      # Add '/' to the end of the URL, otherwise there will be a 404 error.
+      return 301 $scheme://$http_host$request_uri/;
+    }
+    location ~ ^/yarn/([^/]+)/(\d+)/(.*/)?([^/]*)$ {
```
sterowang: I'm thinking of we should limit the user's freedom to compose Yarn urls with arbitrary target_port so that users will not able to use this proxy to probe our backend resources randomly but only able to visit RMs and NMs.

And also we should like WebHDFS not let user to specify RM host name but we pick it up from environment for not exposing deployment specific stuff on API path.

pylon/src/nginx.conf.template
sterowang: Sure, we should retire this general proxy or hide it for internal usage.

---
In reply to: [175000956](https://github.com/Microsoft/pai/pull/377#discussion_r175000956) [](ancestors = 175000956)

pylon/src/nginx.conf.template
hwuu: Definitely. But this requires a redesign of the service configuration mechanism, and also some fundamental addition to the nginx configuration file. I'll create a ticket for this feature.

PR #380, opened 2018-03-13
[fix] change prometheus default port & config

YanjieGao:


ydye:
have you tested this port property yes?

YanjieGao:
yes, test this port on east us

PR #383, opened 2018-03-14
[REST Server] [Job Submit] Fix bug and add unit test cases

hwuu:
Fix bug:
- Job controller could not correctly handle errors and returned wrong results.

Unit test cases:
- 1 positive case.
- 4 negative cases.

In addition, some other parts of the code are also changed to make all above changes work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,15 +45,6 @@ matrix:
       script:
         - npm test
     - language: node_js
-      node_js: 8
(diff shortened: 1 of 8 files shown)


hwuu:
The code passed unit test on my devbox but it didn't on Travis. I've created an issue: https://github.com/Microsoft/pai/issues/384.

PR #390, opened 2018-03-15
Support OpenMPI

abuccts:
Support OpenMPI and runtime ssh.


job-tutorial/cntk/cntk-distributed-example.json
```diff
@@ -0,0 +1,30 @@
+{
+  "jobName": "cntk-distributed-jobguid",
+  "image": "pai.run.cntk",
+
+  // prepare cmudict corpus in CNTK format https://git.io/vbT5A and upload to hdfs
```
hwuu: comments are not supported in standard json?

abuccts: right

abuccts:
Tested on [cntk distributed job](https://docs.microsoft.com/en-us/cognitive-toolkit/Multiple-GPUs-and-machines#3-configuring-parallel-training-in-cntk-in-brainscript), still have issues on [/proc/mount messages](https://stackoverflow.com/questions/46138549/docker-openmpi-and-unexpected-end-of-proc-mounts-line) and [indefinite hung](https://github.com/open-mpi/ompi/issues/3380).

rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -204,14 +204,49 @@ else
...
-# Start ssh service
-start_ssh_service
+chmod 400 ~/.ssh/$APP_ID
 # Generate ssh connect info file in "PAI_CONTAINER_ID-PAI_CURRENT_CONTAINER_IP-PAI_CONTAINER_SSH_PORT" format on hdfs
 hdfs dfs -touchz ${hdfs_ssh_folder}/$PAI_CONTAINER_ID-$PAI_CONTAINER_HOST_IP-$PAI_CONTAINER_SSH_PORT || exit 1
 
+# Generate ssh config
+ssh_config_path=${HDFS_LAUNCHER_PREFIX}/${PAI_USER_NAME}/${PAI_JOB_NAME}/ssh/config
+hdfs dfs -mkdir -p ${ssh_config_path} || exit 1
+hdfs dfs -touchz ${ssh_config_path}/$APP_ID-$PAI_CURRENT_TASK_ROLE_NAME-$PAI_CURRENT_TASK_ROLE_CURRENT_TASK_INDEX-$PAI_CONTAINER_HOST_IP-$PAI_CONTAINER_SSH_PORT || exit 1
```
hwuu: What if `$PAI_CURRENT_TASK_ROLE_NAME` contains a dash (`'-'`)?

rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -204,14 +204,49 @@ else
...
   hdfs dfs -get "${hdfs_ssh_folder}/.ssh/" "/root/" || exit 1
 fi
-# Start ssh service
-start_ssh_service
+chmod 400 ~/.ssh/$APP_ID
 # Generate ssh connect info file in "PAI_CONTAINER_ID-PAI_CURRENT_CONTAINER_IP-PAI_CONTAINER_SSH_PORT" format on hdfs
 hdfs dfs -touchz ${hdfs_ssh_folder}/$PAI_CONTAINER_ID-$PAI_CONTAINER_HOST_IP-$PAI_CONTAINER_SSH_PORT || exit 1
 
+# Generate ssh config
+ssh_config_path=${HDFS_LAUNCHER_PREFIX}/${PAI_USER_NAME}/${PAI_JOB_NAME}/ssh/config
```
hwuu: To avoid the confusion with the config of CloudDocker and increase the ease of code maintenance, I suggest use another directory, say `${HDFS_LAUNCHER_PREFIX}/${PAI_USER_NAME}/${PAI_JOB_NAME}/mpi`.

rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -204,14 +204,49 @@ else
...
+  printf "%s %s\n%s\n%s\n\n" \
+    "[ERROR]" "NodeList" \
+    "${NodeList[@]}" \
+    "ssh services in ${#NodeList[@]} containers are available, not equal to $PAI_JOB_TASK_COUNT, exit ..."
+  exit 2
+fi
+for line in "${NodeList[@]}"; do
+  node=(${line//-/ });
+  printf "%s\n  %s\n  %s\n  %s\n  %s\n  %s\n  %s\n" \
+    "Host ${node[0]}-${node[1]}" \
```
hwuu: You can write IP address here. I did an experiment on my own environment.

rest-server/src/templates/dockerContainerScript.mustache
abuccts: pls refer to the `taskRole.name` schema in [job tutorial](https://github.com/Microsoft/pai/tree/master/job-tutorial#json-config-file-for-job-submission)

rest-server/src/templates/dockerContainerScript.mustache
abuccts: the `Host` field in ssh config is just a hostname, it can be an ip string of course, but it is hard to parse.
and what if two containers are using a same ip?

rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -204,14 +204,49 @@ else
...
 # Generate ssh connect info file in "PAI_CONTAINER_ID-PAI_CURRENT_CONTAINER_IP-PAI_CONTAINER_SSH_PORT" format on hdfs
 hdfs dfs -touchz ${hdfs_ssh_folder}/$PAI_CONTAINER_ID-$PAI_CONTAINER_HOST_IP-$PAI_CONTAINER_SSH_PORT || exit 1
 
-# Write env to bashrc
-env | sed "s/^/export /" >> ~/.bashrc
+# Generate ssh config
+ssh_config_path=${HDFS_LAUNCHER_PREFIX}/${PAI_USER_NAME}/${PAI_JOB_NAME}/ssh/config
+hdfs dfs -mkdir -p ${ssh_config_path} || exit 1
+hdfs dfs -touchz ${ssh_config_path}/$APP_ID-$PAI_CURRENT_TASK_ROLE_NAME-$PAI_CURRENT_TASK_ROLE_CURRENT_TASK_INDEX-$PAI_CONTAINER_HOST_IP-$PAI_CONTAINER_SSH_PORT || exit 1
+while [ `hdfs dfs -ls $ssh_config_path | grep "/$PAI_JOB_NAME/ssh/config/$APP_ID-" | wc -l` -lt  $PAI_JOB_TASK_COUNT ]; do
```
fanyangCS: >while [](start = 0, length = 5)

This is only required when the workload need ssh service. could we make it optional?

job-tutorial/cntk/cntk-distributed-example.json
```diff
@@ -0,0 +1,32 @@
+{
+  "jobName": "cntk-distributed-jobguid",
+  "image": "pai.run.cntk",
```
fanyangCS: >pai.run.cntk [](start = 12, length = 12)

if we don't use pai.build.mpi, why do we build it?

rest-server/src/templates/dockerContainerScript.mustache
abuccts: Then we need an additional field in json config file for user to specify?
This part doesn't matter if user don't use ssh service.

job-tutorial/cntk/cntk-distributed-example.json
abuccts: user can build mpi into their image (`pai.run.cntk`), or build from mpi base image(`pai.build.mpi`).

rest-server/src/templates/dockerContainerScript.mustache
abuccts: That's a problem, I will fix it.

fanyangCS: do we have a checker in rest-server to check whether the name complies with the regex?

---
In reply to: [175660963](https://github.com/Microsoft/pai/pull/390#discussion_r175660963) [](ancestors = 175660963)

abuccts: yes, but the regex is `^[A-Za-z0-9\-._~]+$` which supports `-`

PR #393, opened 2018-03-16
Add quotation to rest-server's password

ydye:


service-deployment/bootstrap/rest-server/rest-server.yaml.template
```diff
@@ -53,7 +53,7 @@ spec:
         - name: LOWDB_ADMIN
           value: {{ clusterinfo['restserverinfo']['lowdb_admin'] }}
```
abuccts: the same for username?

ydye: If we assume some others will create a user name with digits, the answer is yes

ydye: Updated

PR #394, opened 2018-03-19
K8S1.9.4 + ETCD 3.2.17 + K8S dashboard 1.8.3

ydye:


kubernetes-deployment/template/kubelet.sh.template
```diff
@@ -19,7 +19,7 @@
 
 docker run \
     --name=kubelet \
-    --volume=/etc/kubernetes/manifests:/etc/kubernetes/manifests:rw \
```
fanyangCS: >manifests [](start = 29, length = 9)

why do we change it? etcd 3 removes it?

kubernetes-deployment/template/kubelet.sh.template
```diff
@@ -36,7 +36,8 @@ docker run \
     {{ clusterconfig['dockerregistry'] }}/hyperkube:{{ clusterconfig['hyperkubeversion'] }} \
     /hyperkube kubelet --address={{ hostcofig['hostip'] }} \
   --register-node=true \
-  --api-servers={{ clusterconfig['api-servers-ip'] }}:8080 \
+  --kubeconfig=/etc/kubernetes/config \
```
fanyangCS: >kubeconfig=/etc/kubernetes/config \ [](start = 4, length = 35)

if we change to a kubeconfig, should we include an example kube config ?

kubernetes-deployment/template/kubelet.sh.template
ydye: To mount the kubeconfig

kubernetes-deployment/template/kubelet.sh.template
ydye: K8s1.9 has removed the option api-servers. Instead, we should write the ip-address in the kubeconfig file.

ydye: The same configuration of kubectl

fanyangCS: do we have an example kubeconfig for kubectl?

---
In reply to: [175327008](https://github.com/Microsoft/pai/pull/394#discussion_r175327008) [](ancestors = 175327008)

ydye: Yes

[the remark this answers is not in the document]
ydye: yes

Issue #348, opened 2018-03-01
Prometheus watches new servers added to the cluster automatically


fanyangCS:
https://prometheus.io/docs/prometheus/latest/configuration/configuration/#kubernetes_sd_config

YanjieGao:
Please see this #395

PR #395, opened 2018-03-20
prometheus node discovery 

YanjieGao:
From this PR prometheus will leverage kubernetes api server info to monitor node.
When nodes add and leave, prometheus can dynamic detect the live nodes and profile them.
To solve this issue #348

### DIFF
--- a/service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
+++ b/service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
@@ -28,6 +28,8 @@ data:
       kubernetes_sd_configs:
       - api_server: '{{ clusterinfo['webportalinfo']['k8s_api_server_uri'] }}'
         role: node
+      # Extract label __address__ inner ip address and replace the port to exporter's port to build target address. 
+      # Prometheus will fetch data from target address. 
       relabel_configs:
       - source_labels: [__address__]
         regex: '([^:]+)(:\d*)?'
@@ -38,6 +40,8 @@ data:
       kubernetes_sd_configs:
       - api_server: '{{ clusterinfo['webportalinfo']['k8s_api_server_uri'] }}'
         role: node
+      # Extract label __address__ inner ip address and replace the port to exporter's port to build target address.  
+      # Prometheus will fetch data from target address. 
       relabel_configs:
       - source_labels: [__address__]
         regex: '([^:]+)(:\d*)?'
--- a/service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
+++ b/service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
@@ -24,7 +24,7 @@ data:
     # Scrape config for cluster components.
     scrape_configs: 
     - job_name: 'node_exporter'
-      scrape_interval: 15s
+      scrape_interval: 30s
       kubernetes_sd_configs:
       - api_server: '{{ clusterinfo['webportalinfo']['k8s_api_server_uri'] }}'
         role: node
@@ -34,7 +34,7 @@ data:
         replacement: '${1}:{{ clusterinfo['prometheusinfo']['node_exporter_port'] }}'
         target_label: __address__ 
     - job_name: 'cadvisor'
-      scrape_interval: 15s
+      scrape_interval: 30s
       kubernetes_sd_configs:
       - api_server: '{{ clusterinfo['webportalinfo']['k8s_api_server_uri'] }}'
         role: node
--- a/service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
+++ b/service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
@@ -30,7 +30,7 @@ data:
         role: node
       relabel_configs:
       - source_labels: [__address__]
-        regex: '(.*):(.*)'
+        regex: '([^:]+)(:\d*)?'
         replacement: '${1}:{{ clusterinfo['prometheusinfo']['node_exporter_port'] }}'
         target_label: __address__ 
     - job_name: 'cadvisor'
@@ -40,7 +40,7 @@ data:
         role: node
       relabel_configs:
       - source_labels: [__address__]
-        regex: '(.*):(.*)'
+        regex: '([^:]+)(:\d*)?'
         replacement: '${1}:{{ clusterinfo['prometheusinfo']['cadvisor_port'] }}'
         target_label: __address__ 
 
--- a/service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
+++ b/service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
@@ -30,7 +30,7 @@ data:
         role: node
       relabel_configs:
       - source_labels: [__address__]
-        regex: '(.*):10250'
+        regex: '(.*):(.*)'
         replacement: '${1}:{{ clusterinfo['prometheusinfo']['node_exporter_port'] }}'
         target_label: __address__ 
     - job_name: 'cadvisor'
@@ -40,7 +40,7 @@ data:
         role: node
       relabel_configs:
       - source_labels: [__address__]
-        regex: '(.*):10250'
+        regex: '(.*):(.*)'
         replacement: '${1}:{{ clusterinfo['prometheusinfo']['cadvisor_port'] }}'
         target_label: __address__ 
 
--- a/service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
+++ b/service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
@@ -23,18 +23,22 @@ data:
   prometheus.yml: |-
     # Scrape config for cluster components.
     scrape_configs: 
-    - job_name: "node_exporter"
-      scrape_interval: "15s"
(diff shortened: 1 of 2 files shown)


service-deployment/bootstrap/prometheus/prometheus-deployment.yaml.template
```diff
@@ -33,9 +33,9 @@ spec:
         prometheus: "true"
       containers:
       - name: prometheus
-        image: quay.io/coreos/prometheus:v1.1.1
+        image: prom/prometheus:v2.1.0
```
fanyangCS: >prom [](start = 15, length = 4)

is this a local docker registry?

service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
```diff
@@ -23,18 +23,22 @@ data:
   prometheus.yml: |-
     # Scrape config for cluster components.
     scrape_configs: 
-    - job_name: "node_exporter"
-      scrape_interval: "15s"
```
fanyangCS: >scrape_interval: "15s" [](start = 6, length = 22)

if removed, where do you specify the scrape interval?

service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
```diff
@@ -23,18 +23,22 @@ data:
...
-        {% for host in machinelist %}
-        - {{ machinelist[host]['ip']}}:{{ clusterinfo['prometheusinfo']['cadvisor_port'] }}
-        {% endfor %}
+    - job_name: 'node_exporter'
+      kubernetes_sd_configs:
+      - api_server: '{{ clusterinfo['webportalinfo']['k8s_api_server_uri'] }}'
+        role: node
+      relabel_configs:
+      - source_labels: [__address__]
+        regex: '(.*):10250'
```
fanyangCS: >10250 [](start = 21, length = 5)

what does 10250 mean? can you explain the purpose of relabel_configs?

service-deployment/bootstrap/prometheus/prometheus-deployment.yaml.template
YanjieGao: it is official registry

service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
YanjieGao: if removed, will use default value.

service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
YanjieGao: prometheus will defaultly get each node __address__, it contains 10250 port, then replace it with exporter port to fetch metrics.
image

Also ref: https://github.com/prometheus/docs/issues/836

service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
fanyangCS: we need to control the scrape interval. please fix it.

---
In reply to: [176297106](https://github.com/Microsoft/pai/pull/395#discussion_r176297106) [](ancestors = 176297106)

service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
fanyangCS: this means port 10250 is reserved? we should state this in prom deployment service to avoid other service accidently uses it.

---
In reply to: [176298014](https://github.com/Microsoft/pai/pull/395#discussion_r176298014) [](ancestors = 176298014)

YanjieGao: i check 10250 was reserved. this will add a comment at clusterconfig-example.json

service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
YanjieGao: added

service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
fanyangCS: in here: [https://github.com/Microsoft/pai/blob/master/service-deployment/clusterconfig-example.yaml#L121](https://github.com/Microsoft/pai/blob/master/service-deployment/clusterconfig-example.yaml#L121), you should add a field "prometheus_reserved_port" and set it to 10250, and add a comment saying don't change it. This way, a tool can parse all the sections and detect whether there exist port confliction.

---
In reply to: [176301229](https://github.com/Microsoft/pai/pull/395#discussion_r176301229) [](ancestors = 176301229)

sterowang: 10250 is the default Kubelet port. For us, I think we can just ignore any port part in __address__ and add with the exporters port specified in our config file.

---
In reply to: [176300755](https://github.com/Microsoft/pai/pull/395#discussion_r176300755) [](ancestors = 176300755,176298014)

service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
```diff
@@ -23,18 +23,24 @@ data:
...
-        - {{ machinelist[host]['ip']}}:{{ clusterinfo['prometheusinfo']['cadvisor_port'] }}
-        {% endfor %}
+    - job_name: 'node_exporter'
+      scrape_interval: 30s
+      kubernetes_sd_configs:
+      - api_server: '{{ clusterinfo['webportalinfo']['k8s_api_server_uri'] }}'
+        role: node
+      relabel_configs:
+      - source_labels: [__address__]
+        regex: '(.*):(.*)'
```
sterowang: >:(.*) [](start = 20, length = 5)

You can makes this part optional such as (.*)[:\d+] so that whatever __address__ contains port or not, you just pick up the host part.

service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
YanjieGao: yes, prometheus call k8s api to get this port and output the __address__ tag. currently, the 10250 has been remove and use a regex to cover any port

service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
sterowang: For example, if ":" does not exists in __address__, the matching will fail thus the real ports will not be added.

---
In reply to: [178004678](https://github.com/Microsoft/pai/pull/395#discussion_r178004678) [](ancestors = 178004678)

YanjieGao: (.*)[:\d+] cannot match hostip case.
Assume we match 3 test case:
result1 = p.findall('127.0.0.1:1250')
result2 = p.findall('127.0.0.1:')
result3 = p.findall('127.0.0.1')

(.*)[:\d+] will cut result3's tail. 
['127.0.0.1:125']
['127.0.0.1']
['127.0.0.']

Other regex is also meet the same problem, hostip case is so flexible and difficult to combine with hostip: together to match. 

Suggest we could have another direction to think this. How about we only match the IP, this is easy to match the 3 cases.

YanjieGao: Solution2: use ip regex to extract the ip address and ignore other tag
p = re.compile(r'((?:(?:25[0-5]|2[0-4]\d|((1\d{2})|([1-9]?\d)))\.){3}(?:25[0-5]|2[0-4]\d|((1\d{2})|([1-9]?\d))))')

result1 = p.findall('127.0.0.1:1250')
result2 = p.findall('127.0.0.1:')
result3 = p.findall('127.0.0.1')

[('127.0.0.1', '0', '127', '0', '1', '', '1')]
[('127.0.0.1', '0', '127', '0', '1', '', '1')]
[('127.0.0.1', '0', '127', '0', '1', '', '1')]

sterowang: I found this regx works:
([^:]+)(:\d*)?

---
In reply to: [178223066](https://github.com/Microsoft/pai/pull/395#discussion_r178223066) [](ancestors = 178223066)

YanjieGao:
Thanks, test ([^:]+)(:\d*)? work, and refine this to pr code

service-deployment/bootstrap/prometheus/prometheus-configmap.yaml.template
```diff
@@ -23,18 +23,24 @@ data:
...
-        {% endfor %}
+    - job_name: 'node_exporter'
+      scrape_interval: 30s
+      kubernetes_sd_configs:
+      - api_server: '{{ clusterinfo['webportalinfo']['k8s_api_server_uri'] }}'
+        role: node
+      relabel_configs:
+      - source_labels: [__address__]
+        regex: '([^:]+)(:\d*)?'
+        replacement: '${1}:{{ clusterinfo['prometheusinfo']['node_exporter_port'] }}'
```
sterowang: Can you add a line of comment to describe the purpose of the relabel config?

PR #398, opened 2018-03-20
[Rest Server]  Refine job submission api

abuccts:
Refine job submission api, use POST `/api/v1/jobs`.

### DIFF
--- a/rest-server/src/controllers/job.js
+++ b/rest-server/src/controllers/job.js
@@ -54,6 +54,30 @@ const load = (req, res, next, jobName) => {
   });
 };
 
+const init = (req, res, next) => {
+  const jobName = req.body.jobName;
+  new Job(jobName, (job, error) => {
+    if (error) {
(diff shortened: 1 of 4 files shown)


rest-server/src/controllers/job.js
```diff
@@ -54,6 +54,30 @@ const load = (req, res, next, jobName) => {
...
+          message: 'internal server error',
+        });
+      }
+    } else {
+      logger.warn('duplicate job %s', jobName);
+      return res.status(400).json({
+        error: 'DuplicateJobSubmission',
+        message: `job already exists: '${jobName}'`,
+      });
+    }
```
sterowang: If job submit successfully, will this callback not called?

abuccts: callback is called in `if` block, [line 63](https://github.com/Microsoft/pai/pull/398/files#diff-7a88852b42b873aaca542a12ef83bf5bR63)

PR #412, opened 2018-03-26
[VC] Setting up Hadoop queues during deployment

hwuu:
**Note**: This PR is for design discussion and validation at the moment.

Design:
- Step 1. Define virtual clusters (equivalent to Hadoop queues) in `clusterconfig.yaml` file.
- Step 2. Prepare `capacity-scheduler.xml.template` file under the bootstrap folder of the Hadoop service.
- Step 3. Put the file into the list of templates to be rendered during deployment in `service.yaml` file.
- Step 4. In each container of resourcemanager, nodemanager, and jobhistory, copy the `capacity-scheduler.xml` file from configmap folder (mounted from the hadoop-configuration folder by kubernetes) to the hadoop configuration folder, and then start the corresponding Hadoop sub-service.

The purpose of steps 1 ~ 3 is to prepare the `capacity-scheduler.xml` file and put it into the configmap; and then what step 4 does is to start Hadoop sub-services (services inside the container) based on this file.

Currently, if we need to do config reloading without restarting the pod or the service inside each container, we can manually update the `capacity-scheduler.xml` file in the config-map, and then manually trigger the config-reloading command inside each container.

In the future, we need to consider following things:
- Users only need to change the `clusterconfig.yaml` file, i.e. user-level configuration.
- Besides `deploy.py`, we also need a `push-cfg.py` script to publish the new config to the corresponding cluster.
- Inside each container in the cluster, a daemon process is needed to:
    1. Monitor the user-level configuration changes;
    2. Generate new system-level configuration files, i.e. configuration files for the specific service inside the container, according to the new user-level configuration.
    3. Determine what action should be taken (run a reload command or restart the service inside the container) according to the change;
    4. Take the action and guarantee its success.
- That means the daemon process should take care of the whole life cycle of the service inside the container.


service-deployment/bootstrap/hadoop-service/start.sh
```diff
@@ -19,6 +19,12 @@
 
 pushd $(dirname "$0") > /dev/null
 
+# Step 4 of 5 to set up Hadoop queues.
+# Put capacity-scheduler.xml, instantiated from capacity-scheduler.xml.template by deploy.py,
+# into the hadoop-configuration folder, so that the file can be accessed in the container
+# as a part of the configmap.
+cp capacity-scheduler.xml ./hadoop-configuration/
```
ydye: It‘s better directly put capacity-scheduler to /hadoop-configuration

hwuu: Thanks. Changed the code.

service-deployment/prepare_hadoop_config.sh
```diff
@@ -56,7 +56,9 @@ cp hadoop/yarn-site.xml bootstrap/hadoop-service/hadoop-configuration/resourcema
 
 
 # patch for hadoop-run
-patch src/hadoop-run/capacity-scheduler.xml config-patch/capacity-scheduler.xml.patch
+# It is not necessary to patch capacity-scheduler.xml here because this file
+# will be prepared by deploy.py during the process of service deployment.
+#patch src/hadoop-run/capacity-scheduler.xml config-patch/capacity-scheduler.xml.patch
```
ydye: @fanyangCS Should we directly put scheduler.xml to the hadoop-configuration. Or make it as a patch?

service-deployment/clusterconfig-example.yaml
```diff
@@ -63,6 +63,19 @@ clusterinfo:
     configmapname: hadoop-configuration
     hadoop_vip: 10.0.3.9
 
+  # Step 1 of 4 to set up Hadoop queues.
+  # Define all virtual clusters, equivalent concept of Hadoop queues.
+  virtualClusters:
+    default:
+      description: The default virtual cluster for testing purpose only.
+      weight: 20
```
sterowang: >weight [](start = 6, length = 6)

Maybe call it GPUs is simpler to understand for cluster admins?

hwuu: Indeed. But it is better to have an additional mechanism to calculate the percentage automatically according to the # of gpus defined here and the total # of gpus defined in clusterconfig.yaml file. I'll figure out a solution.

service-deployment/prepare_hadoop_config.sh
fanyangCS: yes. make it a patch.

---
In reply to: [177291209](https://github.com/Microsoft/pai/pull/412#discussion_r177291209) [](ancestors = 177291209)

service-deployment/bootstrap/hadoop-service/hadoop-configuration/capacity-scheduler.xml.template
```diff
@@ -0,0 +1,128 @@
+<!--
+TODO: Find a method to deal with the license.
```
fanyangCS: please make this file a patch

service-deployment/bootstrap/hadoop-service/hadoop-configuration/jobhistory-generate-script.sh
```diff
@@ -20,6 +20,20 @@
 cp  /hadoop-configuration/jobhistory-mapred-site.xml $HADOOP_CONF_DIR/mapred-site.xml
 cp  /hadoop-configuration/jobhistory-yarn-site.xml $HADOOP_CONF_DIR/yarn-site.xml
 
+# Step 4(3) of 4 to set up Hadoop queues.
+# Patch the original capacity-scheduler.xml file with the one in the configmap.
+# Three sub-steps:
+#   1) Extract the head part from the original capacity-scheduler.xml file (a.k.a. the target file).
+#   2) Extract the 'configuration' part from the capacity-scheduler.xml in the configmap (a.k.a. the content file).
+#   3) Concatent the above two parts and use it to replace the original capacity-scheduler.xml file.
+# Do the same for resourcemanager and nodemanager.
```
sterowang: Looks this part is common to 3 services, can you put them into a common script and call from 3 services to avoid code duplication?

PR #417, opened 2018-03-27
Add api test for user authentication and management

wangcan0329:


rest-server/test/userToken.js
```diff
@@ -0,0 +1,95 @@
...
+      .set('Authorization', 'Bearer ' + validToken)
+      .send(JSON.parse(global.mustache.render(getTokenTemplate, { 'username': 'new_user', 'password': 'abcdef' })))
+      .end((err, res) => {
+        global.chai.expect(res, 'status code').to.have.status(401);
+        global.chai.expect(res, 'response format').be.json;
+        global.chai.expect(res.body.message, 'response message').equal('authentication failed');
+        done();
+      });
+  });
+
```
hwuu: Add another negative case: user name does not exist.

rest-server/test/userManager.js
```diff
@@ -0,0 +1,388 @@
...
+          'modifiedIndex': 23,
+          'createdIndex': 23
+        }
+      });
+  });
+
+  //
+  // Negative cases
+  //
+
```
hwuu: Add a positive case here?

rest-server/test/userManager.js
```diff
@@ -0,0 +1,388 @@
...
+      .set('Authorization', 'Bearer ' + validToken)
+      .send(JSON.parse(global.mustache.render(newUserTemplate, { 'username': 'new_user', 'password': 'abcdef', 'admin': false, 'modify': true })))
+      .end((err, res) => {
+        global.chai.expect(res, 'status code').to.have.status(201);
+        global.chai.expect(res, 'response format').be.json;
+        global.chai.expect(res.body.message, 'response message').equal('update successfully');
+        done();
+      });
+  });
+
```
hwuu: Add negative cases? Update a non-existing user, etc. And what if password is set to an empty string?

rest-server/test/userManager.js
```diff
@@ -0,0 +1,388 @@
...
+        'prevNode': {
+          'key': '/users/new_user/passwd',
+          'value': '8507b5d862306d5bdad95b3d611b905ecdd047b0165ca3905db0d861e76bce8f3a8046e64379e81f54865f7c41b47e57cec887e5912062211bc9010afea8ab12',
+          'modifiedIndex': 10,
+          'createdIndex': 10
+        }
+      });
+
+      nock(etcdHosts)
+      .put('/v2/keys/users/new_user/admin', { 'value': 'false' })
```
hwuu: Indent 2 spaces.

rest-server/test/userManager.js
```diff
@@ -0,0 +1,388 @@
...
+    global.chai.request(global.server)
+      .put('/api/v1/user')
+      .set('Authorization', 'Bearer ' + validToken)
+      .send(JSON.parse(global.mustache.render(newUserTemplate, { 'username': 'new_user', 'password': '123456', 'admin': true, 'modify': true })))
+      .end((err, res) => {
+        global.chai.expect(res, 'status code').to.have.status(500);
+        global.chai.expect(res, 'response format').be.json;
+        global.chai.expect(res.body.message, 'response message').equal('update failed');
+        done();
+      });
```
hwuu: Add another negative case: what if the user that calls the api is not an admin?

rest-server/test/userToken.js
wangcan0329: Sure

rest-server/test/userManager.js
wangcan0329: There is a positive case behind it

rest-server/test/userManager.js
wangcan0329: if password sets to an empty string will trigger parameter validation error. "password\" is not allowed to be empty.
Already add two negative case:1) update a non-existing user 2)update password to be empty

rest-server/test/userManager.js
wangcan0329: Add negative case to add, update and delete user with non-admin token.

PR #419, opened 2018-03-30
VC feature

hwuu:


Asakuri:
plase update readme and job-tutorial accordingly.

PR #420, opened 2018-03-30
Add TensorFlow mpi job example

abuccts:
Add dockerfile and example for TensorFlow mpi job.


job-tutorial/tensorflow/tensorflow-mpi-example.json
```diff
@@ -0,0 +1,32 @@
...
+  "codeDir": "$PAI_DEFAULT_FS_URI/path/tensorflow-mpi-jobguid/code",
+
+  "taskRoles": [
+    {
+      "name": "ps_server",
+      "taskNumber": 2,
+      "cpuNumber": 2,
+      "memoryMB": 8192,
+      "gpuNumber": 0,
+      "command": "pip --quiet install scipy && python code/tf_cnn_benchmarks.py --local_parameter_device=cpu --num_gpus=4 --batch_size=32 --model=resnet20 --variable_update=parameter_server --data_dir=$PAI_DATA_DIR --data_name=cifar10 --train_dir=$PAI_OUTPUT_DIR --ps_hosts=$PAI_TASK_ROLE_ps_server_HOST_LIST --worker_hosts=$PAI_TASK_ROLE_worker_HOST_LIST --job_name=ps --task_index=$PAI_CURRENT_TASK_ROLE_CURRENT_TASK_INDEX --server_protocol=grpc+mpi"
```
sterowang: >4 [](start = 120, length = 1)

This should be 0. And do we have a env variable for gpu count of current task role? If has, we'd better to use that variable here.

PR #422, opened 2018-03-30
[VC] job rest api support VC field

wangcan0329:


rest-server/src/config/job.js
```diff
@@ -90,6 +90,9 @@ const jobConfigSchema = Joi.object().keys({
   killAllOnCompletedTaskNumber: Joi.number()
     .integer()
     .default(0),
+  queue: Joi.string()
```
hwuu: Do not use 'queue' here; use 'virtualCluster' instead. This is because 'queue' is a concept of Hadoop, but users only need to know 'vc', the user-level concept.

rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -87,6 +87,7 @@ docker run --name $docker_name \
   --volume /tmp/pai-root/alive/$APP_ID:/alive \
   --label PAI_HOSTNAME="$(hostname)" \
   --label PAI_JOB_NAME={{{ jobData.jobName }}} \
+  --label PAI_JOB_QUEUE_NAME={{{ jobData.queue }}} \
```
hwuu: PAI_JOB_VC_NAME

rest-server/test/jobSubmission.js
```diff
@@ -67,31 +68,37 @@ describe('Submit job: POST /api/v1/jobs', () => {
...
-  const validToken = global.jwt.sign({username: 'user1', admin: false}, process.env.JWT_SECRET, {expiresIn: 60});
+  const validToken = global.jwt.sign({ username: 'user1', admin: false }, process.env.JWT_SECRET, { expiresIn: 60 });
   const invalidToken = '';
 
   //
   // Positive cases
   //
 
-  it('Case 1 (Positive): Submit a job.', (done) => {
+  it('Case 1 (Positive): Submit a job with default queue', (done) => {
```
hwuu: ... 'to the default VC'.

rest-server/test/jobSubmission.js
```diff
@@ -100,6 +107,19 @@ describe('Submit job: POST /api/v1/jobs', () => {
       });
   });
 
+  it('Case 2 (Positive): Submit a job which queue=vc1.', (done) => {
```
hwuu: 'Submit a job to vc1'.

rest-server/src/templates/yarnContainerScript.mustache
hwuu: By the way, we need to have the following layers of concepts in mind:
- Layer 1: Virtual cluster;
- Layer 2: Job;
- Layer 3: Task role;
- Layer 4: Task.
So please organize your code lines in the order of above layers.

PR #429, opened 2018-04-04
[Rest Server] Add label for gpu in docker container

abuccts:
Add label for gpu in docker container.

### DIFF
--- a/rest-server/src/templates/yarnContainerScript.mustache
+++ b/rest-server/src/templates/yarnContainerScript.mustache
@@ -48,20 +48,20 @@ while /bin/true; do
   [ -f "/tmp/pai-root/alive/$APP_ID/docker_$CONTAINER_ID" ] && [ ! "$(docker ps | grep $docker_name)" ] && break
 done &
 
-gpu_no=''
+gpu_id=''
 nvidia_devices=''
 {{# taskData.gpuNumber }}
 nvidia_devices+='--device=/dev/nvidiactl --device=/dev/nvidia-uvm'
 gpu_bitmap=$(perl -e 'printf "%b",'$CONTAINER_GPUS)
 for (( i=0,j=$((${#gpu_bitmap}-1)); i<${#gpu_bitmap}; i++,j-- )); do
   if [ ${gpu_bitmap:$i:1} -gt 0 ]; then
-    gpu_no+="$j,"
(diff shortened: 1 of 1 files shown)


rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -85,6 +89,7 @@ docker run --name $docker_name \
   --security-opt apparmor:unconfined \
   --volume /var/drivers/nvidia/current:/usr/local/nvidia:ro \
   --volume /tmp/pai-root/alive/$APP_ID:/alive \
+  --label GPU_NO=$gpu_no \
```
YanjieGao: suggest follow previous style. refine GPU_NO to PAI_TASK_GPU_IDS

abuccts: PAI only sees gpu from 0 to `PAI_GPU_NUM - 1`, don't you need gpu on host?

YanjieGao: Only will use this docker label gpu id and job/task info to join with custom exporter's gpu metrics. Have question. exporter will use nvidia-smi's gpu id series, need confirm with you docker label's id is consist with nvidia-smi's gpu id sequence?

abuccts: `nvidia-smi` from host or docker get different results. `GPU_NO` is for gpus on the host.

rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -48,16 +48,20 @@ while /bin/true; do
   [ -f "/tmp/pai-root/alive/$APP_ID/docker_$CONTAINER_ID" ] && [ ! "$(docker ps | grep $docker_name)" ] && break
 done &
 
+gpu_no=''
 nvidia_devices=''
 {{# taskData.gpuNumber }}
 nvidia_devices+='--device=/dev/nvidiactl --device=/dev/nvidia-uvm'
 gpu_bitmap=$(perl -e 'printf "%b",'$CONTAINER_GPUS)
```
sterowang: >CONTAINER_GPUS [](start = 36, length = 14)

Oops, how about simply export $CONTAINER_GPUS directly?

PR #432, opened 2018-04-04
Enable pagination in rest server

abuccts:
Enable pagination in rest server.

### DIFF
--- a/rest-server/README.md
+++ b/rest-server/README.md
@@ -225,6 +225,14 @@ Configure the rest server ip and port in [service-deployment/clusterconfig.yaml]
     GET /api/v1/jobs/:jobName
     ```
 
+    *Parameters*
+    ```
+    {
+      "page": "page number, starting from 1",
+      "perPage": "entries per page"
+    }
+    ```
+
     *Response if succeeded*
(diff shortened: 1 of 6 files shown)


rest-server/README.md
```diff
@@ -225,6 +225,14 @@ Configure the rest server ip and port in [service-deployment/clusterconfig.yaml]
     GET /api/v1/jobs/:jobName
     ```
 
+    *Parameters*
```
fanyangCS: >Parameters [](start = 5, length = 10)

too complicated. can we just set job number?

wangcan0329: Maybe we will do job pagination and filter in restserver later, so Yifan adds these two parameters.

hwuu: If this feature is to ease the load of web portal to do the pagination, then the web portal hopes the following requirements can be considered:

- Job list may grow very fast (tens of jobs per minute) in the future, so maybe the job list get from the rest-server can be fixed by specifying a timestamp to the API.
- To render the navigation bar correctly, the web portal also needs to know the total number of jobs in the list.

PR #435, opened 2018-04-06
[VC] Add new REST API to request vc info

Asakuri:
Add new REST API to request vc info


rest-server/README.md
```diff
@@ -463,3 +463,86 @@ Configure the rest server ip and port in [service-deployment/clusterconfig.yaml]
...
       "message": "job execute error"
     }
+
+11. `GET vcs/`
+
+    Get the list of vcs.
+
+    *Request*
+    ```
+    GET /api/v1/vcs
```
hwuu: Avoid using abbreviations unless it is really necessary: please use `GET /api/v1/virutal-clusters`.

rest-server/src/config/express.js
```diff
@@ -54,7 +54,7 @@ app.use((err, req, res, next) => {
   logger.warn('%s', err.stack);
   res.status(err.status || 500).json({
     message: err.message,
-    error: config.env === 'development' ? err.stack : {},
+    error: config.env === 'development' ? err.stack : '',
```
hwuu: Why?

rest-server/README.md
Asakuri: "error" field should be string type according to readme, but the default error handler returns a dict instead. Inconsistent json filed type will result to deserialization failure.

rest-server/src/config/express.js
Asakuri: "error" field should be string type according to readme, but the default error handler returns a dict instead. Inconsistent json filed type will result to deserialization failure.

rest-server/src/controllers/vc.js
```diff
@@ -0,0 +1,82 @@
...
+// The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING
+// BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+// module dependencies
+const Vc = require('../models/vc');
```
hwuu: `Vc` --> `VirtualCluster`.

rest-server/src/models/vc.js
```diff
@@ -0,0 +1,93 @@
...
+          error = new Error('VcNotFound');
+        }
+      }
+      next(this, error);
+    });
+  }
+
+  getCapacitySchedulerInfo(queueInfo) {
+    let queues = {};
+    function traverse(queueInfo, queueDict) {
```
hwuu: Could you explain why use a recursive way to iterate all queues?

rest-server/src/job.js
```diff
@@ -0,0 +1,102 @@
...
+// The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING
+// BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+// module dependencies
+const Joi = require('joi');
```
wangcan0329: Please move this content to pai/rest-server/src/config/job.js

rest-server/src/models/vc.js
```diff
@@ -0,0 +1,93 @@
...
+
+// module dependencies
+const unirest = require('unirest');
+const yarnConfig = require('../config/yarn');
+const logger = require('../config/logger');
+
+class Vc {
+  constructor(name, next) {
+    this.getVcList((vcList, error) => {
+      if (error === null) {
```
wangcan0329: use if(!error)? pls refer to [difference between null and undefined](https://stackoverflow.com/questions/5101948/javascript-checking-for-null-vs-undefined-and-difference-between-and/5101991#5101991 
)

rest-server/src/controllers/vc.js
Asakuri: filename

rest-server/src/models/vc.js
Asakuri: The capacity scheduler supports hierarchical queues.

rest-server/src/job.js
Asakuri: add this file by mistake

rest-server/src/models/vc.js
```diff
@@ -0,0 +1,93 @@
...
+          absoluteMaxCapacity: queueInfo.absoluteMaxCapacity,
+          absoluteUsedCapacity: queueInfo.absoluteUsedCapacity,
+          maxApplications: queueInfo.maxApplications,
+          maxApplicationsPerUser: queueInfo.maxApplicationsPerUser,
+          numActiveApplications: queueInfo.numActiveApplications,
+          numApplications: queueInfo.numApplications,
+          numContainers: queueInfo.numContainers,
+          numPendingApplications: queueInfo.numPendingApplications,
+          resourcesUsed: queueInfo.resourcesUsed,
+          state: queueInfo.state,
```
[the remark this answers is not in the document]
Asakuri: This may need more discussion. But if we use dominant calculator, how to get the number of available GPUs for specific VC? you don't know which kind of resource take the lead.

rest-server/src/models/vc.js
Asakuri: I explicitly pass null when no error occurs. And null is more suitable than undefined when it represents a blank object reference. 
"If an API like the DOM in browsers needs an object reference that's blank, we use null, not undefined" is also mentioned in the reference you supplied.

rest-server/src/config/express.js
hwuu: OK. Which part in PAI will deserialize this field? If the existing implementation has not been introducing obvious problem to the system, I suggest we leave it as it is, in case we introduce more problems.

rest-server/src/models/vc.js
hwuu: I got it. Good idea. However, we do not have such a complicated design yet, and we won't have such a design in the foreseeable future. So I suggest we make things simpler -- simplicity usually results in higher performance, higher robustness, and higher maintainability.

rest-server/src/models/vc.js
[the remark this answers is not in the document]
hwuu: You're correct. But I'm talking about user experience instead of implementation. Let's discuss on scrum meeting.

rest-server/README.md
```diff
@@ -463,3 +463,86 @@ Configure the rest server ip and port in [service-deployment/clusterconfig.yaml]
...
+
+    *Request*
+    ```
+    GET /api/v1/virtual-clusters/:vcName
+    ```
+
+    *Response if succeeded*
+    ```
+    {
+      "capacity":0,
```
hwuu: Any field that comes from the user's config?

hwuu: Please define the input/output of the API, and then have some rounds of discussion before the actual implementation.

rest-server/src/controllers/vc.js
```diff
@@ -0,0 +1,82 @@
...
+const logger = require('../config/logger');
+
+/**
+ * Load vc and append to req.
+ */
+const load = (req, res, next, vcName) => {
+  new VirtualCluster(vcName, (vcInfo, error) => {
+    if (error) {
+      if (error.message === 'VcNotFound') {
+        logger.warn('load vc %s error, could not find vc', vcName);
```
hwuu: `vc` --> `virtual cluster`.

PR #446, opened 2018-04-09
fix port not trige preemption issue

qinchen123:
The issue is:
When there is not enough node  (filteredNodes.size() < 1)  to meet the request,  Framework launcher will throw a NotAvailableException if user request ports (getPortNumber() > 0)

We need Framework lunch still submit this job to yarn to preempt other's job. 

The fix is:
1. Allocate the port after Label And GPU filtering.
2. Do resource fit in. 
3. if don't have enough resource( filteredNodes.size() < 1) , still submit the request with the port selected in step1. 
4. If there have enough resource, and Step1 don't successfully select the ports, select the port in the small result set.

### DIFF
--- a/frameworklauncher/src/main/java/com/microsoft/frameworklauncher/applicationmaster/SelectionManager.java
+++ b/frameworklauncher/src/main/java/com/microsoft/frameworklauncher/applicationmaster/SelectionManager.java
@@ -250,19 +250,28 @@ public class SelectionManager { // THREAD SAFE
           "select: reuse pre-selected ports: [%s]", ValueRangeUtils.toString(reusePorts));
       optimizedRequestResource.setPortRanges(reusePorts);
     }
+    if (optimizedRequestResource.getPortNumber() > 0 && ValueRangeUtils.getValueNumber(optimizedRequestResource.getPortRanges()) <= 0) {
+      //If port is required and the portRange is not set in previous steps, allocate port ranges from all candidate nodes.
+      //It is possible to not get the required ports. But it is ok, we will try to select ports again after get the small set of candidate nodes.
+      List<ValueRange> portRanges = selectPortsFromFilteredNodes(optimizedRequestResource);
+      if (ValueRangeUtils.getValueNumber(portRanges) == optimizedRequestResource.getPortNumber()) {
+        optimizedRequestResource.setPortRanges(portRanges);
+      }
+    }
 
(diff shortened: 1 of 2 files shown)


frameworklauncher/src/main/java/com/microsoft/frameworklauncher/applicationmaster/SelectionManager.java
```diff
@@ -250,19 +250,32 @@ public synchronized SelectionResult select(ResourceDescriptor requestResource, S
           "select: reuse pre-selected ports: [%s]", ValueRangeUtils.toString(reusePorts));
       optimizedRequestResource.setPortRanges(reusePorts);
     }
+    if (optimizedRequestResource.getPortNumber() > 0 && ValueRangeUtils.getValueNumber(optimizedRequestResource.getPortRanges()) <= 0) {
+      //If port is required and the portRange is not set in previous steps, allocate port ranges from all candidate nodes.
+      List<ValueRange> portRanges = selectPortsFromFilteredNodes(optimizedRequestResource);
+      if (ValueRangeUtils.getValueNumber(portRanges) == optimizedRequestResource.getPortNumber()) {
+        optimizedRequestResource.setPortRanges(portRanges);
```
yqwang-ms: Seems no need to check

frameworklauncher/src/main/java/com/microsoft/frameworklauncher/applicationmaster/SelectionManager.java
```diff
@@ -286,6 +299,15 @@ public synchronized SelectionResult select(ResourceDescriptor requestResource, S
...
 
+  public synchronized List<ValueRange> selectPortsFromFilteredNodes(ResourceDescriptor optimizedRequestResource) {
+    List<ValueRange> overlapPorts = new ArrayList<ValueRange>();
+    for (String nodeName : filteredNodes) {
+      overlapPorts = ValueRangeUtils.intersectRangeList(overlapPorts, allNodes.get(nodeName).getAvailableResource().getPortRanges());
+    }
+    return ValueRangeUtils.getSubRange(overlapPorts, optimizedRequestResource.getPortNumber(),
+        conf.getAmContainerMinPort());
+  }
+
```
yqwang-ms: selectPorts and selectPortsFromFilteredNodes can be private and overloaded.

PR #447, opened 2018-04-10
mock cluster object model

ydye:


service-deployment/paiLibrary/clusterObjectModel/objectModelFactory.py
```diff
@@ -0,0 +1,63 @@
...
+        self.initializeConifugration()
+
+
+        self.objectModel = dict()
+
+
+
+    def loadYamlConfig(configPath):
+
+        with open(configPath, "r") as file:
```
hwuu: Is `file` a preserved keyword?

ydye: I will change it to f

cluster-configuration/cluster-configuration.yaml
```diff
@@ -50,7 +50,8 @@ machine-sku:
```
hwuu: I suggest we use 3 yaml files, each maps to one of the 3 layers of configurations:

- `infrastructure.yaml` --> SKUs and machines.
- `kubernetes.yaml` --> K8s configurations and role definitions, taking `infrastructure.yaml` only as its dependency.
- `pai.yaml` --> Configuration of services, taking `kubernetes.yaml` only as its dependency.

kubernetes-deployment/bootstrap.py
```diff
@@ -28,6 +28,7 @@
 import time
 import logging
 import logging.config
+from clusterObjectModel import objectModelFactory
```
hwuu: I suggest we follow the coding-style practices of PEP8: https://www.python.org/dev/peps/pep-0008/

service-deployment/paiLibrary/clusterObjectModel/objectModelFactory.py
```diff
@@ -0,0 +1,80 @@
+# Copyright (c) Microsoft Corporation
```
hwuu: Why both of kubernetes-deployment and service-deployment have the same 'clusterObjectModel' folder?

ydye: Because we have no time to integrate them into one module before this release.

cluster-configuration/cluster-configuration.yaml
ydye: Actually, k8s-role-definition should be transparent for user. And user only configure 3 files now.

PR #449, opened 2018-04-10
add docker inspect cmd result parse

YanjieGao:
parse docker labels then used at job metrics exporter

### DIFF
--- a/prometheus/exporter/docker_inspect.py
+++ b/prometheus/exporter/docker_inspect.py
@@ -31,6 +31,10 @@ def inspect(argv):
     except subprocess.CalledProcessError as e:
         raise RuntimeError("command '{}' return with error (code {}): {}".format(e.cmd, e.returncode, e.output))
 
-# execute cmd example: python .\docker_inspect.py 33a22dcd4ba3 
+def main(argv):
+    containerId = argv[0]
+    inspect([containerId])
+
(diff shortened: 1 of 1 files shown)


prometheus/exporter/docker_inspect.py
```diff
@@ -0,0 +1,36 @@
...
+        labels.append("container_label_{0}=\"{1}\"".format(key.replace(".", "_"), jsonObject[0]["Config"]["Labels"][key]))
+
+    for env in jsonObject[0]["Config"]["Env"]:
+        envItem = env.split("=")
+        envs.append("container_env_{0}=\"{1}\"".format(envItem[0].replace(".", "_"), envItem[1]))
+
+    inspectMetrics = {"env": envs, "labels": labels}
+    return inspectMetrics
+    
+def inspect(argv):
```
sterowang: >argv [](start = 12, length = 4)

You should process argv in main and pass specific parameters to inspect such as inspect(containerId).

prometheus/exporter/docker_inspect.py
```diff
@@ -0,0 +1,40 @@
...
```
sterowang: >argv [](start = 12, length = 4)

Just use "containerId" as parameter directly instead of argv here.

PR #451, opened 2018-04-10
[VC] Refine vc view; add yarn link to each vc entry.

hwuu:
image

### DIFF
--- a/webportal/src/app/vc/vc.component.js
+++ b/webportal/src/app/vc/vc.component.js
@@ -43,6 +43,7 @@ const loadData = (specifiedVc) => {
         specifiedVc: specifiedVc,
         data: data,
         formatNumber: formatNumber,
+        yarnWebPortalUri: webportalConfig.yarnWebPortalUri,
         grafanaUri: webportalConfig.grafanaUri,
       });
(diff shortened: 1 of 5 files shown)


webportal/src/app/vc/vc.component.js
```diff
@@ -43,14 +43,15 @@ const loadData = (specifiedVc) => {
...
+        yarnWebPortalUri: webportalConfig.yarnWebPortalUri,
         grafanaUri: webportalConfig.grafanaUri,
       });
       $('#content-wrapper').html(vcHtml);
       table = $('#vc-table').dataTable({
         scrollY: (($(window).height() - 265)) + 'px',
         lengthMenu: [[20, 50, 100, -1], [20, 50, 100, 'All']],
         columnDefs: [
-          { type: 'natural', targets: [0, 1] },
+          { type: 'natural', targets: [0, 1, 2, 3, 4, 5, 6] },
```
Asakuri: What's this for? Why before is only [0, 1]?

hwuu: It was a bug.

hwuu: For defining the method of sorting each column.

PR #471, opened 2018-04-12
Mock Cluster object Model.

ydye:


service-deployment/bootstrap/prometheus/service.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - drivers
+
+templatelist:
+  - node-label.sh
+  - node-exporter-ds.yaml
+  - prometheus-configmap.yaml
+  - prometheus-deployment.yaml
+  - cleanup.sh
+
+startscript: start.sh
```
fanyangCS: >start [](start = 13, length = 5)

pls check w/ Yanjie, I recall he removed it?

ydye: This file wasn't used.

PR #478, opened 2018-04-15
[Service Deployment] Add end-to-end test service

abuccts:
Add end-to-end test service.


service-deployment/bootstrap/end-to-end-test/node-label.sh.template
```diff
@@ -0,0 +1,24 @@
...
+# The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING
+# BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+# DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+{% for host in machinelist %}
+    {% if 'endtoendtest' in machinelist[ host ] and machinelist[ host ][ 'endtoendtest' ] == 'true' %}
```
ydye: I think this endtoendtest2 could be run as a deployment， because different from the namenode and resourcemanger, this service could run everywhere.
Another benefits to run it as a deployment, we can remove this label. And simplify the cluster-configuration.

tests/end_to_end/README.md
```diff
@@ -1,21 +0,0 @@
-# End-to-end Test
```
fanyangCS: I suggest to keep the readme. just stating that we by default run an e2e test in the bed and show how to find it and stop in

service-deployment/bootstrap/end-to-end-test/node-label.sh.template
abuccts: updated

tests/end_to_end/README.md
ydye: When deployed by kubernetes, admin could easily stop it by dashboard or kubectl.

service-deployment/src/end-to-end-test/etc/launcher.json
```diff
@@ -15,7 +15,16 @@
         "resource": {
           "cpuNumber": 1,
           "memoryMB": 512,
-          "portRanges": [],
+          "portDefinitions": {
+            "http": {
+              "start": 0,
+              "count": 1
```
sterowang: If we don't care the port numbers just count, can we just omit this field?

qinchen123: "start": 0 means any port.  You mean only need set count and ignore start ?  

the current code is not support this scenario, the framework will failed at Json mapping to class if "start" is not set. 

I am not sure if the reset-server can auto-fill the "start" if user not set it in user json. 

 

---
In reply to: [181942244](https://github.com/Microsoft/pai/pull/478#discussion_r181942244) [](ancestors = 181942244)

abuccts: This test json is sent to framework launcher directly.

tests/end_to_end/README.md
abuccts: added in `service.yaml`

PR #481, opened 2018-04-16
[Rest Server] Allow admin execute all jobs

abuccts:
Allow admin execute all jobs.

### DIFF
--- a/rest-server/test/jobExecutionType.js
+++ b/rest-server/test/jobExecutionType.js
@@ -102,7 +102,7 @@ describe('Job execution type API /api/v1/jobs/:jobName/executionType', () => {
       });
   });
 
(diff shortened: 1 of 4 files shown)


rest-server/test/jobExecutionType.js
```diff
@@ -102,17 +102,17 @@ describe('Job execution type API /api/v1/jobs/:jobName/executionType', () => {
       });
   });
 
-  it('should not stop other user\'s job', (done) => {
+  it('admin should stop other user\'s job successfully', (done) => {
     chai.request(server)
```
sterowang: Suggest you do not change the normal user unauthorized test case but keep it and add a new admin test case.

PR #498, opened 2018-04-18
[Web Portal] Add logo

hwuu:
image


abuccts:
It's better to have favicons in different resolutions generated by the vector file to fit different devices, instead of a small one. Otherwise, it will be unclear like the icon in your screenshot.

hwuu:
@abuccts : Indeed. Increased the resolution to 256x256.

PR #502, opened 2018-04-19
[REST Server] Save the original job config content

hwuu:
Previous method has an issue: the value of the constant `originData` in `putJob()` function changes after it's initialized, causing the saved job config content still includes unexpected fields such as `username` and other auto-generated ones. This is because `originData` is a *reference* to `data`. So if `data` changes `originData` also changes.

### DIFF
--- a/rest-server/src/middlewares/parameter.js
+++ b/rest-server/src/middlewares/parameter.js
@@ -35,6 +35,7 @@ const validate = (schema) => {
           message: errorMessage,
         });
       } else {
+        req.original_body = req.body;
         req.body = value;
         next();
       }
--- a/rest-server/src/models/job.js
+++ b/rest-server/src/models/job.js
@@ -132,8 +132,7 @@ class Job {
       });
   }
 
-  putJob(name, data, username, next) {
-    const originData = data;
+  putJob(name, data, originalData, username, next) {
     data.username = username;
     if (!data.outputDir.trim()) {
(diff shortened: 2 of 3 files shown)


rest-server/src/models/job.js
```diff
@@ -132,8 +132,7 @@ class Job {
       });
   }
 
-  putJob(name, data, username, next) {
-    const originData = data;
```
abuccts: just use `Object.assign` ?
```
putJob(name, originalData, ...) {
  const data = {};
  Object.assign(data, originalData);
  ...
}
```

hwuu: Actually this `originalData` you proposed is not so original, because it has been processed and altered by middlewares, thus we cannot guarantee that it can pass the input validation.

rest-server/src/middlewares/parameter.js
```diff
@@ -35,6 +35,7 @@ const validate = (schema) => {
           message: errorMessage,
         });
       } else {
+        req.original_body = req.body;
         req.body = value;
         next();
```
sterowang: What changes will made to new body compare to the original body after the validation?

hwuu: The following fields will be added if not explicitly specified: `portList`, `username`, `dataDir`, `outputDir`, `codeDir`, `gpuType`, `virtualCluster`, and `retryCount`. I don't think we should tell users that we have auto-completed and altered their original request.

PR #507, opened 2018-04-21
[REST Server] Prepare job context using WebHDFS

hwuu:


rest-server/src/util/hdfs.js
```diff
@@ -0,0 +1,103 @@
...
+
+  createFolder(path, options, next) {
+    const targetUrl = this._constructTargetUrl(path, options, 'MKDIRS');
+    try {
+      unirest.put(targetUrl)
+      .end((response) => {
+        if (response.status === 200) {
+          next({status: 'succeeded'}, null);
+        } else {
+          next(null, new Error('InternalServerError'));
```
fanyangCS: >InternalServerError [](start = 32, length = 19)

should we include error status code here?

hwuu: Error-handling is another big problem to handle gracefully and systematically in rest-server's code. Will do it step-by-step in other PRs.

rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -161,6 +161,24 @@ function hdfs_upload_atomically()
...
+  echo "[INFO] Creating output directory: ${PAI_OUTPUT_DIR}"
+  hdfs dfs -mkdir ${PAI_OUTPUT_DIR}
+  echo "[INFO] Output directory ready."
+else
+  # Waiting for the master container to create the output direcotry
+  echo "[INFO] Waiting for output directory: ${PAI_OUTPUT_DIR}"
+  while ! hdfs dfs -test -d ${PAI_OUTPUT_DIR} ; do
+    sleep 10
+  done
+  echo "[INFO] Output directory ready."
```
sterowang: Have you considered let rest server to create the output folder thus we can avoid potentially a lots of hdfs access in each task?

rest-server/src/util/hdfs.js
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  createFolder(path, options, next) {
+    const targetUrl = this._constructTargetUrl(path, options, 'MKDIRS');
+    unirest.put(targetUrl)
+    .end((response) => {
+      if (response.status === 200) {
+        next({status: 'succeeded'}, null);
+      } else {
+        next(null, new Error('InternalServerError'));
```
sterowang: >InternalServerError [](start = 30, length = 19)

If there are error message in response body, suggest put it in Error directly instead of a fixed 'InternalServerError' so that we could get more meaningful info if user reported the error to us.

rest-server/test/etcd2Store.js
```diff
@@ -22,6 +22,13 @@ const db = dbUtility.getStorageObject('etcd2', {
   'hosts': etcdHosts,
 });
 describe('etcd2 get function test', () => {
+  afterEach(function() {
+    if (!nock.isDone()) {
+      //TODO: Revamp this file and enable the following error.
+      //this.test.error(new Error('Not all nock interceptors were used!'));
+      nock.cleanAll();
+    }
+  });
```
sterowang: Make this a global function and call in each test to avoid code dup?

rest-server/src/util/hdfs.js
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  createFile(path, data, options, next) {
+    const targetUrl = this._constructTargetUrl(path, options, 'CREATE');
+    unirest.put(targetUrl)
+    .send(data)
+    .end((response1) => {
+      if (response1.status === 201) {
+        next({status: 'succeeded'}, null);
+      } else if (response1.status == 307) {
```
sterowang: >else if (response1.status == 307) [](start = 8, length = 33)

Recursive call to createFile for simply error handling :blush: ?

rest-server/src/util/hdfs.js
```diff
@@ -0,0 +1,86 @@
...
+    });
+  }
+
+  readFile(path, options, next) {
+    const targetUrl = this._constructTargetUrl(path, options, 'OPEN');
+    unirest.get(targetUrl)
+    .end((response) => {
+      if (response.status === 200) {
+        next({status: 'succeeded', content: response.body}, null);
+      } else if (response.status === 404) {
```
sterowang: >else [](start = 8, length = 4)

No 307 in read file case?

rest-server/src/util/hdfs.js
fanyangCS: agree. 

---
In reply to: [183270909](https://github.com/Microsoft/pai/pull/507#discussion_r183270909) [](ancestors = 183270909)

rest-server/src/templates/dockerContainerScript.mustache
hwuu: Yes I have considered. I prefer this method because of two reasons:

1. The value of `outputDir` URI is specified in `hdfs://<user's hdfs uri>` format, instead of `http://<user's webhdfs uri>` which can be used by WebHDFS, in job config file, thus rest-server can't use this URI to create the folder without falling back to use the `hdfs` command.

2. Container won't introduce too much extra load to HDFS, since all other operations such as ssh preparation and mpi preparation also use the similar method to operate HDFS. In addition, these operations happen within job containers distributed in NMs, so it is not very possible to be a bottleneck of memory utilization.

How do you think?

rest-server/src/util/hdfs.js
hwuu: Good idea! Will improve.

rest-server/src/util/hdfs.js
hwuu: Will do.

rest-server/test/etcd2Store.js
hwuu: Thanks! Will do.

rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -161,6 +161,24 @@ function hdfs_upload_atomically()
   hdfs dfs -mv $tempFolder ${2} || exit 1
 }
 
+#
+# Prepare the output directory on HDFS
+#
+echo "[INFO] TaskRole:${task_role_no} Task:${PAI_TASK_INDEX}"
```
fanyangCS: >task_role_no [](start = 24, length = 12)

PAI_OUTPUT_DIR is optional. maybe you should firstly check whether it is being specified or not.

also, can you not use task_role_no? it is likely to be obsoleted later.

hwuu: `outputDir`, which specifies the value of `PAI_OUTPUT_DIR`, is optional in job config json file, but `PAI_OUTPUT_DIR` itself always contains a sound value within job containers. So it is not necessary to do the check.

I saw Can was using this variable in her code. @wangcan0329 : Any alternative we should use?

rest-server/src/models/job.js
```diff
@@ -132,98 +128,99 @@ class Job {
       });
   }
 
-  putJob(name, data, username, next) {
-    const originData = data;
+  putJob(name, data, originalData, username, next) {
```
abuccts: `data`, `originalData`, and `username` are duplicate.

rest-server/src/models/job.js
```diff
@@ -132,98 +128,99 @@ class Job {
...
+            parallelCallback(error);
           }
+        );
+      },
+    ], (parallelError) => {
+      if (parallelError) {
+        return next(parallelError);
+      } else {
+        unirest.put(launcherConfig.frameworkPath(name))
+        .headers(launcherConfig.webserviceRequestHeaders)
```
abuccts: indentation please

rest-server/test/job.js
```diff
@@ -52,36 +63,43 @@ describe('Jobs API /api/v1/jobs', () => {
...
-      .get('/v1/Frameworks')
-      .query({UserName: 'test1'})
-      .reply(200, {'summarizedFrameworkInfos': [frameworkInfos.summarizedFrameworkInfos[0]]});
-  });
-
   // GET /api/v1/jobs
-  it('should return job list', (done) => {
+  it('[P-01] Get job list', (done) => {
+    nock(launcherWebserviceUri)
+    .get('/v1/Frameworks')
```
abuccts: indentation

rest-server/test/jobConfig.js
```diff
@@ -16,90 +16,98 @@
...
 
     nock(webhdfsUri)
-      .get('/webhdfs/v1/Container/test/job1/JobConfig.json?op=OPEN')
-      .reply(
-        200,
-        {
-          'jobName': 'job1',
-        }
-      );
+    .get('/webhdfs/v1/Container/test/job1/JobConfig.json?op=OPEN')
```
abuccts: indentation

service-deployment/src/rest-server/dockerfile.template
```diff
@@ -17,42 +17,29 @@
...
-    HADOOP_INSTALL=/usr/local/hadoop \
-    HADOOP_HDFS_HOME=${HADOOP_INSTALL} \
-    HADOOP_COMMON_LIB_NATIVE_DIR=${HADOOP_INSTALL}/lib/native \
-    HADOOP_OPTS="-Djava.library.path=${HADOOP_INSTALL}/lib/native"
-
-ENV PATH=$PATH:${HADOOP_INSTALL}/bin
-ENV LD_LIBRARY_PATH=$LD_LIBRARY_PATH:${JAVA_HOME}/jre/lib/amd64/server
+RUN apt-get clean
 
+RUN rm -rf /var/lib/apt/lists/*
```
abuccts: run those in a single layer

rest-server/src/models/job.js
hwuu: Thanks. Will do.

rest-server/src/models/job.js
hwuu: What should I do if I hope to save the original data (exactly what the user submitted) here?

service-deployment/src/rest-server/dockerfile.template
hwuu: Could you explain the benefits of doing so?

rest-server/src/models/job.js
hwuu: I'll abandon the previous PR (https://github.com/Microsoft/pai/pull/502). Let's discuss in this thread.

service-deployment/src/rest-server/dockerfile.template
abuccts: please refer to [docker docs](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers)

rest-server/src/models/job.js
abuccts: please also revert #490 to remove `username` param

rest-server/test/job.js
abuccts: please refer to indentation in [nock examples](https://github.com/node-nock/nock#specifying-request-body) or [line wrapping style](https://google.github.io/styleguide/jsguide.html#formatting-line-wrapping)

rest-server/src/util/hdfs.js
hwuu: Done.

rest-server/src/util/hdfs.js
hwuu: Done.

rest-server/src/util/hdfs.js
hwuu: Done.

rest-server/src/models/job.js
abuccts: set `username` in `data`, just use the previous implementation in https://github.com/Microsoft/pai/pull/490/files

hwuu: Got it. Thanks!

service-deployment/src/rest-server/dockerfile.template
```diff
@@ -17,42 +17,29 @@
```
abuccts: It seems this dockerfile doesn't need to be a template anymore.

PR #517, opened 2018-04-24
[Webportal] add user management page and vc access control support

wangcan0329:


webportal/src/app/user/user-register/user-register.component.ejs
```diff
@@ -1,20 +1,80 @@
...
-          <label for="inputPassword" class="sr-only">Password</label>
-          <input type="password" name="password" id="input-password" class="form-control" placeholder="Password" required>
-          <div class="checkbox">
-            <label><input type="checkbox" name="admin">Admin user</label>
+        <div class="user-management-nav nav-tabs-custom">
+          <ul class="nav nav-tabs pull-left">
+            <li class="active">
+              <a href="#add_user" data-toggle="tab" aria-expanded="true">Add User</a>
+            </li>
+            <li class="">
```
abuccts: remove `class` if it's empty

webportal/src/app/user/user-register/user-register.component.js
```diff
@@ -67,4 +90,130 @@ $(document).ready(() => {
...
+    e.preventDefault();
+    const username = $('#form-update-virtual-cluster :input[name=username]').val();
+    const virtualCluster = $('#form-update-virtual-cluster :input[name=virtualCluster]').val();
+    userAuth.checkToken((token) => {
+      $.ajax({
+        url: `${webportalConfig.restServerUri}/api/v1/user/${username}/virtualClusters`,
+        data: {
+          virtualClusters: virtualCluster,
+        },
+        type: 'PATCH',
```
abuccts: `PATCH` may not be supported in jquery or edge, e.g. [ref1](https://stackoverflow.com/questions/13642044/does-the-jquery-ajax-call-support-patch), [ref2](https://stackoverflow.com/questions/11461414/ajax-json-doesnt-get-sent-in-patch-only).

wangcan0329: Thanks! I can use PUT instead.

PR #520, opened 2018-04-24
Fix: add total memory size line at node view and cluster view page

YanjieGao:
image

### DIFF
--- a/service-deployment/bootstrap/grafana/grafana-configuration/pai-nodeview-dashboard.json.template
+++ b/service-deployment/bootstrap/grafana/grafana-configuration/pai-nodeview-dashboard.json.template
@@ -194,6 +194,13 @@
               "legendFormat": "free",
               "refId": "E",
               "step": 600
+            },
+            {
+              "expr": "node_memory_MemTotal{instance=~'$node'}",
+              "format": "time_series",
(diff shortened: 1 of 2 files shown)


service-deployment/bootstrap/grafana/grafana-configuration/pai-nodeview-dashboard.json.template
```diff
@@ -194,6 +194,13 @@
               "legendFormat": "free",
               "refId": "E",
               "step": 600
+            },
+            {
+              "expr": "node_memory_MemTotal{instance=~'$node'}",
```
fanyangCS: >node_memory_MemTotal{instance=~'$node' [](start = 23, length = 38)

this is just a flat line that never changes. can you change the line to "cache"? or just use two lines: free mem, and (total_mem - free_mem)

YanjieGao: change to 2 line:
free mem, and (total_mem - free_mem)

YanjieGao: image

PR #531, opened 2018-04-26
[Web Portal] Fix bug: HW view sometimes show N/A or a black circle

hwuu:
…etrics data is available.

### DIFF
--- a/webportal/src/app/cluster-view/hardware/hardware.component.js
+++ b/webportal/src/app/cluster-view/hardware/hardware.component.js
@@ -78,7 +78,7 @@ const initCells = (idPrefix, instanceList, table) => {
 //
 
 const loadCpuUtilData = (prometheusUri, currentEpochTimeInSeconds, instanceList, table) => {
-  const metricGranularity = '1m';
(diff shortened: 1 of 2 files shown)


webportal/src/app/cluster-view/hardware/hardware.component.js
```diff
@@ -250,7 +252,7 @@ const loadDiskUtilData = (prometheusUri, currentEpochTimeInSeconds, instanceList
 //
 
 const loadEthUtilData = (prometheusUri, currentEpochTimeInSeconds, instanceList, table) => {
-  const metricGranularity = '1m';
+  const metricGranularity = '5m';
```
sterowang: >5m [](start = 29, length = 2)

Suggest define this kind of numbers into a configuration class which contains all tunable parameters.

hwuu: Agree. This number should be consistent with that in Grafana. @YanjieGao : Do you know where I can find the configuration of this value in Grafana? I'll make the code change then.

PR #533, opened 2018-04-26
[rest-server]add all vc to default admin when deploy

wangcan0329:


rest-server/src/models/user.js
```diff
@@ -225,6 +225,12 @@ const setDefaultAdmin = (callback) => {
   update(etcdConfig.adminName, etcdConfig.adminPass, true, false, (res, status) => {
     if (!status) {
       throw new Error('unable to set default admin');
+    } else {
+      updateUserVc(etcdConfig.adminName, '', (errMsg, res) => {
+        if (errMsg || !res) {
+          throw new Error('unable to set default admin virtual cluster');
```
abuccts: just to make sure, rest server will exit and the pod will restart if throws error

wangcan0329: Only when first deployment of rest server will add default admin user's vc here, if updates failed, I think we need to throw error.

abuccts: I see

PR #556, opened 2018-05-03
fix a bug in deploy.py when using cleanup-deployments command

xingtanzjr:
1. The command to cleanup deployments in `README.md` lose a parameter `-c`
2. There is a bug in `deploy.sh` when execute cleanup command. 

    ```
    # line 184, template is the file name rather than the relative path
    if os.path.exists(template):
    ```
   It should be
   ```
   if os.path.exists("bootstrap/{0}/{1}".format(serv,template)):
   ```

### DIFF
--- a/pai-management/README.md
+++ b/pai-management/README.md
@@ -273,7 +273,7 @@ sudo ./deploy.py -d -p /path/to/configuration/directory
 ## Cleanup your previous deployment
 
 ```
-sudo ./deploy.py -p /path/to/configuration/directory
+sudo ./deploy.py -p /path/to/configuration/directory -c
 sudo ./cleanup-service.py
 ```
 
--- a/pai-management/README.md
+++ b/pai-management/README.md
@@ -273,7 +273,7 @@ sudo ./deploy.py -d -p /path/to/configuration/directory
 ## Cleanup your previous deployment
 
 ```
-sudo ./deploy.py -p /path/to/configuration/directory -c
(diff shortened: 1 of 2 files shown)


pai-management/README.md
```diff
@@ -273,7 +273,7 @@ sudo ./deploy.py -d -p /path/to/configuration/directory
 ## Cleanup your previous deployment
 
 ```
-sudo ./deploy.py -p /path/to/configuration/directory
+sudo ./deploy.py -p /path/to/configuration/directory -c
```
ydye: 1) -c option is designed for clean the generated template file.
2) Why we put an command ```sudo ./deploy.py -p /path/to/configuration/directory -c``` here is because when running the ./cleanup-service.py, some template file should be generated. So to prevent user to run ./cleanup-service directly, we recommand user to run the command firstly.

pai-management/deploy.py
```diff
@@ -181,7 +181,7 @@ def clean_up_generated_file(service_config):
 
         for template in template_list:
 
-            if os.path.exists(template):
+            if os.path.exists("bootstrap/{0}/{1}".format(serv,template)):
```
ydye: Yes, this a bug.

pai-management/README.md
xingtanzjr: OK, fine. I know what you means. The command `sudo ./deploy.py -p /path/to/configuration/directory` should be executed firstly to generate the template files which `./cleanup-service.py` need.

PR #570, opened 2018-05-10
[Experimental Jupyter Integration] Basic functionalities

hwuu:


fanyangCS:
Is there a way not to modify web portal to support Jupyter? We will have a lot to modify for things like Tensorboard, AutoML, etc.

abuccts:
Jupyter Notebook is not necessary and most jobs do not need it. You can add it as an job example instead of building it into base image.

rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -226,6 +227,17 @@ done
...
+  --ip=0.0.0.0 \
+  --port=$PAI_CONTAINER_JUPYTER_PORT \
+  --allow-root \
+  --no-browser \
+  --NotebookApp.token="" \
+  --NotebookApp.allow_origin="*" \
+  --NotebookApp.base_url="/jupyter/${PAI_CONTAINER_HOST_IP}/${PAI_CONTAINER_JUPYTER_PORT}/" \
+  &
+
 # Write env to system-wide environment
```
sterowang: Can we make jupyter started per user request?

PR #580, opened 2018-05-16
add watchdog for service health check

YanjieGao:
- watchdog currently is a single instance(or 3 replica), deploy mode is different with node exporter, so make it as a separate pod
- Report metrics: 
service / k8s / node level metrics. 
- Output:
 it will generate log file which could view from k8s watchdog container logs or ssh to this server /datastorage/prometheus/watchdog.log to view the history. 
It also generate a .prom file which could be used by prometheus in the future


pai-management/bootstrap/prometheus/node-label.sh.template
```diff
@@ -24,4 +24,7 @@ kubectl label nodes {{ machinelist[ host ][ 'nodename' ] }} prometheus=true
     {% if 'node-exporter' in machinelist[ host ] and machinelist[ host ][ 'node-exporter' ] == 'true' %}
 kubectl label nodes {{ machinelist[ host ][ 'nodename' ] }} node-exporter=true
     {% endif %}
+    {% if 'watchdog' in machinelist[ host ] and machinelist[ host ][ 'watchdog' ] == 'true' %}
```
ydye: Please be sure, this label could be work in current master branch. Note: the configuration style has been changed. Please modify the mock cluster-object-model.

YanjieGao: test, when set  watchdog: "true" at cluster-configuration.yaml. it will deploy. This file update this config

prometheus/exporter/watchdog.py
```diff
@@ -0,0 +1,227 @@
...
+def parse_pods_status(podsJsonObject, outputFile):
+    # metrics 
+    kube_pod_status_probe_not_ready = 0
+    kube_pod_status_phase_failed = 0
+    kube_pod_status_phase_unknown = 0
+    pod_container_status_waiting = 0
+    pod_container_status_terminated = 0
+    pod_container_status_not_ready = 0
+    pod_container_status_restarted_pod_count = 0 # if one pod restarts > 1, add 1 to this metrics
+
```
sterowang: Suggest we group this information by controllers (ds, deployments, etc.), and in each log, we include controller name, total pods also.

YanjieGao: added

YanjieGao:
I once check this, its default config has a limitation for 10mb of the file. after 10mb will rotate. This doesn't affect current code logic. Only different is whether remove filelog handler. Watchdog also need logger to write logs to console. Other outputfile is in order to output metrics by prometheus and can't leverage k8s.

YanjieGao:
Hao, will remove .log file

fanyangCS:
please ensure the max memory usage

YanjieGao:
added at yaml file
resources:
        limits:
           memory: "1Gi"

PR #581, opened 2018-05-16
remove support to pai-fs

fanyangCS:


abuccts:
need to update other docs accordingly

hwuu:
A thinking. PAI dosen’t have any control to the HDFS now — user can modify anything, including important PAI-system folders, anyone’s data and code, etc. I think it is time we begin thinking about limit hdfs to internal use only.

fanyangCS:
there are dependencies in test code, rest server, etc. I will close this PR and do it later.

PR #584, opened 2018-05-17
[HOTFIX] Fix permission issue when creating job folders in HDFS

hwuu:


rest-server/src/models/job.js
```diff
@@ -404,6 +410,32 @@ class Job {
     return frameworkDescription;
   }
 
+  _initializeJobContextRootFolders(next) {
+    const hdfs = new Hdfs(launcherConfig.webhdfsUri);
+    async.parallel([
+      (parallelCallback) => {
+        hdfs.createFolder(
+          '/Output',
+          {'user.name': 'root', 'permission': '777'},
```
abuccts: what if the username is omitted, can we use the default username and permission?

rest-server/src/models/job.js
```diff
@@ -138,18 +138,24 @@ class Job {
     }
     userModel.checkUserVc(data.userName, data.virtualCluster, (error, result) => {
       if (!error) {
-        this._prepareJobContext(name, data, (error, result) => {
+        this._initializeJobContextRootFolders((error, result) => {
```
abuccts: do not need to mkdir on every job submission request, only initialize those dirs at the beginning

rest-server/src/models/job.js
hwuu: Default username is 'dr.who' and default permission is 755 (please see the [API doc](https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-hdfs/WebHDFS.html#Make_a_Directory)) -- other users won't have permission to create sub-folders.

rest-server/src/models/job.js
hwuu: Yep.

rest-server/src/models/job.js
abuccts: I mean `/Container` dir where user do not need to access could have permission 755.
As for default user, native hdfs lib will use `root`.

fanyangCS:
please merge this PR to master as well

rest-server/src/models/job.js
hwuu: @abuccts : I think the current implementation, in which all HDFS operations happen when and only when a job is submitted, is easier to maintain. In addition, mkdir operation is light weighted; it will introduce very little overhead to HDFS.

PR #591, opened 2018-05-21
[Examples] Add MXNet example

abuccts:
Add MXNet example.

### DIFF
--- a/example/mxnet/mxnet.autoencoder.json
+++ b/example/mxnet/mxnet.autoencoder.json
@@ -1,0 +1,14 @@
+{
+  "jobName": "mxnet-autoencoder",
+  "image": "paiexample/pai.example.mxnet",
+  "taskRoles": [
+    {
+      "name": "main",
+      "taskNumber": 1,
+      "cpuNumber": 4,
+      "memoryMB": 8192,
+      "gpuNumber": 1,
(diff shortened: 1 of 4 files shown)


example/mxnet/mxnet.autoencoder.json
```diff
@@ -0,0 +1,14 @@
...
+  "jobName": "mxnet-autoencoder",
+  "image": "paiexample/pai.example.mxnet",
+  "taskRoles": [
+    {
+      "name": "main",
+      "taskNumber": 1,
+      "cpuNumber": 4,
+      "memoryMB": 8192,
+      "gpuNumber": 1,
+      "command": "pip install scikit-learn && cd incubator-mxnet/example/autoencoder && python mnist_sae.py --gpu"
```
fanyangCS: >scikit-learn [](start = 30, length = 12)

you mean mxnet? we don't need scikit learn here, right?

abuccts: no, mxnet has been installed in the Docker image.
this [example](https://github.com/apache/incubator-mxnet/blob/master/example/autoencoder/data.py#L23) downloads mnist through `sklearn.datasets`, so `scikit-learn` needs to be installed to run the example

PR #599, opened 2018-05-22
remove pai-fs in readme

fanyangCS:


abuccts:
I have published an image [paiexample/pai.example.hdfs](https://hub.docker.com/r/paiexample/pai.example.hdfs/).

PR #601, opened 2018-05-22
[Launcher]: Best effort to resolve external IP and Host for local machine

yqwang-ms:


frameworklauncher/src/main/java/com/microsoft/frameworklauncher/common/utils/DnsUtils.java
```diff
@@ -17,21 +17,87 @@
...
+
+      localAddress = socketLocalAddress;
     } catch (Exception e) {
-      return null;
+      LOGGER.logWarning(e, "Failed to resolve local address by DatagramSocket");
+    }
+
+    if (localAddress == null) {
+      try {
+        localAddress = Inet4Address.getLocalHost();
```
fanyangCS: should we consider a case where local IP address may change after the DnsUtil initialized? Your later implementation get the IP on the fly, so doesn't have this issue.

yqwang-ms: good catch. 
It would be better to restart service if its external IP changed. 
Anyway from DnsUtils view, it should not have this assumption. 

---
In reply to: [190150781](https://github.com/Microsoft/pai/pull/601#discussion_r190150781) [](ancestors = 190150781)

yqwang-ms: But from caller sider, even I do not cache here, we still do not have mechanism to detect the ip changes and then updated it accordingly. 
So, let's keep it another work item if future required, ok?

---
In reply to: [190167929](https://github.com/Microsoft/pai/pull/601#discussion_r190167929) [](ancestors = 190167929,190150781)

fanyangCS: yes. I suggest we just state our assumption here (by adding some comments). no need for further improvement.

---
In reply to: [190170025](https://github.com/Microsoft/pai/pull/601#discussion_r190170025) [](ancestors = 190170025,190167929,190150781)

PR #636, opened 2018-05-30
[Deployment]Add paictl cluster k8s-bootup command

wangcan0329:


pai-management/paictl.py
```diff
@@ -335,6 +338,61 @@ def pai_service():
         service_management_refrasher.run()
 
 
+def maintain_cluster_k8s(cluster_config, **kwargs):
```
ydye: We'd better remove this function, and package them into a python module. for example ```k8sPaiLibrary/machineApi``` or ```paiLibrary/paiMachine```. Because less logic in this script, easier to maintain this script.

PR #657, opened 2018-06-05
Remove hardcode of user dir. And get it through shell.

ydye:


hwuu:
Did you test it? If you did I will approve.

PR #658, opened 2018-06-06
add k8s console log show watchdog log path

YanjieGao:
also increase watchdog log file max size

### DIFF
--- a/prometheus/exporter/watchdog.py
+++ b/prometheus/exporter/watchdog.py
@@ -25,7 +25,9 @@ from logging.handlers import RotatingFileHandler
 import time
 import common
 
+loggerRoot = logging.getLogger()  
 logger = logging.getLogger("watchdog")  
+loggerCommon = logging.getLogger("common")  
 
 class Service:
     name = ""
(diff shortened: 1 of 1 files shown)


prometheus/exporter/watchdog.py
```diff
@@ -321,14 +323,26 @@ def main(argv):
...
+
+    loggerRoot.setLevel(logging.INFO)  
+    loggerRoot.propagate = False
+    ch = logging.StreamHandler()
+    ch.setLevel(logging.INFO)
+    # create formatter and add it to the handlers
+    formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
+    ch.setFormatter(formatter)
+    loggerRoot.addHandler(ch) 
+    loggerRoot.info("Please view log trace:")
```
fanyangCS: >loggerRoot.info("Please view log trace:")  [](start = 4, length = 42)

remove

prometheus/exporter/watchdog.py
```diff
@@ -321,14 +323,26 @@ def main(argv):
...
+    loggerRoot.propagate = False
+    ch = logging.StreamHandler()
+    ch.setLevel(logging.INFO)
+    # create formatter and add it to the handlers
+    formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
+    ch.setFormatter(formatter)
+    loggerRoot.addHandler(ch) 
+    loggerRoot.info("Please view log trace:") 
+    loggerRoot.info("Log trace path {}/watchdog.log".format(logDir))
+
```
fanyangCS: in loggerRoot, please explain why you do not out log in k8s log directly. 
also, please output the current configuration of logger (e.g.,, maxBytes, etc.)

prometheus/exporter/watchdog.py
```diff
@@ -25,7 +25,9 @@
 import time
 import common
 
+loggerRoot = logging.getLogger()  
 logger = logging.getLogger("watchdog")  
+loggerCommon = logging.getLogger("common")
```
sterowang: Console displays "No handlers could be found for logger "common"", why we need to deal with "common"?

YanjieGao: use pai/pai-management/k8sPaiLibrary/maintainlib/common.py   ssh method. common.py also call logging lib and will generate "No handlers could be found for logger "common""

YanjieGao: If not deal with it, it log will also up to root logger and keep generate to console

PR #660, opened 2018-06-06
[Rest Server] Refine error handling in rest server

abuccts:
Refine error handling in rest server.

### DIFF
--- a/rest-server/src/controllers/job.js
+++ b/rest-server/src/controllers/job.js
@@ -36,6 +36,7 @@ const load = (req, res, next, jobName) => {
         next(error);
       }
     } else {
+      // will be handled by framework launcher in the future
       if (req.method === 'PUT' && req.path === `/${jobName}` || req.method === 'POST' && req.path === '/') {
         const error = new Error(`Submit job error: job ${jobName} already exists.`);
         error.status = httpStatus.BAD_REQUEST;
--- a/rest-server/src/controllers/job.js
+++ b/rest-server/src/controllers/job.js
@@ -18,7 +18,6 @@
 // module dependencies
 const httpStatus = require('http-status');
 const Job = require('../models/job');
-const logger = require('../config/logger');
 
 /**
  * Load job and append to req.
@@ -26,10 +25,10 @@ const logger = require('../config/logger');
 const load = (req, res, next, jobName) => {
   new Job(jobName, (job, error) => {
     if (error) {
-      if (error.message === 'JobNotFound') {
+      if (error.message === 'Could not find job.') {
         if (req.method !== 'PUT' && req.method !== 'POST') {
           error.status = httpStatus.NOT_FOUND;
-          error.message = `could not find job ${jobName}`;
+          error.message = `Get job error: could not find job ${jobName}.`;
           next(error);
         }
       } else {
@@ -38,7 +37,7 @@ const load = (req, res, next, jobName) => {
       }
     } else {
       if (req.method === 'PUT' && req.path === `/${jobName}` || req.method === 'POST' && req.path === '/') {
-        const error = new Error(`job already exists: '${jobName}'`);
+        const error = new Error(`Submit job error: job ${jobName} already exists.`);
         error.status = httpStatus.BAD_REQUEST;
         next(error);
       }
@@ -55,11 +54,11 @@ const list = (req, res, next) => {
   Job.prototype.getJobList(req._query, (jobList, err) => {
     if (err) {
       err.status = httpStatus.INTERNAL_SERVER_ERROR;
-      err.message = 'get job list error';
+      err.message = 'List job error: error occurs in getting job list.';
       next(err);
     } else if (jobList === undefined) {
       err.status = httpStatus.INTERNAL_SERVER_ERROR;
-      err.message = 'could not find job list';
+      err.message = 'List job error: get empty job list.';
       next(err);
     } else {
       return res.status(httpStatus.OK).json(jobList);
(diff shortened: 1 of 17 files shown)


rest-server/src/controllers/job.js
```diff
@@ -26,77 +27,42 @@ const load = (req, res, next, jobName) => {
...
+        next(error);
       }
     } else {
-      if (job.jobStatus.state !== 'JOB_NOT_FOUND' && req.method === 'PUT' && req.path === `/${jobName}`) {
-        logger.warn('duplicate job %s', jobName);
-        return res.status(400).json({
-          error: 'DuplicateJobSubmission',
-          message: `job already exists: '${jobName}'`,
-        });
+      if (req.method === 'PUT' && req.path === `/${jobName}` || req.method === 'POST' && req.path === '/') {
```
fanyangCS: >(req.method [](start = 9, length = 11)

please add a comment: in the future, will remove this check and let Launcher handles this.

rest-server/src/controllers/job.js
```diff
@@ -26,77 +27,42 @@ const load = (req, res, next, jobName) => {
...
+          error.message = `could not find job ${jobName}`;
+          next(error);
         }
       } else {
-        logger.warn('internal server error');
-        return res.status(500).json({
-          error: 'InternalServerError',
-          message: 'internal server error',
-        });
+        error.status = httpStatus.INTERNAL_SERVER_ERROR;
```
fanyangCS: >error [](start = 8, length = 5)

in this case, internal server error is not helpful. can you include the error stack in error.message?

rest-server/src/controllers/job.js
abuccts: ok

rest-server/src/controllers/job.js
abuccts: This status equals to setting `500` http code, the original error will be returned directly. If `NODE_ENV` env var is set to `development`, the stack will be included.

fanyangCS: please change the default env to development then.
here: [https://github.com/Microsoft/pai/blob/8cff97104b6becd9dd761afaceef6e6df4632c20/pai-management/src/rest-server/dockerfile.template#L32](https://github.com/Microsoft/pai/blob/8cff97104b6becd9dd761afaceef6e6df4632c20/pai-management/src/rest-server/dockerfile.template#L32)


---
In reply to: [193369399](https://github.com/Microsoft/pai/pull/660#discussion_r193369399) [](ancestors = 193369399)

hwuu: Dumping error messages to log is not sufficient. Front-end applications can’t access the log, so we need to output something more meaningful directly in API response body.

rest-server/src/controllers/job.js
```diff
@@ -120,23 +86,19 @@ const update = (req, res) => {
...
-        });
+        err.status = httpStatus.UNAUTHORIZED;
+        err.message = `job update error: no virtual cluster right to access ${data.virtualCluster}`;
+        next(err);
       } else {
-        return res.status(500).json({
-          error: 'JobUpdateError',
-          message: err.message,
-        });
+        err.status = httpStatus.INTERNAL_SERVER_ERROR;
```
hwuu: Showing ‘JobUpdateError’ only usually makes users frustrated and confused (we have many reports about this). Suggest include more detailed error description.

rest-server/src/controllers/job.js
abuccts: “returned directly” means it will be returned to http response

rest-server/src/controllers/job.js
abuccts: "JobUpdateError" has been removed already

rest-server/src/models/job.js
```diff
@@ -112,16 +107,19 @@ class Job {
...
               requestRes.body :
               JSON.parse(requestRes.body);
           if (requestRes.status === 200) {
             next(this.generateJobDetail(requestResJson), null);
           } else if (requestRes.status === 404) {
-            next(null, new Error('JobNotFound'));
+            next(null, new Error('Could not find job.'));
           } else {
-            next(null, new Error('InternalServerError'));
+            next(null, new Error('Get job error.'));
```
fanyangCS: >next [](start = 12, length = 4)

can the webportal know the status code in this case?

rest-server/src/controllers/job.js
fanyangCS: still, is it possible to include more detailed err msg here?

---
In reply to: [193630715](https://github.com/Microsoft/pai/pull/660#discussion_r193630715) [](ancestors = 193630715)

rest-server/src/controllers/token.js
```diff
@@ -17,37 +17,31 @@
...
         username: username,
         admin: admin,
       }, tokenConfig.secret, {expiresIn: expiration}, (signError, token) => {
         if (signError) {
-          logger.warn('sign token error\n%s', signError.stack);
-          return res.status(500).json({
-            error: 'SignTokenFailed',
-            message: 'sign token failed',
-          });
+          signError.status = httpStatus.INTERNAL_SERVER_ERROR;
```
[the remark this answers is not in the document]
abuccts: `logger.warn` has already been done in the final handler at https://github.com/Microsoft/pai/blob/master/rest-server/src/config/express.js#L54-L58

rest-server/src/models/job.js
abuccts: yes, http status will be set in controllers if error is not null in callback

fanyangCS:
please resolve the conflict before merge.

rest-server/src/controllers/job.js
```diff
@@ -16,87 +16,53 @@
...
   new Job(jobName, (job, error) => {
     if (error) {
-      if (error.message === 'JobNotFound') {
-        if (req.method !== 'PUT') {
-          logger.warn('load job %s error, could not find job', jobName);
-          return res.status(404).json({
-            error: 'JobNotFound',
-            message: `could not find job ${jobName}`,
-          });
+      if (error.message === 'Could not find job.') {
```
sterowang: >message [](start = 16, length = 7)

Be sure to use "error" instead of "message" anywhere to ensure we work on structured data instead of unstructured human read text.

rest-server/src/controllers/user.js
```diff
@@ -32,84 +32,74 @@ const update = (req, res) => {
...
       if (err || !state) {
-        logger.warn('remove user %s failed', username);
-        return res.status(500).json({
-          error: 'RemoveFailed',
-          message: 'remove failed',
-        });
+        const error = err || new Error('remove failed');
+        error.status = httpStatus.INTERNAL_SERVER_ERROR;
+        next(error);
       } else {
```
sterowang: We should provide "error.message" always in our whole project to give users a clear information includes What and Why, better with How to hint user solve the problem.

abuccts:
Rest server error handling refined in #807. Closed

PR #661, opened 2018-06-06
Fix Bug. 

ydye:
If the command with the sudo need password, then without the pasword write, the process will be blocked。

### DIFF
--- a/pai-management/k8sPaiLibrary/maintainlib/common.py
+++ b/pai-management/k8sPaiLibrary/maintainlib/common.py
@@ -188,9 +188,11 @@ def ssh_shell_paramiko(host_config, commandline):
 
 
 def ssh_shell_paramiko_with_result(host_config, commandline):
+
     hostip = str(host_config['hostip'])
     if ipv4_address_validation(hostip) == False:
         return False
+
     username = str(host_config['username'])
     password = str(host_config['password'])
     port = 22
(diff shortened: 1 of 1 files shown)


pai-management/k8sPaiLibrary/maintainlib/common.py
```diff
@@ -188,20 +188,25 @@ def ssh_shell_paramiko(host_config, commandline):
...
     if 'sshport' in host_config:
         if port_validation(host_config['sshport']) == False:
             return (None, None)
         port = int(host_config['sshport'])
+
     ssh = paramiko.SSHClient()
     ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
     ssh.connect(hostname=hostip, port=port, username=username, password=password)
     stdin, stdout, stderr = ssh.exec_command(commandline, get_pty=True)
+    stdin.write(password + '\n')
```
hwuu: Did you test it? If my memory is correct, the content of stdin will also be included in stdout, making result_stdout contain more lines than expected, thus all code logic that depends on this function should be changed accordingly.

ydye: Yes，on the test bed， if without password output， some node will be blocked.

ydye: Yes, this issue will block the fix.

PR #663, opened 2018-06-06
Write password in stdin when ssh due to the need of sudo. And workaro…

ydye:
…und the password issue.

### DIFF
--- a/pai-management/k8sPaiLibrary/maintainlib/common.py
+++ b/pai-management/k8sPaiLibrary/maintainlib/common.py
@@ -188,9 +188,11 @@ def ssh_shell_paramiko(host_config, commandline):
 
 
 def ssh_shell_paramiko_with_result(host_config, commandline):
+
     hostip = str(host_config['hostip'])
     if ipv4_address_validation(hostip) == False:
         return False
+
     username = str(host_config['username'])
     password = str(host_config['password'])
     port = 22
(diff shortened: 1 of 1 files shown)


pai-management/k8sPaiLibrary/maintainlib/common.py
```diff
@@ -188,28 +188,40 @@ def ssh_shell_paramiko(host_config, commandline):
...
     stdin, stdout, stderr = ssh.exec_command(commandline, get_pty=True)
+    stdin.write(password + '\n')
+    stdin.flush()
     logger.info("Executing the command on host [{0}]: {1}".format(hostip, commandline))
     result_stdout = ""
     for response_msg in stdout:
+        res_msg = response_msg.replace('\n', '')
+        res_msg = res_msg.replace('\r', '')
+        if password == res_msg:
+            # TODO: We should change this behavior.
```
ydye: @hwuu About this comment, what's your opinion?

hwuu: We can simply ignore the first line of the stdout, since it is definitely the content of stdin. And by the way, shall we require that the 'username' have password-less sudo privilege?

ydye: Is it a known behavior ?

ydye:
we'd better revert this pr, if we can't solve this issue asap.

https://github.com/Microsoft/pai/pull/657

pai-management/k8sPaiLibrary/maintainlib/common.py
hwuu: Known behavior, at least according to my own test. We need to do more extensive tests after all. I mean, we may assume that the user does not need to input the password when doing sudo on target machines. Is it a rational requirement?

ydye: - Fixed by filter the first line.
- But in some situation, user will have to input the password. At least, I come across this issue.

hwuu:
Did you test it? If you did I will approve.

ydye:
@hwuu tested. But only deploy part.

PR #668, opened 2018-06-06
[Deploy] Refine the logic of defining VCs.

hwuu:


ydye:
@hwuu We should update corresponding part in mangement-dev.  Because deploy.py will be deleted after the code merging.

hwuu:
@ydye Sure I will change accordingly. Can we merge mgmt-dev to master today?

PR #674, opened 2018-06-07
Revert "Write password in stdin when ssh due to the need of sudo. And workaro…"

ydye:
Reverts Microsoft/pai#663


Gerhut:
Try `sudo --stdin --prompt "" mkdir -p {0}`, it may works.

PR #678, opened 2018-06-07
avoiding None translate to str('None')

hao1939:
@hwuu ,

Please merge this fix, it will check "None" before convert to string.

### DIFF
--- a/pai-management/paictl.py
+++ b/pai-management/paictl.py
@@ -84,15 +84,14 @@ def login_docker_registry(docker_registry, docker_username, docker_password):
 
 def generate_secret_base64code(docker_info):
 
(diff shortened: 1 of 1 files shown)


pai-management/paictl.py
```diff
@@ -110,8 +110,8 @@ def generate_secret_base64code(docker_info):
 
 def generate_docker_credential(docker_info):
```
hwuu: Not the latest code? Please see: https://github.com/Microsoft/pai/blob/b6161ef164abc4d60916c7e54615786b4b564816/pai-management/paictl.py#L85

hwuu: Please merge the latest code from `management-dev` branch and have a try again.

PR #682, opened 2018-06-07
Solve Textfile busy issue when deploying service.

ydye:
In the [SO problem: What generates the “text file busy” message in Unix?
](https://stackoverflow.com/questions/16764946/what-generates-the-text-file-busy-message-in-unix?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa), they says, ```The Text file busy error in specific is about trying to modify an executable while it is executing. The "Text" here refers to the fact that the file being modified is the text segment for a running program. This is a very special case, and not the generic one that your answer seems to suggest. Even so, your answer isn't entirely incorrect. – ArjunShank```

And we have a lot of shell and code will first set the permission of the script with the command ``` chmod u+x```, and with high potential, it will break the process of deployment.

This afternoon, @hao1939 , comes across this issue.
```
Error2: 
/bin/sh: 1: ./bootstrap/end-to-end-test/start.sh: Text file busy
2018-06-07 09:18:25,723 [ERROR] - paiLibrary.common.linux_shell : Failed to execute the start script of service end-to-end-test
``` 

## My solution

Remove all code which will call ```chmod``` to make script executable. And instead, run the script with the command following ```/bin/bash script.sh ```

### DIFF
--- a/pai-management/bootstrap/cluster-configuration/delete.sh
+++ b/pai-management/bootstrap/cluster-configuration/delete.sh
@@ -19,6 +19,6 @@
 
 pushd $(dirname "$0") > /dev/null
 
-sh stop.sh
+/bin/bash stop.sh
 
 popd > /dev/null
--- a/pai-management/bootstrap/cluster-configuration/start.sh
+++ b/pai-management/bootstrap/cluster-configuration/start.sh
@@ -19,8 +19,8 @@
 
 pushd $(dirname "$0") > /dev/null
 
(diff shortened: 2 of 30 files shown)


hwuu:
Still don’t quite understand. Did you mean, for example, foo.sh, was being modified by ‘chmod’ while being executed by shell? These two operations are carried out in sequence in the code, why did they overlap with each other?

PR #689, opened 2018-06-08
Refine VC configuration; fix bugs

hwuu:
Logic:
- Each VC will be assigned with (capacity / total_capacity * 100%) of the resources in the system.
- The system will automatically create the 'default' VC with 0 capacity, if 'default' VC has not been explicitly specified in the configuration file.
- The system will split resources to each VC evenly if the total capacity is 0.
- The capacity of each VC will be automatically set to 0 if it is a negative number in the configuration file.

Test cases (all passed):
- Define capacities in float numbers.
- Do not include virtualClusters section in service-configurations.yaml file.
- Do not include 'default' VC in virtualClusters section.
- Set the capacity of a VC as a negative number.
- Set the capacities of all VCs as 0.


hao1939:
It works with the default config. 👍

PR #695, opened 2018-06-10
[fix] rm the gpu log file probe log

YanjieGao:
rm the gpu file check as lspci will check no gpu machine


ydye:
After this change, could we disable the readness proble failed info if there is no GPU on the target node?

YanjieGao:
yes  if no gpu machine will not show probe log. if contains gpu but driver error will show error log

PR #696, opened 2018-06-11
job metrics network metric change data source

YanjieGao:
as docker status cmd netIO always shows zero for job containers. replace use docker status cmd get network metrics to cadvisor get container network metrics.

### DIFF
--- a/prometheus/exporter/cadvisor_parse.py
+++ b/prometheus/exporter/cadvisor_parse.py
@@ -5,16 +5,16 @@ import re
 import socket,fcntl,struct
 
 def get_ip_address(ifname):
-    s=socket.socket(socket.AF_INET,socket.SOCK_DGRAM)
+    s = socket.socket(socket.AF_INET,socket.SOCK_DGRAM)
     return socket.inet_ntoa(fcntl.ioctl(s.fileno(), 0x8915, struct.pack('256s',ifname[:15]))[20:24])
 
 def extractData(regex, content, index=1): 
-  r = ''
-  p = re.compile(regex) 
-  m = p.search(content) 
-  if m: 
-    r = m.group(index) 
-  return r 
+  result = ''
+  pattern = re.compile(regex) 
+  match = pattern.search(content) 
+  if match: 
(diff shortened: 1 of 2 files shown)


prometheus/exporter/docker_stats.py
```diff
@@ -52,21 +53,34 @@ def convert_to_byte(data):
...
+                netOutMetric = outItem["value"]
+
+        netMetric =  {"in": netInMetric, "out": netOutMetric}
+    
         id = data[i][0]
         containerInfo = {
             "id": data[i][0],
             "CPUPerc": parse_percentile(data[i][1]),
             "MemUsage_Limit": parse_usage_limit(data[i][2]),
-            "NetIO": parse_io(data[i][3]),
```
fanyangCS: >data[i][3] [](start = 30, length = 10)

why doesn't  the data[i][3]  of docker stats work?

prometheus/exporter/cadvisor_parse.py
```diff
@@ -0,0 +1,47 @@
...
+def parseLine(line, inOrOut):
+    line = line.split(" ")
+    metric = line[1]
+    regex = r',name=\"(.*)\",n'
+    index = 1
+    name = extractData(regex, line[0], index)
+    return {"name": name, "value": metric, "inOut": inOrOut}
+
+def get_network_metric():
+    cadvisor_metrics = requests.get("http://{}:4194/metrics".format(get_ip_address('eth0'))).text
```
fanyangCS: >eth0 [](start = 84, length = 4)

we can't do this. we cannot assume the name "eth0" is always there.

prometheus/exporter/docker_stats.py
```diff
@@ -22,6 +22,7 @@
 import re
 import logging  
 logger = logging.getLogger("gpu_expoter")  
+import cadvisor_parse
```
fanyangCS: >cadvisor_parse [](start = 7, length = 14)

you assume cadvisor is always deployed?

PR #706, opened 2018-06-11
Update eslint command

abuccts:
Update eslint command in package.json.


Gerhut:
Why not just `eslint .` ?

Gerhut:
I think directly `eslint .` is better, Is there any reason for avoid linting test files?

abuccts:
`eslint .` will also check files under `dist` and `node_modules`

Issue #711, opened 2018-06-12
Build service image 'frameworklauncher' failed


hao1939:
Step to reproduce:
`sudo docker build -t frameworklauncher ./src/frameworklauncher/`

Error logs:

> 2018-06-12 08:10:31,352 [INFO] - paiLibrary.common.docker_handler : Begin to execute the command: docker build -t frameworklauncher ./src/frameworklauncher/
Sending build context to Docker daemon   1.31MB
Step 1/9 : FROM hadoop-run
 ---> 68ccfe23ba58
Step 2/9 : RUN apt-get -y update &&     apt-get -y install maven
 ---> Running in 752370bfa242
Get:1 http://security.ubuntu.com/ubuntu xenial-security InRelease [107 kB]
Hit:2 http://archive.ubuntu.com/ubuntu xenial InRelease
Get:3 http://archive.ubuntu.com/ubuntu xenial-updates InRelease [109 kB]
Get:4 http://archive.ubuntu.com/ubuntu xenial-backports InRelease [107 kB]
Get:5 http://security.ubuntu.com/ubuntu xenial-security/universe Sources [81.2 kB]
Get:6 http://archive.ubuntu.com/ubuntu xenial-updates/universe Sources [256 kB]
Get:7 http://archive.ubuntu.com/ubuntu xenial-updates/main amd64 Packages [1024 kB]
Get:8 http://security.ubuntu.com/ubuntu xenial-security/main amd64 Packages [650 kB]
Get:9 http://archive.ubuntu.com/ubuntu xenial-updates/restricted amd64 Packages [13.1 kB]
Get:10 http://archive.ubuntu.com/ubuntu xenial-updates/universe amd64 Packages [817 kB]
Get:11 http://archive.ubuntu.com/ubuntu xenial-updates/multiverse amd64 Packages [18.8 kB]
Get:12 http://archive.ubuntu.com/ubuntu xenial-backports/universe amd64 Packages [8068 B]
Get:13 http://security.ubuntu.com/ubuntu xenial-security/restricted amd64 Packages [12.7 kB]
Get:14 http://security.ubuntu.com/ubuntu xenial-security/universe amd64 Packages [448 kB]
Get:15 http://security.ubuntu.com/ubuntu xenial-security/multiverse amd64 Packages [3735 B]
Fetched 3656 kB in 2s (1644 kB/s)
Reading package lists...
[...]

PR #714, opened 2018-06-12
fix docker build apt-get related error

hao1939:
This fix if for #710 #711 .

Hi @abuccts,   @ydye ,

Please help reviewing this Dockerfile fix.
As discussed before, we'd better fix the base image.
Bad lucky, the Dockerfile of base image (which is src/hadoop-run/dockerfile) is out of control.

Also noticed that the 'src/hadoop-run/dockerfile' was generated on the fly, maybe someone know how we can update it?

Clean up the apt-get cache before run any apt-get command.

### DIFF
--- a/pai-management/src/end-to-end-test/dockerfile
+++ b/pai-management/src/end-to-end-test/dockerfile
@@ -17,7 +17,6 @@
 
 FROM hadoop-run
 
-RUN rm -rf /var/lib/apt/lists/*
 RUN apt-get -y update && \
     apt-get -y install python git jq && \
     apt-get clean && \
--- a/pai-management/src/end-to-end-test/dockerfile
+++ b/pai-management/src/end-to-end-test/dockerfile
@@ -17,6 +17,7 @@
 
 FROM hadoop-run
 
+RUN rm -rf /var/lib/apt/lists/*
 RUN apt-get -y update && \
     apt-get -y install python git jq && \
     apt-get clean && \
--- a/pai-management/src/frameworklauncher/dockerfile
+++ b/pai-management/src/frameworklauncher/dockerfile
@@ -19,7 +19,7 @@ FROM hadoop-run
 
 RUN apt-get -y update && \
     apt-get -y install maven
-RUN rm -rf /var/lib/apt/lists
+RUN rm -rf /var/lib/apt/lists/*
 
 COPY copied_file/ /usr/local/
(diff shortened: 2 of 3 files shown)


pai-management/src/end-to-end-test/dockerfile
```diff
@@ -17,6 +17,7 @@
 
 FROM hadoop-run
 
+RUN rm -rf /var/lib/apt/lists/*
 RUN apt-get -y update && \
```
sterowang: Why not to fix this problem in hadoop-run image? Is this one required by other images inherited from hadoop-run?

hao1939: I believe we should fix it in both.
But the Dockerfile of hadoop-run was generated on the fly, in this case, we'd better put some redundancy here for protecting.

hao1939: Yes, I think so, other images inherited from hadoop-run will need this.

fanyangCS:
can you start with the base dockerfile?

hao1939:
Hi @abuccts, @ydye @sterowang ,

I have fixed it in the base image 'hadoop-run'.
In the same time, I place the 'apt-get cache cleanup' after the 'apt-get' command. (Since the base image is under control, the protecting is no longer necessary.)

hao1939:
@fanyangCS,
Yes, I have appended a new commit.

pai-management/src/frameworklauncher/dockerfile
```diff
@@ -19,6 +19,7 @@ FROM hadoop-run
 
 RUN apt-get -y update && \
     apt-get -y install maven
+RUN rm -rf /var/lib/apt/lists
```
hwuu: Why do you need this line here even when hadoop-run image has already included it? And why do you delete the whole directory (rm -rf .../lists) here, while you keep the directory itself (rm -rf .../list/*****) in hadoop-run?

hao1939: Well, clean the apt-get could reduce the image size.

hao1939: @hwuu ,
Good point, I haven't noticed this difference yet.
I will update it to (rm -rf .../list/*) in both.
It's OK to remove the whole folder if you won't use apt-get again, but maybe there is difference.
Because 'hadoop-run' is base, can we can build image from it, so I believe the (rm -rf .../list/*) is more safe.

PR #715, opened 2018-06-12
[Examples] Move job examples from job tutorial to examples dir

abuccts:
Move job examples from job tutorial to examples dir.


examples/cntk/cntk.g2p.json
```diff
@@ -0,0 +1,22 @@
+{
+  "jobName": "cntk-g2p",
+  "image": "paiexample/pai.example.cntk",
+
+  // prepare cmudict corpus in CNTK format https://git.io/vbT5A and upload to hdfs
```
fanyangCS: >// prepare cmudict corpus in CNTK format https://git.io/vbT5A and upload to hdfs [](start = 2, length = 80)

can you use a readme.md to explain all these and remove comments in the json file?

abuccts: ok

PR #735, opened 2018-06-14
[Rest Server] Fix Docker container OOM issue

abuccts:
Kill Docker container when yarn container exit.


abuccts:
Should we set [`--oom-kill-disable` flag](https://docs.docker.com/engine/reference/run/#runtime-constraints-on-resources) or not?

And it's better to figure out Docker container OOM issue after changing container log directory.

rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -31,16 +34,14 @@ function exit_handler()
   exec 13>&-
   hdfs dfs -put /tmp/pai_yarncontainer_$CONTAINER_ID.log \
     {{{ hdfsUri }}}/Container/{{{ jobData.userName }}}/{{{ jobData.jobName }}}/log/
+  kill -9 $(docker inspect --format="{{.State.Pid}}" $docker_name)
```
fanyangCS: >docker inspect [](start = 12, length = 14)

actually, when a docker container runs into OOM and set oom-kill-disable, docker inspect will hang indefinitely

fanyangCS: >$docker_name [](start = 53, length = 12)

I hope it does not return 0 if the docker_name doesn't exist.

abuccts: removed

abuccts: It won't.
- If there's no error in trap handler, the exit code is the same as the last command before trap handler.
- If there's an error in trap handler, the exit code should be the same as that.

fanyangCS: I am not talking about exit code. I am wondering What is the result of "docker inspect --format="{{.State.Pid}}" $docker_name" when $docker_name does *not* exist (i.e., already exited)

---
In reply to: [195936120](https://github.com/Microsoft/pai/pull/735#discussion_r195936120) [](ancestors = 195936120)

PR #737, opened 2018-06-17
[Rest Server] Refine container log

abuccts:
Refine yarn container and docker container debug log. Enable debug log on yarn portal.

### DIFF
--- a/rest-server/src/templates/dockerContainerScript.mustache
+++ b/rest-server/src/templates/dockerContainerScript.mustache
@@ -20,7 +20,7 @@
 
 # Bootstrap script for docker container.
 
-exec 17>/tmp/pai_dockercontainer_$PAI_CONTAINER_ID.log
+exec 17>/pai/log/DockerContainerDebug.log
 BASH_XTRACEFD=17
 
 function exit_handler()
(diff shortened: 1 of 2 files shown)


rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -180,6 +179,11 @@ ENV
 echo PAI_CURRENT_CONTAINER_PORT="$(cut -f 1 -d"," <<< $PAI_CONTAINER_HOST_http_PORT_LIST)" >> $ENV_LIST
 
 
+# Prepare docker debug log
+mkdir -p "/tmp/pai-root/log/$APP_ID/$CONTAINER_ID"
+ln -s /tmp/pai-root/log/$APP_ID/$CONTAINER_ID/DockerContainerDebug.log $LOG_DIRS/DockerContainerDebug.log
```
abuccts: @yqwang-ms mounting `/var/lib/yarn/userlogs/...` as a volume in `docker run` does not work, so using symbolic link instead

rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -20,7 +20,7 @@
 
 # Bootstrap script for docker container.
 
-exec 17>/tmp/pai_dockercontainer_$PAI_CONTAINER_ID.log
+exec 17>/pai/log/DockerContainerDebug.log
```
fanyangCS: >DockerContainerDebug.log [](start = 17, length = 24)

do we have a way to control the size of the log?

PR #738, opened 2018-06-17
[Rest Server] Remove hdfs get in docker container

abuccts:
Remove hdfs get in docker container.


rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -205,5 +220,5 @@ docker run --name $docker_name \
   --label PAI_CURRENT_TASK_ROLE_NAME={{{ taskData.name }}} \
   --env-file $ENV_LIST \
   --entrypoint '/bin/bash' {{{ jobData.image }}} \
-  '-c' 'hdfs dfs -get {{{ hdfsUri }}}/Container/$HADOOP_USER_NAME/$FRAMEWORK_NAME/DockerContainerScripts/{{{ idx }}}.sh bootstrap.sh && /bin/bash bootstrap.sh' \
+  '/pai/bootstrap/docker_bootstrap.sh' \
   || exit 1
```
fanyangCS: >exit 1 [](start = 5, length = 6)

can we use (exit $?)

rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -54,9 +54,7 @@ task_role_no={{{ idx }}}
 
 printf "%s %s\n%s\n\n" "[INFO]" "ENV" "$(printenv | sort)"
 
-if [[ -n $PAI_CODE_DIR ]]; then
-  hdfs dfs -get $PAI_CODE_DIR || exit 1
-fi
+mv /pai/code/* ./
```
fanyangCS: >./ [](start = 15, length = 2)

this is a breaking change. can you preserve the previous folder?

rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -195,6 +208,8 @@ docker run --name $docker_name \
   $nvidia_devices \
   --device=/dev/fuse \
   --security-opt apparmor:unconfined \
+  --volume /tmp/pai-root/bootstrap/$APP_ID/$CONTAINER_ID:/pai/bootstrap \
+  --volume /tmp/pai-root/code/$APP_ID/$CONTAINER_ID:/pai/code \
   --volume /var/drivers/nvidia/current:/usr/local/nvidia:ro \
```
sterowang: Is it possible to mount to "WORK_DIR" directly?

rest-server/src/templates/yarnContainerScript.mustache
abuccts: sure

rest-server/src/templates/dockerContainerScript.mustache
abuccts: this isn't a breaking change, user's code is placed under WORK_DIR on the both cases.

rest-server/src/templates/yarnContainerScript.mustache
abuccts: no, user may store some files under WORK_DIR in their Docker image, mounting WORK_DIR will overwrite the whole directory.

PR #739, opened 2018-06-19
Yuan/add doc for k8s config

hao1939:
Hi @ydye @hwuu ,

Please helping review my doc update.
This PR include:
1. Explanation about kubernetes-configuration.yaml/docker-registry.
2. Update about how to use prebuild 'dev-box' image.

### DIFF
--- a/pai-management/doc/how-to-setup-dev-box.md
+++ b/pai-management/doc/how-to-setup-dev-box.md
@@ -1,11 +1,40 @@
 # Setup dev-box
 
-## Recommended: In docker Container
+## Install docker on your server
 
-- Install docker on your host.
-- Finish configuring the cluster-configuration
+```dev-box``` is a docker container you could use to maintain PAI cluster. And we provide a prebuild image hosted on Docker Hub for your convenience.
 
-#### Build the Dev-Box's docker image.
+## Use prebuild dev-box image
+
+```bash
+
+# Pull the dev-box image from Docker Hub
+sudo docker pull docker.io/openpai/dev-box
+
+# Run your dev-box
+sudo docker run -itd \
+        -e COLUMNS=$COLUMNS -e LINES=$LINES -e TERM=$TERM \
+        -v /pathConfiguration:/cluster-configuration  \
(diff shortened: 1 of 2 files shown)


pai-management/doc/how-to-setup-dev-box.md
```diff
@@ -1,11 +1,40 @@
...
 
-#### Build the Dev-Box's docker image.
+## Use prebuild dev-box image
+
+```bash
+
+# Pull the dev-box image from Docker Hub
+sudo docker pull docker.io/openpai/dev-box
+
+# Run your dev-box
```
fanyangCS: please add a comment:

1. Assume the path of ```custom-hadoop-binary-path``` in your service-configuration is ```/pathHadoop```
2. Assume the directory path of your cluster-configuration is ``/pathConfiguration`

hao1939: Thanks, I will fix it with a new commit.

pai-management/doc/how-to-setup-dev-box.md
```diff
@@ -1,11 +1,43 @@
...
+# Assume the path of custom-hadoop-binary-path in your service-configuration is /pathHadoop.
+# Assume the directory path of your cluster-configuration is /pathConfiguration.
+# By the way, both of the above could be an empty directory.
```
fanyangCS: >By the way, both of the above could be an empty directory. [](start = 2, length = 58)

if you would like to say so. you better explain what empty directory means (what the behavior? if they are empty).

fanyangCS: I suggest we do not say so. do not give users the freedom they are not sure how to leverage.

---
In reply to: [196300305](https://github.com/Microsoft/pai/pull/739#discussion_r196300305) [](ancestors = 196300305)

hao1939: Yes, you are right. My express is misleading.
Actually I want to tell user that they don't need to modify it.

PR #740, opened 2018-06-19
Install Cron in base-image

ydye:
Ref: https://serverfault.com/questions/853364/cron-and-crontab-are-missing-in-docker-image-of-ubuntu-16-04


fanyangCS:
@ydye @qinchen123  do you know why our deployment works without cron?

PR #757, opened 2018-06-23
Build hadoop only when needed

### LINKED ISSUE #720: [PAI Deployment]Image build need add check to see if need to build hadoop ai.
When build target image it also build hadoop ai image. Need to add check to see if need to build hadoop ai image.

### LINKED ISSUE #747: base-image, hadoop-run is not properly tagged, so can't push to registry
For example, here's your `services-configuration.yaml`:
```
cluster:
 ...
  docker-registry-info:
    docker-namespace: openpai
    docker-registry-domain: docker.io
    docker-tag: your-tag
```

Then the base-image should be tagged as `openpai/base-image:your-tag`, but it is `base-image:latest`.
And the same as `hadoop-run`.

hao1939:
1. Check image.yaml, when `src/hadoop-run/hadoop` is in the copy-list, build hadoop-ai, else skip hadoop-ai. Fixes #720 .
2. Move 'tag()' into the 'build()', now the base_image could be tagged. Fixes #747 .

@ydye , @haowu, please take a look.

### DIFF
--- a/pai-management/src/hadoop-run/image.yaml
+++ b/pai-management/src/hadoop-run/image.yaml
@@ -17,5 +17,11 @@
 
 prerequisite: base-image
 
+copy-list:
+  # TODO It's a bit tricky here.
+  # This two line indicates it need hadoop-ai. For now, I don't want to make much change on code, so.
+  - src: src/hadoop-run/hadoop
+    dst: src/hadoop-run/tmp
+
 template-list:
   - dockerfile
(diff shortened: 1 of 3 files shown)


pai-management/src/hadoop-run/image.yaml
```diff
@@ -17,5 +17,11 @@
 
 prerequisite: base-image
 
+copy-list:
+  # TODO It's a bit tricky here.
```
fanyangCS: >TODO [](start = 4, length = 4)

1. add a comment saying sth. like: please comment the two lines below if you would like to skip hadoop build and use the image published in docker.io
2. please explain exactly what are things todo here. what makes you uncomfortable and wish to change in the future.

hao1939:
Will figure it out in another PR, close it now.

PR #760, opened 2018-06-25
[Pylon] Refine and fix bugs

hwuu:
Refine 1: Change the urls of yarn web pages from `/yarn/<ip>/<port>/...` to `/yarn/<ip>:<port>/...`, more consistent with the format of native yarn web page urls.

Refine 2: Support job history server (port 8188), thus the logs of finished jobs can now be viewed via pylon.

Fix bug: PAI REST API returns incorrect yarn url.

### DIFF
--- a/pylon/src/nginx.conf.template
+++ b/pylon/src/nginx.conf.template
@@ -54,7 +54,7 @@ http {
     # The proxy to WebHDFS API depends on it.
     #
 
-    location ~ ^/a/([^/]+)/(\d*)(.*)$ {
+    location ~ ^/a/([^/]+):(\d*)(.*)$ {
       set $target_host   $1;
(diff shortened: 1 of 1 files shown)


pylon/src/nginx.conf.template
```diff
@@ -76,11 +76,11 @@ http {
       subs_filter_types *;
       subs_filter
         \"containerLog\":\"http://([^/]*):8042/(.*)\"
```
fanyangCS: >8042 [](start = 42, length = 4)

in the future, we should change this hardcode port setting

hwuu: Exactly. Noted.

hwuu:
https://github.com/Microsoft/pai/issues/816

PR #763, opened 2018-06-25
Update openpai examples

abuccts:
Update openpai examples, change registry namespace from `paiexample` to `openpai`.

### DIFF
--- a/examples/mpi/cntk-mpi.json
+++ b/examples/mpi/cntk-mpi.json
@@ -1,6 +1,6 @@
 {
   "jobName": "cntk-mpi",
-  "image": "paiexample/pai.example.cntk-mpi",
+  "image": "openpai/pai.example.cntk-mpi",
 
   // prepare cmudict corpus in CNTK format https://git.io/vbT5A and upload to hdfs
   "dataDir": "$PAI_DEFAULT_FS_URI/path/cntk-mpi/data",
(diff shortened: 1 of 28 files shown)


fanyangCS:
>     "image": "aiplatform/pai.run.tensorflow",

should we change this as well?

---
Refers to: examples/jupyter/README.md:86 in 6208169. [](commit_id = 62081697b3083e06cea4859ff1bcf1fd85859986, deletion_comment = False)

examples/mpi/cntk-mpi.json
```diff
@@ -1,6 +1,6 @@
 {
   "jobName": "cntk-mpi",
-  "image": "paiexample/pai.example.cntk-mpi",
+  "image": "openpai/pai.example.cntk-mpi",
 
   // prepare cmudict corpus in CNTK format https://git.io/vbT5A and upload to hdfs
   "dataDir": "$PAI_DEFAULT_FS_URI/path/cntk-mpi/data",
```
sterowang: What about we remove comments from the all example jsons so that users can use it directly?

abuccts: Not all of the examples can be submitted directly, we can not put large dataset in images.
[Here](https://github.com/Microsoft/pai/tree/master/examples#off-the-shelf-examples) lists off-the-shelf examples.

fanyangCS: I suggest we support yaml file in the future. single line comment is supported in yaml

---
In reply to: [197685206](https://github.com/Microsoft/pai/pull/763#discussion_r197685206) [](ancestors = 197685206)

abuccts: ok

PR #764, opened 2018-06-25
[exporter] job view network metrics

YanjieGao:
This PR will make monitor to get job network bandwidth
Cause: As PAI job container use host network and launched from yarn nm (which is also running inner container), let tools like docker stats or other tools can't directly get container network traffic metrics. 

Solution:
This solution will get approximate job container network traffic. 

- exporter compute the metrics periodically
- get each container current connections [cmd lsof]
- get each connection last duration transform datasize [cmd iftop] 
- accumulate each container's connection transform size and compute the approximate bandwidth

This will let exporter does not need to maintain state and as a stateless service.
note: the disappeared connections  between 2 time interval may not be accumulated let this solution to be approximate results.

### DIFF
--- a/pai-management/src/gpu-exporter/dockerfile
+++ b/pai-management/src/gpu-exporter/dockerfile
@@ -42,6 +42,7 @@ RUN cd infilter && make && cd ..
 RUN cp infilter/infilter /usr/bin
 #
 # start
-#
-
-CMD python /usr/local/job_exporter.py /datastorage/prometheus 30
+#  
+# The start cmd will be called at K8s config yaml file.
+# local container test cmd example: 
+# CMD python /usr/local/job_exporter.py /datastorage/prometheus 30
--- a/pai-management/src/gpu-exporter/dockerfile
+++ b/pai-management/src/gpu-exporter/dockerfile
@@ -40,9 +40,3 @@ RUN cp -r /usr/local/docker/* /usr/bin/
 RUN git clone https://github.com/yadutaf/infilter
 RUN cd infilter && make && cd ..
 RUN cp infilter/infilter /usr/bin
-#
-# start
-#  
-# The start cmd will be called at K8s config yaml file.
-# local container test cmd example: 
-# CMD python /usr/local/job_exporter.py /datastorage/prometheus 30
--- a/pai-management/src/gpu-exporter/dockerfile
+++ b/pai-management/src/gpu-exporter/dockerfile
@@ -29,7 +29,7 @@ ENV PATH=$PATH:$NV_DRIVER/bin
 WORKDIR /root/
 
 RUN apt-get update && \
-    apt-get -y install wget build-essential python python-pip git pciutils
+    apt-get -y install wget build-essential python python-pip git pciutils iftop lsof
 
 COPY copied_file/exporter/* /usr/local/
 
@@ -37,7 +37,9 @@ RUN wget https://download.docker.com/linux/static/stable/x86_64/docker-17.06.2-c
 RUN cp docker-17.06.2-ce.tgz /usr/local
 RUN tar xzvf /usr/local/docker-17.06.2-ce.tgz -C /usr/local/
 RUN cp -r /usr/local/docker/* /usr/bin/
-
+RUN git clone https://github.com/yadutaf/infilter
+RUN cd infilter && make && cd ..
+RUN cp infilter/infilter /usr/bin
 #
 # start
 #
--- a/prometheus/exporter/job_exporter.py
+++ b/prometheus/exporter/job_exporter.py
@@ -24,6 +24,7 @@ import docker_inspect
 import gpu_exporter
 import time
 import logging  
+import network
 from logging.handlers import RotatingFileHandler
 
 logger = logging.getLogger("gpu_expoter")  
@@ -61,13 +62,14 @@ def parse_from_env(envs):
 
     return envStr
 
-def gen_job_metrics(logDir, gpuMetrics):
+def gen_job_metrics(logDir, gpuMetrics, connectionDic):
     stats = docker_stats.stats()
     outputFile = open(logDir + "/job_exporter.prom", "w")
     for container in stats:
         inspectInfo = docker_inspect.inspect(container)
         if not inspectInfo["labels"]:
             continue
+        pid = inspectInfo["pid"]
         gpuIds, labelStr = parse_from_labels(inspectInfo["labels"])
         envStr = parse_from_env(inspectInfo["env"])
         labelStr = labelStr + envStr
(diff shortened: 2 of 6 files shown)


pai-management/src/gpu-exporter/dockerfile
```diff
@@ -29,15 +29,17 @@ ENV PATH=$PATH:$NV_DRIVER/bin
...
+    apt-get -y install wget build-essential python python-pip git pciutils iftop lsof
 
 COPY copied_file/exporter/* /usr/local/
 
 RUN wget https://download.docker.com/linux/static/stable/x86_64/docker-17.06.2-ce.tgz
 RUN cp docker-17.06.2-ce.tgz /usr/local
 RUN tar xzvf /usr/local/docker-17.06.2-ce.tgz -C /usr/local/
 RUN cp -r /usr/local/docker/* /usr/bin/
-
+RUN git clone https://github.com/yadutaf/infilter
```
fanyangCS: >t clone https://github.com/yadutaf/infilter [](start = 6, length = 43)

can we use nsenter instead?

YanjieGao: after i and yibo test. 

nsenter also need each job container install lsof (like docker exec cmd) if container does not contains lsof cmd. [The installation can called by exporter container]

Then yibo recommend infilter.

fanyangCS:
> CMD python /usr/local/job_exporter.py /datastorage/prometheus 30

can we make it a command line in docker run, instead of putting it in a dockerfile.

---
Refers to: pai-management/src/gpu-exporter/dockerfile:47 in 07b407e. [](commit_id = 07b407e0fc49db9a91e62590f96e11f2b13bb1e5, deletion_comment = False)

YanjieGao:
seems can't, need k8s to launch this image, need a dockerfile and the entrypoint script.

prometheus/exporter/job_exporter.py
```diff
@@ -82,8 +78,11 @@ def gen_job_metrics(logDir, gpuMetrics):
         containerCPUPerc = 'container_CPUPerc{{{0}}} {1}\n'.format(labelStr, stats[container]["CPUPerc"])
         containerMemUsage = 'container_MemUsage{{{0}}} {1}\n'.format(labelStr, stats[container]["MemUsage_Limit"]["usage"])
         containerMemLimit = 'container_MemLimit{{{0}}} {1}\n'.format(labelStr, stats[container]["MemUsage_Limit"]["limit"])
-        containerNetIn = 'container_NetIn{{{0}}} {1}\n'.format(labelStr, stats[container]["NetIO"]["in"])
-        containerNetOut = 'container_NetOut{{{0}}} {1}\n'.format(labelStr, stats[container]["NetIO"]["out"])
+        inSize, outSize = network.acc_per_container_network_metrics(connectionDic, pid)
+        inBandwidth = inSize / 40
+        outBandwidth = outSize / 40
```
YitongFeng: Why divided by 40 ?

YanjieGao: use the each last 40s accumulate bytes from iftop cmd / 40 will get bytes per second

prometheus/exporter/network.py
```diff
@@ -0,0 +1,163 @@
...
+# DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+import subprocess
+import json
+import sys
+import re
+import time
+import logging  
+logger = logging.getLogger("gpu_expoter")
```
xudifsd: seems all script in exporter get logger with "gpu_expoter"(I mean the same name, not misspell), is this by design?

prometheus/exporter/network.py
```diff
@@ -0,0 +1,163 @@
...
+    elif "K" in data:
+        return number * 1024
+    else: 
+        return number
+
+def parse_iftop(stats):
+    connectionDic = {}
+    data = [line.split(',') for line in stats.splitlines()]
+
+    while True and len(data) > 0:
```
xudifsd: isn't this same with `while len(data) > 0`?

pai-management/src/gpu-exporter/dockerfile
```diff
@@ -29,17 +29,20 @@ ENV PATH=$PATH:$NV_DRIVER/bin
...
+RUN cp infilter/infilter /usr/bin
 #
 # start
-#
-
-CMD python /usr/local/job_exporter.py /datastorage/prometheus 30
+#  
+# The start cmd will be called at K8s config yaml file.
+# local container test cmd example: 
+# CMD python /usr/local/job_exporter.py /datastorage/prometheus 30
```
YitongFeng: As you've changed to start it with k8s yaml, this line seems duplicate

YanjieGao: for remind developer local test docker container cmd. If test doesn't want to submit to k8s, then it could run this container at local host.

prometheus/exporter/network.py
```diff
@@ -0,0 +1,163 @@
...
+            logger.error("exporter acc_per_container_network_metrics error {}".format(e))
+        accInBytes = 0
+        accOutBytes = 0
+        return 0, 0
+
+def main(argv):
+    timeSleep = 3
+    iter = 0
+    pid = [argv]
+    while(True):
```
xudifsd: well, do not around True with parentheses

prometheus/exporter/network.py
YanjieGao: currently, the scripts running in one container, container name is gpu_exporter, so let them use same name

prometheus/exporter/network.py
YanjieGao: good suggestion, will update it

prometheus/exporter/network.py
YanjieGao: will update it

prometheus/exporter/job_exporter.py
```diff
@@ -82,8 +78,11 @@ def gen_job_metrics(logDir, gpuMetrics):
         containerCPUPerc = 'container_CPUPerc{{{0}}} {1}\n'.format(labelStr, stats[container]["CPUPerc"])
         containerMemUsage = 'container_MemUsage{{{0}}} {1}\n'.format(labelStr, stats[container]["MemUsage_Limit"]["usage"])
         containerMemLimit = 'container_MemLimit{{{0}}} {1}\n'.format(labelStr, stats[container]["MemUsage_Limit"]["limit"])
-        containerNetIn = 'container_NetIn{{{0}}} {1}\n'.format(labelStr, stats[container]["NetIO"]["in"])
-        containerNetOut = 'container_NetOut{{{0}}} {1}\n'.format(labelStr, stats[container]["NetIO"]["out"])
+        inSize, outSize = network.acc_per_container_network_metrics(connectionDic, pid)
+        inBandwidth = inSize / 40
```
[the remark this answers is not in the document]
YanjieGao: iftop only give last 2s   last 10s   last 40s choice (Not customized), suggest use the max one. will add a note to explain this. should we expose this config to user?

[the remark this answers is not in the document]
fanyangCS: yes.

---
In reply to: [200541208](https://github.com/Microsoft/pai/pull/764#discussion_r200541208) [](ancestors = 200541208)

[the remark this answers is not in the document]
YanjieGao: ok will update it

PR #769, opened 2018-06-26
Lock the image tag for release 0.6.

hao1939:
Lock the docker images tags used for release 0.6 to 'pai-0.6.y'.

@fanyangCS @hwuu please take a look.

### DIFF
--- a/pai-management/doc/how-to-write-pai-configuration.md
+++ b/pai-management/doc/how-to-write-pai-configuration.md
@@ -185,7 +185,7 @@ cluster:
     - ```docker-registry-domain```: E.g., gcr.io. If public，fill docker_registry_domain with the word "public".
     - ```docker-username```: The account of the docker registry
     - ```docker-password```: The password of the account
-    - ```docker-tag```: The image tag of the service. You could set the version here. Or just set latest here.
+    - ```docker-tag```: The image tag of the service, which should match the PAI release version, for example: `pai-0.6.y` for the release 0.6. Or use `latest` for the latest code.
     - ```secret-name```: Must be lower case, e.g., regsecret. The name of the secret in Kubernetes will be created for your cluster.
 
 Note that we provide a read-only public docker registry on DockerHub for official releases. To use this docker registry, th `docker-registry-info` section should be configured as follows, leaving `docker-username` and `docker-password` commented:
@@ -196,7 +196,7 @@ docker-registry-info:
   - docker-registry-domain: docker.io
   #- docker-username: <n/a>
   #- docker-password: <n/a>
-  - docker-tag: latest # or a specific version, i.e. 0.5.0.
+  - docker-tag: pai-0.6.y # the image tag for PAI release 0.6
   - secret-name: <anything>
 ```
 
--- a/pai-management/doc/how-to-write-pai-configuration.md
+++ b/pai-management/doc/how-to-write-pai-configuration.md
@@ -185,7 +185,7 @@ cluster:
     - ```docker-registry-domain```: E.g., gcr.io. If public，fill docker_registry_domain with the word "public".
     - ```docker-username```: The account of the docker registry
     - ```docker-password```: The password of the account
-    - ```docker-tag```: The image tag of the service, which should match the PAI release version, for example: `v0.6.1` for the release 0.6. Or use `latest` for the latest code.
+    - ```docker-tag```: The image tag of the service, which should match the PAI release version, for example: `v0.6.1` for the release 0.6.*. Or use `latest` for the latest code.
     - ```secret-name```: Must be lower case, e.g., regsecret. The name of the secret in Kubernetes will be created for your cluster.
 
 Note that we provide a read-only public docker registry on DockerHub for official releases. To use this docker registry, th `docker-registry-info` section should be configured as follows, leaving `docker-username` and `docker-password` commented:
@@ -196,7 +196,7 @@ docker-registry-info:
   - docker-registry-domain: docker.io
   #- docker-username: <n/a>
   #- docker-password: <n/a>
-  - docker-tag: v0.6.1 # the image tag for PAI release 0.6
+  - docker-tag: v0.6.1 # the image tag for PAI release 0.6.*
   - secret-name: <anything>
 ```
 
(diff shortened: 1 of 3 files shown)


pai-management/quick-start/services-configuration.yaml.template
```diff
@@ -43,7 +43,8 @@ cluster:
     #docker-username: <username>
     #docker-password: <password>
 
-    docker-tag: latest
+    # Should match the PAI version, pai-0.6.y for the PAI release 0.6.
+    docker-tag: pai-0.6.y
```
fanyangCS: >pai-0.6.y [](start = 16, length = 9)

I suggest you put a specific version in comment. and leave the default value to latest. people may not care about specific version.

hao1939: In my opinion, user should always use a latest stable release.
The latest code/image may break.

fanyangCS: that I agree. Maybe we should always release the latest stable docker image that allows user to use the latest tag,  and push a pre-release docker images somewhere else, instead of the "lastest"?

---
In reply to: [198000722](https://github.com/Microsoft/pai/pull/769#discussion_r198000722) [](ancestors = 198000722)

hao1939: It works, but maybe a bit weird.
When 'latest' is not always the latest, what does it mean?

And should we encourage user to use the edge version?
When we recommend the usage of 'latest', we are encouraging them to use the newest feature.
When we set the specific 'version' as default, we are encouraging them to use stable version.

fanyangCS: got it. make sense.

---
In reply to: [198007885](https://github.com/Microsoft/pai/pull/769#discussion_r198007885) [](ancestors = 198007885)

pai-management/doc/how-to-write-pai-configuration.md
```diff
@@ -185,7 +185,7 @@ cluster:
     - ```docker-registry-domain```: E.g., gcr.io. If public，fill docker_registry_domain with the word "public".
     - ```docker-username```: The account of the docker registry
     - ```docker-password```: The password of the account
-    - ```docker-tag```: The image tag of the service. You could set the version here. Or just set latest here.
+    - ```docker-tag```: The image tag of the service, which should match the PAI release version, for example: `pai-0.6.y` for the release 0.6. Or use `latest` for the latest code.
```
hwuu: `pai-0.6.y` is the name of a temporary 'release candidate' branch that may eventually be deleted after the release (am I correct @sterowang ?) So I suggest we organize official images in this way:
- During the version release process: Create images with tag name **'pai-0.6.y'** (release candidate name) using the latest release candidate branch. Update these images whenever necessary. The `docker-tag` field in config file of the release candidate branch is set to be this tag name as well.
- Before exiting (announcing) the version release process: Create images with tag name **'pai-0.6.0'** (release name) using the code to be released (packaged to .tar.gz file). These images will never be updated. The `docker-tag` field in config file of the release package is set to this tag name as well.
- In master branch, the `docker-tag` field in config file is always set to **'latest'**. All images with tag 'latest' are built from the latest code in the master branch.

pai-management/quick-start/services-configuration.yaml.template
hao1939: Yeah, good idea.

hwuu: I suggest we don't put any release version dependency in the code. We can use 'latest' and add some comments describing how to specify the version in the yaml file.

hao1939: I see, what about don't merge this into master branch?
Master branch should always use the 'latest', but this branch is for 0.6, it's reasonable to lock it to specific version.

pai-management/doc/how-to-write-pai-configuration.md
sterowang: My bad. I just realized RCs is not means to the intermediate states for images being verified, but some published versions for public testing. For example, when we want release a very big change version v0.6.0, maybe we can release v0.6.0-rc0, v0.6.0-rc1. Once we confidence enough for the last RC version, we publish the same version as gold version without RC surfix.

So for each image build, we can tag docker images with some build number based on the last commits to represent a specific git version. Once we publish the image, we can retag this build tag with release version. For example, we tag images this time with tag pai-0.6.0-12323123. Once we tagged code, we tag images with pai-0.6.0 and anounch it.

---
In reply to: [198025403](https://github.com/Microsoft/pai/pull/769#discussion_r198025403) [](ancestors = 198025403,198014564)

pai-management/doc/how-to-write-pai-configuration.md
```diff
@@ -185,7 +185,7 @@ cluster:
     - ```docker-registry-domain```: E.g., gcr.io. If public，fill docker_registry_domain with the word "public".
     - ```docker-username```: The account of the docker registry
     - ```docker-password```: The password of the account
-    - ```docker-tag```: The image tag of the service. You could set the version here. Or just set latest here.
+    - ```docker-tag```: The image tag of the service, which should match the PAI release version, for example: `v0.6.1` for the release 0.6. Or use `latest` for the latest code.
```
hwuu: 0.6 —> 0.6.1;
the latest code —> the latest code in master branch.

pai-management/doc/how-to-write-pai-configuration.md
```diff
@@ -196,7 +196,7 @@ docker-registry-info:
   - docker-registry-domain: docker.io
   #- docker-username: <n/a>
   #- docker-password: <n/a>
-  - docker-tag: latest # or a specific version, i.e. 0.5.0.
+  - docker-tag: v0.6.1 # the image tag for PAI release 0.6
```
hwuu: 0.6 —> 0.6.1

pai-management/doc/how-to-write-pai-configuration.md
hao1939: Yes, you noticed the difference.
Here's something I want to express: when user choose release 0.6 (so emit the third version), I want to recommend the latest tag(that's why it include the third).

PR #777, opened 2018-06-26
[Example]Add caffe example

Asakuri:


examples/caffe/Dockerfile.example.caffe
```diff
@@ -0,0 +1,39 @@
...
+        liblmdb-dev \
+        libopencv-dev \
+        libprotobuf-dev \
+        libsnappy-dev \
+        protobuf-compiler \
+        python-dev \
+        python-numpy \
+        python-pip \
+        python-setuptools \
+        python-scipy
```
abuccts: python dependencies will be installed using `pip` in line 31?

PR #782, opened 2018-06-27
[Example]Add caffe2 resnet50 example

Asakuri:


examples/caffe2/Dockerfile.example.caffe2
```diff
@@ -0,0 +1,45 @@
...
+      libgtest-dev \
+      libiomp-dev \
+      libleveldb-dev \
+      liblmdb-dev \
+      libopencv-dev \
+      libopenmpi-dev \
+      libsnappy-dev \
+      libprotobuf-dev \
+      libgflags-dev \
+      openmpi-bin \
```
abuccts: Is it possible to use openmpi base image?

Asakuri: Is it necessary? I'm not sure whether the environment is consistent.

PR #785, opened 2018-06-28
Only build hadoop-ai, when required.

ydye:


pai-management/paiLibrary/paiBuild/build_center.py
```diff
@@ -46,6 +46,22 @@ def __init__(self, cluster_object_model, build_target = None, os_type="ubuntu16.
 
         self.os_type = os_type
 
+        self.image_depend_on_hadoop_ai = self.resolve_hadoop_dependency()
+
+
+
+    def resolve_hadoop_dependency(self):
+
+        if "hadoop-run" in self.image_list:
```
hwuu: If another new component also requires hadoop, should the code here be changed accordingly? I prefer putting these dependency info to a configuration file, or a `modules.dep` file in the code folder of each component, instead of hard-code the logic here.

ydye: please build from hadoop-run

PR #791, opened 2018-06-28
[Rest Server] Add --shm-size in docker run

abuccts:
Add `--shm-size` in docker run. Set shared memory size the same as cpu memory.


rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -208,6 +208,7 @@ docker run --name $docker_name \
   --network=host \
   --cpus={{{ taskData.cpuNumber }}} \
   --memory={{{ taskData.memoryMB }}}m \
+  --shm-size={{{ taskData.memoryMB }}}m \
   $nvidia_devices \
```
sterowang: What's specific problem need to solve to add this switch?

abuccts: pytorch needs to use shm when reading data

abuccts:
It's ok to expose this field in json file, but we'd better check the value, e.g. `shmsize <= memory * 0.5`?

fanyangCS:
> # PAI 深度学习任务指南

this file will be removed. pls exclude it in this PR.

---
Refers to: job-tutorial/README-zh.md:19 in f9ea61b. [](commit_id = f9ea61b1678a76fadf2562ec173c9cfb2963719c, deletion_comment = False)

job-tutorial/README.md
```diff
@@ -112,6 +113,7 @@ Below please find the detailed explanation for each of the parameters in the con
 | `taskRole.taskNumber`          | Integer, required          | Number of tasks for the task role, no less than 1 |
 | `taskRole.cpuNumber`           | Integer, required          | CPU number for one task in the task role, no less than 1 |
 | `taskRole.memoryMB`            | Integer, required          | Memory for one task in the task role, no less than 100 |
+| `taskRole.shmMB`               | Integer, optional          | Shared memory for one task in the task role, no more than memory size |
```
fanyangCS: >Shared [](start = 64, length = 6)

please state the default value here.

abuccts: updated

PR #798, opened 2018-06-29
[Launcher]: Refine GangAllocation

yqwang-ms:
1. Decouple with generateContainerIpList and Deprecate generateContainerIpList
2. Decouple with Task No Retry
3. Decouple with minFailedTaskCount=1
4. Decouple with yarn.resourcemanager.rm.container-allocation.expiry-interval-ms
5. Deprecate samePortAllocation since it is coupled with GangAllocation and useless
6. Deprecate client side validation since validation depends on server side runtime env


frameworklauncher/README.md
```diff
@@ -26,7 +26,7 @@ FrameworkLauncher (or Launcher for short) is built to enable running Large-Scale
 * **High Availability**
   * All Launcher and Hadoop components are Recoverable and Work Preserving. So, User Services is by designed No Down Time, i.e. always uninterrupted when our components shutdown, crash, upgrade, or even any kinds of outage for a long time.
```
qinchen123: >Recoverable [](start = 43, length = 11)

recoverable ? shouldn't upper cases ?

yqwang-ms: This is by design as aligned with Work Preserving

---
In reply to: [199691277](https://github.com/Microsoft/pai/pull/798#discussion_r199691277) [](ancestors = 199691277)

frameworklauncher/conf/frameworklauncher.yml
```diff
@@ -25,6 +25,7 @@ amRmResyncFrequency: 6
 amRequestPullIntervalSec: 30
 amStatusPushIntervalSec: 30
 amFrameworkInfoPublishIntervalSec: 30
+amGangAllocationTimeoutSec: 500
```
qinchen123: >500 [](start = 28, length = 3)

500， different 1100 in pai-management, which one is final number ?

yqwang-ms: 1100 is the final number on k8s

---
In reply to: [199710525](https://github.com/Microsoft/pai/pull/798#discussion_r199710525) [](ancestors = 199710525)

PR #811, opened 2018-07-04
Add utf-8 encoding for stdout in pai-management/k8sPaiLibrary/maintainlib/common.py

Beyyes:
In ubuntu environment which uses Chinese language, ```python paictl.py cluster k8s-bootup -p /path/to/cluster-configuration/dir``` will cause the encoding error as the picture shown below. 
So I remove the unnecessary output code.
Or maybe it could be resolved by using utf-8 encoding, but i think it's not a universal solution.

image

### DIFF
--- a/pai-management/k8sPaiLibrary/maintainlib/common.py
+++ b/pai-management/k8sPaiLibrary/maintainlib/common.py
@@ -160,8 +160,6 @@ def sftp_paramiko(src, dst, filename, host_config):
     stdin, stdout, stderr = ssh.exec_command("sudo mkdir -p {0}".format(dst), get_pty=True)
     stdin.write(password + '\n')
     stdin.flush()
-    for response_msg in stdout:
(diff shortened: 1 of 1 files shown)


hwuu:
Using correct encoding for each text being processed is the solution indeed. Please propose a PR of a complete solution (for example, to radically solve the encoding issue) or create an issue so that we can officially make a schedule :)

ydye:
Yes, agree with @hwuu 
@Beyyes, I think  executing command in a silent way is unfriendly for issue debugging. And we'd better find a solution based on encoding.

Beyyes:
@hwuu 
I have changed the ``` print(response_msg.strip('\n'))``` to ```print(response_msg.encode('utf-8').strip('\n'))```, is this ok?

hwuu:
@beyyes: Sure. Thanks for making this change!

PR #815, opened 2018-07-05
While generating config, mark pai-master as worker too

hao1939:
Mark the first ip as pai-work too.
@ydye , @hwuu , please take a look.


hwuu:
Putting master and worker roles into one node will introduce some potential resource conflicts. So it should be discouraged.

hao1939:
It's not suitable for all cases. Close the PR.

PR #832, opened 2018-07-09
REST Server: refine user errors.

Gerhut:
#807

### DIFF
--- a/rest-server/src/config/user.js
+++ b/rest-server/src/config/user.js
@@ -26,7 +26,8 @@ const userPutInputSchema = Joi.object().keys({
   password: Joi.string()
     .min(6)
     .required(),
-  admin: Joi.boolean(),
+  admin: Joi.boolean()
+    .required(),
   modify: Joi.boolean()
     .required(),
 }).required();
--- a/rest-server/src/config/user.js
+++ b/rest-server/src/config/user.js
@@ -27,7 +27,8 @@ const userPutInputSchema = Joi.object().keys({
     .min(6)
     .required(),
   admin: Joi.boolean(),
-  modify: Joi.boolean(),
+  modify: Joi.boolean()
+    .required(),
 }).required();
 
 // define the input schema for the 'remove user' api
--- a/rest-server/src/config/user.js
+++ b/rest-server/src/config/user.js
@@ -26,8 +26,7 @@ const userPutInputSchema = Joi.object().keys({
   password: Joi.string()
     .min(6)
     .required(),
-  admin: Joi.boolean()
-    .required(),
+  admin: Joi.boolean(),
   modify: Joi.boolean()
     .required(),
 }).required();
--- a/rest-server/src/controllers/user.js
+++ b/rest-server/src/controllers/user.js
@@ -18,11 +18,12 @@
 // module dependencies
 const userModel = require('../models/user');
 const logger = require('../config/logger');
+const createError = require('../util/error');
 
 /**
  * Create / update a user.
  */
-const update = (req, res) => {
+const update = (req, res, next) => {
   const username = req.body.username;
   const password = req.body.password;
   const admin = req.body.admin;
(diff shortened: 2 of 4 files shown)


rest-server/src/config/user.js
```diff
@@ -27,7 +27,8 @@ const userPutInputSchema = Joi.object().keys({
     .min(6)
     .required(),
   admin: Joi.boolean(),
-  modify: Joi.boolean(),
+  modify: Joi.boolean()
+    .required(),
```
hwuu: Why is `modify` required while `admin` is not?

rest-server/src/controllers/user.js
```diff
@@ -18,63 +18,49 @@
...
           message: 'update successfully',
         });
       }
     });
   } else {
-    return res.status(401).json({
-      error: 'NotAuthorized',
-      message: 'not authorized',
-    });
+    next(createError('Forbidden', 'ERR_FORBIDDEN_USER', `Non-admin is not allow to do this operation.`));
```
hwuu: allow --> allowed.

hwuu: allow --> allowed

rest-server/src/controllers/user.js
```diff
@@ -18,63 +18,49 @@
...
-          error: 'RemoveFailed',
-          message: 'remove failed',
-        });
-      } else {
-        return res.status(200).json({
-          message: 'remove successfully',
-        });
+    userModel.remove(username, (err) => {
+      if (err) {
+        return next(createError.unknown(err));
```
hwuu: Is it 'internal server error'?

rest-server/src/controllers/user.js
```diff
@@ -131,31 +117,16 @@ const updateUserVc = (req, res) => {
...
       }
+      return res.status(200).json(userList);
     });
   } else {
-    return res.status(401).json({
-      error: 'NotAuthorized',
-      message: 'not authorized',
-    });
+    next(createError('Forbidden', 'ERR_FORBIDDEN_USER', `Non-admin is not allow to do this operation.`));
```
hwuu: allow --> allowed

rest-server/src/models/user.js
```diff
@@ -224,24 +195,23 @@ const checkUserVc = (username, virtualCluster, callback) => {
...
-          let userName = key.replace(etcdConfig.storagePath() + '/', '');
-          userInfoList.push({
-            username: userName,
-            admin: res.get(etcdConfig.userAdminPath(userName)),
-            virtualCluster: res.has(etcdConfig.userVirtualClusterPath(userName)) ? res.get(etcdConfig.userVirtualClusterPath(userName)) : 'default',
-          });
-        }
-      });
-      next(null, userInfoList);
+const getUserList = (callback) => {
```
hwuu: What is the purpose to change `next` to `callback`?

rest-server/src/config/user.js
Gerhut: This change is just a simple substitution of https://github.com/Microsoft/pai/pull/832/files#diff-3dd2c15619e4f4cdb338de066fddb092L45

rest-server/src/models/user.js
Gerhut: In my understand of the coding style, when using `callback` as callback name, error will be the first argument, while using `next` as callback name, error will be the second argument (https://github.com/Microsoft/pai/blob/2e55ab6b9a48c4ac0767b82e7f904f6498a895c0/rest-server/src/models/vc.js#L76)

This `next` is the only exception and confuse me a while.

rest-server/src/config/user.js
Gerhut: Found admin validation in model code, will move it to param schema

rest-server/src/models/user.js
Gerhut: Node.js [error document](https://nodejs.org/api/errors.html#errors_error_first_callbacks) and [util document](https://nodejs.org/api/util.html#util_util_callbackify_original) encourage "error-first callbacks".

rest-server/src/controllers/user.js
Gerhut: `createError.unknown` will not transform known error objects like

https://github.com/Microsoft/pai/blob/qixcheng/rest-server-refine-error/user/rest-server/src/models/user.js#L90

or 

https://github.com/Microsoft/pai/blob/qixcheng/rest-server-refine-error/user/rest-server/src/models/user.js#L97

It will only transform unknown upstream errors (usually from HDFS/etcd utilities) into general 500 errors.

At present, most of `UnknownError`s are unhandled `UpstreamServerError`s. Like other non-`UpstreamServerError` `UnknownError`s, it can only be resolved manually by developers. So when any `UnknownError` happened, report `error.message` to developers may be the only solution, no matter it is from upstream server or rest-server itself. So I think different `error.message` in `UnknownError` is helpful enough.

Gerhut:
@abuccts admin parameter back to optional

PR #839, opened 2018-07-09
Refine job errors

Gerhut:
#807

### DIFF
--- a/rest-server/src/controllers/job.js
+++ b/rest-server/src/controllers/job.js
@@ -17,6 +17,7 @@
 
 // module dependencies
 const Job = require('../models/job');
+const createError = require('../util/error');
 const logger = require('../config/logger');
 
 /**
@@ -25,28 +26,16 @@ const logger = require('../config/logger');
 const load = (req, res, next, jobName) => {
   new Job(jobName, (job, error) => {
     if (error) {
-      if (error.message === 'JobNotFound') {
+      if (error.code === 'ERR_NO_JOB') {
         if (req.method !== 'PUT') {
-          logger.warn('load job %s error, could not find job', jobName);
-          return res.status(404).json({
-            error: 'JobNotFound',
-            message: `could not find job ${jobName}`,
-          });
+          return next(error);
         }
       } else {
-        logger.warn('internal server error');
-        return res.status(500).json({
-          error: 'InternalServerError',
-          message: 'internal server error',
-        });
+        return next(createError.unknown(error));
       }
     } else {
       if (job.jobStatus.state !== 'JOB_NOT_FOUND' && req.method === 'PUT' && req.path === `/${jobName}`) {
-        logger.warn('duplicate job %s', jobName);
-        return res.status(400).json({
-          error: 'DuplicateJobSubmission',
-          message: `job already exists: '${jobName}'`,
-        });
+        return next(createError('Conflict', 'ERR_CONFLICT_JOB', `Job name ${jobName} already exists`));
       }
     }
     req.job = job;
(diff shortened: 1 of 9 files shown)


rest-server/src/controllers/job.js
```diff
@@ -25,28 +26,16 @@ const logger = require('../config/logger');
...
     if (error) {
-      if (error.message === 'JobNotFound') {
+      if (error.code === 'ERR_NO_JOB') {
         if (req.method !== 'PUT') {
-          logger.warn('load job %s error, could not find job', jobName);
-          return res.status(404).json({
-            error: 'JobNotFound',
-            message: `could not find job ${jobName}`,
-          });
+          return next(error);
```
hwuu: error --> createError(...) ?

rest-server/src/controllers/job.js
```diff
@@ -58,38 +47,27 @@ const init = (req, res, next) => {
...
   Job.prototype.getJobList(req._query, (jobList, err) => {
     if (err) {
-      logger.warn('list jobs error\n%s', err.stack);
-      return res.status(500).json({
-        error: 'GetJobListError',
-        message: 'get job list error',
-      });
+      return next(createError.unknown(err));
     } else if (jobList === undefined) {
+      // Unreachable
```
hwuu: If these lines are unreachable, why do we still preserve them?

rest-server/src/controllers/job.js
```diff
@@ -111,35 +89,14 @@ const get = (req, res) => {
...
-          error: 'VirtualClusterNotFoundInDatabase',
-          message: `job update error: search virtual cluster from db failed`,
-        });
-      } else {
-        return res.status(500).json({
-          error: 'JobUpdateError',
-          message: err.message,
-        });
-      }
+      return next(createError.unknown(err));
```
hwuu: I think your method covers too many details.

hwuu: I suggest we split errors into two categories:

- Internal errors, which are raised by models and handled by controllers;
- Exposed errors, which will be returned to API callers.

Errors in the first category contain many details, for example HDFS error, launcher error, VC error, etc., which are not necessary to be exposed to API callers. After the processed by controllers, these internal errors will eventually be reorganized and translated into exposed errors, which contain meaningful information relevant to external developers.

rest-server/src/controllers/job.js
Gerhut: Since this error has `code` property in this branch, assume that it has been already transformed into `HttpError` in `utils/error.js`

rest-server/src/models/job.js
```diff
@@ -119,9 +123,9 @@ class Job {
           if (requestRes.status === 200) {
             next(this.generateJobDetail(requestResJson), null);
           } else if (requestRes.status === 404) {
-            next(null, new Error('JobNotFound'));
+            next(null, createError('Not Found', 'ERR_NO_JOB', `Job ${name} not found.`));
```
abuccts: move error to the first parameter?

rest-server/src/controllers/job.js
abuccts: I think those errors have been handled in models instead of here.

rest-server/src/controllers/job.js
abuccts: @Gerhut only refined error handling in this PR.

rest-server/src/controllers/job.js
Gerhut: - `VirtualClusterNotFound` raised in https://github.com/Microsoft/pai/blob/qixcheng/rest-server-refine-error/job/rest-server/src/models/user.js#L204
- `NoRightAccessVirtualCluster` raised in https://github.com/Microsoft/pai/blob/qixcheng/rest-server-refine-error/job/rest-server/src/models/user.js#L216
- `VirtualClusterNotFoundInDatabase` covers too many details in https://github.com/Microsoft/pai/blob/qixcheng/rest-server-refine-error/job/rest-server/src/models/user.js#L208 so I throwed a `UnknownError` with HDFS Error message.

PR #848, opened 2018-07-11
avoid hanging nvidia-smi block job metrics generation

xudifsd:
We observe in some cases, nvidia-smi call will hang indefinitely, this will block job metrics generation.

So we use thread to call nvidia-smi and if there is already one thread doing that, ignore gpu metrics.

### DIFF
--- a/prometheus/exporter/job_exporter.py
+++ b/prometheus/exporter/job_exporter.py
@@ -16,6 +16,7 @@
 # DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
+import functools
 import subprocess
 import threading
 import json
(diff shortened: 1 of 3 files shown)


prometheus/exporter/job_exporter.py
```diff
@@ -96,22 +101,55 @@ def gen_job_metrics(logDir, gpuMetrics):
...
+
+    semaphore = threading.Semaphore(1)
+
+    while True:
         try:
             logger.info("job exporter running {0} iteration".format(str(iter)))
             iter += 1
             # collect GPU metrics
-            gpuMetrics = gpu_exporter.gen_gpu_metrics_from_smi(logDir)
+            gpuMetrics = try_get_gpu_metrics(semaphore, logDir)
```
YanjieGao: 1 consideration: Suggest judge the nvidia-smi launch time and drop  nvidia-smi long duration ago old metrics? if it is hanging long time then it return results after duration d [Could consider or set a duration threshold ]. The metric is the older gpu metrics, maybe mistake operator?

prometheus/exporter/job_exporter.py
```diff
@@ -96,22 +101,55 @@ def gen_job_metrics(logDir, gpuMetrics):
...
+
+    return None
+
 def main(argv):
     logDir = argv[0]
     timeSleep = int(argv[1])
     iter = 0
-    while(True):
+
+    semaphore = threading.Semaphore(1)
```
YanjieGao: If exception occurs, does this could release? Do we need a try catch and release this at catch block?

xudifsd: semaphore is decreased on success acquire call. Its release is handled in wrapper function.

prometheus/exporter/job_exporter.py
```diff
@@ -96,22 +101,55 @@ def gen_job_metrics(logDir, gpuMetrics):
...
+        queue = Queue(1)
+        t = threading.Thread(target=wrapper, args=(queue,), name="gpu-metrices-getter")
+        t.start()
+        try:
+            result = queue.get(block=True, timeout=3)
+            return result
+        except Empty:
+            return None
+    else:
+        logger.warn("another thread is running")
```
YanjieGao: how about give a detailed log. such as get gpu metrics thread is running etc.

xudifsd:
@YanjieGao changed according to comments, also add test cases

PR #857, opened 2018-07-12
fix error in parsing docker stats

xudifsd:
`tb` and `tib` is different, means 10^12 and 2^40 respectively.


xudifsd:
@YanjieGao fixed broken unit test.

YanjieGao:
if tested and passed, could merge.

PR #873, opened 2018-07-13
fix hard coding api server into watchdog image

### LINKED ISSUE #866: Cluster specific configuration built into image. This image could only be used in one specific cluster.
https://github.com/Microsoft/pai/blob/7c7065d878e64a4a9df2bc92ac24e0b360af7ae1/pai-management/src/watchdog/dockerfile.template#L39

xudifsd:
fix #866


examples/kafka/kafka.json
```diff
@@ -1,26 +1,26 @@
...
-          "beginAt": 0,
-          "portNumber": 1
-        }
-      ]
-    }
-  ]
-}
+{
+  "jobName": "kafka",
+  "image": "openpai/pai.example.kafka",
```
YanjieGao: Why affect this file?

xudifsd: this file's line ending is \r\n, so I use dos2unix to converted this file.

YanjieGao: ok, this is better not include at this PR, may confuse others.

xudifsd:
@YanjieGao done

PR #879, opened 2018-07-13
[Rest Server]  Expose task role completion policy

abuccts:
Obsolete `killOnAnyComplete` and use task role completion policy.


job-tutorial/README.md
```diff
@@ -88,41 +88,43 @@ A json file describe detailed configuration required for a job submission. The d
...
+| `taskRole.shmMB`                 | Integer, optional          | Shared memory for one task in the task role, no more than memory size. The default value is 64MB |
+| `taskRole.gpuNumber`             | Integer, required          | GPU number for one task in the task role, no less than 0 |
+| `taskRole.portList`              | List, optional             | List of `portType` to use                |
+| `taskRole.portType.label`        | String in `^[A-Za-z0-9._~]+$` format, required | Label name for the port type |
+| `taskRole.portType.beginAt`      | Integer, required          | The port to begin with in the port type, 0 for random selection |
+| `taskRole.portType.portNumber`   | Integer, required          | Number of ports for the specific type    |
+| `taskRole.command`               | String, required           | Executable command for tasks in the task role, can not be empty |
+| `taskRole.minFailedTaskCount`    | Integer, optional          | Number of failed tasks to kill the entire job, no less than 1   |
+| `taskRole.minSucceededTaskCount` | Integer, optional          | Number of succeeded tasks to kill the entire job, no less than 1 |
```
yqwang-ms: Better to link to launcher user manual for details, such as 
[https://github.com/Microsoft/pai/blob/master/frameworklauncher/doc/USERMANUAL.md#ApplicationCompletionPolicy](https://github.com/Microsoft/pai/blob/master/frameworklauncher/doc/USERMANUAL.md#ApplicationCompletionPolicy)

And at the bottom link to the entire doc:
[https://github.com/Microsoft/pai/blob/master/frameworklauncher/doc/USERMANUAL.md](https://github.com/Microsoft/pai/blob/master/frameworklauncher/doc/USERMANUAL.md)

rest-server/src/config/job.js
```diff
@@ -74,15 +74,20 @@ const jobConfigSchema = Joi.object().keys({
...
         command: Joi.string()
           .required(),
+        minFailedTaskCount: Joi.number()
+          .integer()
+          .min(1)
+          .optional(),
+        minSucceededTaskCount: Joi.number()
+          .integer()
+          .min(1)
+          .optional(),
```
yqwang-ms: To distinguish field NOT EXIST and field is set to NULL (To allow use can set minFailedTaskCount = null).

Can we make:
  minFailedTaskCount.default(1)
  minSucceededTaskCount.default(null)

And 
        'applicationCompletionPolicy': {
          'minFailedTaskCount': data.taskRoles[i].minFailedTaskCount,
          'minSucceededTaskCount': data.taskRoles[i].minSucceededTaskCount,
        },

yqwang-ms:
Seems there are other places in PAI using "killAllOnCompletedTaskNumber", could you please search and clean up them?

rest-server/src/config/job.js
abuccts: Default value means if user doesn't set the field then the field will be set to default value, which is the same as current code.
We can make `minFailedTaskCount.default(1)` but can't make `minSucceededTaskCount.default(null)` because `minSucceededTaskCount` is limited to integer in schema.

yqwang-ms:
How about [https://github.com/Microsoft/pai/blob/a8fe51ed5d4b1203998e6dc5d094b4d1a64eb9a2/webportal/src/app/job/job-submit/job-submit.schema.js#L203](https://github.com/Microsoft/pai/blob/a8fe51ed5d4b1203998e6dc5d094b4d1a64eb9a2/webportal/src/app/job/job-submit/job-submit.schema.js#L203)

---
In reply to: [405002761](https://github.com/Microsoft/pai/pull/879#issuecomment-405002761) [](ancestors = 405002761)

rest-server/src/config/job.js
yqwang-ms: If user set the minFailedTaskCount = null, from below line:

'minFailedTaskCount': data.taskRoles[i].minFailedTaskCount || 1,

What will you send to launcher? null or 1?
I expect you also send null to launcher.




---
In reply to: [202507760](https://github.com/Microsoft/pai/pull/879#discussion_r202507760) [](ancestors = 202507760)

yqwang-ms:
@Gerhut If without #839, this PR will still work (backward compatiable), it is fine. If not, we would better make the update here.

yqwang-ms:
@Gerhut Thanks for confirm

rest-server/src/config/job.js
abuccts: Because the minFailedTaskCount is limited to integer, user can not set it to null. I suggest we can use 0 to replace null in config file.
And what's difference between minFailedTaskCount=null and minFailedTaskCount=taskNumber?

yqwang-ms: Do we have any way to let integer support null?
Such as value is null, undefined or NaN.


[https://stackoverflow.com/questions/16474796/set-javascript-variable-null-or-leave-undefined](https://stackoverflow.com/questions/16474796/set-javascript-variable-null-or-leave-undefined)
"undefined" means this value does not exist. typeof returns "undefined"

"null" means this value exists with an empty value. When you use typeof to test for "null", you will see that it's an object. Other case when you serialize "null" value to backend server like asp.net mvc, the server will receive "null", but when you serialize "undefined", the server is unlikely to receive a value.

---
In reply to: [203658064](https://github.com/Microsoft/pai/pull/879#discussion_r203658064) [](ancestors = 203658064,203646504)

abuccts: Can we use 0 to support null? Because minFailedTaskCount is greater or equal than 1 at present.

Gerhut:
@abuccts Since #807 is merge into master, if you have trouble with resolving conflict, feel free to contact me.

rest-server/src/config/job.js
yqwang-ms: Please use -1 to substitute null if js really cannot accept null as int value.

---
In reply to: [203690975](https://github.com/Microsoft/pai/pull/879#discussion_r203690975) [](ancestors = 203690975)

job-tutorial/README.md
abuccts: updated

rest-server/src/config/job.js
abuccts: updated, allow `null` in `minFailedTaskCount` and `minSucceededTaskCount`

yqwang-ms:
Pls check:
[https://github.com/Microsoft/pai/blob/a8fe51ed5d4b1203998e6dc5d094b4d1a64eb9a2/webportal/src/app/job/job-submit/job-submit.schema.js#L203](https://github.com/Microsoft/pai/blob/a8fe51ed5d4b1203998e6dc5d094b4d1a64eb9a2/webportal/src/app/job/job-submit/job-submit.schema.js#L203)

rest-server/src/config/job.js
```diff
@@ -74,15 +74,22 @@ const jobConfigSchema = Joi.object().keys({
...
           .required(),
+        minFailedTaskCount: Joi.number()
+          .integer()
+          .min(1)
+          .allow(null)
+          .default(1),
+        minSucceededTaskCount: Joi.number()
+          .integer()
+          .min(1)
+          .allow(null)
```
yqwang-ms: So, js allow null to be int value?

rest-server/src/config/job.js
```diff
@@ -74,15 +74,22 @@ const jobConfigSchema = Joi.object().keys({
           .default([]),
         command: Joi.string()
           .required(),
+        minFailedTaskCount: Joi.number()
+          .integer()
+          .min(1)
+          .allow(null)
+          .default(1),
```
yqwang-ms: Can you test below things:

If user set the minFailedTaskCount = null, from below line:

'minFailedTaskCount': data.taskRoles[i].minFailedTaskCount,

What will you send to launcher? 
I expect you also send null to launcher.

rest-server/src/config/job.js
```diff
@@ -74,15 +74,22 @@ const jobConfigSchema = Joi.object().keys({
...
+        minSucceededTaskCount: Joi.number()
+          .integer()
+          .min(1)
+          .allow(null)
+          .default(null),
       }))
```
yqwang-ms: If we have default, do we still need optional?

fanyangCS:
Can you support backward compatibility?

PR #880, opened 2018-07-13
Chenqi/marketplace

MaggieQi:
fix job submit bugs


abuccts:
Please use `npm run lint` to check code style, for example, 2 spaces indent instead of 4.

Issue #915, opened 2018-07-18
k8s kill pods/containers due to protection machanism


DongZhaoYu:
K8s can kill pods or containers on a machine to protect its health.
This could be triggered by the shortage of disk space.
Need to configure the hdfs to avoid consume too much disk to trigger k8s's protection mechanism.

DongZhaoYu:
It seems the disk pressure is not  caused by the hdfs. The pulled image seems consumed much disk storage. But need further investigation.

fanyangCS:
Is there a way to limit docker cache?

DongZhaoYu:
Is k8s aware of the docker container we started from Yarn container? If not the k8s GC may cannot clean the cache of the docker container.

fanyangCS:
@DongZhaoYu  No, k8s is not aware of those docker containers.

fanyangCS:
https://k8smeetup.github.io/docs/tasks/configure-pod-container/quality-service-pod/.
For k8s Pods, we can assign a proper QoS level to reduce the chance of critical pods being killed.

fanyangCS:
https://github.com/Microsoft/pai/issues/827

hao1939:
Here's stackoverflow session about limiting storage usage of a container: https://stackoverflow.com/a/33017556/500268.

Recently, docker support limit the storage on some kind of storage drivers.

https://docs.docker.com/engine/reference/commandline/run/#set-storage-driver-options-per-container
```
Set storage driver options per container
$ docker run -it --storage-opt size=120G fedora /bin/bash
This (size) will allow to set the container rootfs size to 120G at creation time. This option is only available for the devicemapper, btrfs, overlay2, windowsfilter and zfs graph drivers. For the devicemapper, btrfs, windowsfilter and zfs graph drivers, user cannot pass a size less than the Default BaseFS Size. For the overlay2 storage driver, the size option is only available if the backing fs is xfs and mounted with the pquota mount option. Under these conditions, user can pass any size less than the backing fs size.
```

hao1939:
#1777 is for the disk restriction.

fanyangCS:
fix in 0.10.y by: 1 disable pod eviction (both kubelet and cluster level). 2 deploy a watchdog to kill job containers when it consumes too much resource. 3. enforce max memory usage for each PAI service pod.

PR #929, opened 2018-07-19
Update kubelet version to fix issue of Resource Handling Configuration

ydye:
#915

Update:
After investigating, I found the build version 1.9.4 of kubelet docker image has some issue which may lead to image-gc-threshold failure.


pai-management/k8sPaiLibrary/template/kubelet.sh.template
```diff
@@ -44,7 +44,6 @@ docker run \
   --pod-manifest-path=/etc/kubernetes/manifests \
   --allow-privileged=true \
   --logtostderr=true \
-  --image-gc-high-threshold=98 \
-  --image-gc-low-threshold=95 \
   --pod-infra-container-image {{ clusterconfig['dockerregistry'] }}/pause-amd64:3.0 \
+  --eviction-hard=memory.available<5%,nodefs.available<5%,imagefs.available<5%,nodefs.inodesFree<5%,imagefs.inodesFree<5% \
```
fanyangCS: do you know why the previous config doesn't work? can we add a comment to state which version of kubelet support the flag?

ydye: Flag --image-gc-high-threshold has been deprecated, This parameter should be set via the config file Flag --image-gc-low-threshold has been deprecated, This parameter should be set via the config file

ydye: Some kubelet Garbage Collection features in this doc will be replaced by kubelet eviction in the future.

REF: https://kubernetes.io/docs/concepts/cluster-administration/kubelet-garbage-collection/

YitongFeng:
May I ask what's the issue? @ydye

ydye:
@YitongFeng  Didn't find it on the changelog of 1.9

PR #938, opened 2018-07-20
refactor watchdog and add test cases

### LINKED ISSUE #883: watchdog will get error in each iteration
```
2018-07-16 05:30:21,658 - watchdog - ERROR - watchdog error <type 'exceptions.TypeError'>
2018-07-16 05:30:21,658 - watchdog - ERROR - watchdog error 'bool' object is not iterable
2018-07-16 05:30:21,658 - watchdog - ERROR - watchdog error <traceback object at 0x7fd0a5e8a098>
2018-07-16 05:30:21,658 - watchdog - ERROR - node_current_docker_error{instance="IP"} 1
```

this exception may raised in `check_docker_daemon_status` fn in watchdog.py, should use logger.exception(ex) to locate root cause.

xudifsd:
current watchdog build still depends `cluster-configuration/cluster-configuration.yaml` be valid and configured right. Will remove this dependence later.

fix #883


YanjieGao:
please keep the test json data files simpler and let it has small size. 
remove the data is not used by the test case.

PR #942, opened 2018-07-20
simplify image-build related code

wangcan0329:


pai-management/paiLibrary/paiBuild/build_center.py
```diff
@@ -112,7 +123,7 @@ def build(self, image_name):
             self.docker_cli
         )
 
-        base_image = image_build_worker.get_dependency()
+        base_image = self.get_base_image(image_name)
```
ydye: Only a suggestion. Maybe ```get_base_image``` should be moved into image_build.py

PR #945, opened 2018-07-20
[Pai-build] refine end-to-end-test, drivers, hadoop-run and rest-server images.yaml and dockerfile

wangcan0329:


pai-management/doc/add-service.md
```diff
@@ -88,19 +74,7 @@ prerequisite: hadoop-run
...
-    - Note: If your base image could be pulled from a docker registry, just remove this. If your base image is pai's customized image or yourself's customized image added into pai, you must set this field.
-
-
-- ```template-list``` part:
-    - You should list your template file in this list, then paictl will generate the template according the list for you. The configuration data comes from cluster configuration, and the template engine is [jinja2](http://jinja.pocoo.org/). After cluster-object-model module is developed, a more detailed guide will be write.
-    - A ```filename``` in ```template-list``` is corresponding to  the template file named ```filename.template``` in your image's directory. And paictl will generate the template with the file named ```filename```.
-    - This part is not recommended to use. Any cluster-specific configuration shouldn't be set into docker image.
-    - An use case in pai, [hadoop-run's dockerfile](../src/hadoop-run/dockerfile.template)
-
+Configuration only consists copy-list part. if you don't need you can just ignore it.
```
hwuu: Does 'ignore it' mean that you provide an empty `image.yaml` file or just don't provide this file at all?

pai-management/paiLibrary/paiBuild/build_center.py
```diff
@@ -67,7 +67,7 @@ def get_base_image(self, image_name):
         file_path = "src/{0}/dockerfile".format(image_name)
         with open(file_path, 'r') as fin:
             for line in fin:
-                if line.startswith("FROM"):
+                if line.strip().startswith("FROM"):
```
hwuu: It looks Yundong has added this 'get base image' functionality already :)

Issue #606, opened 2018-05-23
watch disk usage on host machine


fanyangCS:
better alert when usage is close to full

ydye:
Can we add logic to triggle email-alerts on prom?

xudifsd:
`node_filesystem_avail_bytes` already exported via node-exporter, for example in [INT bed](http://10.151.40.234:9091/graph?g0.range_input=15m&g0.expr=node_filesystem_avail_bytes&g0.tab=0).

xudifsd:
closed as already solved in above PR.

PR #971, opened 2018-07-24
add email alert

### LINKED ISSUE #921: Provide email notification
As OpenPAI is an async platform. It's better to notify users when job started, completed or error. So that users don't need to visit web portal frequently.

xudifsd:
add alert manager, user could config email notification for now, default monitor cpu, mem, disk usage, will add more monitoring in the future.

fix #921 and #606


cluster-configuration/services-configuration.yaml
```diff
@@ -109,6 +109,13 @@ prometheus:
   prometheus-port: 9091
   # port for node exporter
   node-exporter-port: 9100
+  alerting:
+    alert_manager_port: 9093
+    alert_receiver: alert@example.com
```
YanjieGao: suggest Comment out the default config as an example. User maybe use quick config and not aware this config. The fake mail may cause error

fanyangCS:
please add readme and design code accordingly.

xudifsd:
@fanyangCS ok, in another PR

PR #987, opened 2018-07-25
change metrics name due to node_exporter upgrade

xudifsd:
node_exporter upgrade introduced many metrics name change, change grafana metrics to address no data point in font page.


fanyangCS:
please test it in INT bed after merge

xudifsd:
@fanyangCS deployed in INT bed, it looks all good.

PR #997, opened 2018-07-27
[k8s deployment] Replace return with exit code.

ydye:


pai-management/k8sPaiLibrary/maintainlib/add.py
```diff
@@ -55,32 +55,24 @@ def __init__(self, cluster_config, node_config, clean):
...
             self.logger.error("[{0}] Error: {1} is an undefined role, quit add job in host [{2}]".format(time.asctime(), node_config['k8s-role'], node_config['nodename']))
+            sys.exit(1)
 
 
 
     def prepare_package(self, node_config, job_name):
 
-        if self.jobname == 'error':
-            return
-
```
hao1939: Hi @ydye ,

When would `self.jobname == 'error'`?

pai-management/k8sPaiLibrary/maintainlib/kubectl_install.py
```diff
@@ -71,11 +72,16 @@ def kubectl_configuration_generate(self):
...
+
         while True:
             res = common.execute_shell_return("kubectl get node", "There will be a delay after installing, please wait.")
-
+            times = times + 1
             if res == True:
                 break
+            if times == 10:
+                self.logger.error("kubectl ready test failed. Exit paictl.")
+                sys.exit(1)
```
hao1939: What would happen after the retry count down?
How long will it wait?

pai-management/k8sPaiLibrary/maintainlib/add.py
ydye: In the function init(), this error has been detected and exit from the program.

pai-management/k8sPaiLibrary/maintainlib/kubectl_install.py
ydye: A short time, I think it's ok.

hao1939: Would it quit the deployment if fails?
Should it be ready in a short time?

hao1939: I'm worry about it would fail because of time out.
It's fine if we are sure it would ready in a short time.

ydye: The command failed is ok. Because sometimes, the api-server of the master node may be not ready. And if every thing is fine, it will started soon. And then the kubectl could work well. 
But if this command fail 10 times, I will assume something wrong happened.

PR #998, opened 2018-07-27
Add service healthcheck

### LINKED ISSUE #350: Add service health check in cluster wide.

### LINKED ISSUE #972: restart prometheus will prompt error
```
Error from server (AlreadyExists): error when creating "watchdog-ds.yaml": daemonsets.apps "watchdog" already exists
```

but start script already deleted
```
daemonset.extensions "watchdog" deleted
```

xudifsd:
fix #350 , also fix #972 

refactored previous service metric:
from `pod_current_probe_not_ready 1` to `pai_pod_count{ready="false", name="node_exporter"} 1`

for pai service pod, it will emit metric like:
```
pai_pod_count{host_ip="10.151.40.4",initialized="true",instance="10.151.40.4",job="node_exporter",name="zookeeper",phase="running",podscheduled="true",ready="true"}
```

pai_pod_count relays on all pai service labeled as `app: $service_name`, I found grafana not satisfy this, so I changed grafana's label.

for node, it will emit metric like:
```
pai_node_count{disk_pressure="false",instance="10.151.40.4",job="node_exporter",memory_pressure="false",name="10.151.40.4",out_of_disk="false",ready="true"}
```

this refactoring will make alert rule much simpler.

Also added some rule we frequently encountered when operating our cluster. Will then deployed in INT bed to ease the burden of DRI.

### DIFF
--- a/prometheus/doc/watchdog-metrics.md
+++ b/prometheus/doc/watchdog-metrics.md
@@ -1,4 +1,4 @@
-# Watchdog 
+# Watchdog
 watchdog will check service / K8S / Node level health. It is a long running service and will continues report & record health metrics to log files.
 # How to use
 For example
@@ -8,7 +8,7 @@ ssh watchdogPodHostIP
 
 cd /var/log/containers
 
-vi watchdog-xx.log 
+vi watchdog-xx.log
 
 log example:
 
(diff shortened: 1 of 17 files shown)


xudifsd:
@YitongFeng I changed grafana's label key from k8s-app to app to keep consistent with other components, I search its usage, hopefully this change will not break anything?

prometheus/exporter/watchdog.py
```diff
@@ -33,288 +34,280 @@
...
         logger.info(metric)
 
+
+def load_machine_list(configFilePath):
+    cluster_config = common.load_yaml_file(configFilePath)
+    return cluster_config['hosts']
+
+
 def main(argv):
     logDir = argv[0]
```
sterowang: Could you add some comments here to describe what things we are monitored and what metrics we generated?

prometheus/test/data/pods_list.json
```diff
@@ -5,68 +5,17 @@
...
     {
       "metadata": {
         "name": "drivers-one-shot-ndldh",
         "generateName": "drivers-one-shot-",
         "namespace": "default",
-        "creationTimestamp": "2018-07-18T06:27:22Z"
+        "creationTimestamp": "2018-07-18T06:27:22Z",
+        "labels": {
+            "app": "drivers-one-shot",
+            "controller-revision-hash": "2607302775",
```
sterowang: >2607302775 [](start = 41, length = 10)

What does this hash means? Is it will break tests in future potentially?

xudifsd: this copied from real world server, I do not think this will break tests, I only used label with app key.

prometheus/exporter/watchdog.py
xudifsd: sure, I was intended to describe in doc. but describe in code should be also helpful.

prometheus/doc/watchdog-metrics.md
```diff
@@ -37,47 +37,32 @@ go to k8s container log folder
...
-| node_current_notready| node status, value 1 is error|
-| notready_node_count| all nodes not ready count|
-| node_current_docker_error| per node docker daemon occurs error, 1 is error |
-| docker_error_node_count| all nodes docker occurs error node count |
+| k8s_api_server_count | has error key in label, if error != "ok", means api server is not functioning correctly |
+| k8s_etcd_count | has error key in label, if error != "ok", means etcd is not functioning correctly |
+| k8s_kubelet_count | has error key in label, if error != "ok", means kubelet is not functioning correctly |
 
+# Alerting
+Alerting rules are under `[prometheus/prometheus-alert](../prometheus-alert)`, we added some basic healthcheck rules for pai service and node.
```
YanjieGao: suggest we tell end user more details, such as how to add the rule at which file of pai.

xudifsd: Added basic info. I intended to rewrite some parts of doc. Put metrics & alerting in one doc file, and add how to emit metric from other pai service.

YanjieGao: that's useful

YanjieGao: as your plan, list the rule file path here can make user feel smooth here and minimal cost. otherwise user read alert manager document but can't find the config file

xudifsd: yeah, I will submit another doc PR. Let this PR focus on functionality.

PR #999, opened 2018-07-27
skip the error message from docker inspect since it is misleading

DongZhaoYu:
Keep the exit handler as it is and only skip its error output.


fanyangCS:
> function exit_handler()

please sync with the latest code first.

---
Refers to: rest-server/src/templates/yarnContainerScript.mustache:29 in f49704a. [](commit_id = f49704a004bc75ded3d1557e9ca837275c47676a, deletion_comment = False)

rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -30,7 +30,7 @@ function exit_handler()
 {
   printf "%s %s\n" \
     "[DEBUG]" "Yarn container exit handler: EXIT signal received in yarn container, exiting ..."
-  pid=$(docker inspect --format={{{ inspectFormat }}} $docker_name) && kill -9 $pid ||\
+  pid=$(docker inspect --format={{{ inspectFormat }}} $docker_name 2>/dev/null) && kill -9 $pid ||\
     printf "%s %s\n"  "[DEBUG]" "Yarn container exit handler tries to kill the container $docker_name that does not exist. It probably has already exited."
```
fanyangCS: >printf "%s %s\n [](start = 4, length = 15)

I suggest to break this into multiple steps.
1. get pid (and drop the log in case of failure)
2. if cannot get pid, go to 4
3. if get the pid, kill -9. log the result.  
4. kill 0

rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -28,10 +28,13 @@ BASH_XTRACEFD=13
...
 function exit_handler()
 {
-  printf "%s %s\n" \
-    "[DEBUG]" "Yarn container exit handler: EXIT signal received in yarn container, exiting ..."
-  pid=$(docker inspect --format={{{ inspectFormat }}} $docker_name) && kill -9 $pid ||\
-    printf "%s %s\n"  "[DEBUG]" "Yarn container exit handler tries to kill the container $docker_name that does not exist. It probably has already exited."
+  pid=$(docker inspect --format={{{ inspectFormat }}} $docker_name 2>/dev/null)
+  if [ $pid ]; then
+    printf "%s %s\n" \
+      "[DEBUG]" "Yarn container exit handler: EXIT signal received in yarn container, exiting ..."
```
fanyangCS: I think this should be moved to the beginning ?

rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -28,10 +28,13 @@ BASH_XTRACEFD=13
...
+    kill -9 $pid ||\
+      printf "%s %s\n"  "[DEBUG]" "Yarn container exit handler tries to kill the container $docker_name that does not exist. It probably has already exited."
```
fanyangCS: >that does not exist. It probably has already exited. [](start = 104, length = 52)

but failed. Maybe it has already exited.

rest-server/src/templates/yarnContainerScript.mustache
sterowang: Yes, the more readable and trackable log messages should looks like:

[DEBUG] Yarn container exit handler: EXIT signal received in yarn container, performing clean up actions ...
[DEBUG] Trying to killing Docker container "xxx" ...
[DEBUG] $docker_name that does not exist. It probably has already exited.
or 
[DEBUG] $docker_name killed.

---
In reply to: [205736949](https://github.com/Microsoft/pai/pull/999#discussion_r205736949) [](ancestors = 205736949)

DongZhaoYu: refine the log to make it more readable and clear. done.

rest-server/src/templates/yarnContainerScript.mustache
DongZhaoYu: log changed. done

PR #1002, opened 2018-07-27
[pai-build] add pai-build doc.

wangcan0329:


hwuu:
PAI build is a special component of PAI compared with those dockerized micro-service modules. So it is neither necessary nor correct to follow the same document structure as they do. I suggest you re-organize the doc to make the following items clear:
- What is PAI's process of building docker images?
- What standard a module should follow to integrate with PAI's build mechanism?
- How to configure the behavior of PAI build? (For example, using a build-level configuration file.)
- What is the future plan?

hwuu:
Btw, if you remove this functionality from paictl, please find somewhere to place the usage of 'build' that end-users can follow.

wangcan0329:
I agree we can add a section to describe "What is PAI's process of building docker images?". But I think this document aims to describe what is **current**  PAI build process, **not a design** doc for future build process. 
I think this document should match our current implementation, actually we don't have build-level configuration file and have not removed build from pai-management. The document will be updated continuously when we refine our process, or have new features.

Issue #411, opened 2018-03-26
Link metrics averaging timespan (Grafana and web portal) with Prometheus's scrapping interval


hwuu:
Timespan should be larger than the scrapping interval.

PR #1011, opened 2018-07-30
Fix grafana timespan config hardcode 

YitongFeng:
Fix issue #411 
After sync with Di, we think that if two services share the configuration, using Object model is better than sharing the same config-map.


cluster-configuration/services-configuration.yaml
```diff
@@ -101,6 +101,7 @@ prometheus:
   prometheus-port: 9091
   # port for node exporter
   node-exporter-port: 9100
+  scrape_interval: 30
```
xudifsd: You should also modify `pai-management/quick-start/services-configuration.yaml.template`.

Otherwise quick start user will get jinja2 undefined variable exception.

YitongFeng: Solved

pai-management/bootstrap/grafana/grafana-configuration/pai-clusterview-dashboard.json.template
```diff
@@ -388,7 +388,7 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "100 - avg (irate(node_cpu_seconds_total{mode=\"idle\"}[5m])) * 100",
+              "expr": "100 - avg (irate(node_cpu_seconds_total{mode=\"idle\"}[{{clusterinfo['prometheusinfo']['scrape_interval']}}0s])) * 100",
```
hwuu: Is there an extra '0'? -- {{clusterinfo['prometheusinfo']['scrape_interval']}}**0**s

hwuu:
Would you like to fix web portal as well?

pai-management/bootstrap/grafana/grafana-configuration/pai-clusterview-dashboard.json.template
YitongFeng: Yes, as we can't do arithmetic in this grafana metrics, so I did a hack that add a "0", let it times 10.

YitongFeng:
Ok, let me do it. @hwuu

pai-management/bootstrap/grafana/grafana-configuration/pai-clusterview-dashboard.json.template
sterowang: We can follow the mock clusterinfo currently, but should refactor it in future with formal solutions.

YitongFeng: Remind by Di, the Jinja support arithmetic, so in webportal fix, I changed it to " * 10 ", I will change the grafana settings in this way as well.

PR #1000, opened 2018-07-27
Web portal: Refine document

Gerhut:

PR #1021, opened 2018-07-31
Web Portal: Optimize speed of `npm start`

### LINKED ISSUE #936: Optimize boot speed of Web Portal
At present, it will compile all assets files, including all CSS, JavaScript files when web portal container boots.

https://github.com/Microsoft/pai/blob/bab0081b567caec29c59ced6ab7d9c1f209f411d/webportal/package.json#L79

https://github.com/Microsoft/pai/blob/bab0081b567caec29c59ced6ab7d9c1f209f411d/webportal/package.json#L71

https://github.com/Microsoft/pai/blob/bab0081b567caec29c59ced6ab7d9c1f209f411d/webportal/package.json#L81

Some of the compile processes can be moved to build stage, except supply runtime configuration(environment variables, like REST_SERVER_URI, etc.) to pages.

There are 3 possible solutions for the configuration supplement:

1. By AJAX, write an API in express, which responses current environment variables, requested when each page opened.
2. Inline to pages, make an extra render process to render a JSON formatted environment object(dictionary) in each page when booted.
3. Generate a JavaScript file, make an extra render process to generate a JavaScript file including environment when booted, referenced in each page, by `<script>`.

At present, the environment object generating process runs just before code compiling, as a JSON module. It cannot works after moving compiling process to build stage.

Gerhut:
1. Move `npm run build` run in image build stage, decouple with env.
2. Inject env to `window.ENV` in `env.js`, refernced by all pages.
3. Rename `webportal.config.json` to js, which exports `window.ENV`.

Closes #936


Gerhut:
Docs will be updated after #1000 merged

webportal/package.json
```diff
@@ -67,18 +68,14 @@
...
     "lint": "eslint \"src/**/*.js\"",
-    "node": "node",
-    "preprocess": "node config/preprocess.js",
-    "rimraf": "rimraf",
-    "server": "node server/index.js",
-    "start": "npm run build && npm run server",
     "test": "npm run lint",
-    "webpack": "webpack --config webpack.config.js --progress --profile",
+    "clean": "rimraf dist",
+    "prebuild": "npm run clean && node config/preprocess.js",
```
hwuu: When are `prebuild` and `prestart` called?

Gerhut: @hwuu They runs just before `npm run build` & `npm start` (shorthand of `npm run start`)

<https://docs.npmjs.com/misc/scripts>

webportal/src/app/layout/layout.component.ejs
```diff
@@ -114,6 +114,6 @@
     <div class="control-sidebar-bg"></div>
 
   </div>
-
+  <script src="scripts/env.js"></script>
```
hwuu: Why do you add this to this .ejs file? I'm wondering whether 'require' the .js file in job-view.component.js is sufficient?

webportal/src/app/env.js
```diff
@@ -0,0 +1,10 @@
+/* eslint-disable */
+window.ENV = {
```
hwuu: I'm curious about how your method works. I guess this env.js is a template, and it will be dynamically instantiated by envsub inside the container before the service starts (the 'prestart' command of npm). I'm I right?

webportal/src/app/cluster-view/hardware/hardware-detail.component.js
```diff
@@ -19,7 +19,7 @@
 
 const hardwareDetailComponent = require('./hardware-detail.component.ejs');
 const breadcrumbComponent = require('../../job/breadcrumb/breadcrumb.component.ejs');
-const webportalConfig = require('../../config/webportal.config.json');
+const webportalConfig = require('../../config/webportal.config.js');
```
hwuu: Is it a common practice that we'd better 'require' a .js file other than a .json file?

webportal/src/app/env.js
Gerhut: @hwuu You are right, this makes the environment variables injected to window.ENV in the runtime, not related to webpack build.

webportal/src/app/layout/layout.component.ejs
Gerhut: @hwuu `layout.component.ejs` is the base template of all html pages, it makes all page scripts can access `window.ENV` by `require('webportal.config.js')`

webportal/src/app/cluster-view/hardware/hardware-detail.component.js
Gerhut: @hwuu In fact `require('webportal.config')` is better, it will make webpack to choose the most suitable extension, even `.ts` if exists and related plugin is used.

Gerhut:
Benchmarks:


Optimization \ Stage | Build | Run
-|-|-
without |56.738s|30.9s
with | 1:32.97s | 0.3s


Build:
- `time docker build --tag webportal .`
- Record total seconds

Run:
- `docker run --detach --name container_name image_tag`
- `docker logs --timestamp container_name`
- Record timestamp of first line & listening line, calculate difference

Without optimization:
- Build: **56.738s**
- Run: **8:19:18.7 - 8:18:47.8 = 30.9s**

webportal/package.json
```diff
@@ -74,7 +74,7 @@
     "prebuild": "npm run clean && node config/preprocess.js",
     "build": "webpack --config webpack.config.js --progress --profile",
     "build:dev": "npm run build -- --debug --devtool eval-cheap-module-source-map --output-pathinfo --watch",
-    "prestart": "envsub --env-file .env --system src/app/env.js dist/scripts/env.js",
+    "prestart": "envsub --env-file .env --system src/app/env.js dist/scripts/env.js && ping -c 1 -W 1 172.17.0.1",
```
hwuu: Explain a little?

Gerhut: It's a workaround to pass the cluster deploy test in Jenkins, not related to this optimization. I will revert it and raise a new PR to discuss it.

PR #1023, opened 2018-08-01
[WIP]Web portal: optimize service view

### LINKED ISSUE #605: speed up service view loading

Gerhut:
Fixes #605 

Currently `/api/v1/nodes` & `/api/v1/namespaces/{namespace}/pods` API of K8S are requested in sequence, this PR make them requested in parallel.

Sequence:
image

Parallel:
image

There is a request slowdown when in parallel, may caused by limited bandwidth or service performance.

Since `nodes` response has less size than `pods` but costs more time, I think service performance issue may be more likely.

Ping @ydye @YitongFeng for some help, thank you.


Gerhut:
Request / update pods page by page may works, will have a try.

PR #1029, opened 2018-08-02
[Pylon] Add readiness probe

hwuu:


pai-management/bootstrap/pylon/pylon.yaml.template
```diff
@@ -58,5 +58,11 @@ spec:
             - name: pylon
               containerPort: 80
               hostPort: {{ clusterinfo['pyloninfo']['port'] }}
+          readinessProbe:
+            httpGet:
+              path: /healthz
+              port: pylon
+            initialDelaySeconds: 10
```
YanjieGao: not sure the init time cost of nginx, initialDelaySeconds maybe longer and not so tight would be better.  Will avoid first time readiness probe log confused user for weak server deployment env.

hwuu: To my experience, the time cost is very short, less than 1 second. I put a very conservative value here.

YanjieGao: ok, np

PR #1030, opened 2018-08-02
Enable exit code in service deployment

ydye:


hwuu:
How about other modules such as rest-server?

ydye:
@hwuu 
Wait @DongZhaoYu to replace create with apply. Then I will make this change to all service.

pai-management/bootstrap/cluster-configuration/delete.sh
```diff
@@ -19,6 +19,6 @@
 
 pushd $(dirname "$0") > /dev/null
 
-/bin/bash stop.sh
+/bin/bash stop.sh || exit $?
```
sterowang: I remembered there some shell handler used to exit bash if some commands exit code is none 0, can that handler be used in paictl scripts so that we needn't to add "exit $?" everywhere?

ydye: The script ```set -e``` could make bash handle exit code every where. But in some specific place, the exit code is none zero is expected. So I handle exit in this way.

PR #1033, opened 2018-08-03
[WIP]Fix k8s cluster network connection issues.

Gerhut:
Currently, web portal (as well as rest server sometimes) cannot be connected in Jenkins test of cluster mode, `Connection reset by peer.` raised. However, when web portal pings outside once (whatever before or after server is started), it could be connected normally. This PR tries to find the cause of this wired behavior.

- Must ping outside, `127.0.0.1` / `127.0.0.11`(Docker's DNS) / container IP not works.


Gerhut:
Added a connection check to REST server before server start, which makes the web portal "ping outside", pass the cluster test, and not looks like a workaround.

webportal/server/server.js
```diff
@@ -17,13 +17,28 @@
...
+    }
+  })
+};
+
+checkRestServer(() => {
+  // start the server
+  app.listen(config.serverPort, () => {
+    logger.info('Webportal server starts on port %d', config.serverPort);
+  });
+})
```
sterowang: Synced with Qixiang offline, we can add a "ping restserver" to workaround the problem on CI/CD bed, but we should still not introduce dependency on RestAPI here. Webportal has his own error handling and can give user correct response when RestServer is not avaliabile.

Gerhut:
Disappeared after reboot the node.

PR #1038, opened 2018-08-04
upgrade k8s without deleting etcd data

DongZhaoYu:
The etcd data will not be deleted by default when clean the cluster. New configurations will applied to the live objects if the objects is restored by k8s when the cluster is started.


pai-management/bootstrap/cluster-configuration/configmap-create.sh
```diff
@@ -17,6 +17,6 @@
 # DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-kubectl create configmap  host-configuration --from-file=host-configuration/
-kubectl create configmap  docker-credentials --from-file=docker-credentials/
-kubectl create configmap  gpu-configuration --from-file=gpu-configuration/
\ No newline at end of file
+kubectl create configmap  host-configuration --from-file=host-configuration/ --dry-run -o yaml | kubectl apply --overwrite=true -f -
```
ydye: What‘s effect of --dry-run here? And what's effect of --overwrite here?
If I remove --overwrite, what's will happen? I know when we label the node with overwrite, if the key already exists, it will be ovewrite. But when applying some object, without the --overwrite, it still works.

DongZhaoYu: The --dry-run here is to let create command generate the configuration without executing it. Since we need to read config from host-configuration file, the apply command cannot support reading config from a file. So I leverage the create dry run to generate the config and then apply it to the object.

The --overwrite flag is to resolve the conflict when applying new configuration by overwriting the old config with the new one. If we don't enable overwrite here, when merge the new config with the live config, if conflicts occur, the apply will fail and kubectl will report an error. Here I enable --overwrite is to make sure the new config can be successfully applied to the live object in case the object has been restored by k8s.

PR #1045, opened 2018-08-07
[SSH Access]add hdfs and ssh check in dockerContainerScript

wangcan0329:


mzmssg:
Could you explain why don't we replace hdfs with webhdfs?

wangcan0329:
This is a low priority work item to use webhdfs instead, so we just want to mitigate this issue now.

PR #1056, opened 2018-08-10
generate pai_service_name label from app label

xudifsd:
make prometheus generate pai_service_name lable for scraped metrics.

### DIFF
--- a/prometheus/doc/exporter-for-other-services.md
+++ b/prometheus/doc/exporter-for-other-services.md
@@ -148,4 +148,5 @@ If service decide to leverage node-exporter to expose metrics, service should mo
 located in `/datastorage/prometheus` and write file with extension `.prom`.
 
 For expose using port, service should annotate pod with `prometheus.io/scrape`, `prometheus.io/path`
-and `prometheus.io/port`.
+and `prometheus.io/port`. For service with label key `app`, prometheus will generate `pai_service_name`
+label for generated metrics.
--- a/prometheus/doc/exporter-for-other-services.md
+++ b/prometheus/doc/exporter-for-other-services.md
@@ -151,7 +151,7 @@ For expose using port, service should annotate pod with `prometheus.io/scrape`, 
 and `prometheus.io/port`. For service with label key `app`, prometheus will generate `pai_service_name`
 label for generated metrics.
 
-prometheus will list all pods and get those pods with `prometheus.io/scrape` is true, and then,
+Prometheus will list all pods and get those pods with `prometheus.io/scrape` is true, and then,
 prometheus will issue request to `http://${pod_host_ip}:${prometheus.io/port}${prometheus.io/path}` to
 get metrics.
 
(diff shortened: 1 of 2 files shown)


prometheus/doc/exporter-for-other-services.md
```diff
@@ -148,4 +148,5 @@ If service decide to leverage node-exporter to expose metrics, service should mo
 located in `/datastorage/prometheus` and write file with extension `.prom`.
 
 For expose using port, service should annotate pod with `prometheus.io/scrape`, `prometheus.io/path`
-and `prometheus.io/port`.
+and `prometheus.io/port`. For service with label key `app`, prometheus will generate `pai_service_name`
+label for generated metrics.
```
YanjieGao: it's better give a simplest example service config this label at which position.

prometheus/doc/exporter-for-other-services.md
```diff
@@ -148,4 +148,57 @@ If service decide to leverage node-exporter to expose metrics, service should mo
...
+      name: foo
+      labels:
+        app: foo
+      annotations:
+        prometheus.io/scrape: "true"
+        prometheus.io/path: "/"
+        prometheus.io/port: "80"
+    spec:
+      containers:
+      - name: nginx
```
YanjieGao: Question:
1 nginx is sidecar container for other services to expose metrics at specific port? if this , it's better tell user this info 
2 Or nginx is a exmple for nginx service expose metrics?
3 If suggest to use port to expose, how about tell owners to Unified definition of port usage to avoid conlict

xudifsd: This is a simple example to use port to expose something. I can update doc to illustrate it more clearly.

I really do not know how to avoid port conflict. Maybe coordinate with other to use different port. The best solution for this is enable kubernetes network model and make service use private network. I will update doc to suggest coordinate with others.

prometheus/doc/exporter-for-other-services.md
```diff
@@ -148,4 +148,66 @@ If service decide to leverage node-exporter to expose metrics, service should mo
...
+      hostNetwork: true
+```
+
+will generate metrics like:
+
+`pai_service_count{instance="10.151.40.4:80",job="pai_serivce_exporter",k8s_pod_name="foo-69869cc4d6",pai_service_name="foo"} 1`
+
+metric labels `k8s_pod_name` and `pai_service_name` are generated from pod name and pod label value which has key `app`.
+
+In above example, we use nginx to accept reqests from prometheus and expose metrics we write in
```
YanjieGao: reqest

xudifsd: @YanjieGao done

YanjieGao: i mean spell error at reqest. see this sentence "we use nginx to accept "reqests" from prometheus"

xudifsd: yes, already fixed

PR #1068, opened 2018-08-13
restruct hadoop folder

wangcan0329:


src/hadoop-run/build/component.dep
```diff
@@ -0,0 +1 @@
+hadoop-ai
```
ydye: If some component is specified here, what will happen?
1) build-pre.sh
2) Copy

or other order

wangcan0329: Dependency wrote in components.dep should just be component name and make sure it can be found under src folder.

ydye: And then copy all the folder into current compnent?

ydye: So, how can we confirm that the hadoop binary will be built when building hadoop-run? I mean the building order.

wangcan0329: 1). Hadoop-run's component.dep writes dependency of hadoop-ai.
2). Resolve component dependency according to “components.dep” and Dockerfile “From” statement then will get a topological graph. find we need to build hadoop-ai first.
3). Build hadoop-ai first.
4). Copy the hadoop-ai folder to hadoop-run.

ydye: Ok

PR #1069, opened 2018-08-13
Fix scrape time has no default value

YitongFeng:


xudifsd:
Does other references in grafana have the same problem?

YitongFeng:
Have add default value as well. @xudifsd

PR #1073, opened 2018-08-14
Chenqi paishare

MaggieQi:
Fix Outputdir cannot write due to permission error


shishaochen:
Please use "Squash Merge" for better history track.

PR #1080, opened 2018-08-15
Refactor pai-management document

ydye:


YanjieGao:
which service use hbase?

ydye:
No service use hbase

ydye:
An example here

PR #1106, opened 2018-08-20
refine faq for resource allocation

YanjieGao:
add default configuration

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -14,9 +14,17 @@ A: Users can find historical job logs through yarn. Please check [issue-1072](ht
 
 ### Q: To improve the cluster usage, user would like to see a VC can use up all cluster resource if others don’t use it.
 
-A: By default, a VC can use up all cluster resource if others don’t use it. OpenPAI use [capacity scheduler](https://hadoop.apache.org/docs/r1.2.1/capacity_scheduler.html) of YARN for resource allocation. maximum-capacity defines a limit beyond which a queue cannot use the capacity of the cluster. This provides a means to limit how much excess capacity a queue can use. By default, there is no limit. [OpenPAI capacity scheduler](../pai-management/bootstrap/hadoop-resource-manager/hadoop-resource-manager-configuration/capacity-scheduler.xml.template) not set this item and there is no limit.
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -14,9 +14,17 @@ A: Users can find historical job logs through yarn. Please check [issue-1072](ht
...
 
 ### Q: To ensure one user cannot occupy excessive resource, operator would like to set a quota constraint for individual users. 
 
 A: OpenPAI use capacity scheduler of YARN for resource allocation. User can configure the item "[yarn.scheduler.capacity.root.{{ queueName }}.minimum-user-limit-percent](https://hadoop.apache.org/docs/r1.2.1/capacity_scheduler.html)" to control the user's resource quota. This configuration item is in this file [capacity-scheduler.xml.template](../pai-management/bootstrap/hadoop-resource-manager/hadoop-resource-manager-configuration/capacity-scheduler.xml.template) of OpenPAI.
 
+```xml
+<property>
+    <name>yarn.scheduler.capacity.root.{{ queueName }}.minimum-user-limit-percent</name>
```
mzmssg: In my test, a single user can still use up all resource in spite of lower this item, and other user cannot preempt the allocated resource. A better limitation item is `yarn.scheduler.capacity.root.{{ queueName }}.user-limit-factor` , which can prevent user occupy excessive resource.

YanjieGao:
The item could added.

For use test case:
I guess : 
- not use many user to test and make the percent to touch the config percent ? (see official intro )

The offical intro:
mapred.capacity-scheduler.queue.<queue-name>.minimum-user-limit-percent | 
Each queue enforces a limit on the percentage of resources allocated to a user at any given time, if there is competition for them. This user limit can vary between a minimum and maximum value. The former depends on the number of users who have submitted jobs, and the latter is set to this property value. For example, suppose the value of this property is 25. If two users have submitted jobs to a queue, no single user can use more than 50% of the queue resources. If a third user submits a job, no single user can use more than 33% of the queue resources. With 4 or more users, no user can use more than 25% of the queue's resources. A value of 100 implies no user limits are imposed.
-- | --

Issue #1095, opened 2018-08-17
node exporter failed


shawntian90:
Hi,

When node exporter container launched, the [healthy_check.py](https://github.com/Microsoft/pai/blob/master/prometheus/exporter/healthy_check.py) will failed when it was running on a `Master` node, which has NVIDIA GPU card. 

This GPU is not configured to used by pai cluster, also not configured in `/cluster-configuration/cluster-configuration.yaml`.

It failed my because the NVIDIA driver is not properly installed or the driver version is not `384.111`.

I'd like to help get out of this bug via PR, which may fixed by check this server is GPU node or CPU node. But I do not find a good way to fix it.

Pls help!

YitongFeng:
Hi, Shawn, could you please provide the node-exporter's error log and your cluster-configuration?

shawntian90:
## cluster-configuration
In the following settings, the node `master-node` has one NVIDIA GPU card, which won't be used in openpai cluster.

```
default-machine-properties:
  # Account with root permission
  username: ****
  password: ****
  sshport: ****


machine-sku:

  WORKER:
    mem: 128
    gpu:
      type: geforcegtxtitanx
      count: 8
    cpu:
      vcore: 32
    dataFolder: "/data"
    #Note: Up to now, the only supported os version is Ubuntu16.04. Please do not change it here.
    os: ubuntu16.04

  CPUNODE:
    mem: 16
    cpu:
      vcore: 16
    dataFolder: "/data"
    #Note: Up to now, the only supported os version is Ubuntu16.04. Pls don't change it here.
    os: ubuntu16.04



machine-list:

    - hostname: master-node
      hostip: IP
      machine-type: CPUNODE
      etcdid: etcdid1
      k8s-role: master
      dashboard: "true"
      zkid: "1"
      pai-master: "true"
      watchdog: "true"
      alert-manager: "true"


    - hostname: worker-node
      hostip: IP
      machine-type: WORKER
      k8s-role: worker
      pai-worker: "true"
```

shawntian90:
## node-exporter error log

```
Readiness probe failed: /bin/sh: 1: nvidia-smi: not found Traceback (most recent call last): File "/usr/local/healthy_check.py", line 88, in <module> main() File "/usr/local/healthy_check.py", line 85, in main raise RuntimeError("gpu-exporter readiness probe failed, error component:" + exception) RuntimeError: gpu-exporter readiness probe failed, error component:| nvidiasmi
```

[...]

PR #1117, opened 2018-08-22
bugfix: check nvidia-smi exists before run `nvidia-smi -q -x`

shawntian90:
Bugfix for #1095 

Run command `which nvidia-smi` to check exists of `nvidia-smi`, preventing the health-check error.

### DIFF
--- a/prometheus/exporter/gpu_exporter.py
+++ b/prometheus/exporter/gpu_exporter.py
@@ -67,8 +67,8 @@ def collect_gpu_info():
     except OSError as e:
         if e.errno == os.errno.ENOENT:
             logger.warning("nvidia-smi not found")
-        else:
-            raise
+
+        return None
 
 
 
--- a/prometheus/exporter/gpu_exporter.py
+++ b/prometheus/exporter/gpu_exporter.py
@@ -23,6 +23,7 @@ import os
 import logging
 
 import utils
+
 from utils import Metric
 
 logger = logging.getLogger(__name__)
@@ -54,12 +55,9 @@ def collect_gpu_info():
     try:
         logger.info("call nvidia-smi to get gpu metrics")
 
-        nvidia_smi_exists = len(utils.check_output(["which", "nvidia-smi"])) > 0
-        if nvidia_smi_exists:
-            smi_output = utils.check_output(["nvidia-smi", "-q", "-x"])
-            return parse_smi_xml_result(smi_output)
-        else:
-            logger.info('nvidia-smi not found')
+        smi_output = utils.check_output(["nvidia-smi", "-q", "-x"])
+
+        return parse_smi_xml_result(smi_output)
     except subprocess.CalledProcessError as e:
         if e.returncode == 127:
             logger.exception("nvidia cmd error. command '%s' return with error (code %d): %s",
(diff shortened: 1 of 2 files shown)


shawntian90:
#1095

prometheus/exporter/gpu_exporter.py
```diff
@@ -54,9 +54,12 @@ def collect_gpu_info():
     try:
         logger.info("call nvidia-smi to get gpu metrics")
 
-        smi_output = utils.check_output(["nvidia-smi", "-q", "-x"])
-
-        return parse_smi_xml_result(smi_output)
+        nvidia_smi_exists = len(utils.check_output(["which", "nvidia-smi"])) > 0
```
xudifsd: catch OSError directly like [this](https://stackoverflow.com/a/11210185) so you do not have to call `which`

prometheus/exporter/gpu_exporter.py
```diff
@@ -54,9 +54,12 @@ def collect_gpu_info():
...
+        if nvidia_smi_exists:
+            smi_output = utils.check_output(["nvidia-smi", "-q", "-x"])
+            return parse_smi_xml_result(smi_output)
+        else:
+            logger.info('nvidia-smi not found')
```
xudifsd: maybe using logger.warning.

For events we expected it happen and not happen a lot, we use info.
For events we do not expect but reasonable, we use warning.
For events indication configuration/hardware/environment error that may cause the program not working/act wrongly, we use error.

prometheus/exporter/gpu_exporter.py
shawntian90: refactored

prometheus/exporter/gpu_exporter.py
shawntian90: refactored

prometheus/exporter/gpu_exporter.py
```diff
@@ -64,6 +64,13 @@ def collect_gpu_info():
         else:
             logger.exception("command '%s' return with error (code %d): %s",
                     e.cmd, e.returncode, e.output)
+   except OSError as e:
+        if e.errno == os.errno.ENOENT:
+            logger.warning("nvidia-smi not found")
+        else:
+            raise
```
xudifsd: Please do not raise here.

If command not exists, this means the user does not want gpu being used by pai, so exporter and health checker will *pretend* gpu not exist in this node. So we just return default value which is None.

Same applies to `healthy_check.py`

shawntian90: gotcha and fixed with new commit

shawntian90:
@xudifsd 
I found that source of pai, `gpu_exporter.py`, mixed with camel style and snake case, even in one file. 

**CamelCase**
```python
for gpu in gpuList:
        minorNumber = gpu.getElementsByTagName('minor_number')[0].childNodes[0].data
        utilization = gpu.getElementsByTagName("utilization")[0]

        gpuUtil = utilization.getElementsByTagName('gpu_util')[0].childNodes[0].data.replace("%", "").strip()
        gpuMemUtil = utilization.getElementsByTagName('memory_util')[0].childNodes[0].data.replace("%", "").strip()
```

**snake_case**
```python
logger.info("call nvidia-smi to get gpu metrics")

smi_output = utils.check_output(["nvidia-smi", "-q", "-x"])

return parse_smi_xml_result(smi_output)
```

Is there any reason for that?

xudifsd:
@shawntian90 Hi, it is not a deliberate result, previous service maintainer uses camel style, but since PEP-8 recommended underscore style and I personally favor it, I gradually change them into underscore style.

PR #1118, opened 2018-08-22
[Pai build] refactor pai build

wangcan0329:


build/model/dependency_graph.py
```diff
@@ -0,0 +1,110 @@
...
+    def topology(self):
+        prev_count = dict()
+        ret = list()
+        search_queue = list()
+        for name, node in self.services.items():
+            prev_count[name] = len(node.inedges)
+            if prev_count[name] == 0:
+                search_queue.append(name)
+
+        while search_queue:
```
ydye: A loop check here will be better.

wangcan0329: done.

Issue #227, opened 2018-01-15
in clusterview, show allocated GPU count instead of used GPU count


fanyangCS:

scarlett2018:
cleaning old issues,  @fanyangCS   please help to evaluate whether this is still an issue, thanks.

fanyangCS:
making the reserved GPU count consistent

YitongFeng:
Is on the way, should we add this feature in v0.7 release? @fanyangCS

fanyangCS:
please link the PR to this issue. If the change involves multiple components, I would suggest to test it thoroughly and go with the next release.

YitongFeng:
#1035

PR #1140, opened 2018-08-24
modify grafana view to keep consistent with yarn

YitongFeng:
Test on my VM: http://10.169.8.247:3000/dashboard/db/pai_clusterview?refresh=30s&orgId=1

image

Solve issue #227


fanyangCS:
can we not show NotAvailableGPUs in the first page? it would be confusing to users. ok to hide it somewhere

fanyangCS:
@YitongFeng , I do not suggest to remove it. I suggest not to show it in the landing page. Plus, I suggest you explain to user what does NotAvailableGPU mean to user.

PR #1144, opened 2018-08-24
refactor paictl.py

xudifsd:
use `add_subparser` and `add_parser` to handle subcmd in paictl.

This can print help msg for both `paictl.py -h` and `paictl.py machine -h` and `paictl.py machine add -h`.

### DIFF
--- a/pai-management/k8sPaiLibrary/maintaintool/kubectl-install.sh
+++ b/pai-management/k8sPaiLibrary/maintaintool/kubectl-install.sh
@@ -19,12 +19,11 @@
 
 mkdir -p ~/.kube
 
-[[ -f "/usr/local/bin/kubectl" ]] &&
-{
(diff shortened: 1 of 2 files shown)


pai-management/k8sPaiLibrary/maintaintool/kubectl-install.sh
```diff
@@ -19,12 +19,11 @@
 
 mkdir -p ~/.kube
 
-[[ -f "/usr/local/bin/kubectl" ]] &&
-{
+if which kubectl > /dev/null; then
```
ydye: what's the difference between these two different style.

xudifsd: I put kubectl in differnt location, but in PATH. Previous method failed in this case.

PR #1149, opened 2018-08-24
Zhaoyu/hdfs access

DongZhaoYu:
add ways to access hdfs data

### DIFF
--- a/pai-management/doc/hdfs.md
+++ b/pai-management/doc/hdfs.md
@@ -148,7 +148,8 @@ curl -i -X PUT "http://hdfs-name-node-address:50070/webhdfs/v1/test/test_file?op
 ```
 This command will return the data node where the file should be written. The location URI would be like
 >http://hdfs-name-node-address:50075/webhdfs/v1/test/test_file?op=CREATE&namenoderpcaddress=hdfs-data-node-address:9000&createflag=&createparent=true&overwrite=false
-<br>Then run following command with this URI to write file data:
+
+Then run following command with this URI to write file data:
 ```bash
 curl -i -X PUT -T file-data-to-write returned-location-uri
 ```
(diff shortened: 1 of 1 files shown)


pai-management/doc/hdfs.md
```diff
@@ -91,6 +91,66 @@ curl http://NAME_NODE_ADDRESS:50070/jmx
 Currently OpenPAI management tool doesn't deploy HDFS in a High Availability (HA) fashion. This will be added in a future release.
 For solution about the HA feature please refer [HDFS High Availability](https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-hdfs/HDFSHighAvailabilityWithNFS.html).
 
+# Access HDFS Data
```
YanjieGao: Hi Zhaoyu, currently pai suggest user use webhdfs access data. Please add web hdfs api of PAI to tell user how to 
add / delete data etc.

https://github.com/Microsoft/pai/blob/1e8db911a4cb115dab3423fe7a261860966128c9/rest-server/test/hdfs.js

YanjieGao:
Suggest adjust api order, adjust restful webhdfs content as the first one for user to see [min cost to operate data], list cmd contend after webhdfs

pai-management/doc/hdfs.md
```diff
@@ -91,6 +91,93 @@ curl http://NAME_NODE_ADDRESS:50070/jmx
...
+
+Data on HDFS can be accessed by various ways. Users can choose the proper way according to there needs.
+
+## HDFS Command
```
YanjieGao: should add table of contents to link to each items. As later this doc maybe linked by other docs

pai-management/doc/hdfs.md
```diff
@@ -86,16 +118,37 @@ curl http://DATA_NODE_ADDRESS:50075/jmx
...
+```bash
+curl -i -X PUT "http://hdfs-name-node-address:50070/webhdfs/v1/test/test_file?op=CREATE"
+```
+This command will return the data node where the file should be written. The location URI would be like
+>http://hdfs-name-node-address:50075/webhdfs/v1/test/test_file?op=CREATE&namenoderpcaddress=hdfs-data-node-address:9000&createflag=&createparent=true&overwrite=false
+
+Then run following command with this URI to write file data:
+```bash
+curl -i -X PUT -T file-data-to-write returned-location-uri
+```
```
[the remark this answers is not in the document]
DongZhaoYu: missing example added. done.

PR #1167, opened 2018-08-28
Zhaoyu/env support

DongZhaoYu:
support username and jobname env in codeDir, authFile, outputDir and dataDir


Gerhut:
Have you tried using regular expression and what's the problem?

DongZhaoYu:
I want to avoid to replace in names like "$PAI_USER_NAME_LIST", in this case PAI_USER_NAME_LIST should be viewed as the env name. We shouldn't replace the $PAI_USER_NAME here. 
I find it is hard to do this by regx.

Gerhut:
Perhaps `/\$PAI_USER_NAME(?![\w\d])/g` may help you, it's fast and easier to read.

You could test on <https://regexr.com/3ui82>

PR #1168, opened 2018-08-28
[PAIShare] add modal for data, docker and script

MaggieQi:
add modal for data, docker and script


shishaochen:
There are errors from TSLint at https://travis-ci.org/Microsoft/pai/jobs/421393013.

PR #1170, opened 2018-08-28
minify grafana config and exclude template files to avoid too large configmap

### LINKED ISSUE #1147: grafana failed to start with too large configmap
start grafana service failed with:

```
2018-08-24 16:26:41,135 [INFO] - paiLibrary.paiService.service_management_start : Begin to start service: [ grafana ]
2018-08-24 16:26:41,135 [INFO] - paiLibrary.paiService.service_start : Begin to execute service grafana's start script.
The ConfigMap "grafana-configuration" is invalid: metadata.annotations: Too long: must have at most 262144 characters
2018-08-24 16:26:45,624 [ERROR] - paiLibrary.common.linux_shell : Failed to execute the start script of service grafana 
```

xudifsd:
fixed #1147

with following observation:
* previous method of creating grafana configmap puts all files under grafana-configuration directory, including template files. This doubles configMap size.
* json files can be easily compressed using minify tools.

This PR reduced configMap size from 277920 byte to 61082.

### DIFF
--- a/pai-management/bootstrap/grafana/minify.py
+++ b/pai-management/bootstrap/grafana/minify.py
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env python
 
 # Copyright (c) Microsoft Corporation
 # All rights reserved.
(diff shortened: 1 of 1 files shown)


YitongFeng:
Configmap create from files and from directory seems get the same result, why you said doubles configMap size?

YitongFeng:
I am worry about that although the json file is compressed, if the number of dashboards increase in the future, the size may also be exceeded.

xudifsd:
create configMap from directory will include `.template` file. This obviously double size.

Yeah, current size is 21% of limit. We might still exceed size limit if we add 30 more json files, I expect this will not happen in short term.

We can treat this as hot fix. You can add a full-fledged solution(split into different configMaps) in future.

Issue #781, opened 2018-06-27
Failed create pod sandbox. ! Change docker default config may lead to PAI services failure.


hao1939:
Thanks for @YanjieGao , we found one cause of the issue:
1. Change docker daemon config, for example "graph" as below.
```
~$ cat /etc/docker/daemon.json
{
  "graph": "/mnt/var/lib/docker"
}
```
This will change the storage base of docker images/containers.

2. Our service container mount the path, for example `/var/log/pods` and `/var/lib/docker`.
When app in container try to access file under directory `/var/log/pods`, error would raise.

As below.
`failed to open log file "/var/log/pods/3b18fb62-79b1-11e8-baf6-000d3a1af972/hadoop-node-manager_0.log": open /var/log/pods/3b18fb62-79b1-11e8-baf6-000d3a1af972/hadoop-node-manager_0.log: no such file or directory`
![image](https://user-images.githubusercontent.com/1547343/41969265-6ce07108-7a39-11e8-86d8-51e1bbd1be41.png)

### **3. Noticed!!! `/var/log/pods/3b18fb62-79b1-11e8-baf6-000d3a1af972/hadoop-node-manager_0.log` is a symblink to a file under `/mnt/var/log/containers/`.**
But the `/mnt/var/log/containers/` was not mounted into docker!

![image](https://user-images.githubusercontent.com/1547343/41955308-16689a7a-7a12-11e8-8699-9f7c4b9b793e.png)

hao1939:
After remove the manually installed docker and install PAI again, it works.

Conclusion:

PAI is tightly coupled on Docker. ### **Please don't change any Docker default configuration!**

fanyangCS:
are we still having this issue. can we close it?

fanyangCS:
Close this issue and track it in #1338

PR #1188, opened 2018-08-28
Remove unnecessary volume of docker root dir in pai's service

ydye:


hao1939:
@ydye , per my understanding, after this change, we still don't support customized docker config?
For example, if user made the docker config below (see the #781), PAI would fail.
```
~$ cat /etc/docker/daemon.json
{
  "graph": "/mnt/var/lib/docker"
}

```

xudifsd:
ok for node-exporter removal.

ydye:
@mzmssg please check this pr, to ensure whether this change will have a side effect on your work.

mzmssg:
@ydye You can remove it first, I will test and add it back if it's necessary for docker executor.

PR #1198, opened 2018-08-29
[Paishare] Set/Get Github Pernal Access Token

yanli2017:


rest-server/src/config/user.js
```diff
@@ -46,9 +46,18 @@ const userVcUpdateInputSchema = Joi.object().keys({
     .optional(),
 }).required();
 
+// define the input schema for the 'update user github token' api
+const userGithubTokenUpdateInputSchema = Joi.object().keys({
+  githubToken: Joi.string()
```
shishaochen: How about using "Joi.token()"?

shishaochen:
I suggest using "GitHubPAT" isntead of "Token".
Because OAuth also use the "Token" concept.

shishaochen:
Please fix CI errors as well.

PR #1202, opened 2018-08-29
Add k8s retry

YitongFeng:


pai-management/k8sPaiLibrary/maintainlib/common.py
```diff
@@ -56,6 +56,23 @@ def execute_shell(shell_cmd, error_msg):
 
 
 
+def execute_shell_retry(shell_cmd, clean_cmd, error_msg):
+    count = 0
+    success_flag = False
+    while count < 5 and not success_flag:
+        try:
+            subprocess.check_call( shell_cmd, shell=True )
+            success_flag = True
```
ydye: Remove ```suceess_flag``` here. And instead of it, add ```break``` here

pai-management/k8sPaiLibrary/maintainlib/common.py
```diff
@@ -56,6 +56,23 @@ def execute_shell(shell_cmd, error_msg):
 
 
 
+def execute_shell_retry(shell_cmd, clean_cmd, error_msg):
```
ydye: Add a parameter ```retries_times``` with default value ```5```

pai-management/k8sPaiLibrary/maintainlib/common.py
```diff
@@ -56,6 +56,23 @@ def execute_shell(shell_cmd, error_msg):
...
+    count = 0
+    success_flag = False
+    while count < 5 and not success_flag:
```
ydye: Replace hardcode number with parameter.

pai-management/k8sPaiLibrary/maintainlib/common.py
```diff
@@ -56,6 +56,23 @@ def execute_shell(shell_cmd, error_msg):
...
+        try:
+            subprocess.check_call( shell_cmd, shell=True )
+            success_flag = True
+        except subprocess.CalledProcessError:
+            count += 1
+            logger.error(error_msg)
+            logger.info("%s error, retrying %d", shell_cmd, count)
+            logger.info("Cleaning failed Pods...")
```
ydye: This function could be reused in other module too. Please remove the specific infomation.
Or, you shouldn't put it in the comm.py module.

pai-management/k8sPaiLibrary/maintainlib/common.py
```diff
@@ -56,6 +56,23 @@ def execute_shell(shell_cmd, error_msg):
...
+            execute_shell_return(clean_cmd, "Can't fine the Pod, cleaning needless...")
+            if count == 5:
```
hao1939: 👍 Very readable!
One hint only, `while count < 5` and `if count == 5:` may be duplicated. Can remove one of it.

YitongFeng: I think no duplicates, it means when it failed 5 times, it will abort the deployment process.

pai-management/k8sPaiLibrary/maintainlib/deploy.py
```diff
@@ -118,8 +118,9 @@ def create_kube_proxy(self):
         generated_data = common.generate_from_template_dict(template_data, dict_map)
 
         common.write_generated_file(generated_data, "kube-proxy.yaml")
-        common.execute_shell(
+        common.execute_shell_retry(
             "kubectl apply --overwrite=true -f kube-proxy.yaml",
+            "kubectl delete -f kube-proxy.yaml",
```
ydye: A little confuse about this, if we failed to create a daemonset and the the object may not exists on k8s. Why should we still delete it?

YitongFeng: I am considering when Daemonset is created but no pod in it. 
Remind by you, I am confused that whether this condition will raise "kube-proxy created failed" error?

pai-management/k8sPaiLibrary/maintainlib/common.py
hao1939: I mean, you could `break if count == 5`, or let the `while` loop quit because `count < 5 == false`.

pai-management/k8sPaiLibrary/maintainlib/common.py
```diff
@@ -56,6 +56,21 @@ def execute_shell(shell_cmd, error_msg):
...
+        try:
+            subprocess.check_call( shell_cmd, shell=True )
+            break
+        except subprocess.CalledProcessError:
+            count += 1
+            logger.error(error_msg)
+            logger.info("%s error, retrying %d", shell_cmd, count)
+            if count == retry_count:
+                sys.exit(1)
+
```
ydye: Add a sleep time here.

YitongFeng: Resolved

Issue #1153, opened 2018-08-27
`/paictl.py cluster k8s-bootup -p /cluster-configuration` randomly failed


hao1939:
The k8s-bootup randomly failed because of `kube-proxy`.

> ...
> 2018-08-27 04:10:31,663 [INFO] - k8sPaiLibrary.maintainlib.kubectl_install : Generate the configuation file of kubectl.
> NAME       STATUS     ROLES     AGE       VERSION
> 10.0.1.5   Ready      <none>    21s       v1.9.4
> 10.0.1.7   NotReady   <none>    4s        v1.9.4
> 10.0.1.8   Ready      <none>    20s       v1.9.4
> 2018-08-27 04:10:35,939 [INFO] - k8sPaiLibrary.maintainlib.kubectl_install : Successfully install kubectl and configure it!
> 2018-08-27 04:10:35,939 [INFO] - k8sPaiLibrary.maintainlib.deploy : Create kube-proxy daemon for kuberentes cluster.
> error: unable to recognize "kube-proxy.yaml": no matches for kind "DaemonSet" in version "apps/v1"
> 2018-08-27 04:10:36,241 [ERROR] - k8sPaiLibrary.maintainlib.common : Failed to create kube-proxy

YitongFeng:
@fanyangCS OK, others are closed but can't be deleted. Reference of the similar issues is for easy tracking.

mzmssg:
update in #1166

YitongFeng:
For no matches for kind "DaemonSet" in version "apps/v1", it maybe the k8s API version's problem. 
Our k8s version is 1.9, refer to [k8s 1.9 API docs](https://v1-9.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.9/#daemonset-v1-apps), it shows that v1.9 DaemontSet support "apps/v1", "apps/v1beta1" and "apps/v1beta2". The "beta" is testing version, if this version runs well k8s group will upgrade it to "apps/v1" and marked as stable version.

Our single-node k8s version is 1.9.4, maybe can't totally support the stable version "apps/v1", #1126 change the version to 1.9.9, and have not appear the error now. Or you can use "apps/v1beta1" and "apps/v1beta2" in your daemonset yaml file.

hao1939:
More details:

- In the dev-box which raise `error: unable to recognize "kube-proxy.yaml": no matches for kind "DaemonSet" in version "apps/v1beta2"`:
[...]

PR #1210, opened 2018-08-30
check api resources before install proxy

YitongFeng:
Related issue #1153

### DIFF
--- a/pai-management/k8sPaiLibrary/maintainlib/deploy.py
+++ b/pai-management/k8sPaiLibrary/maintainlib/deploy.py
@@ -203,6 +203,9 @@ class deploy:
         kubectl_install_instance = kubectl_install.kubectl_install(self.cluster_config)
         kubectl_install_instance.run()
 
+        # check the registerd api resources
+        common.execute_shell("kubectl api-resources", "kubectl command failed!")
+
         self.create_kube_proxy()
         self.create_k8s_dashboard()
 
--- a/pai-management/k8sPaiLibrary/maintainlib/deploy.py
+++ b/pai-management/k8sPaiLibrary/maintainlib/deploy.py
@@ -204,7 +204,10 @@ class deploy:
         kubectl_install_instance.run()
 
         # check the registerd api resources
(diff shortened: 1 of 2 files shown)


pai-management/k8sPaiLibrary/maintainlib/deploy.py
```diff
@@ -197,6 +197,9 @@ def run(self):
         kubectl_install_instance = kubectl_install.kubectl_install(self.cluster_config)
         kubectl_install_instance.run()
 
+        # check the registerd api resources
+        common.execute_shell("kubectl api-resources", "kubectl command failed!")
```
ydye: Maybe you should retry util successful

YitongFeng: Good point

pai-management/k8sPaiLibrary/maintainlib/deploy.py
```diff
@@ -203,6 +203,12 @@ def run(self):
         kubectl_install_instance = kubectl_install.kubectl_install(self.cluster_config)
         kubectl_install_instance.run()
 
+        # check the registerd api resources
+        common.execute_shell_retry("kubectl api-resources", "kubectl command failed!", 5)
+
+        # create kube-proxy until daemonset resource is registerd
+        common.execute_shell_retry("kubectl api-resources | grep -q daemonsets", "Controller manager hasn't create daemonset object!", 5)
```
hao1939: 1. Will `kubectl api-resources | grep -q daemonsets`  return none zero when daemonsets not ready?
2. Does the grep -q case-sensitive?

YitongFeng: yes, only when it grep nothing it return 0, I have test

YitongFeng: kubectl api-resources results are all lowercase, the "grep" is case-sensitive

hao1939:
Fixes #1153

PR #1220, opened 2018-08-30
Handle UnicodeEncodeError of convert unicode string to ascii string

shawntian90:
`docker_username` and `docker_password` may be unicode string.

### DIFF
--- a/pai-management/paictl.py
+++ b/pai-management/paictl.py
@@ -87,12 +87,21 @@ def login_docker_registry(docker_registry, docker_username, docker_password):
     logger.info("docker registry login successfully")
 
 
+def to_str(value):
+    try:
+        return str(value)
+    except UnicodeEncodeError:
+        return str(value.encode("utf8"))
+    except:
+        raise
+
+
 
 def generate_secret_base64code(docker_info):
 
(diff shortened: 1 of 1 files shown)


hao1939:
Hi @shawntian90 ,

Could you share the background of the PR?
Are you using unicode characters in the configuration file?

shawntian90:
@hao1939 

I wanna to enable my private docker registry, which username is unicode string, such as "容器管理员".

There will be an exception `UnicodeEncodeError` when run commands like `python paictl.py` inside dev-box.

xudifsd:
similar to #809 

I suggest change all `with open("foo") as f` to `with `[`codecs.open`](https://docs.python.org/2/library/codecs.html#codecs.open)`("foo", "utf-8")` should also change output method too.

It's better to search keyword `open` to find input/output calls.

pai-management/paictl.py
```diff
@@ -87,12 +87,21 @@ def login_docker_registry(docker_registry, docker_username, docker_password):
     logger.info("docker registry login successfully")
 
 
+def to_str(value):
```
ydye: Can you move this function to the common module?

shawntian90:
@xudifsd @ydye 
I will make a patch  to all `open` statement and read configuration file via `codecs.open` this week.

PR #1088, opened 2018-08-16
[Deployment] Secure kubelet public api

abuccts:
Secure kubelet public api, add authentication for https port 10250 and close http port 10255.

Please refer to https://github.com/kubernetes/kubernetes/issues/7965, https://github.com/kubernetes/kubernetes/pull/59666 for reasons and https://github.com/docker/for-linux/issues/324, [Backdooring through kubelet](https://medium.com/handy-tech/analysis-of-a-kubernetes-hack-backdooring-through-kubelet-823be5c3d67c) for hacking examples.

PR #1222, opened 2018-08-30
secure kubelet public api, close http port, add auth to https port

xudifsd:
cherry-picked from #1088
 
also since that PR closed http port of kubelet, watchdog can not read metrics from kubelet anymore, remove from watchdog.


xudifsd:
already deployed in INT BED, and checked can not hack using method provided by [this](https://medium.com/handy-tech/analysis-of-a-kubernetes-hack-backdooring-through-kubelet-823be5c3d67c).

PR #1230, opened 2018-08-31
Add support for random docker root dir for kubelet

ydye:


pai-management/k8sPaiLibrary/template/kubelet.sh.template
```diff
@@ -19,13 +19,18 @@
 
 set -e
 
+
+DOCKER_ROOT_DIR_FOR_KUBELET=`docker info | grep "Docker Root Dir" | awk -F' ' '{print $NF}'`
```
hao1939: Hi @ydye ,
Could we using Docker client api for this purpose?

ydye: What's the benefit?

hao1939: API is more readable and compatible.
Shell script heavily dependent on the command line output format, it's hard to assume the result without execute it.

ydye: But we will have to install docker-client into the remote host.

hao1939: I see.
In case that docker client could connect remote daemon, we may not need to execute script on remote host.

ydye: This will introduce another dependency, and make deployment more complicated

YitongFeng: The command `docker info` on my VM or in the dev-box always show a warning `no swap limit support`, so I think use python api is better:
image

ydye: This message is okay.

PR #1232, opened 2018-08-31
[PAIShare] Add local cache for GitHub access.

shishaochen:


rest-server/src/controllers/template.js
```diff
@@ -93,20 +155,30 @@ const share = (req, res) => {
       'message': 'Failed to parse template content.',
     });
   }
-  template.save(type, name, item, req.body.contributor, function(err, saved) {
+  let account = req.user.name;
```
yanli2017: change to 'let account = req.user.username;'

rest-server/src/controllers/template.js
```diff
@@ -93,20 +155,30 @@ const share = (req, res) => {
...
+  console.log(`${account} is tring to share template.`);
```
yanli2017: logger.info ?

PR #1282, opened 2018-09-05
change email template

xudifsd:
Change alert email template.

The default email template contains many info not useful even obscure to end users, for example, `instance` label is generated by prometheus master which represent where this metrics scraped from, not problematic service's IP.

This PR changed email template for OpenPAI, only display alert's summary if applicable, and fallback to original version if summary not exist.

image

I copied default template, and only changed logic with [this](https://github.com/Microsoft/pai/pull/1282/files#diff-c0a60ee34697d1a458ef7fd020e41020R70) if branch.


xudifsd:
@YitongFeng In the last commit, I changed you template generation logic a little. If template file contains nothing, return original content.

YanjieGao:
could you give us a eample of "only display alert's summary" screenshot? the image above seems no summary ?

xudifsd:
@YanjieGao for service not ready alert, it looks like:

image

PR #1289, opened 2018-09-05
Add feedback with current version

### LINKED ISSUE #1122: the release version of the current deployment
How does an operator know which pai version the cluster runs on? Can you show version number somewhere in web portal?

Gerhut:
Closed #1122 

image

When clicking the "Feedback", a GitHub "new issue" page will be opened with version label already set, since the labels of an issue are not easier to be removed by the issuer than the title or body. However, the label of version must be created before, otherwise, there will be no label assigned automatically.

image

I edit the `version` field in package info of REST server and web portal to match the version of the system, does pom.xml of launcher also need? @yqwang-ms 

There will be 2 more steps in the release process:

1. Modify the version field in the `package.json` of REST server and web portal.
2. Create the version label on GitHub issue.

Theoretically, a Travis validation task could be added on `pai-v`-prefixed branch to ensure them.


rest-server/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "pai-rest-server",
-  "version": "0.1.0",
+  "version": "0.8.0",
```
fanyangCS: I will approve the PR first. 
is it possible to get the version automatically from the release branch name?

Gerhut: @fanyangCS Currently there is no third part "y" in branch name "pai-0.7.y", so we can only *validate* the version field on release branches but cannot auto *generate* it.

BTW, I used `git describe HEAD` as temporary version names in some early practices, which includes latest tag, commits count since latest tag and current commit hash. Unfortunately, it is not match our git flow, a little, since our release tags is not on the master branch.

Our `git describe --tags master` now looks like `v0.5.0-822-ga17f1548`, but `git describe --tags v0.7.1` is `v0.7.1`(add `--tags` because we are using lightweight tags but not annotated tags for release versions), does that look fine?

PR #1292, opened 2018-09-06
Zhaoyu/disk cleaner/clean docker cache

DongZhaoYu:
1. definition for condition/action/rule
2. executor of rules
3. docker cache clean scripts


cleaner/cleaner_main.py
```diff
@@ -0,0 +1,65 @@
...
+            self.logger.info("worker %s started.", k)
+
+    def sync(self):
+        try:
+            for w in self.workers.values():
+                w.join()
+        except:
+            self.logger.error("cleaner interrupted and will exit.")
+            for w in self.workers.values():
+                w.terminate()
```
sterowang: >w.terminate() [](start = 16, length = 13)

Suggest to add try/except here also so that if some worker's terminate() failed, we can continue on to the next.

cleaner/worker.py
```diff
@@ -0,0 +1,66 @@
...
+        exc = None
+        method_name = self.method.__name__
+        try:
+            self.logger.info("start to execute method %s.", method_name)
+            with CountdownTimer(duration=self.timeout):
+                self.method(self.arg)
+        except Timeout as e:
+            self.logger.error("command %s timeout.", method_name)
+            exc = e
+        except Exception as e:
```
sterowang: Shall we force to kill the timed out command here?

cleaner/scripts/reclaimable_docker_cache.sh
```diff
@@ -0,0 +1,23 @@
...
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+# DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+#!/bin/bash
+
+docker system df | \
+gawk 'BEGIN {s=0}
+      END {print s}
+      match($5, /([0-9]+\.?[0-9]*)/, a) {if(a[2] == "M") s += a[1]/1024; else s += a[1]}'
```
sterowang: >match($5, /([0-9]+\.?[0-9]*)/, a) {if(a[2] == "M") s += a[1]/1024; else s += a[1]}' [](start = 6, length = 83)

Can you explain a little bit of the rule a little bit here? For example to give some examples?

cleaner/utils/logger.py
```diff
@@ -0,0 +1,33 @@
...
+# THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING
+# BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+# DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+import multiprocessing
+
+
+class LoggerMixin(object):
```
sterowang: Can you explain the purpose of this utility as comments?

cleaner/cleaner_main.py
DongZhaoYu: add a new method cleaner.terminate() to kill a worker and all its subprocesses.
This method will catch all exceptions when killing each worker.

cleaner/worker.py
DongZhaoYu: Yes. But the better way to kill the command is in common.run_cmd method. I have catch the exception in this method and terminate the command when timeout or other errors occur.

cleaner/utils/logger.py
DongZhaoYu: Yes, the explanation is added to the comments.

cleaner/cleaner_main.py
```diff
@@ -0,0 +1,73 @@
...
+    def start(self):
+        for k, w in self.workers.items():
+            w.start()
+            self.logger.info("worker %s started.", k)
+
+    def terminate(self):
+        for k, w in self.workers.items():
+            try:
+                # terminate the worker and all its subprocesses
+                common.kill_process_tree(w.pid, 1, self.logger)
```
sterowang: >1 [](start = 48, length = 1)

Is 1 second enough? How about 5 seconds to give more graceful time to the processes being killed?

cleaner/scripts/reclaimable_docker_cache.sh
sterowang: I saw GB also there, does your code handle it?

---
In reply to: [215850149](https://github.com/Microsoft/pai/pull/1292#discussion_r215850149) [](ancestors = 215850149)

cleaner/scripts/reclaimable_docker_cache.sh
```diff
@@ -0,0 +1,41 @@
...
+#
+# We summer up the result in column 5 (RECLAIMABLE) and return the size in gigabytes.
+
+docker system df | \
+gawk 'BEGIN {s=0}
+      END {print s}
+      match($5, /([0-9]+\.?[0-9]*)(M|G|B)/, a) {
+          if(a[2] == "M")
+              s += a[1]/1024;
+          else if(a[2] == "B")
```
sterowang: Should be "G"?

Issue #1295, opened 2018-09-06
Is it possible to deploy it in AWS ec2?


liudongqing:
I tried and found it must use ssh username and password, but looks like EC2 doesn't allow that, do you have the plan to support ssh with pem file?

liudongqing:
I modified the code of common.py:
`ssh.connect(hostname=hostip, port=port, username=username, key_filename="my_keyfile")`
adn 
` k = paramiko.RSAKey.from_private_key_file('id_rsa')
   transport.connect(username=username, pkey=k)`

Looks like it works.

fanyangCS:
we welcome contribution.

liudongqing:
@ydye, I created a PR, could you please help to review it?  also could you please help me to check the CI result? It failed but I don't think my change has the impact on that part. 

Thanks!

fanyangCS:
https://github.com/Microsoft/pai/issues/825

fanyangCS:
@liudongqing is this the only blocking issue on EC2?

PR #1308, opened 2018-09-08
Support ssh auth with private key file to enable to deploy pai to AWS…

liudongqing:
… EC2 (#1295)

### DIFF
--- a/pai-management/doc/cluster-bootup.md
+++ b/pai-management/doc/cluster-bootup.md
@@ -141,7 +141,9 @@ machines:
 # that the username/password pair or username/key-filename is valid and has sudo privilege.
 ssh-username: pai
 ssh-password: pai-password
-ssh-key-filename: key-filename
+
+# (Optional, default=None) the key file that ssh client uses, that has higher priority then password.
+#ssh-keyfile-path: <keyfile-path>
 
 # (Optional, default=22) Port number of ssh service on each machine.
 #ssh-port: 22
--- a/pai-management/doc/cluster-bootup.md
+++ b/pai-management/doc/cluster-bootup.md
@@ -138,9 +138,10 @@ machines:
   - 192.168.1.13
 
 # (Required) Log-in info of all machines. System administrator should guarantee
-# that the username/password pair is valid and has sudo privilege.
+# that the username/password pair or username/key-filename is valid and has sudo privilege.
 ssh-username: pai
 ssh-password: pai-password
+ssh-key-filename: key-filename
 
 # (Optional, default=22) Port number of ssh service on each machine.
 #ssh-port: 22
@@ -337,9 +338,10 @@ machines:
   - 192.168.1.11
 
 # (Required) Log-in info of all machines. System administrator should guarantee
-# that the username/password pair is valid and has sudo privilege.
+# that the username/password pair or username/key-filename is valid and has sudo privilege.
 ssh-username: pai
 ssh-password: pai-password
+ssh-key-filename: key-filename
 
 # (Optional, default=22) Port number of ssh service on each machine.
 #ssh-port: 22
--- a/pai-management/k8sPaiLibrary/maintainlib/common.py
+++ b/pai-management/k8sPaiLibrary/maintainlib/common.py
@@ -167,11 +167,16 @@ def sftp_paramiko(src, dst, filename, host_config):
         if port_validation(host_config['sshport']) == False:
             return False
         port = int(host_config['sshport'])
-
+    key_filename = None
+    if 'key-filename' in host_config:
+        if os.path.isfile(str(host_config['key-filename'])):
+            key_filename = str(host_config['key-filename'])
+        else:
+            logger.warn("The key file: {0} specified doesn't exist".format(host_config['key-filename']))
     # First make sure the folder exist.
     ssh = paramiko.SSHClient()
     ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
-    ssh.connect(hostname=hostip, port=port, username=username, password=password)
+    ssh.connect(hostname=hostip, port=port, key_filename=key_filename, username=username, password=password)
 
     stdin, stdout, stderr = ssh.exec_command("sudo mkdir -p {0}".format(dst), get_pty=True)
     stdin.write(password + '\n')
@@ -183,7 +188,10 @@ def sftp_paramiko(src, dst, filename, host_config):
 
     # Put the file to target Path.
     transport = paramiko.Transport((hostip, port))
-    transport.connect(username=username, password=password)
+    pkey = None
+    if key_filename is not None:
+        pkey = paramiko.RSAKey.from_private_key_file(key_filename)
+    transport.connect(username=username, pkey=pkey, password=password)
 
     sftp = paramiko.SFTPClient.from_transport(transport)
     sftp.put('{0}/{1}'.format(src, filename), '{0}/{1}'.format(dst, filename))
@@ -216,9 +224,15 @@ def ssh_shell_paramiko_with_result(host_config, commandline):
         if port_validation(host_config['sshport']) == False:
             return (None, None)
         port = int(host_config['sshport'])
+    key_filename = None
+    if 'key-filename' in host_config:
+        if os.path.isfile(str(host_config['key-filename'])):
+            key_filename = str(host_config['key-filename'])
+        else:
+            logger.warn("The key file: {0} specified doesn't exist".format(host_config['key-filename']))
     ssh = paramiko.SSHClient()
     ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
-    ssh.connect(hostname=hostip, port=port, username=username, password=password)
+    ssh.connect(hostname=hostip, port=port, key_filename=key_filename, username=username, password=password)
     stdin, stdout, stderr = ssh.exec_command(commandline, get_pty=True)
     logger.info("Executing the command on host [{0}]: {1}".format(hostip, commandline))
     result_stdout = ""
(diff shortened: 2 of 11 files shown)


ydye:
@hao1939

pai-management/doc/cluster-bootup.md
```diff
@@ -337,9 +338,10 @@ machines:
   - 192.168.1.11
 
 # (Required) Log-in info of all machines. System administrator should guarantee
-# that the username/password pair is valid and has sudo privilege.
+# that the username/password pair or username/key-filename is valid and has sudo privilege.
 ssh-username: pai
 ssh-password: pai-password
+ssh-key-filename: key-filename
```
ydye: Should be a path to the keyfile？

pai-management/doc/cluster-bootup.md
```diff
@@ -138,9 +138,10 @@ machines:
   - 192.168.1.13
 
 # (Required) Log-in info of all machines. System administrator should guarantee
-# that the username/password pair is valid and has sudo privilege.
+# that the username/password pair or username/key-filename is valid and has sudo privilege.
 ssh-username: pai
 ssh-password: pai-password
+ssh-key-filename: key-filename
```
ydye: Should be a path to the keyfile？

pai-management/k8sPaiLibrary/maintainlib/common.py
```diff
@@ -216,9 +224,15 @@ def ssh_shell_paramiko_with_result(host_config, commandline):
         if port_validation(host_config['sshport']) == False:
             return (None, None)
         port = int(host_config['sshport'])
+    key_filename = None
+    if 'key-filename' in host_config and host_config['key-filename'] is not None:
+        if os.path.isfile(str(host_config['key-filename'])):
+            key_filename = str(host_config['key-filename'])
+        else:
+            logger.warn("The key file: {0} specified doesn't exist".format(host_config['key-filename']))
```
ydye: If the admin configure the field ```ssh-key-filename```  or ```key_filename```, but paictl failed to find it. Is it better to exit from script?

hao1939:
Hi @liudongqing ,

Noticed your PR is `out-of-data with the base branch`, could you rebase on the master branch?

ydye:
We'd better add the logic for user to choose.
They can choose ```password``` or ```key_filename``` to set. And they must set one of them. For another one, it's optional. 
And in current code base, the password is the necessary field. And the ```key_filename``` is optional. I think we can make it more flexible

prometheus/exporter/watchdog.py
```diff
@@ -125,12 +125,18 @@ def ssh_exec(host_config, command, histogram=ssh_histogram):
         port = 22
         if "sshport" in host_config:
             port = int(host_config["sshport"])
+        key_filename = None
```
xudifsd: Well, watchdog have different runtime environment compare to paictl.py: watchdog is running in container and pai-master node. The key-filename exist in host executing paictl.py may not necessary exist in watchdog container.

You should maybe copy the key-filename into configMap like `pai-management/bootstrap/watchdog/watchdog-configmap.yaml.template` and then mount into watchdog container

pai-management/doc/cluster-bootup.md
liudongqing: Yes, it is a path to keyfile, Maybe better to name it to  keyfile-path?

pai-management/doc/cluster-bootup.md
liudongqing: same as above

pai-management/k8sPaiLibrary/maintainlib/common.py
liudongqing: I am not sure the principle of PAI development. If user provide both keyfile and password, but keyfile paht is wrong, maybe we should try password as well?

hao1939:
Hi @liudongqing ,

Your PR can't pass CI build, it's because that CI scripts make some config change hardcoded.
Could you make it to work with your PR?

Here's code change needed:
https://github.com/Microsoft/pai/blob/a59f05fe32934dd1164c26caafc0676f0763b692/Jenkinsfile#L82
https://github.com/Microsoft/pai/blob/a59f05fe32934dd1164c26caafc0676f0763b692/Jenkinsfile#L172
https://github.com/Microsoft/pai/blob/a59f05fe32934dd1164c26caafc0676f0763b692/Jenkinsfile#L276

prometheus/exporter/watchdog.py
liudongqing: Thanks for the information, I will try your suggestions.

pai-management/k8sPaiLibrary/maintainlib/common.py
```diff
@@ -167,11 +167,16 @@ def sftp_paramiko(src, dst, filename, host_config):
...
+    if 'key-filename' in host_config:
+        if os.path.isfile(str(host_config['key-filename'])) and host_config['key-filename'] is not None:
+            key_filename = str(host_config['key-filename'])
+        else:
+            logger.warn("The key file: {0} specified doesn't exist".format(host_config['key-filename']))
     # First make sure the folder exist.
     ssh = paramiko.SSHClient()
     ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
-    ssh.connect(hostname=hostip, port=port, username=username, password=password)
+    ssh.connect(hostname=hostip, port=port, key_filename=key_filename, username=username, password=password)
```
sterowang: >key_filename=key_filename, username=username, password=password [](start = 44, length = 63)

If both key and password exists, which one has higher priority?

ydye:
```
Authentication is attempted in the following order of priority:

The pkey or key_filename passed in (if any)
key_filename may contain OpenSSH public certificate paths as well as regular private-key paths; when files ending in -cert.pub are found, they are assumed to match a private key, and both components will be loaded. (The private key itself does not need to be listed in key_filename for this to occur - just the certificate.)
Any key we can find through an SSH agent
Any “id_rsa”, “id_dsa” or “id_ecdsa” key discoverable in ~/.ssh/
When OpenSSH-style public certificates exist that match an existing such private key (so e.g. one has id_rsa and id_rsa-cert.pub) the certificate will be loaded alongside the private key and used for authentication.
Plain username/password auth, if a password was given
```

REF: http://docs.paramiko.org/en/2.4/api/client.html

pai-management/doc/cluster-bootup.md
ydye: Yes, ```ssh-keyfile-path``` make sense.

pai-management/k8sPaiLibrary/maintainlib/common.py
ydye: @hao1939  What's your opinion?

prometheus/exporter/watchdog.py
liudongqing: use kubernetes secret to achieve this.

liudongqing:
@hao1939 rebased. And removed the added line from cluster-configuration.yaml to pass CI. As ssh-keyfile-path is an optional configuration, I think it is OK for its absence in quick-start-example.yaml

hao1939:
Hi @liudongqing ,

Yeah, agree with you, `ssh-keyfile-path` could be optional.

Now, the  ` ./paictl.py config generate` is broken, could you fix it?

```
+ ./paictl.py config generate -i /var/lib/jenkins/bed1/singlebox/quick-start/quick-start.yaml -o /var/lib/jenkins/bed1/singlebox/cluster-configuration
  File "./paictl.py", line 639, in <module>
    main(sys.argv[1:])
  File "./paictl.py", line 634, in main
    args.handler(args)
  File "./paictl.py", line 603, in generate_configuration
    args.force)
  File "/mnt/jenkins/workspace/pai_PR-1308-U3ZJB3VWLXGSM2FZ6D3H7KOYIUH6QB3RAS3JQPEYATIFBYYP26NA/pai-management/paiLibrary/paiCluster/cluster_util.py", line 80, in generate_configuration
    "key-filename": quick_start_config["ssh-keyfile-path"],
KeyError: 'ssh-keyfile-path'
script returned exit code 1
```

prometheus/exporter/watchdog.py
```diff
@@ -125,12 +125,18 @@ def ssh_exec(host_config, command, histogram=ssh_histogram):
...
+        if 'keyfile_path' in host_config:
+            if os.path.isfile(str(host_config['keyfile_path'])) and host_config['keyfile_path'] is not None:
+                key_filename = str(host_config['keyfile_path'])
+            else:
+                logger.warn("The key file: {0} specified doesn't exist".format(host_config['keyfile_path']))
```
xudifsd: please use `logger.error` and increase `error_counter` like `error_counter.labels(type="ssh_key_file").inc()`

pai-management/quick-start/services-configuration.yaml.template
```diff
@@ -117,6 +117,10 @@ prometheus:
   #    smtp_auth_username: user@gmail.com
   #    smtp_auth_password: gmail_password
 
+  # if you want to use key file to login nodes:
+  # 1. create secret with command 'kubectl create secret generic ssh-key-secret --from-file=<secret name>=/path/to/keyfile'
```
xudifsd: Please add `kubectl create secret xxx` command in `pai-management/bootstrap/prometheus/start.sh` to create secret instead of requiring user read the comments and create secret.

In that case, you should change `pai-management/bootstrap/prometheus/start.sh` to `pai-management/bootstrap/prometheus/start.sh.template` and add entry `start.sh` in `template-list field` of `pai-management/bootstrap/prometheus/service.yaml`, since not all users need to create that secret.

hao1939:
Hi @liudongqing , The PR has already passed the CI.

prometheus/exporter/watchdog.py
```diff
@@ -125,12 +125,18 @@ def ssh_exec(host_config, command, histogram=ssh_histogram):
         port = 22
         if "sshport" in host_config:
             port = int(host_config["sshport"])
+        key_filename = None
+        if 'keyfile-path' in host_config:
+            if os.path.isfile(str(host_config['keyfile-path'])) and host_config['keyfile-path'] is not None:
+                key_filename = str(host_config['keyfile-path'])
+            else:
+                logger.warn("The key file: {0} specified doesn't exist".format(host_config['keyfile-path']))
```
xudifsd: please use `logger.error` and increase `error_counter` like `error_counter.labels(type="ssh_key_file").inc()`

xudifsd:
We are planning to freeze master code change in several days to do large code structure change. To minimize impact on this PR, we decided to merge this PR and resolve possible conflict internally.

PR #1331, opened 2018-09-11
New datascientist onboard process doc

YitongFeng:
I think a "quick start" here is reasonable, for someone want to run example to get familiar with PAI first, someone want to customize their own job, each take what he need.

### DIFF
--- a/docs/datascientist_onboard.md
+++ b/docs/datascientist_onboard.md
@@ -11,8 +11,8 @@
 
 ## Prerequisites
 This document assumes:
-	• The PAI system has already been deployed properly and a docker registry is available to store docker images. 
-	• You have done the debug of your code on your local machine to ensure it works well.
+- The PAI system has already been deployed properly and a docker registry is available to store docker images. 
+- You have done the debug of your code on your local machine to ensure it works well.
 
 
 ## Basic Workflow 
@@ -56,7 +56,8 @@ We assumes that you've prepared your code on your local machine and done the deb
 #### 1.2 Operate HDFS data
 **Note:** The relative positions of your code and the data downloaded into the container may be changed, debug your code and make sure it can get access to the dataset.
 
-##### 1.2.1 Small dataset
+**1.2.1 Small dataset**
+
 If your data size is small, you have the following ways to handle your code and data:
 
 1) You can use HDFS commands in your code to download data to the container when it is launched.
@@ -97,11 +98,11 @@ You can specify the download commands in [job JSON configuration file](#3.-Write
     [HDFS mount](https://github.com/Microsoft/hdfs-mount.git) is a tool to mount HDFS as a local Linux file system. Use CNTK job for example, to use this tool, firstly install it in your job container, refer to [CNTK job container Dockerfile](https://github.com/Microsoft/pai/blob/45fb9385db65538d16b5feb2929e67773bc0eeb8/examples/cntk/Dockerfile.example.cntk#L36). Excecute `hdfs-mount` command in your code, e.g. [CNTK job's start script](https://github.com/Microsoft/pai/blob/45fb9385db65538d16b5feb2929e67773bc0eeb8/examples/cntk/cntk-g2p.sh#L31), then you can use the mounted HDFS directory as local file system.
 
 
-**1.2.2 Large dataset or amount of smallfiles**
+**1.2.2 Large dataset or amount of small files**
 
 If your dataset is large, solutions in the above section may get worse performance, this section provide some solutions to read HDFS data directlly.
 
-If your dataset consist of smallfiles, it can cause namenode memory management problem and compute framework performance problem. People usually pack multiple small files and download it in batches to mitigate this issue.
+If your dataset consist of small files, it can cause namenode memory management problem and compute framework performance problem. People usually pack multiple small files and download it in batches to mitigate this issue.
 
 **Pack data solutions:**
 - User compact small files through scripts or opensources tools. (For example using TFRecord for Tensorflow.)
--- a/docs/datascientist_onboard.md
+++ b/docs/datascientist_onboard.md
@@ -1,0 +1,227 @@
+[TOC]
+
+# Getting Started with PAI
+
+## Overview
+- OpenPAI is an open source platform that provides complete AI model training and resource management capabilities. 
+- It supports multiple types of jobs including Deep Learning, Machine Learning, Big Data, etc. 
+- It runs jobs in Docker containers and provides container schedule mechanism and user-friendly monitor system.
+- It use distributed file system HDFS to store and manage data.
+- For Deep Learning jobs, OpenPAI supports all the DL frameworks (including CNTK, TensorFlow, etc.) as long as you install the corresponding packages in your customized job container.  
+- Users need to prepare a config file in JSON format and submit it for a job submission 
+
+
+## Prerequisites
+This document assumes:
+	• The PAI system has already been deployed properly and a docker registry is available to store docker images. 
+	• You have done the debug of your code on your local machine to ensure it works well.
+
+
+## Basic Workflow 
+
+To run a job on PAI, you should follow the follow the following basic workflow:
+1. Prepare your data and code, upload them to HDFS.
+2. Prepare your job container
+3. Write a job JSON configuration file
+4. Register an account, then submit job
+5. Debug job
+	
+	
+## Quick start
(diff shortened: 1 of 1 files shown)


docs/datascientist_onboard.md
```diff
@@ -0,0 +1,230 @@
...
+4. Register an account, then submit job
+5. Debug job
+	
+	
+## Quick start
+This part will guide you getting started  from running our example jobs  by submitting the job's json configuration file straightly without any modification. Thus you don't need to prepare the code, data and job container, which means you just need to go with 4, 5 of Basic Workflow.
+
+- [Quick start with submitting a CIFAR-10 job](https://github.com/Microsoft/pai/blob/master/examples/README.md#quickstart)
+- [Learn more example jobs](https://github.com/Microsoft/pai/tree/master/examples#offtheshelf)
+
```
mzmssg: If the CIFAR-10 example could be run end-to-end and break down step-by-step, rather than a ready-made json, it would be more understandable for me. For people who have no domain knowledge about PAI and docker(i.e. me when check in ), even I submit such an example, I don't know what happened and how could I run my job on PAI. During my DRI, some user asks me the similar question.

docs/datascientist_onboard.md
```diff
@@ -0,0 +1,230 @@
...
+    **Note:** We don't recommend to pack code into container unless your code won't change. Because every time you change your code you will build a new image. You can upload your code to HDFS, then specify the HDFS url of your code in [job JSON configuration file](#3.-Write-a-job-json-configuration-file) `codedir` field, PAI will automatically download it to the job container's path `/pai/code/`.
+
+
+3) Use HDFS mount.    
+    [HDFS mount](https://github.com/Microsoft/hdfs-mount.git) is a tool to mount HDFS as a local Linux file system. Use CNTK job for example, to use this tool, firstly install it in your job container, refer to [CNTK job container Dockerfile](https://github.com/Microsoft/pai/blob/45fb9385db65538d16b5feb2929e67773bc0eeb8/examples/cntk/Dockerfile.example.cntk#L36). Excecute `hdfs-mount` command in your code, e.g. [CNTK job's start script](https://github.com/Microsoft/pai/blob/45fb9385db65538d16b5feb2929e67773bc0eeb8/examples/cntk/cntk-g2p.sh#L31), then you can use the mounted HDFS directory as local file system.
+
+
+**1.2.2 Large dataset or amount of small files**
```
mzmssg: Would it be better of putting hdfs performance improvement under a  `Advanced Usage` section? As a new user, the first stage is "how to run my work on PAI", after going through whole workflow, then "how to run it better" seems to be a question.

docs/datascientist_onboard.md
```diff
@@ -0,0 +1,230 @@
...
+For your convenience, we have pre-built some images for you:
+  - openpai/pai.example.tensorflow
+  - openpai/pai.example.tensorflow-serving
+  - openpai/pai.example.keras.tensorflow
+  - openpai/pai.example.sklearn
+  - openpai/pai.example.pytorch
+  - openpai/pai.example.kafka
+  - openpai/pai.example.mxnet
+  - openpai/pai.example.caffe2
+You can find their Dockerfile in the corresponding directory of [PAI job examples](https://github.com/Microsoft/pai/tree/master/examples).
```
mzmssg: I think we could add some tutorial here to teach scientists how to debug their container locally, avoiding a bad usage experience at the beginning, although we can't hide the docker details completely. 

Eg: Jobs may fail due to some little bugs(typo, path errors), and it would take scientists much time to fix them (finding bugs, fixing, rebuilding, pushing, pulling), especially in the case of users packing data into container for convenience. In my experience, testing such containers locally before submitting is very helpful. If `docker run -v ...:codeDir -v ...:dataDir -v ...:outputDir image_name /bin/bash -c command ` works well locally, then such job should success in PAI as well.

docs/datascientist_onboard.md
```diff
@@ -0,0 +1,230 @@
...
+Please refer to [debug job document](https://github.com/Microsoft/pai/blob/master/docs/job_tutorial.md#how-to-debug-the-job-) for details.
+	
+## Getting started by scenarios [TODO]
+	1. Trainning a Model
+	Training your model on PAI
+	
+	2. Serving a Model
+	Run your model as a long-running service on PAI.
+	
+	3. Model Prediction
```
mzmssg: I think few people would run a separate inference job on PAI. But another scenario is model evaluation, which is also a long-run service, accepting a model, running it on a given testset and scoring the model. A little difference from model serving, model evaluation only uses gpu as model coming, but model serving takes up GPU memory all the time, to avoid load model everytime.

docs/datascientist_onboard.md
YitongFeng: How about the data is a a bit large? Such as 20G data, not big but also not small, download into the container all at once will take longtime, pack into container makes the pulling image time long, and these two methods users will do a download everytime he runs a job; or use hdfs mount, it is low efficient.

docs/datascientist_onboard.md
YitongFeng: I think taking GPUs or not is DL framework's things, not PAI's feature. So I suggest to just split into two types: train a model and serving a model. 
As train a model seems much duplicate with the current doc, I think just add a serving a Model doc is reasonable.

docs/datascientist_onboard.md
```diff
@@ -0,0 +1,230 @@
...
+### 4. Register an account, then submit job
+Now you are ready to summit a job, find your admin to register a username and password, open PAI homepage, login to summit your job, refer to [Submit a job in Web Portal](https://github.com/Microsoft/pai/blob/master/docs/submit_from_webportal.md) for more details.
+
+Other submit ways:
+- [Submit a job in Visual Studio](https://github.com/Microsoft/vs-tools-for-ai/blob/master/docs/pai.md)
+
+- [Submit a job in Visual Studio Code](https://github.com/Microsoft/vscode-tools-for-ai/blob/master/docs/quickstart-05-pai.md)
+
+### 5. Debug job 
+Please refer to [debug job document](https://github.com/Microsoft/pai/blob/master/docs/job_tutorial.md#how-to-debug-the-job-) for details.
```
mzmssg: Should we explain how to debug from logs? I don't think users know what the `YarnContainerScript.log` and `DockerContainerScript.log` mean. For me, they are very confusing until I read related scripts, but it is impossible for users.

I think we need to teach users how to judge the exit reason, if it's system reason, users should ask help from admin or DRI, otherwise where could users find their code stdout and stderr info.

docs/datascientist_onboard.md
mzmssg: Agree with you, such scenario is over complicated, beyond what PAI cares.

docs/datascientist_onboard.md
```diff
@@ -188,6 +188,11 @@ docker tag pai.run.tensorflow ${your_user_name}/pai.run.tensorflow
 docker push ${your_user_name}/pai.run.tensorflow
 ```
  
+ **Tips:** You can debug your container locally, ensure it works well then go on the submit job process. Excecute the following command:
+ ```
+ docker run -v ${local_codedir}:${container_codedir} -v ${local_datadir}:${container_datadir} ${image_name_you_built} -c ${commands_to_exec}
```
mzmssg: `/bin/bash -c`

docs/datascientist_onboard.md
```diff
@@ -0,0 +1,235 @@
...
+    You can pack your code and dataset into your container when you are [building your job container](#3.1-Need-to-modify-the-DL-framework's-version,-or-add-some-dependencies).
+
+    Use `COPY` command in Dockerfile to copy files from local into container, then build it. 
+    
+    For more information please refer to [Dockerfile reference](https://docs.docker.com/engine/reference/builder/#copy).
+
+    **Note:** We don't recommend to pack code into container unless your code won't change. Because every time you change your code you will build a new image. You can upload your code to HDFS, then specify the HDFS url of your code in [job JSON configuration file](#3.-Write-a-job-json-configuration-file) `codedir` field, PAI will automatically download it to the job container's path `/pai/code/`.
+
+
+3) Use HDFS mount.
```
mzmssg: Seems `hadoop-hdfs-fuse` is another mount tools(https://github.com/Microsoft/pai/blob/27be2f781498f16f67c44ebf2cfd9199a55fdcf9/pai-management/doc/hdfs.md#mountable-hdfs-), should we keep consistent with our docs

YitongFeng: But our CNTK example which used HDFS mount is using https://github.com/Microsoft/hdfs-mount.git I think it's OK

mzmssg: Get it

fanyangCS:
outdated.

PR #1346, opened 2018-09-12
Folder refactor

hao1939:


yqwang-ms:
We would better merge without squash to keep each owners changes for tracking.

PR #1394, opened 2018-09-17
Zhaoyu/deleted files

DongZhaoYu:
list the deleted files and show warning in log


src/cleaner/scripts/check_deleted_files.py
```diff
@@ -0,0 +1,45 @@
...
+    files = common.run_cmd("source ./scripts/list_deleted_files.sh 2>/dev/null", log)
+    if len(files) <= 1:
+        log.info("no deleted files found.")
+        return
+    else:
+        # skip the field names from the command
+        files = files[1:]
+
+    for f in files:
+        f_fields = f.split(" ")
```
xudifsd: Is output of the first column guaranteed not having space?

src/cleaner/scripts/check_deleted_files.py
```diff
@@ -0,0 +1,45 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+import multiprocessing
+from cleaner.utils import common
+
+logger = multiprocessing.get_logger()
+
+
+def list_and_check_files(arg, log=logger):
+    files = common.run_cmd("source ./scripts/list_deleted_files.sh 2>/dev/null", log)
```
xudifsd: If the script is not too complicated, call command directly would be more readable and efficient(without calling bash)

src/cleaner/scripts/check_deleted_files.py
DongZhaoYu: Yes, the first column of the output is process id and it has no space.

src/cleaner/scripts/check_deleted_files.py
DongZhaoYu: Consider the performance. It is reasonable to call it directly. Will change it.

PR #1399, opened 2018-09-18
[Rest-server]Add OS check for ssh-keygen and fix code size bug

wangcan0329:


src/rest-server/src/models/job.js
```diff
@@ -575,24 +575,30 @@ class Job {
...
-              parallelCallback(error);
-            });
-          }
-        });
+        // Add OS platform check
+        // Only generate ssh key files on Linux
+        if (process.platform.toUpperCase() === 'LINUX') {
+          this.generateSshKeyFiles(name, (error, sshKeyFiles) => {
+            if (error) {
+              logger.error('Generated ssh key files failed');
```
Gerhut: `parallelCallback()`

PR #1413, opened 2018-09-20
[aks] deploy dev-box as a daemonset choice for user

YanjieGao:
Reason: As AKS kubernetes can't access directly from external. PAI deploy need ssh to server. 
Need dev-box container run inside cluser.
deploy as a pod will reduce 2 times ssh cost than "ssh to  aks server and launch dev-box container at that server".

This is also useful for deploy pai on existing k8s cluster. 

How to use:
as dev-box step is before deploy service. so need let user label dev-box node and create the pod.


src/dev-box/dev-box.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: apps/v1
+kind: DaemonSet
```
sterowang: Suggest to rename this file to dev-box-k8s-deploy.yaml

YanjieGao: updated

src/dev-box/dev-box-k8s-deploy.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: apps/v1
+kind: DaemonSet
```
hao1939: Could you deploy dev-box as a 'POD'?
I'm not sure why it should be a daemonset.

YanjieGao: Sure, it could be, need some test to test it
 use daemonset is consist to yundong's previous deployment, most of deployed service as daemonset.

src/dev-box/dev-box-k8s-deploy.yaml
```diff
@@ -0,0 +1,44 @@
...
+        dev-box: "true"
+      containers:
+      - image: docker.io/openpai/dev-box
+        imagePullPolicy: Always
+        securityContext:
+          privileged: true
+        volumeMounts:
+        - mountPath: /root/.docker
+          name: docker-cred-volume
+        - mountPath: /bin/docker
```
ydye: Can you ensure the binary will always be here?

src/dev-box/dev-box-k8s-deploy.yaml
```diff
@@ -0,0 +1,44 @@
...
+          name: docker-bin
+        - mountPath: /var/run/docker.sock
+          name: docker-socket
+        name: dev-box
+        command: [ "/bin/bash", "-c", "--" ]
+        args: [ "while true; do sleep 3000; done;" ]
```
ydye: ```/bin/bash``` is ok

YanjieGao: thx, will try

src/dev-box/dev-box-k8s-deploy.yaml
YanjieGao: will remove and keep  -v /var/run/docker.sock:/var/run/docker.sock \

Issue #721, opened 2018-06-13
job container network traffic metrics monitor


YanjieGao:
phenomenon: 
As job container use host network, job container's network traffic get 0 from docker status cmd. Job metrics page network traffics will always shows zero.

Job submit cmd & config: https://github.com/Microsoft/pai/tree/master/rest-server/src/templates

Have tried other source to get this per container metrics of host network container, but failed:
(1) docker status cmd
network zero
(2) docker api
not return network traffic item
(3) /proc/container_pid/net/dev
show host network container same eth0 traffics, its host traffics
(4) kubelet build in cadvisor
show host network container same eth0 traffics, its host traffics

YanjieGao:
https://github.com/Microsoft/pai/pull/764

PR #1418, opened 2018-09-20
collect network metrics for containers

xudifsd:
almost same with #764 

This functionality requires job-exporter container to be privileged and has host pid namespace. Since this will have security issue, I enabled admission control in api server with recommended policy and `DenyEscalatingExec`. @ydye please verify this.

It also requires other service to have non-host pid namespace to be accurate.


xudifsd:
fixed #721

PR #1421, opened 2018-09-20
update hadoop version to fix NodeManger GPU detect issue when ecc is turn off.

qinchen123:


ydye:
Please change the version tag in this file: https://github.com/Microsoft/pai/blob/master/src/hadoop-ai/build/build-pre.sh

ydye:
@qinchen123

mzmssg:
need to update here
 https://github.com/Microsoft/pai/blob/master/src/hadoop-ai/build/build-pre.sh#L25

Issue #834, opened 2018-07-09
Not able to see the previous retry from the job view


hao1939:
This job has been retried 35 times.
http://10.151.40.179/view.html?jobName=tensorflow-cifar10-test_v8

I don’t seem to be able to see the previous retry from the job view.

yqwang-ms:
This is planed to be handled by:
Launcher's HistoryServer:
Store all the history of Framework Request and Status for each previous versions and attempts.

fanyangCS:
will do some quick fix in job view by @Gerhut

wangcan0329:
Already do quick fix in job view. if job retry count > 0 you can click retry number in job detail page then will jump to Yarn portal which will show job history automatically.

PR #1425, opened 2018-09-20
[Rest-server] Add job retry history

wangcan0329:
if job retry count > 0 we can click retry number in job detail page then jump to Yarn portal which will show job history automatically.

image

image


fanyangCS:
address this issue: https://github.com/Microsoft/pai/issues/834

PR #1427, opened 2018-09-21
[PAIShare Doc] How-to-config-gitHubPAT.md

yanli2017:


wangcan0329:
Is it OK to show your personal token directly in Copy the new generated token step?

PR #1428, opened 2018-09-21
[aks] add aks deploy switch / api-server uri config / k8s dashboard uri config

YanjieGao:
note: 
- aks api-server format is:   https://string:443
for aks config will let kube/config contains old and new cluster server info. so change to a for loop for all cluster server info.
- k8sdashboard access use a ip which is different with node inner ip address

Future:
user config k8s-type + load-balance-ip (api-server string). Will make api-server uri config / k8s dashboard uri config work

### DIFF
--- a/docs/pai-management/doc/example/add-service/hbase/deploy/hbase-master.yaml.template
+++ b/docs/pai-management/doc/example/add-service/hbase/deploy/hbase-master.yaml.template
@@ -30,8 +30,6 @@ spec:
     spec:
       hostNetwork: true
       hostPID: true
-      nodeSelector:
-        hbasemaster: "true"
       containers:
       - name:  hbase-master
         image: {{ clusterinfo['dockerregistryinfo']['prefix'] }}hbase:{{ clusterinfo['dockerregistryinfo']['docker_tag'] }}
--- a/docs/pai-management/doc/example/add-service/hbase/deploy/hbase-regionserver.yaml.template
+++ b/docs/pai-management/doc/example/add-service/hbase/deploy/hbase-regionserver.yaml.template
@@ -30,8 +30,6 @@ spec:
     spec:
       hostNetwork: true
       hostPID: true
-      nodeSelector:
-        hbaseregionserver: "true"
       containers:
       - name:  hbase-regionserver
         image: {{ clusterinfo['dockerregistryinfo']['prefix'] }}hbase:{{ clusterinfo['dockerregistryinfo']['docker_tag'] }}
--- a/docs/pai-management/doc/example/add-service/hbase/deploy/start.sh
+++ b/docs/pai-management/doc/example/add-service/hbase/deploy/start.sh
@@ -37,31 +37,17 @@ pushd $(dirname "$0") > /dev/null
 # hbase master
 kubectl create -f hbase-master.yaml
 
-# A tool to check whether a node was labeled with hbasemaster="true"
-PYTHONPATH="../.." python -m  k8sPaiLibrary.monitorTool.check_node_label_exist -k hbasemaster -v "true"
-ret=$?
 
-if [ $ret -ne 0 ]; then
-    echo "No hbase-master Pod in your cluster"
-else
-    # A tool to wait all pod which as label app=hbase-master to be ready.
-    PYTHONPATH="../.." python -m  k8sPaiLibrary.monitorTool.check_pod_ready_status -w -k app -v hbase-master
-fi
+# A tool to wait until the service is ready.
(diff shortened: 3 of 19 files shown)


deployment/paiLibrary/clusterObjectModel/paiObjectModel.py
```diff
@@ -311,13 +312,26 @@ def getPaiWebPortalUri(self):
     def getK8sApiServerUri(self):
 
         ip = self.rawData["kubernetesConfiguration"]["kubernetes"]["load-balance-ip"]
-        ret = "http://{0}:{1}".format(ip, 8080)
+        ret = ""
+        if self.rawData["kubernetesConfiguration"]["kubernetes"]["k8s-type"] == "aks":
+            ret = "https://{0}:{1}".format(ip, 443)
+        else:
```
sterowang: Actually this section is used to determin the K8S API server port. On-prem it configured as 8080, on AKS it configured as 443. So I suggest we can add a kubernetes/api-server-port config item and configure it as 443 instead of adding a so called "k8s-type" as "aks". This should be more general and direct.

deployment/paiLibrary/clusterObjectModel/paiObjectModel.py
```diff
@@ -311,13 +312,26 @@ def getPaiWebPortalUri(self):
...
+        if self.rawData["kubernetesConfiguration"]["kubernetes"]["k8s-type"] == "aks":
+            cmd = "kubectl describe pods kubernetes-dashboard --namespace=kube-system | grep IP"
+            output = linux_shell.execute_shell_with_output(cmd, "Failed to exec {0}".format(cmd))
+            # extract ip address
+            dashboardInfo = re.findall(r"\b(?:[0-9]{1,3}\.){3}[0-9]{1,3}\b", output)
+            if dashboardInfo:
+                ret = "http://{0}:9090".format(dashboardInfo[0])
+                return ret            
+
         vip = ""
```
sterowang: Can this way applies to our on-prem cluster also? If so, suggest to replace the whole logic with the new code universally. We should keep our code simple and unified in maximum extend instead of adding more and more branches unless really no good ways to simplify the code.

deployment/k8sPaiLibrary/maintainlib/kubectl_conf_check.py
```diff
@@ -59,16 +59,22 @@ def check(self):
 
         try:
             local_kubectl_conf = common.load_yaml_file("{0}/config".format(self.kube_conf_path))
-            api_server_address = local_kubectl_conf['clusters'][0]['cluster']['server']
-
             api_server_address_pai_conf = "http://{0}:8080".format(self.cluster_config['clusterinfo']['api-servers-ip'])
+	    api_server_address_aks_conf = "https://{0}:443".format(self.cluster_config['clusterinfo']['api-servers-ip'])
```
fanyangCS: 8080 and 443 are just ad hoc config info that should go to cluster config file

YanjieGao: move it as api-server-port config at pai k8s config file

deployment/paiLibrary/clusterObjectModel/paiObjectModel.py
YanjieGao: test it could used for on-prem, unify use the same logic

deployment/paiLibrary/clusterObjectModel/paiObjectModel.py
YanjieGao: move it as api-server-port config and remove k8s-type config. for 8080 to 443, as serveral files have dependecy and use 8080 hard code config, need confirm with yundong, currently just config 8080 as a temp and work config

YanjieGao: pai use http, aks use https, also make it as a config.

deployment/k8sPaiLibrary/maintainlib/kubectl_conf_check.py
```diff
@@ -59,16 +59,19 @@ def check(self):
 
         try:
             local_kubectl_conf = common.load_yaml_file("{0}/config".format(self.kube_conf_path))
+            http = self.cluster_config['clusterinfo']['api-server-http']
```
xudifsd: maybe `api-server-scheme`?

YanjieGao: updated

PR #1431, opened 2018-09-25
Add an operator [ paictl conf update ] to upload configuration to k8s 

ydye:
#### Progress
- [x] paictl config update -p /path/to/local/configuration [-c kubeconfig-path ]
- [x] paictl config update -e external-storage-config [-c kubeconfig-path]
- [x] paictl config update [-c kubeconfig-path]

image

- [x] paiclt.py config external-config-update -e external-configuration [-c kube-config]

image

- [x] paictl.py config get -o /path/to/output [-c kube-config]

#### Change
1) With the provided external-configuration-storage conf, you could get your cluster-configuration from the service like git, and then upload it to k8s cluster.
2) The cluster-configuration will be updated into a configmap. And all service in the k8s cluster could get it from configmap volume.
3) Tool ```paictl``` will get the configuration from k8s-configmap. So next time, admin should update their cluster configuration from external storage to the cluster. Then manage the openpai cluster.


Jenkinsfile
```diff
@@ -64,6 +64,9 @@ sudo chown core:core -R /pathHadoop/
 QUICK_START_PATH=${JENKINS_HOME}/${BED}/singlebox/quick-start
 CONFIG_PATH=${JENKINS_HOME}/${BED}/singlebox/cluster-configuration
 
+# Install necessary python package
+sudo pip install GitPython
```
hao1939: Could we remove the line? Noticed that you have updated the dependencies in `dev-box`.

ydye: No we can't. paictl need import it. You will call paictl to generate cluster-configuration out of the dev-box.

ydye:
#1440

deployment/confStorage/external_version_control/git_storage.py
```diff
@@ -0,0 +1,87 @@
...
+            self.logger.error("Path: {0}".format(self.path))
+            self.logger.error("Repo: {0}".format(self.repo_url))
+            sys.exit(1)
+
+        pai_temp_path = "./tmp-config-{0}".format(self.time)
+        directory_handler.directory_copy("{0}/*".format(configuation_path), pai_temp_path)
+
+
+
+    def open(self):
```
xudifsd: maybe you can use `__enter__` and `__exit__`, which is more elegant, and can clean storage in exit.

Same applies to `local_storage`.

deployment/confStorage/external_version_control/git_storage.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def __init__(self, storage_configuration, local_store = "/tmp/pai-conf-sync/git-storage"):
+
+        self.repo_url = storage_configuration["url"]
+        self.branch = storage_configuration["branch"]
+        self.path = storage_configuration["path"]
+
+        self.time = str(int(time.time()))
+        self.logger = logging.getLogger(__name__)
+        self.local_store = "{0}-{1}".format(local_store, self.time)
```
xudifsd: why not using [`tempfile`](https://docs.python.org/2/library/tempfile.html)

deployment/confStorage/external_version_control/git_storage.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+
+    def git_clone(self):
+        try:
+            self.repo = git.Repo.clone_from(self.repo_url, self.local_store, branch=self.branch)
```
xudifsd: can we just clone with `--depth=1`, since we requires no history.

ydye: Changed.

deployment/confStorage/external_version_control/git_storage.py
ydye: Changed

deployment/confStorage/external_version_control/git_storage.py
ydye: We will have to remove the temp directory created by ```tempfile.mkdtemp([suffix=''[, prefix='tmp'[, dir=None]]]) ```.  And I think it's same as what we have done.

PR #1435, opened 2018-09-26
add init option in docker run, help reaping zombie processes

mzmssg:
hanging container, keep alive after `exit_handle`
image

only script process and 2 zombie processes. `/bin/bash` doesn't reap them correctly.
image

`--init` will overwrite `entrypoint` with `/dev/init`, thus the pid 1 process looks like `/dev/init -- /bin/bash dockerboostrap.sh`. `init` process have a right behavior in reaping zombie processes.

### DIFF
--- a/src/rest-server/src/templates/yarnContainerScript.mustache
+++ b/src/rest-server/src/templates/yarnContainerScript.mustache
@@ -251,6 +251,7 @@ docker pull {{{ jobData.image }}} \
 ## ATTENTION: do not specify `--pid=host` when run job, otherwise job-exporter can not get right
 ## network consumption
 docker run --name $docker_name \
+  --init \
   --rm \
   --tty \
   --privileged=false \
--- a/src/rest-server/src/templates/yarnContainerScript.mustache
+++ b/src/rest-server/src/templates/yarnContainerScript.mustache
@@ -276,6 +276,6 @@ docker run --name $docker_name \
   --label PAI_USER_NAME={{{ jobData.userName }}} \
   --label PAI_CURRENT_TASK_ROLE_NAME={{{ taskData.name }}} \
   --env-file $ENV_LIST \
-  --entrypoint '/bin/bash' {{{ jobData.image }}} \
-  '/pai/bootstrap/docker_bootstrap.sh'
+   {{{ jobData.image }}} \
+  '/bin/bash /pai/bootstrap/docker_bootstrap.sh'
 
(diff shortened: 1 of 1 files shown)


fanyangCS:
can you point us to the reference about your findings?

mzmssg:
@fanyangCS 

This is a blog talking about PID 1 issue in docker run, it also mentions that bash have some issues in handling signal.
https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/

This is `--init` behavior in `docker run`
https://stackoverflow.com/questions/43122080/how-to-use-init-parameter-in-docker-run

src/rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -275,6 +276,6 @@ docker run --name $docker_name \
   --label PAI_USER_NAME={{{ jobData.userName }}} \
   --label PAI_CURRENT_TASK_ROLE_NAME={{{ taskData.name }}} \
   --env-file $ENV_LIST \
-  --entrypoint '/bin/bash' {{{ jobData.image }}} \
-  '/pai/bootstrap/docker_bootstrap.sh'
+   {{{ jobData.image }}} \
+  /bin/bash '/pai/bootstrap/docker_bootstrap.sh'
```
fanyangCS: this will set the pid of "bash docker_bootstrap.sh" to 1? it seems previously it is already the case.

mzmssg: Not actually:
Without `--init`, pid 1 process is `bash docker_bootstrap.sh`.
With `--init`, pid 1 process is `/dev/init`, and `bash docker_bootstrap.sh` will be its subprocess.

fanyangCS: yeah. https://docs.docker.com/engine/reference/run/#specify-an-init-process

mzmssg: In my testing, `--init` can fix the suicide failure as well.  it is a good solution to such issues.

https://stackoverflow.com/questions/49162358/docker-init-zombies-why-does-it-matter
https://github.com/krallin/tini/issues/8#issuecomment-146135930

fanyangCS: https://github.com/krallin/tini#understanding-tini. if the child of process exits, init will exit

PR #1436, opened 2018-09-26
PAIShare opt-in

Gerhut:


src/webportal/src/app/user/user-logout/user-logout.component.js
```diff
@@ -20,6 +20,7 @@ const userLogout = () => {
   cookies.remove('user');
   cookies.remove('token');
   cookies.remove('admin');
+  cookies.remove('paishareOptIn');
```
yanli2017: Is this should be "hasGitHubPAT" ?

PR #1441, opened 2018-09-27
Zhaoyu/cleaner build deploy

DongZhaoYu:
cleaner build and deploy scripts.


fanyangCS:
how do we use it?

DongZhaoYu:
For the docker cache, we check the reclaimable cache size every 2 seconds and if the size is great than 50G, we will clean cache.
For the deleted files, we also do the check every 2 seconds, if there are deleted files opened by a live process, currently we only log warnings.

src/cleaner/cleaner_main.py
```diff
@@ -49,12 +51,28 @@ def terminate(self):
...
+                        self.logger.error("worker %s exit with code %s", k, w.exitcode)
+                        self.workers.pop(k)
+                if len(self.workers) == 0:
+                    self.logger.info("all workers are stopped and exit cleaner.")
+                    break
+                self.update_liveness()
+                time.sleep(2)
+        except Exception as e:
             self.logger.error("cleaner interrupted and will exit.")
+            self.logger.exception(e)
```
xudifsd: [logger.exception](https://docs.python.org/2/library/logging.html#logging.Logger.exception) method accept string msg, and will print msg and exception.
use like `self.logger.exception(cleaner interrupted and will exit.)`

src/cleaner/deploy/cleaner.yaml.template
```diff
@@ -0,0 +1,76 @@
...
+    spec:
+      hostPID: true
+      hostNetwork: true
+      containers:
+      - name: docker-cache-cleaner
+        image: {{ clusterinfo['dockerregistryinfo']['prefix'] }}cleaner:{{ clusterinfo['dockerregistryinfo']['docker_tag'] }}
+        args:
+        - 'docker_cache'
+        imagePullPolicy: Always
+        securityContext:
```
xudifsd: do you really need the privileged setting?

src/cleaner/build/cleaner.dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+RUN apt-get -y update && \
+    apt-get -y install lsof gawk
+
+RUN pip install psutil
+
+RUN mkdir -p /cleaner
+COPY scripts /cleaner/scripts
+COPY utils /cleaner/utils
+COPY ./*.py /cleaner/
```
xudifsd: Put COPY command at the end. In this way, docker build can reuse download docker-ce layer, this can speed up the building process.

src/cleaner/build/cleaner.dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+RUN curl -SL https://download.docker.com/linux/static/stable/x86_64/docker-17.06.2-ce.tgz \
+    | tar -xzvC /usr/local \
+    && mv /usr/local/docker/* /usr/bin
```
xudifsd: remove the tgz, to minimize image size.

src/cleaner/deploy/cleaner.yaml.template
DongZhaoYu: yes, docker command and lsof need privilege access.

src/cleaner/build/cleaner.dockerfile
DongZhaoYu: curl -SL will not write the tar ball to disk. The content will be written to stdout and tar will decompress it via pipeline.

src/cleaner/cleaner_main.py
DongZhaoYu: change done

src/cleaner/build/cleaner.dockerfile
DongZhaoYu: change done

DongZhaoYu:
I have changed the cleaner to check docker cache and deleted files every 30 minutes. The setting of 2 seconds seems too heavy.

PR #1444, opened 2018-09-28
refine refresh

YitongFeng:
In the past, execute kubectl label node regardless of whether the node has the label, it will output confused log, refine it by check the existence of the label first, if already has label, do nothing.

- [x] Check label globally
Now refresh each service all need to scan the whole machinelist once, if refresh all services and the cluster is big it will take longtime, should change the scan step to global way.

### DIFF
--- a/deployment/paiLibrary/paiService/service_refresh.py
+++ b/deployment/paiLibrary/paiService/service_refresh.py
@@ -25,13 +25,13 @@ from ..common import linux_shell
 class service_refresh:
 
 
-    def __init__(self, service_conf, serivce_name, machinelist):
+    def __init__(self, service_conf, service_name, role_map):
 
         self.logger = logging.getLogger(__name__)
 
         self.service_conf = service_conf
-        self.service_name = serivce_name
-        self.machinelist = machinelist
+        self.service_name = service_name
+        self.role_map = role_map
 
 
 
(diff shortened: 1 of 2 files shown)


deployment/paiLibrary/paiService/service_refresh.py
```diff
@@ -41,51 +41,59 @@ def start(self):
...
-                        if linux_shell.execute_shell_return(cmd, ""):
-                             self.logger.info("Remove Node " + nodename + " label " + rule['notin'] + ", due to the cluster-configuration machinelist doesn't specify this label")
-                             cmd = "kubectl label nodes " + nodename + " " + rule['notin'] +"- || exit $?"
+            if 'deploy-rules' in self.service_conf:
+                for rule in self.service_conf['deploy-rules']:
+                    if 'in' in rule:
+                        # If machinelist config has defined the label
+                        if rule['in'] in self.machinelist[host]:  
+                            # If the node did't have the label, label it
+                            cmd_checklabel = "kubectl describe node " + nodename + " | grep -q " + rule['in'] + "='true'"
```
ydye: Have you evaluated the time cost？

deployment/paiLibrary/paiService/service_refresh.py
```diff
@@ -41,51 +41,59 @@ def start(self):
...
-                    if rule['notin'] in self.machinelist[host]:
-                        self.logger.info("Role defined in cluster-configuration machinelist, label the node of this role...")
-                        cmd = "kubectl label --overwrite=true nodes " + nodename + " " + rule['notin'] + "='true' || exit $?"
-                        linux_shell.execute_shell(cmd, err_msg_prefix + cmd)     
-                    else:
-                        cmd = "kubectl describe node " + nodename + " | grep -q " + rule['notin'] + "='true'"
-                        if linux_shell.execute_shell_return(cmd, ""):
-                             self.logger.info("Remove Node " + nodename + " label " + rule['notin'] + ", due to the cluster-configuration machinelist doesn't specify this label")
-                             cmd = "kubectl label nodes " + nodename + " " + rule['notin'] +"- || exit $?"
+            if 'deploy-rules' in self.service_conf:
```
ydye: Can you split this part into several function and try to reuse the code？

PR #1445, opened 2018-09-28
[PAIShare] Set Custom GitHub Repo

yanli2017:


src/rest-server/deploy/rest-server.yaml.template
```diff
@@ -52,6 +52,12 @@ spec:
           value: {{ clusterinfo['restserverinfo']['default_pai_admin_password'] }}
         - name: ETCD_URI
           value: {{ clusterinfo['restserverinfo']['etcd_uri'] }}
```
wangcan0329: 1.You need to parse clusterinfo['restserverinfo']['github_owner'] in paiObjectModel.py , you can refer to restserverinfo section. 
2. I'm not sure where the new envs come from. if it's user configuration user need to config it in config file, but if PAI decides you need to add logic to get the value.
3. Please refer to https://github.com/Microsoft/pai/blob/master/deployment/paiLibrary/clusterObjectModel/paiObjectModel.py
Line 205 to 222.

yanli2017: Thanks wangcan. The new envs need user to config.

PR #1447, opened 2018-09-29
Modify the paictl cluster command to set up kubernetes environment.

ydye:
Remove ```paictl.py cluster install-kubectl -p /path/to/cluster/config ```.
Add ```paictl.py cluster k8s-set-env [-p /path/to/cluster/config]```

### DIFF
--- a/deployment/k8sPaiLibrary/maintainlib/kubectl_install.py
+++ b/deployment/k8sPaiLibrary/maintainlib/kubectl_install.py
@@ -52,6 +52,7 @@ class kubectl_install:
             commandline,
             "Failed to install kubectl on your dev-box"
         )
+        self.logger.info("Successfully install kubectl on the dev-box.")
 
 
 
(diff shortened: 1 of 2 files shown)


docs/paictl/paictl-manual.md
```diff
@@ -185,16 +187,17 @@ python paictl.py cluster k8s-clean -p /path/to/cluster-configuration/dir
...
 ```
 
-- The machine list should be provided in CSV format.
-- All configuration files of cluster-configuration will be generated to the local folder, and then be used for bootstrapping the whole cluster.
-- By default, in the generated configuration, a single-master Kubernetes is configured by default.
-- Advanced users or developers can fine-tune the content of the generated configuration files according to specific environments.
+- Install kubectl in the deployment box.
+- Configure KUBECONFIG for kubectl or other kubernetes' client.
+- The args ```-p /path/to/cluster/configuration``` is optional. If admin don't pass this args, paictl will ask admin to input the necessary infomation to generate KUBECONFIG.
+- When in a clean dev-box, please run this command firstly.
```
hao1939: Will it prompt hints if user forget to run it first?

ydye: If user forget to setup environment, all command will failed. I will add information in the future.

hao1939: Could we provide a command displaying the currently config?
For example `python paictl.py config info`.

So user can double check which cluster they are working on.

ydye: I have provided a command ```paictl config get``` to get the latest configuration

PR #1450, opened 2018-09-29
Disk config

DongZhaoYu:
add new config entry to allow user to config the docker data location


ydye:
@DongZhaoYu 
Could this PR be compatible with this [script](https://github.com/Microsoft/pai/blob/7e35691704be5e8e230f62e6e39d1ad9cc53378d/deployment/k8sPaiLibrary/template/kubelet.sh.template#L33)?

DongZhaoYu:
@ydye it's compatible, the data root path will be configured before kubelet is started. And we will restart docker daemon to consume the latest configuration. This script call "docker info" to get the data root and it is exactly the path we have configured.

PR #1453, opened 2018-10-06
Upgrade Node.js version of  webportal & rest server to 8 (Carbon)

Gerhut:
Replace `npm run yarn` (which stands for `yarn --ignore-engines`) to `yarn` directly

- [x] Document


.travis.yml
```diff
@@ -66,7 +66,7 @@ matrix:
         - npm run coveralls
 
     - language: node_js
-      node_js: 7
+      node_js: node
```
Gerhut: Ref: <https://github.com/creationix/nvm/blob/master/README.md#usage> `node` version means the latest version of node.js

.travis.yml
```diff
@@ -55,7 +55,7 @@ matrix:
         - mvn clean test jacoco:report coveralls:report
 
     - language: node_js
-      node_js: 6
+      node_js: lts/carbon
```
Gerhut: Unify the travis test version as LTS name version, which has been already used in Dockerfile (`FROM node:carbon`)

PR #1458, opened 2018-10-08
[AKS] prometheus config leverage service account access api-server

YanjieGao:
prometheus deployment leverage service account. this is maybe useful for deploy pai on existing k8s which api-server access need auth.

when use this on aks:

- user config use_tls
- config api-server url. 
note: this has the dependency with pr (https://github.com/Microsoft/pai/pull/1428)  for api-server url configuration.

### DIFF
--- a/src/prometheus/deploy/prometheus-configmap.yaml.template
+++ b/src/prometheus/deploy/prometheus-configmap.yaml.template
@@ -36,6 +36,9 @@ data:
     scrape_configs:
     - job_name: 'node_exporter'
       scrape_interval: {{ prom_info['scrape_interval']|default(30) }}s
+      tls_config:
+        ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+      bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
       kubernetes_sd_configs:
       - api_server: '{{ clusterinfo['webportalinfo']['k8s_api_server_uri'] }}'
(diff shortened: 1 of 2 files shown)


src/prometheus/deploy/prometheus-configmap.yaml.template
```diff
@@ -36,6 +36,13 @@ data:
     scrape_configs:
     - job_name: 'node_exporter'
       scrape_interval: {{ prom_info['scrape_interval']|default(30) }}s
+      {% if clusterinfo['prometheusinfo']['use_tls'] -%}
+      tls_config:
+        ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
```
xudifsd: This path is a volume or what?

YanjieGao: a volume mounted by k8s for service account auth

xudifsd: Ok, then, I think it's inappropriate to config under `prometheus`.

YanjieGao: yes, watchdog's method is a little different with prometheus. it need reconfig http url access.

YanjieGao: make config as a global cluster wide config which prometheus watchdog webportal future use

PR #1463, opened 2018-10-09
Expanding unit test coverage for ApplicationMaster, Configuration, and RMResyncHandler classes

christopher-fredregill:
Noticed the coveralls % coverage has been decreasing lately. This just adds a few test cases to get the coverage back up.

### DIFF
--- a/subprojects/frameworklauncher/yarn/pom.xml
+++ b/subprojects/frameworklauncher/yarn/pom.xml
@@ -165,6 +165,18 @@
       <version>${mockito.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.github.stefanbirkner</groupId>
+      <artifactId>system-rules</artifactId>
+      <version>1.18.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <version>3.11.1</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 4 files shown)


subprojects/frameworklauncher/yarn/pom.xml
```diff
@@ -165,6 +165,18 @@
       <version>${mockito.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.github.stefanbirkner</groupId>
+      <artifactId>system-rules</artifactId>
```
yqwang-ms: Thanks for this PR.
Could you use some more well-known libs? Or just do not use it?

christopher-fredregill: Probably not, but I'll also point out that this _is_ a fairly well-known library, at least as far as testing utils are concerned (see https://github.com/search?l=Maven+POM&p=1&q=stefanbirkner%3Asystem-rules&type=Code or https://mvnrepository.com/artifact/com.github.stefanbirkner/system-rules/usages, for example).

yqwang-ms: but, anyway, it is owned by a private person, instead of an org, can we believe it will be continue to maintain?

---
In reply to: [224299238](https://github.com/Microsoft/pai/pull/1463#discussion_r224299238) [](ancestors = 224299238)

christopher-fredregill: There are several contributors on the main repo, not just one individual. It's also been forked many times. Not sure what else I can suggest, other than to look into it.

yqwang-ms: Could you please only use the lib which is also used by [https://github.com/apache/hadoop](https://github.com/apache/hadoop)

---
In reply to: [224308608](https://github.com/Microsoft/pai/pull/1463#discussion_r224308608) [](ancestors = 224308608)

christopher-fredregill: OK, can do. Unfortunately, the tests I submitted were pretty dependent on that library, so we're basically talking about a total re-write. I probably wont get to it for a while, so I'll just go ahead and close this one. Thanks for the review.

yqwang-ms: weclome :)

PR #1466, opened 2018-10-09
New command: paictl machine update

ydye:
1) Add a new configmap ```pai-node-config```, to store the configuration of each node. Keep this configmap in the namespace ```kube-system```.

2) ```pai machine update```. If machine is removed from the cluster-configuration stored in the configmap, it will be removed. If machine is added into the cluster-configuration stored in the configmap, it will be added.

3) Fix some bug.

image


deployment/k8sPaiLibrary/maintainlib/deploy.py
```diff
@@ -106,6 +107,26 @@ def job_executer(self, node_config, job_name):
 
 
 
+    def update_node_config(self):
+        cluster_config = self.cluster_config
+        node_config_from_cluster_conf = dict()
+
+        for role in cluster_config["remote_deployment"]:
```
hao1939: En, will need the configuration layout and explanation in order to understand.

deployment/k8sPaiLibrary/maintainlib/update.py
```diff
@@ -0,0 +1,253 @@
...
+        objectModel = objectModelFactory.objectModelFactory(self.tmp_path)
+        ret = objectModel.objectModelPipeLine()
+
+        directory_handler.directory_delete(self.tmp_path)
+        self.logger.info("Successfully get latest configuration from Kubernetes.")
+        return ret["k8s"]
+
+
+
+    def get_node_list_from_k8s_api(self):
```
hao1939: I'd prefer the style likes `ClusterConfiguration.get_latest()`,  `ClusterConfiguration.get_node_list()`, the `from` indicate that we may be able to extract a object.
What do you think?

deployment/k8sPaiLibrary/maintainlib/deploy.py
ydye: No, It's by design. For different k8s role.

deployment/k8sPaiLibrary/maintainlib/update.py
ydye: ```node_list``` comes from kubernetes api. So I don't think it should be part of the cluster configuration.
The obect you say, should be a future feature. Cluster object model. And now, we handle it as the easiest way. Just get all configuration from a dictionary.

deployment/k8sPaiLibrary/maintainlib/deploy.py
```diff
@@ -106,6 +107,26 @@ def job_executer(self, node_config, job_name):
 
 
 
+    def update_node_config(self):
+        cluster_config = self.cluster_config
+        node_config_from_cluster_conf = dict()
```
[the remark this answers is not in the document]
ydye: Little confused about this comment.

PR #1475, opened 2018-10-11
change nm to hostpid

mzmssg:


src/hadoop-node-manager/deploy/hadoop-node-manager.yaml.template
```diff
@@ -29,7 +29,8 @@ spec:
         app: hadoop-node-manager
     spec:
       hostNetwork: true
-      hostPID: false
+      # hostPID is design for monitor job resources, but it causes network traffic mistakes.
```
xudifsd: maybe `node manager needs hostPID to be true to be able to show process name in nvidia-smi result`?

Original comment is not that obvious.

mzmssg: update

PR #1481, opened 2018-10-12
expose zombie count and add alert

xudifsd:
Monitor two types of zombie container:
1. container with log displaying `USER COMMAND END` but still exist for a long period.
2. those containers whose yarn container not exist but job container still exist.

We only send out alert email when the count above 0 for 1 hour, because we just want to find out the root cause of zombie container.

### DIFF
--- a/src/job-exporter/src/job_exporter.py
+++ b/src/job-exporter/src/job_exporter.py
@@ -59,6 +59,40 @@ pai_services = map(lambda s: "k8s_" + s, [
 ])
 
 
+class ZombieRecorder(object):
+    def __init__(self):
+        self.zombies = {} # key is container id, value is enter zombie time
+
+        # When we first meet zombie container, we only record time of that meet,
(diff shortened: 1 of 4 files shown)


src/job-exporter/src/job_exporter.py
```diff
@@ -72,14 +80,99 @@ def parse_from_labels(labels):
...
+    There are two types of zombie:
+        1. container which outputed "USER COMMAND END" but did not exist for a long period of time
+        2. yarn container exited but job container didn't
+    """
+    exited_containers = set(filter(is_container_exited, stats.keys()))
+    logger.debug("exited_containers is %s", exited_containers)
+
+    zombie_count1 = generate_zombie_count_type1(zombie_containers,
+            exited_containers, datetime.datetime.now())
+    zombie_count2 = generate_zombie_count_type2(stats)
```
mzmssg: For type2, a time interval is needed as well.

PR #1491, opened 2018-10-15
Remove old label check & refine hbase example

YitongFeng:


src/zookeeper/deploy/start.sh
```diff
@@ -24,16 +24,8 @@ pushd $(dirname "$0") > /dev/null
...
-ret=$?
-
-if [ $ret -ne 0 ]; then
-    echo "No Zookeeper Pod in your cluster"
-else
-    # wait until all zookeeper are ready.
-    PYTHONPATH="../../../deployment" python -m  k8sPaiLibrary.monitorTool.check_pod_ready_status -w -k app -v zookeeper || exit $?
-fi
-
+# wait until all zookeepers are ready.
```
ydye: Before the ready check, we'd better sleep a minutes

YitongFeng: ok I think a minutes is a bit long, I set it to 30s

PR #1525, opened 2018-10-16
[doc] fix errors

ivanchean:


yqwang-ms:
Thanks @ivanchean for this PR, however it is not an error.
The row is for "Batch Job without **Task** Fault Tolerance", so we disabled the retry for **Task**, i.e. 
FancyRetryPolicy = false
MaxRetryCount = 0

However, **Framework** retry still can be enabled as:
FancyRetryPolicy = true
MaxRetryCount = 3

PR #1534, opened 2018-10-16
[Web Portal] support select and stop multiple jobs

sunqinzheng:
[Web Portal] support select and stop multiple jobs

- joi 13.0.2 needs node >= 8.9.0, so I remove the engines requirements

### DIFF
--- a/src/webportal/src/app/job/job-view/job-view.component.js
+++ b/src/webportal/src/app/job/job-view/job-view.component.js
@@ -23,6 +23,10 @@ require('datatables.net/js/jquery.dataTables.js');
 require('datatables.net-bs/js/dataTables.bootstrap.js');
 require('datatables.net-bs/css/dataTables.bootstrap.css');
 require('datatables.net-plugins/sorting/natural.js');
+require('datatables.net-buttons-bs/js/buttons.bootstrap.js');
+require('datatables.net-buttons-bs/css/buttons.bootstrap.css');
+require('datatables.net-select-bs/js/select.bootstrap.js');
+require('datatables.net-select-bs/css/select.bootstrap.css');
 require('./job-view.component.scss');
 const url = require('url');
 // const moment = require('moment/moment.js');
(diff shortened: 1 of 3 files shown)


Gerhut:
Hi @sunqinzheng 
You could restore `engines` field and use `npm run yarn` (which stands for `yarn --ignore-engines`) as a workaround.

I have already opened #1453 to solve this.

src/webportal/src/app/job/job-view/job-view.component.js
```diff
@@ -286,11 +290,43 @@ const loadJobs = (specifiedVc) => {
     ],
     'deferRender': true,
     'autoWidth': false,
+    'dom': '<\'row\'<\'col-sm-6\'B><\'col-sm-6\'f>>' +
+      '<\'row\'<\'col-sm-12\'tr>>' +
+      '<\'row\'<\'col-sm-5\'i><\'col-sm-7\'p>>',
+    'buttons': [
+      {
+        text: 'Select all',
```
Gerhut: I think the "Select all" button label is confused here, since it actually select all jobs in current page.

src/webportal/src/app/job/job-view/job-view.component.js
```diff
@@ -286,11 +290,43 @@ const loadJobs = (specifiedVc) => {
...
+        action: function(e, dt) {
+          dt.rows({page: 'current'}).select();
+        },
+      },
+      'selectNone',
+      {
+        text: 'Stop',
+        action: function( e, dt, node, config ) {
+          const rows = dt.rows({selected: true});
+          const res = confirm(`Are you sure to stop ${rows.count()} jobs?`);
```
Gerhut: In fact, enable multi-selection in paged list is hard to "sync the UI state with user" in a limited interface, especially multi-stop such a dangerous operation. User often do not know which jobs they selected after a long page traversing/ordering/filtering. Although there is a count is displayed, when the count becomes large, it turns less and less meaningful to user, unless they already have a target count number.

Therefore, could you describe the common case that multi-stop is helpful? I propose we could add some limits/notices to this operation:

1. User can only stop jobs in current page;
2. Limit the maximum job count user can stop;
3. List all selected jobs in the confirm dialog;
4. Selected job always displayed in the table.

sunqinzheng: Some users reported that they need a job management page to stop all jobs with a specific tag like interns' jobs, low priority jobs...
We can implement this by
  1. add tag to the job name
  2. search jobs with the tag
  3. stop all jobs in the search result.

sunqinzheng: Now all jobs will be deselected when page/search changed.

src/webportal/src/app/job/job-view/job-view.component.js
sunqinzheng: Renamed as 'Select displayed jobs'

src/webportal/src/app/job/job-view/job-view.component.js
```diff
@@ -322,7 +322,14 @@ const loadJobs = (specifiedVc) => {
     'select': {
       style: 'multi',
     },
+    'search': {
+      regex: true,
```
Gerhut: Shell we open a new PR to discuss this? At least it should be recorded to changelog separately, so that user may be confused when they are searching jobs including dots which are legal in job names.

PR #1540, opened 2018-10-17
Guide message for admin who first time use machine update.

ydye:
If kubernetes is deployed by the new paictl, the node configuration will be created after the deploymend.
But for the cluster deployed by the old version paictl, the node configuration may be missed. So add this messge to help admin to operate the cluster.

### DIFF
--- a/deployment/k8sPaiLibrary/maintainlib/update.py
+++ b/deployment/k8sPaiLibrary/maintainlib/update.py
@@ -107,6 +107,10 @@ class update:
     def get_node_config_from_k8s(self):
         self.logger.info("Try to get node configuration from kubernetes' configmap.")
         configmap_data = kubernetes_handler.get_configmap(self.kube_config_path, "pai-node-config", "kube-system")
+
+        if configmap_data == None:
+            return None
+
         pai_node_list = configmap_data["data"]["node-list"]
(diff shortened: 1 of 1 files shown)


deployment/k8sPaiLibrary/maintainlib/update.py
```diff
@@ -245,9 +249,33 @@ def run(self):
         self.node_list_from_k8s = self.get_node_list_from_k8s_api()
         self.node_config_from_cluster_conf = self.get_node_config_from_cluster_configuration()
         self.node_config_from_k8s = self.get_node_config_from_k8s()
+        if self.node_config_from_k8s == None:
+            self.logger.warning("No node configuration is found in your kubernetes cluster.")
+            self.logger.warning("Paictl will replace it with the node configuration from cluster-configuration.")
+            self.logger.warning("Suggestion: Due to the first you execute this command.")
+            self.logger.warning("Suggestion: This time please don't change (Remove or add) the machine list in your configuration.")
```
hao1939: User may be confused, it's may need to clarify the difference between`node configuration` and `cluster-configuration`.

ydye: User will never know there is a machine-list.

PR #1544, opened 2018-10-17
add job Envs

### LINKED ISSUE #1527: Feature request: support job parameters
Add an entry for user customized job parameters, so user can pass parameters to jobs.
For example: 

- batch_size:
- epoch:
- any other custom parameters

hao1939:
fix #1527.

### DIFF
--- a/src/rest-server/src/models/job.js
+++ b/src/rest-server/src/models/job.js
@@ -403,6 +403,14 @@ class Job {
     for (let i = 0; i < data.taskRoles.length; i ++) {
       tasksNumber += data.taskRoles[i].taskNumber;
     }
+    let jobEnvs = '';
+    if (data.jobEnvs) {
+        for (let key in data.jobEnvs) {
+            if (data.jobEnvs.hasOwnProperty(key)) {
+                jobEnvs = jobEnvs.concat(key, '=', data.jobEnvs[key], '\n');
+            }
+        }
+    }
     const yarnContainerScript = mustache.render(
         yarnContainerScriptTemplate, {
           'idx': idx,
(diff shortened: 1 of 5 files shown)


hao1939:
The UI looks like below:
image

src/rest-server/src/models/job.js
```diff
@@ -403,6 +403,14 @@ class Job {
     for (let i = 0; i < data.taskRoles.length; i ++) {
       tasksNumber += data.taskRoles[i].taskNumber;
     }
+    let jobEnvs = '';
+    if (data.jobEnvs) {
+        for (let key in data.jobEnvs) {
+            if (data.jobEnvs.hasOwnProperty(key)) {
+                jobEnvs = jobEnvs.concat(key, '=', data.jobEnvs[key], '\n');
```
Gerhut: No encode needed?

hao1939: There are indeed injection risks, javascript injection and bash injection.
I will take more investigation.

hao1939: It's some kind of complicated.
We would pass the env.list to docker command, which would quote. 

So if we quote it with single/double quote, the value would corrupt, as below:
https://github.com/Microsoft/pai/blob/4491119258197ccd6700f295b3ff518ed4a83c81/src/rest-server/src/templates/yarnContainerScript.mustache#L277


```
hayua@stcvl-131:~$ cat env.list
double_quote="with double quote"
single_quote='with single quote'
hayua@stcvl-131:~$ sudo docker run -it --env-file env.list ubuntu env |grep quote
double_quote="with double quote"
single_quote='with single quote'
```

hao1939: Hi @Gerhut,
Now the "env list' was passing to 'docker run', which would take care of 'encoding and escaping'.
We should pass it as it is.

hao1939:
Not yet, I will add doc.

PR #1549, opened 2018-10-18
OCR opensource example

mzmssg:


examples/ocr-serving/copied_file/request_ocr/ocrLibrary/OcrHandle.py
```diff
@@ -0,0 +1,111 @@
...
+    elif methodName == "pyocr":
+        return useOcrPyocr
+    elif methodName == "api":
+        return useOcrApi
+    else:
+        logger.error("invalid methodName: {}".format(methodName))
+        return None
+
+
+def useOcrTesserocr(imgPath):
```
YitongFeng: tesserocr & pyocr are all python wrapper of Tesseract ocr model, why use two of them?

examples/ocr-serving/copied_file/request_ocr/app.py
```diff
@@ -0,0 +1,97 @@
...
+    return None
+
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument("-p", "--port", dest="port", default=5000, help="Listen port, default is 5000")
+    parser.add_argument("-m", "--method", dest="method", default="pyocr",
+                        help="Ocr method, choose from [tesserocr, pyocr, api], default is pyocr")
```
YitongFeng: It's better to explain the arguments in the document as well.

examples/ocr-serving/copied_file/request_ocr/app.py
```diff
@@ -0,0 +1,97 @@
...
+               }
+    return render_template("index.html", **context)
+
+
+@app.route("/", methods=["POST"])
+def uploadimage():
+    try:
+        img = request.files["filechoose"]
+        base, ext = os.path.splitext(img.filename)
+        base += time.strftime("_%Y%m%d%H%M%S_", time.localtime(time.time())) + getRandomStr(8)
```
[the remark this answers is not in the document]
mzmssg: A request may be sent multi times, second level filename isn't enough to distinguish them. There might be a writing race condition.

examples/ocr-serving/copied_file/request_ocr/ocrLibrary/OcrHandle.py
mzmssg: There is a little performance difference for this two wrapper, I think it is caused by some default values, but no detailed docs for it, so I keep them both.

examples/ocr-serving/copied_file/request_ocr/app.py
mzmssg: Added

PR #1564, opened 2018-10-19
[hotfix]increase prometheus memory limits (data keep 1 week)

hao1939:
Increase prometheus memory limits to 10, keep data for one week.


hao1939:
Hi @fanyangCS ,

Yes, the required memory for SingleBox won't increase, because we set the 'requests' to '256Mi'.
Yeah, 10G won't guarantee to keep 1-week data, the estimation is based on our prod-bed.

hao1939:
Hi @YanjieGao ,
I updated the wiki, the prometheus memory limits was increased to 10G.

PR #1584, opened 2018-10-24
[WIP] add a utility to convert pai share format to frameworkcontroller

xudifsd:
References:
* [Protocol description](https://github.com/Microsoft/pai/blob/50a427b258f1f37b171e63672f26ab1f0aeb6c3f/docs/marketplace-and-submit-job-v2/marketplace-and-submit-job-v2.md)
* [Current implementation for pai](https://github.com/Microsoft/pai/blob/master/src/rest-server/src/middlewares/jobV2.js)

We are building frameworkcontroller. The controller needs different job format to pai and pai-share, to utilize pai-share format, we need a converter to convert job description to one that frameworkcontroller can recognize.

Some fields in pai-share is not elegant enough, for example, current implementation for parsing uri from prerequisite is ugly and not general(Using , to partition multiple uri. Using github to trigger git clone).

Also I think we should merge this converter with [current converter](https://github.com/Microsoft/pai/blob/master/src/rest-server/src/middlewares/jobV2.js), this can reduce code we maintain and possible converter behavior conflicts.

Finished functionality of converter, needs to test some pai-share jobs in frameworkcontroller. Also need to add more test cases.

### DIFF
--- a/src/utilities/to_crd.py
+++ b/src/utilities/to_crd.py
@@ -385,7 +385,7 @@ class Job(object):
         restart_policy = "Never"
 
         result = {
-                "apiVersion": "launcher.microsoft.com/v1",
+                "apiVersion": "frameworkcontroller.microsoft.com/v1",
                 "kind": "Framework",
                 "metadata": {"name": self.name},
                 "spec": {
--- a/src/utilities/to_crd.py
+++ b/src/utilities/to_crd.py
@@ -1,0 +1,423 @@
+#!/usr/bin/env python
+
+import sys
+import yaml
+import json
+import codecs
+import re
+import logging
+import attr
+import urlparse
+
+logger = logging.getLogger(__name__)
+
+
+def get_required_field(obj, field, err_msg):
+    if obj.get(field) is None:
+        raise RuntimeError(err_msg)
+    return obj[field]
+
+
+@attr.s
+class Prerequisite(object):
+    name = attr.ib() # required
+    type = attr.ib() # required
+    contributor = attr.ib() # optional
+    desc = attr.ib() # optional
+    uri = attr.ib() # required
+
+    @staticmethod
+    def get_or_fail(obj, key):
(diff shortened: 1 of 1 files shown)


yqwang-ms:
So, what is the roadmap for the conversion tool?
Seems this PR have not converted PAI Share API to CRD?

src/utilities/to_crd.py
```diff
@@ -0,0 +1,423 @@
...
+    def to_crd(self):
+        desc = self.desc
+        if desc is None:
+            desc = "job of " + self.name
+
+        retry_policy = {"fancyRetryPolicy": False}
+        restart_policy = "Never"
+
+        result = {
+                "apiVersion": "launcher.microsoft.com/v1",
```
yqwang-ms: API is changed:
apiVersion: frameworkcontroller.microsoft.com/v1
, please refer https://github.com/Microsoft/frameworkcontroller/tree/master/example/framework

I will update int bed controller now

xudifsd:
> So, what is the roadmap for the conversion tool?
> Seems this PR have not totally converted PAI Share API to CRD.
> Such as cannot find volume related spec.

I'm not sure when to mount a volume, controller uses volume in NFS, but pai-share doesn't support NFS, I thinks this is what launcher's format doesn't compatible with pai-share's, since from pai-share's perspective, NFS is not reproducible.

For roadmap, I think we should be able to run all jobs described in pai-share, and later we should decide whether let this conversion tool and jobV2.js coexist or not. Maybe we will modify pai-share's format to make it more elegant, and being able to express NFS mount.

yqwang-ms:
NFS is not reproducible? Do they use storage, like HDFS, HDFS is reproducible?
We can sync all these in the meeting.

xudifsd:
I do not know their opinion, if we want to express NFS volume, we need to modify protocol of pai-share, like adding nfs prerequisite.

PR #1589, opened 2018-10-24
docs: tutorial of autobuild docker

debuggy:
A tutorial for autobuild docker which could make user maintain a dockerfile to build docker image automatically.


examples/Dockerfiles/autobuild_docker/README.md
```diff
@@ -0,0 +1,38 @@
+# How to maintain docker image by using dockerfile 
+
```
YanjieGao: it is better to add a table of contents

debuggy: added

PR #1601, opened 2018-10-26
[AKS] api-server RBAC access for prometheus & watchdog

YanjieGao:
As api-server over AKS need RBAC access, make prometheus & watchdog leverage service acccount of K8s to access api-server for this case.


src/watchdog/src/watchdog.py
```diff
@@ -241,11 +241,17 @@ def _map_fn(item):
...
 
 
-def collect_healthz(gauge, histogram, service_name, address, port, url):
+def collect_healthz(gauge, histogram, service_name, address, port, url, tls, ca_path, bearer_path):
     with histogram.time():
         error = "ok"
         try:
-            error = requests.get("http://{}:{}{}".format(address, port, url)).text
+            if tls:
+                with open(bearer_path, 'r') as bearer_file:
```
xudifsd: Please, read the bearer in main, and pass content to `collect_healthz` function.

If you implement this way, it will read the file many times in runtime, which incur IO.

src/watchdog/src/watchdog.py
```diff
@@ -340,9 +346,15 @@ def load_machine_list(configFilePath):
         return yaml.load(f)["hosts"]
 
 
-def request_with_histogram(url, histogram):
+def request_with_histogram(url, histogram, tls, ca_path, bearer_path):
     with histogram.time():
-        return requests.get(url).json()
+        if tls:
+            with open(bearer_path, 'r') as bearer_file:
```
xudifsd: Same above, read the file in main and pass content to this function. Since two function needs the same content, I think maybe you should pass headers directly, this save the cost of creating map over and over again.

src/watchdog/src/watchdog.py
```diff
@@ -418,6 +433,9 @@ def main(args):
     parser.add_argument("--interval", "-i", help="interval between two collection", default="30")
     parser.add_argument("--port", "-p", help="port to expose metrics", default="9101")
     parser.add_argument("--hosts", "-m", help="yaml file path contains host info", default="/etc/watchdog/config.yml")
+    parser.add_argument("--tls", "-t", help="access api-server auth or not", default="False")
```
xudifsd: Maybe remove tls argument? Since if user call with --ca and --bearer it implies they want to use tls.

src/watchdog/src/watchdog.py
```diff
@@ -363,6 +375,9 @@ def main(args):
     parse_result = urlparse.urlparse(address)
     api_server_ip = parse_result.hostname
     api_server_port = parse_result.port or 80
+    tls = args.tls
+    ca_path = args.ca
+    bearer_path = args.bearer
```
xudifsd: maybe should check if ca_path bearer_path is either both provided or not.

YanjieGao: check at each method

YanjieGao: checked

src/watchdog/src/watchdog.py
YanjieGao: remove

src/watchdog/src/watchdog.py
YanjieGao: pass headers

src/watchdog/src/watchdog.py
YanjieGao: pass content

src/watchdog/deploy/watchdog.yaml.template
```diff
@@ -60,11 +60,19 @@ spec:
         command:
           - "python"
           - "/watchdog.py"
+          - "--tls"
+          - "{{ clusterinfo['api-server']['tls_access'] }}"
```
xudifsd: remove this

deployment/quick-start/services-configuration.yaml.template
```diff
@@ -41,6 +41,10 @@ cluster:
     # Must be lower case, e.g., regsecret.
     secret-name: pai-secret
 
+  api-server:
+    # use tls for prometheus access
+    tls_access: false
```
xudifsd: maybe should set a default value in pai object module to false, without default value, user with previous config may get exception in expanding template of prometheus?

src/watchdog/src/watchdog.py
```diff
@@ -241,11 +241,14 @@ def _map_fn(item):
...
 
 
-def collect_healthz(gauge, histogram, service_name, address, port, url):
+def collect_healthz(gauge, histogram, service_name, address, port, url, ca_path, headers):
     with histogram.time():
         error = "ok"
         try:
-            error = requests.get("http://{}:{}{}".format(address, port, url)).text
+            if (ca_path is not None) and (headers is not None):
+                error = requests.get("https://{}:{}{}".format(address, port, url), headers = headers, verify = ca_path).text
```
xudifsd: Maybe `headers` and `verify` have default value None? If this is the case, I thinke we may not need an if else branch, `error = requests.get("https://{}:{}{}".format(address, port, url), headers = headers, verify = ca_path).text` should be ok for both cases.

src/watchdog/src/watchdog.py
```diff
@@ -340,9 +343,12 @@ def load_machine_list(configFilePath):
         return yaml.load(f)["hosts"]
 
 
-def request_with_histogram(url, histogram):
+def request_with_histogram(url, histogram, ca_path, headers):
     with histogram.time():
-        return requests.get(url).json()
+        if (ca_path is not None) and (headers is not None):
```
xudifsd: Same above

src/watchdog/src/watchdog.py
xudifsd: Oh, I mean, you should check before entering the while loop of the main. That way the configuration error can be found before any logic began, it's easier for debug.

YanjieGao:
updated and make it through warning log

src/watchdog/src/watchdog.py
YanjieGao: rm and test it work

src/watchdog/src/watchdog.py
YanjieGao: checked

deployment/quick-start/services-configuration.yaml.template
YanjieGao: added at pai object model

deployment/paiLibrary/clusterObjectModel/paiObjectModel.py
```diff
@@ -194,7 +194,10 @@ def serviceParse(self):
             serviceDict["clusterinfo"]["docker-registry-info"]["docker-tag"]
         serviceDict["clusterinfo"]["dockerregistryinfo"]["secretname"] = \
             serviceDict["clusterinfo"]["docker-registry-info"]["secret-name"]
-
+        serviceDict["clusterinfo"]["api-server"]["tls_access"] = False
```
xudifsd: what this means? `serviceDict["clusterinfo"]` has `api-server` as a map?

deployment/paiLibrary/clusterObjectModel/paiObjectModel.py
```diff
@@ -194,7 +194,10 @@ def serviceParse(self):
...
+        if ("api-server" in serviceDict["clusterinfo"]) and ("tls_access" in serviceDic["clusterinfo"]["api-server"]):
+            serviceDict["clusterinfo"]["apiserver"] = serviceDic["clusterinfo"]["api-server"]
```
xudifsd: what this means? change the name? I see template uses key of `api-server` instead of `apiserver`

src/watchdog/src/watchdog.py
```diff
@@ -363,6 +363,15 @@ def main(args):
     parse_result = urlparse.urlparse(address)
     api_server_ip = parse_result.hostname
     api_server_port = parse_result.port or 80
+    ca_path = args.ca
+    bearer_path = args.bearer
+    if (ca_path is None and bearer_path is not None) or (ca_path is not None and bearer_path is None):
+        log.warning("please probide bearer_path and ca_path at the same time or not")
```
xudifsd: `probide` -> `provide`

src/watchdog/src/watchdog.py
```diff
@@ -267,7 +267,7 @@ def collect_k8s_componentStaus(k8s_gauge, api_server_ip, api_server_port, nodesJ
         ip = name["metadata"]["name"]
 
         collect_healthz(k8s_gauge, kubelet_healthz_histogram,
-            "k8s_kubelet", ip, 10255, "/healthz")
+            "k8s_kubelet", ip, 10255, "/healthz", None, None)
```
xudifsd: are you sure AKS kubelet doesn't requires authentication?

YanjieGao: Sure

src/watchdog/src/watchdog.py
```diff
@@ -340,9 +340,9 @@ def load_machine_list(configFilePath):
         return yaml.load(f)["hosts"]
 
 
-def request_with_histogram(url, histogram):
+def request_with_histogram(url, histogram, ca_path, headers):
     with histogram.time():
-        return requests.get(url).json()
+        requests.get(url, headers = headers, verify = ca_path).json()
```
xudifsd: why removing return?

YanjieGao: added

deployment/quick-start/services-configuration.yaml.template
```diff
@@ -41,6 +41,10 @@ cluster:
     # Must be lower case, e.g., regsecret.
     secret-name: pai-secret
 
+  api-server:
```
ydye: Service configuration? I don't think this section should be here.

YanjieGao: which position you think is better?

PR #1605, opened 2018-10-26
move vc configuration to a separate configmap

mzmssg:
then other service, such as rest-server could update vc by changing this configmap


fanyangCS:
like the user info, maybe you should provide a tool to export VC config info for backup purpose.

mzmssg:
>           like the user info, maybe you should provide a tool to export VC config info for backup purpose.

Yeah, this PR only include configmap reading parts, there will be a configmap writing PR to complete vc runtime updating.

mzmssg:
Close firstly, because paictl have some refactor, this PR doesn't work and will **block Jenkins**. Will reopen it when fixing

Issue #1581, opened 2018-10-23
Cannot mount hdfs


qyyy:
According to this [document](https://github.com/Microsoft/pai/blob/3f7190dd185c75ca42a38a03a3c9def172798106/docs/hadoop/hdfs.md#mountable-hdfs-), I try to install the Hadoop-hdfs-fuse by `apt-get install hadoop-hdfs-fuse` when I build a docker image. But I failed. The error message is following:
```
E: The method driver /usr/lib/apt/methods/https could not be found.
W: http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1604/x86_64/Release.gpg: The key(s) in the keyring /etc/apt/trusted.gpg.d/cloudera-cdh5.gpg are ignored as the file is not readable by user '_apt' executing apt-key.
W: http://developer.download.nvidia.com/compute/machine-learning/repos/ubuntu1604/x86_64/Release.gpg: The key(s) in the keyring /etc/apt/trusted.gpg.d/cloudera-cdh5.gpg are ignored as the file is not readable by user '_apt' executing apt-key.
W: http://security.ubuntu.com/ubuntu/dists/xenial-security/InRelease: The key(s) in the keyring /etc/apt/trusted.gpg.d/cloudera-cdh5.gpg are ignored as the file is not readable by user '_apt' executing apt-key.
W: http://archive.ubuntu.com/ubuntu/dists/xenial/InRelease: The key(s) in the keyring /etc/apt/trusted.gpg.d/cloudera-cdh5.gpg are ignored as the file is not readable by user '_apt' executing apt-key.
W: http://archive.ubuntu.com/ubuntu/dists/xenial-updates/InRelease: The key(s) in the keyring /etc/apt/trusted.gpg.d/cloudera-cdh5.gpg are ignored as the file is not readable by user '_apt' executing apt-key.
W: http://archive.ubuntu.com/ubuntu/dists/xenial-backports/InRelease: The key(s) in the keyring /etc/apt/trusted.gpg.d/cloudera-cdh5.gpg are ignored as the file is not readable by user '_apt' executing apt-key.
E: Failed to fetch https://archive.cloudera.com/cdh5/ubuntu/trusty/amd64/cdh/dists/trusty-cdh5/InRelease
E: Some index files failed to download. They have been ignored, or old ones used instead.
[...]

PR #1609, opened 2018-10-29
change hadoop config directory

DongZhaoYu:
change the hadoop config directory path to avoid the conflict with hdfs-fuse tool.


DongZhaoYu:
> Could you help explain what the conflict is and why renaming this dir resolve it?

@mzmssg the /etc/hadoop is mounted as readonly. When installing hdfs-fuse tool it needs to write its own config under this path. So it will fail now.  Currently the configuration files are only needed by spark job. And the path can be customized by the HADOOP_CONF_DIR env variable in spark. So it is OK to change the dir path. And fuse tool can create the /etc/hadoop dir by its own and write the its config files there and wouldn't conflict with spark job.

fanyangCS:
@DongZhaoYu do you have an issue created for this problem?

DongZhaoYu:
> @DongZhaoYu do you have an issue created for this problem?

@fanyangCS , the related issue is https://github.com/Microsoft/pai/issues/1581

fanyangCS:
> > Could you help explain what the conflict is and why renaming this dir resolve it?
> 
> @mzmssg the /etc/hadoop is mounted as readonly. When installing hdfs-fuse tool it needs to write its own config under this path. So it will fail now. Currently the configuration files are only needed by spark job. And the path can be customized by the HADOOP_CONF_DIR env variable in spark. So it is OK to change the dir path. And fuse tool can create the /etc/hadoop dir by its own and write the its config files there and wouldn't conflict with spark job.

@DongZhaoYu and @mzmssg, 

Is it possible to make the HADOOP_CONFIG_DIR RW instead of RO? 
In order not to let a docker instance to change the content on host file, we can copy the hadoop file into the Docker instance in the docker container script.

src/rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -243,6 +243,11 @@ fi
 mkdir -p "/tmp/pai-root/log/$APP_ID/$CONTAINER_ID"
 ln -s /tmp/pai-root/log/$APP_ID/$CONTAINER_ID/DockerContainerDebug.log $LAUNCHER_LOG_DIR/DockerContainerDebug.log
 
+# copy the hadoop configuration file
+hadoop_config_dir="hadoop_config"
+mkdir -p $hadoop_config_dir
```
fanyangCS: Do you need to remove it after the job is done?

DongZhaoYu: @fanyangCS I think we would better keep the configure files. First the configure file dir is in the yarn local folder. Yarn cleaner will delete it after a period of time. Currently in our setting it will be deleted after 86400 seconds. This allow us to debug the configure if the job fails. Secondly the configure file size is small, about 200K. Unlike the code, it doesn't take much space.

fanyangCS: Make sense. by the way, the current approach has one side effect: if we change the Hadoop config after the job launches, the Docker container's Hadoop will be outdated. If the Hadoop read failed, it may not retry since the system thinks this is user failures.
@mzmssg and @sterowang , should we handle this case?

sterowang: Another potential practice that we still mount the original hadoop config as read only, but we mount it into another path such as /etc/hadoop-dist, and in docker container script, we copy it to /etc/hadoop. By this way, in docker container, user would have possibilities to update hadoop config changes again if /etc/hadoop-dist get chance updated.

src/rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -276,7 +281,7 @@ docker run --name $docker_name \
   --volume $container_local_dir/$bootstrap_dir:/pai/bootstrap:ro \
   --volume $container_local_dir/$code_dir:/pai/code:ro \
   --volume /var/drivers/nvidia/current:/usr/local/nvidia:ro \
-  --volume /etc/hadoop-configuration-for-jobs:/etc/hadoop:ro \
+  --volume $container_local_dir/$hadoop_config_dir:/etc/hadoop \
   --volume $container_local_dir/$ssh_config_file:/pai/ssh_config/ \
```
sterowang: Just noticed this line of code, should we make /pai/ssh_config also readonly @wangcan0329?

src/rest-server/src/templates/yarnContainerScript.mustache
DongZhaoYu: > Another potential practice that we still mount the original hadoop config as read only, but we mount it into another path such as /etc/hadoop-dist, and in docker container script, we copy it to /etc/hadoop. By this way, in docker container, user would have possibilities to update hadoop config changes again if /etc/hadoop-dist get chance updated.

copy the configures to /etc/hadoop in docker container is hard for the user to debug the misconfigure problem when job fails. Since if the the configure is changed by the job and when it fails and exit, the configure files will lost. If we want the user job to sync the latest configure, we can further mount the host configure dir as readonly into the docker container as /etc/hadoop-configure-host. So user job can decide how to do the sync. But this is tricky. Do we need to support this?

src/rest-server/src/templates/yarnContainerScript.mustache
DongZhaoYu: it makes sense to mount it as readonly.

src/rest-server/src/templates/yarnContainerScript.mustache
mzmssg: I think current solution is a little tricky(Copy from NM pod to host, then host to job container). Because NM config does only contain NM related fields, which may be not enough for users.

I'm not sure which field Spark on Yarn needed, if only RM address and HDFS address needed, I think we could pass them by ENVs, then users generate config by themselves.

src/rest-server/src/templates/yarnContainerScript.mustache
wangcan0329: Sure. It's better to mount it as readonly, I will fix it.

src/rest-server/src/templates/yarnContainerScript.mustache
DongZhaoYu: From the doc https://github.com/apache/spark/blob/master/docs/running-on-yarn.md . The configuration file directory is required to run spark on yarn. If we store a copy on hdfs. User still need to sync it if we change it.

PR #1613, opened 2018-10-29
update jobEnvs doc

hao1939:
Update job tutorial, add doc about `jobEnvs`.


fanyangCS:
did you test the job example?

hao1939:
Sure, I will test it on int-bed.

PR #1615, opened 2018-10-29
[example]horovod example

qyyy:
Add horovod example.
Add auto-test support.


examples/mpi/prepare.sh
```diff
@@ -1,7 +1,7 @@
 #mpi cntk prepare
 echo "Prepare for the mpi example!"
 
-function prepare_data(){
+function mpi_cntk_prepare_data(){
```
fanyangCS: do you need cntk to run the horovod example?

PR #1632, opened 2018-10-31
[Drivers] How to upgrade the drivers version in OpenPAI cluster

ydye:
Before the upgrading process, all service, which may occupy gpu resource, should be stopped.

### DIFF
--- a/docs/drivers/drivers.md
+++ b/docs/drivers/drivers.md
@@ -14,13 +14,13 @@ In the [path](../../src/drivers/build), you can find the available version in th
 drivers-${version}.dockerfile
 ``` 
 
-###### Cluster Configuration
+###### 1 Cluster Configuration
 
 - Choose an available version in Openpai, and change the ```drivers``` section in ```service-configuration.yaml```.  
 
 - Update the configuration in your cluster with the command ```paictl config push```. If you wanna investigate more detail about this command, please refer to this [link](../paictl/paictl-manual.md#Config_Push) 
 
-###### Stop corresponding service
+###### 2 Stop corresponding service
 
 - Delete hadoop-node-manager
     - Note: all running jobs will be killed and retry after upgrading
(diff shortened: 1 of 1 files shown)


docs/drivers/drivers.md
```diff
@@ -0,0 +1,59 @@
...
+
+- Choose an available version in Openpai, and change the ```drivers``` section in ```service-configuration.yaml```.  
+
+- Update the configuration in your cluster with the command ```paictl config push```. If you wanna investigate more detail about this command, please refer to this [link](../paictl/paictl-manual.md#Config_Push) 
+
+###### Stop corresponding service
+
+- Delete hadoop-node-manager and delete all jobs
+    - Note: all jobs will be killed
+
```
mzmssg: I think `killed` is a little confusing, actually these job will retry when node-manager is available.

mzmssg: Add more words? Such as `all running jobs will be killed and retry after upgrading`

ydye: Done

docs/drivers/drivers.md
```diff
@@ -0,0 +1,59 @@
+## Drivers Maintenance 
+
+#### How to upgrade the Nvidia drivers in your cluster
+
+
+###### Available GPU drivers version in OpenPAI
+
+```
+"384.111"  (Default Value)
```
YanjieGao: it is better to tell user how to check

docs/drivers/drivers.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+
+###### Cluster Configuration
```
YanjieGao: pls add steps number to let user known he is in which steps

docs/drivers/drivers.md
```diff
@@ -0,0 +1,59 @@
...
+
+- Delete hadoop-node-manager
+    - Note: all running jobs will be killed and retry after upgrading
+       
+```
+./paictl service delete -n hadoop-node-manager
+```
+
+   
+- Stop monitoring service
```
YanjieGao: [optional suggestion] seem later paictl support paictl stop -n [service1, service2, service3 ..], batch start and stop services. Could be useful for admin.

docs/drivers/drivers.md
ydye: Done

docs/drivers/drivers.md
ydye: Done

docs/drivers/drivers.md
ydye: Sure

ydye:
@YanjieGao  Can you check whether the process cover all the service which may occupy gpu resource in the monitoring level?

PR #1642, opened 2018-11-01
[Drivers] Add two new versions

ydye:
Add 2 version drivers.
- 390.25 : This version will become the default drivers version. And it will be deployed and tested into int-bed.
- 410.73 : This the neweast version. Won't test it at this release.


deployment/paiLibrary/clusterObjectModel/paiObjectModel.py
```diff
@@ -323,7 +323,7 @@ def serviceParse(self):
             serviceDict["clusterinfo"]["driversinfo"] = dict()
 
         if "version" not in serviceDict["clusterinfo"]["driversinfo"]:
-            serviceDict["clusterinfo"]["driversinfo"]["version"] = "384.111"
+            serviceDict["clusterinfo"]["driversinfo"]["version"] = "390.25"
```
hao1939: As discussed, we could keep the default as it is.

PR #1646, opened 2018-11-02
Update yarnContainerScript.mustache

xwzheng1020:
add entrypoint=""

in #1435 the `entrypoint` was removed.
but in my experiment, the `--init` will not override the `--entrypoint=""` defaultly
so, if the docker image has set ENTRYPOINT (sth. like ls)，the docker run will fail。


mzmssg:
Do a simple experiment, `--init` will not overwrite `entrypoint`, but append `entrypoint` to the init process. Thus the pid 1 process looks like `/dev/init -- entrypoint command`.

To be more elegant, I think we could add the option `--entrypoint /bin/bash`, then command field keep only `'/pai/bootstrap/docker_bootstrap.sh'`

xwzheng1020:
emm, but i think `--entrypoint="" CMD="/bin/bash /pai/bootstrap/docker_bootstrap.sh" `maybe more clearable than `--entrypoint=/bin/bash CMD="/pai/bootstrap/docker_bootstrap.sh"`

PR #1670, opened 2018-11-07
Dixu/docker daemon check

### LINKED ISSUE #1600: [aks] watchdog execute ssh to node cmd hanging
i test script inner watchdog container.

it is hanging at this funtion [ssh to each node exec docker check cmd]. comment this will not hanging
https://github.com/Microsoft/pai/blob/4f23c079cc4b233a24485af01e5777a8a391ce70/src/watchdog/src/watchdog.py#L400

logs: 
image

i test the cmd directly on that host, not hanging, maybe the python ssh lib lead to this issue?

This is my working branch: https://github.com/Microsoft/pai/blob/yanjga/aks-prometheus/src/watchdog/src/watchdog.py#L416

xudifsd:
close #1600 
Remove hanging by offload docker daemon check from watchdog to job-exporter.


docs/alerting/exporter-metrics.md
```diff
@@ -30,3 +30,4 @@ Some important metrics are listed below.
 | `service_net_out_byte`| `gpu_exporter` | Network out traffic by pai service detectived by docker stats (byte) |
 | `service_block_in_byte`| `gpu_exporter` | Block io in traffic by pai service detectived by docker stats (byte) |
 | `service_block_out_byte`| `gpu_exporter` | Block io out traffic by pai service detectived by docker stats (byte) |
+| `docker_daemon_count` | `gpu_exporter` | Count of docker daemon, count is always 1, but error label indicate what type of error this docker daemon has |
```
YanjieGao: why count is always 1? 1 represent correct state? if this "Count of docker daemon" is wrong?

xudifsd: this is similar to watchdog's metrics, use label to indicate error, the value of metrics is useless in single metrics. This way we can use `sum(docker_daemon_count{"error"!="ok"}) > 0` to do alert.

YanjieGao: Got it, Maybe change to similar explain will be easy understand for user "Used to summary healthy count of docker daemon,  this metrics per server value is always 1, leverage label to filter health metrics and accumulate them to a sum count for watchdog to use ...."

PR #1675, opened 2018-11-07
[Ssh log refine] Add ssh log.

wangcan0329:


src/rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -147,6 +147,10 @@ if service --status-all 2>&1 | grep -q ssh; then
   mv /pai/ssh_config/config /root/.ssh/
   # Start ssh service
   start_ssh_service
+else
+  printf "%s %s\n %s %s \n" \
+    "[WARNING]" "Open-ssh server not found in image will skip prepare ssh info." \
```
fanyangCS: Open-ssh server not found, will *not* enable ssh service.

src/rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -147,6 +147,10 @@ if service --status-all 2>&1 | grep -q ssh; then
...
+    "[INFO]" "You can refer to PAI dockerfile example to get full support: https://github.com/Microsoft/pai/tree/master/examples/Dockerfiles"
```
fanyangCS: Please refer to PAI dockerfile example to get ssh service enabled.

PR #1680, opened 2018-11-08
Zhaoyu/hdfs readme fix

DongZhaoYu:
change the webhdfs request schema

### DIFF
--- a/docs/hadoop/hdfs.md
+++ b/docs/hadoop/hdfs.md
@@ -219,7 +219,7 @@ Please refer [HdfsCLI](https://hdfscli.readthedocs.io/en/latest/) for the detail
 
 # FAQ <a name="FAQ"></a>
 
-1. Why cannot upload data to OpenPAI cluster deployed on Azure? <br>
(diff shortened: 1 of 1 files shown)


docs/hadoop/hdfs.md
```diff
@@ -131,27 +131,28 @@ Data on HDFS can be accessed by various ways. Users can choose the proper way ac
...
+curl -i -X PUT "http://master-node-address/webhdfs/api/v1/test/test_file?op=CREATE"
 ```
 This command will return the data node where the file should be written. The location URI would be like
->http://hdfs-name-node-address:5075/webhdfs/v1/test/test_file?op=CREATE&namenoderpcaddress=hdfs-data-node-address:9000&createflag=&createparent=true&overwrite=false
+>http://master-node-address/a/data-node-address:5075/webhdfs/v1/test/test_file?op=CREATE&namenoderpcaddress=hdfs-name-node-address:9000&createflag=&createparent=true&overwrite=false
 
 Then run following command with this URI to write file data:
 ```bash
-curl -i -X PUT -T file-data-to-write returned-location-uri
+curl -i -X PUT -T file-data-to-write "returned-location-uri"
```
YanjieGao: for returned-location-uri , tell which url is better

Issue #1658, opened 2018-11-02
Some datanode is killed for unknown reason


mzmssg:
171 node
![image](https://user-images.githubusercontent.com/11887940/47908246-c7fe7380-dec7-11e8-9fc9-74f3b42534ac.png)

173 node
![image](https://user-images.githubusercontent.com/11887940/47907610-8ae5b180-dec6-11e8-86f8-b97e2030a0ce.png)

175 node
![image](https://user-images.githubusercontent.com/11887940/47907635-9638dd00-dec6-11e8-87a8-d8d60708b18b.png)

Not sure who killed them. But it might result in hdfs access failure.

fanyangCS:
Is it due to OOF?

mzmssg:
For a failed datanode, it's killed for readiness probe failure
![image](https://user-images.githubusercontent.com/11887940/48131083-cbc83680-e2c9-11e8-84b2-594b8ceaf235.png)

mzmssg:
Seems due to our resource limitation, docker events show it's in oom status
```
Type=container  Status=exec_create: cat /jobstatus/jobok  ID=edf6b4b172edb0b46f36e2d5caa2b89e5bdd6b8403661593d9a9d1e50ca5a28f
Type=container  Status=exec_start: cat /jobstatus/jobok  ID=edf6b4b172edb0b46f36e2d5caa2b89e5bdd6b8403661593d9a9d1e50ca5a28f
Type=container  Status=oom  ID=2dc13c49a71342b0826940f6a9247e8120dba06f97a98d96117117226bf754c8
Type=container  Status=die  ID=2dc13c49a71342b0826940f6a9247e8120dba06f97a98d96117117226bf754c8
Type=container  Status=destroy  ID=8fc03a7ab925eaf703302af02caf218041b0dcdc2d7fbf357b6b0eb079840ebb
Type=container  Status=exec_create: cat /jobstatus/jobok  ID=edf6b4b172edb0b46f36e2d5caa2b89e5bdd6b8403661593d9a9d1e50ca5a28f
```

fanyangCS:
@mzmssg , can you explain it in details? whose readiness probe? I am not aware that we have a readiness check in DataNode. will k8s kill a pod when it is not ready? I think k8s will just show that pod is not ready, but not kill it?
where is this the log from,  kubelet? Also, where is the oom log from?

mzmssg:
@fanyangCS
[...]

PR #1689, opened 2018-11-09
[hot-fix] increase data-node memory limits to 4Gi

hao1939:
Increase data-node memory limits to 4Gi.

### DIFF
--- a/src/hadoop-node-manager/deploy/hadoop-node-manager-configuration/nodemanager-generate-script.sh
+++ b/src/hadoop-node-manager/deploy/hadoop-node-manager-configuration/nodemanager-generate-script.sh
@@ -62,8 +62,8 @@ mem_total=`cat /proc/meminfo | grep "MemTotal" | awk '{print $2}'`
 # memory size to nodemanager is (mem_total - mem_reserved)
 if [ $(grep 'ip:' /host-configuration/host-configuration.yaml|wc -l) -gt 1 ]
 then
-    echo "Node role is 'Worker'. Reserve 12G for os and k8s."
-    let mem_reserved=12*1024
+    echo "Node role is 'Worker'. Reserve 16G for os and k8s."
+    let mem_reserved=16*1024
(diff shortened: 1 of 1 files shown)


src/hadoop-data-node/deploy/hadoop-data-node.yaml.template
```diff
@@ -52,6 +52,8 @@ spec:
           periodSeconds: 3
         resources:
           limits:
+            memory: "4Gi"
+          requests:
             memory: "1Gi"
         env:
         - name: HADOOP_DATANODE_OPTS
```
mzmssg: Need to raise JVM memory opts?

mzmssg:
I think we also need to modify here?
https://github.com/Microsoft/pai/blob/6cfcb8a3553a6142ebde8a6a07f4cda6d7239179/src/hadoop-node-manager/deploy/hadoop-node-manager-configuration/nodemanager-generate-script.sh#L66

src/hadoop-data-node/deploy/hadoop-data-node.yaml.template
hao1939: Not sure, currently we haven't found any java limit was hit.

hao1939:
@mzmssg ,

It's node manager limit. How much should it be?

src/hadoop-data-node/deploy/hadoop-data-node.yaml.template
mzmssg: #1658 For 175node screenshot, jave OOM is recorded. So I think a larger JVM opts is needed.

hao1939: Hi @mzmssg , 
What about increase the max heap size to 2G?

hao1939:
Hi @mzmssg , increase the reserved memory on worker to 16G.

PR #1696, opened 2018-11-13
[Deployment] Ignore the error when encoding string to `utf-8` 

ydye:


ydye:
Close this PR. Because we have to propose a complete design to solve all these issues.

PR #1700, opened 2018-11-14
PAI webportal plugin, make marketplace a plugin.

Gerhut:
Some further works to do

1. The `WEBPORTAL_PLUGINS` environment variable needs to be a JSON formatted list, currently we can only serialize the list to JSON string in `paiObjectModel.py`. I will move the serialization to template `webportal.yaml.template` when jinja's `tojson` filter (introduced in jinja 2.9, currently we are using 2.8) is available to use. (Perhaps mounting a configmap is better.)
2. The built-in plugins (now only marketplace) is hard-coded in `paiObjectModel.py`, will move to web portal's default config when new object model is landed.
3. The GitHub owner / repository / path config now belongs to REST server, but not web portal / web portal plugin. So currently I remain the config as it was. Moving the config from REST server to web portal needs more designs in Marketplace.


Gerhut:
Rebased on current object model

PR #1701, opened 2018-11-15
Add a tutorial to guide how to add new service configuraiton into com

ydye:


docs/pai-management/doc/how-to-add-new-conm.md
```diff
@@ -0,0 +1,200 @@
...
+    <td>com["cluster"]["common"]["cluster-id"]</td>
+    <td>string</td>
+</tr>
+</table>
+
+#### How to write your template with cluster object model <a name="#use_link"></a>
+
+
+- You could find all service configuration in corresponding services' document in the path ```src/${service_name}/config/${service_name}.md```.
+- Please refer to [jinja2 document](http://jinja.pocoo.org/docs/2.10/), and learn how to write the final tempalte with the  ```Data in Jinja2 Template```
```
Gerhut: I have something else to discuss, which may be not related to this PR.

The python-jinja2 package installed is from [ubuntu 16.04 official repo](https://packages.ubuntu.com/xenial/python-jinja2) and the version is 2.8, doc link may should be pointed to version 2.8. Unfortunately, it has been removed from [jinja2 official site](http://jinja.pocoo.org/).

Moreover, the [`tojson filter`](http://jinja.pocoo.org/docs/2.10/templates/#tojson) introduced in jinja 2.9 might be useful to some other feartures. So are there some plans to update the version of jinja2?

docs/pai-management/doc/how-to-add-new-conm.md
```diff
@@ -0,0 +1,200 @@
...
+    def validation_pre(self):
+        #   if error: 
+        #      return False, "message"
+
+        return True, None
+    
+    #### Generate the final service object model
+    def run(self):
+        # parse your service object model here, and return a generated dictionary
+        return cluster_com
```
mzmssg: A question, origin ObjectModel contains conversion from `vc` to `hadoopQueue`, now I need to move this logic to the parser?

docs/pai-management/doc/how-to-add-new-conm.md
```diff
@@ -0,0 +1,200 @@
...
+${service_name}:
+    service-mandatory-key1: usable-value1
+    
+```
+
+#### Service Parser <a name="Service"></a>
+
+###### path
+
+Please create the service.parser in the following path.
```
xudifsd: Is this mandatory ?

docs/pai-management/doc/how-to-add-new-conm.md
```diff
@@ -0,0 +1,200 @@
...
+
+```bash
+src/${service_name}/config/${service_name}.py
+```
+###### Specification
+
+You should implement the parse class like following.
+
+```python
+class cluster:
```
xudifsd: The exact name should be matched?

docs/pai-management/doc/how-to-add-new-conm.md
ydye: Path?

docs/pai-management/doc/how-to-add-new-conm.md
ydye: Yes

docs/pai-management/doc/how-to-add-new-conm.md
xudifsd: No, service.parser. If I didn't need parser, should I also create this?

ydye: No. If you don't need parser, you could ignore this

docs/pai-management/doc/how-to-add-new-conm.md
xudifsd: Then maybe use `Cluster`? Since more compliant with [pep8](https://www.python.org/dev/peps/pep-0008/#class-names)

ydye: Update

Issue #1551, opened 2018-10-18
[AKS] api-server auth access 


YanjieGao:
## Problem: 
PAI services need query api-server data for monitor and metrics. api-server permission & auth problem caused some services cannot directly read the api-server data.
Problem screenshot:
Prometheus: x509: certificate is valid for apiserver
Service view no data
Watchdog can't access api-server
 
## Challenge:
It is better to provide a common solution that each service can share. If not, each service is configured by itself.
 
## Pai service involved: 
Prometheus & Weportal & watchdog
 
## Proposal & current status: 

Leverage k8s RBAC & service (see reference) account to access api-server.
o	For prometheus:
Tested and create a PR
Code location: https://github.com/Microsoft/pai/blob/master/src/prometheus/deploy/prometheus-configmap.yaml.template
Solution: https://github.com/Microsoft/pai/pull/1458
o	For webportal
Code location: https://github.com/Microsoft/pai/blob/master/src/webportal/src/app/cluster-view/services/services.component.js#L44
Solution: service account + curl (attach token file) tested passed, haven't test under js script.
o	For watchdog
Code location: https://github.com/Microsoft/pai/blob/master/src/watchdog/src/watchdog.py#L244
Solution: service account + curl (attach token file) tested passed, haven't test under python script.
•	Need to do:
 
1.	Try weather contains more general solutions to avoid configuration for each service component
2.	If there is no better general method, implement the corresponding solution for each service.
•	Reference:
o	K8s RBAC: https://kubernetes.io/docs/reference/access-authn-authz/rbac/
o	K8s service account: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/

## Dev plan
- yanjie will create a pai/aks branch as master branch for receive all aks related pr
https://github.com/Microsoft/pai/tree/pai/aks
- some knowledge could refer yanjga/aks-test driver related change
- feature dev folk & submit pr to pai/aks branch
- If all tests passed, then cherry pick pai/aks pr to master branch.

YanjieGao:
After discussion with @Gerhut , we prefer method1 to solve this problem and qixiang is help to try this solution.

 the program summary:
 
3 proposals:
all use service account token.
method1: rest-server as proxy to attach token to api-server request call
method2: give token to client side, client side call service-view request attach tokens 
method3: remove service-view page.


this is also related role binding issue. Related discussion is at:
https://github.com/Microsoft/pai/issues/1709

YanjieGao:
[...]

PR #1710, opened 2018-11-16
[AKS] api-server RBAC access for watchdog

YanjieGao:
As api-server over AKS need RBAC access, make prometheus & watchdog leverage service acccount of K8s to access api-server for this case.

@xudifsd @ydye ,mv the previous reviewd pr to submit to pai/aks

- Previous reviewed PR:
https://github.com/Microsoft/pai/pull/1601
- Related tracking issue:
https://github.com/Microsoft/pai/issues/1551


deployment/quick-start/services-configuration.yaml.template
```diff
@@ -41,6 +41,10 @@ cluster:
     # Must be lower case, e.g., regsecret.
     secret-name: pai-secret
 
+  api-server:
```
ydye: Don't put this configuration here. It's not a service.

YanjieGao: this is flag will be used by prometheus / watchdog / webportal. Do you have suggestions it is better to put this config to which location and file?

ydye: kubernetes-configuraiton

src/prometheus/deploy/prometheus-configmap.yaml.template
```diff
@@ -36,6 +36,13 @@ data:
     scrape_configs:
     - job_name: 'node_exporter'
       scrape_interval: {{ prom_info['scrape_interval']|default(30) }}s
+      {% if clusterinfo['api-server']['tls_access'] -%}
+      tls_config:
+        ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
```
fanyangCS: >/var/run/secrets/kubernetes.io/serviceaccount/ca.crt [](start = 17, length = 52)

maybe you should make the path to to keep the ca file and secretes configurable. Remember AKS is not the only k8s deployment you need to support.

fanyangCS: This change not only affect Prometheus and watchdog. It makes some assumption to the underlying host environment. 

I suggest you to break this into multiple PRs. 1. one makes the CA/secretes path configurable. 2. the other one allows Prometheus to specify the path to CA/Secretes. 3. the final one does the configuration for watchdog.

---
In reply to: [234093754](https://github.com/Microsoft/pai/pull/1710#discussion_r234093754) [](ancestors = 234093754)

YanjieGao: Will split it and resubmit

YanjieGao: @ydye how about define the path env and which i can get the path from the env.

YanjieGao:
@xudifsd, @ydye @fanyangCS  
Seperate current PR to 2:
1 watchdog api-server access related is this PR
2 prometheus related is https://github.com/Microsoft/pai/pull/1875

will add unified token path after merge to new cluster object model branch

src/watchdog/src/watchdog.py
```diff
@@ -353,6 +362,9 @@ def main(args):
     parser.add_argument("--log", "-l", help="log dir to store log", default="/datastorage/prometheus")
     parser.add_argument("--interval", "-i", help="interval between two collection", default="30")
     parser.add_argument("--port", "-p", help="port to expose metrics", default="9101")
+    parser.add_argument("--hosts", "-m", help="yaml file path contains host info", default="/etc/watchdog/config.yml")
```
xudifsd: what's this? I didn't see any usage.

YanjieGao: removed

src/watchdog/src/watchdog.py
```diff
@@ -204,11 +204,11 @@ def _map_fn(item):
     map(_map_fn, podsJsonObject["items"])
 
 
-def collect_healthz(gauge, histogram, service_name, address, port, url):
+def collect_healthz(gauge, histogram, service_name, address, port, url, ca_path, headers):
     with histogram.time():
         error = "ok"
         try:
-            error = requests.get("http://{}:{}{}".format(address, port, url)).text
+            error = requests.get("https://{}:{}{}".format(address, port, url), headers = headers, verify = ca_path).text
```
fanyangCS: this assumes the k8s deployment will enable https. and this won't work with existing k8s deployment, right?

src/watchdog/src/watchdog.py
```diff
@@ -217,20 +217,20 @@ def collect_healthz(gauge, histogram, service_name, address, port, url):
...
     # check kubelet
     nodeItems = nodesJsonObject["items"]
 
     for name in nodeItems:
         ip = name["metadata"]["name"]
-
+        
         collect_healthz(k8s_gauge, kubelet_healthz_histogram,
-            "k8s_kubelet", ip, 10255, "/healthz")
+            "k8s_kubelet", "http", ip, 10255, "/healthz", None, None)
```
fanyangCS: where do you detect whether this should be "http" or "https"?

fanyangCS:
neither Jenkins nor any unit test covers this part of code. we need to test on this PR.

src/watchdog/src/watchdog.py
YanjieGao: For on-prem existing PAI it use http. 

For AKS I haven't find a method to access this kubelet port. I also suggest this metrics is better to be skipped for pai on AKS, as this responsibility should be taken by AKS and not PAI.

we could add a pending issue for future test.

YanjieGao:
after no issue and approved, will retest it

src/watchdog/src/watchdog.py
fanyangCS: in this case you will need to discuss with the team for this decision before submitting a PR.

YanjieGao: Not aware this before, so not mentioned. Currently i haven't make any decision. and keep previous code as much as possible

YanjieGao: @xudifsd , kubelet url access issue created and assigned to you, needs your help to investigate, thanks. 
https://github.com/Microsoft/pai/issues/1878

deployment/quick-start/services-configuration.yaml.template
YanjieGao: Hi @ydye, after refactor, this config move to kubernetes config .  make it as user configure according to their needs

Issue #1716, opened 2018-11-16
[AKS] aks deployment config IP list, ssh/user/password


YanjieGao:
## Problem: 
For existing K8s, Such as AKS env k8s need config IP list, ssh/user/password, please update document if this is different from existing approach 

## Dependency:
paictl & dev-box

## Reference Solution & Plan:
refer branch yanjga/aks-test

## Dev plan
- yanjie will create a pai/aks branch as master branch for receive all aks related pr. https://github.com/Microsoft/pai/tree/pai/aks
- some knowledge could refer yanjga/aks-test driver related change
- feature dev folk & submit pr to pai/aks branch
- If all tests passed, then cherry pick pai/aks pr to master branch.

YanjieGao:
After discussion, suggestions as a document prompt user. Not maintain scripts in PAI

Issue #1715, opened 2018-11-16
[AKS] deploy the devbox to k8s cluster as a pod. This should become an optional PAI service 


YanjieGao:
## Problem: 
For existing K8s, Such as AKS env nodes only contains inner ip. Current deploy maybe contains some part need execute ssh command. On AKS need a jump box pod to execute. For genernal purpose, deployer and admin also could leverage this dev-box as jump box to execute ssh cmd.

## Dependency:
dev-box

## Reference Solution & Plan:
refer branch yanjga/aks-test

## Dev plan
- yanjie will create a pai/aks branch as master branch for receive all aks related pr. https://github.com/Microsoft/pai/tree/pai/aks
- some knowledge could refer yanjga/aks-test driver related change
- feature dev folk & submit pr to pai/aks branch
- If all tests passed, then cherry pick pai/aks pr to master branch.

YanjieGao:
proposal: This could be a common service for ssh / deployment (jump box) inner intranet ip cluster scenarios. Currently as service deployment config is not flexible, suggest use a config to configure this service's start and stop.

YanjieGao:
proposal: This could be a common service for ssh / deployment (jump box) inner intranet ip cluster scenarios. Currently as service deployment config is not flexible, suggest use a config to configure this service's start and stop.

Issue #1714, opened 2018-11-16
[AKS] dev-box kubectl configuration 


YanjieGao:
## Problem: 
for AKS or other existing k8s env, the nodes maybe only contains inner ip. We need operate k8s cluster out of k8s cluster. And for aks it also need auth then config kubectl, so we should enhance previous kubectl install part.

## Dependency:
dev-box & paictl.py

## Reference Solution & Plan:
refer branch yanjga/aks-test

## Dev plan
- yanjie will create a pai/aks branch as master branch for receive all aks related pr. https://github.com/Microsoft/pai/tree/pai/aks
- some knowledge could refer yanjga/aks-test driver related change
- feature dev folk & submit pr to pai/aks branch
- If all tests passed, then cherry pick pai/aks pr to master branch.

YanjieGao:
After discussion, suggestions as a document prompt user

PR #1722, opened 2018-11-19
[AKS] add deploy scripts for dev-box env configuration on AKS

YanjieGao:
## Problem & Goal
In order to meet the pre-requirement of quick start (ssh env and get node ip list), user need exeucte this step. 

These additional configuration are used for Azure CLI login, kubectl config, SSH env config and get node list. 

## Solve & doc

Organize the PR into a form that provides the information needed for the next quick start steps.

After performing the previous common prepare the dev-box step,  add a new step for user to config dev-box

Docs pls refer:
docs/aks/pai_aks_deploy.md#c-step-0

## Results
 
After performing the new steps, it will have the following effects on the cluster:
- node will be set the new ssh user name and password [which will be used by quick start config]
- user could get the node ip list [which will be used by quick start config]
- azure cli and kubectl can be used at dev-box [which will be used for service deployment, note: as ssh depend this box, so the dev-box service need this step to be launched, other maintain scripts will be provided at next PR]
- a jump box dev-box pod will be deployed [Which will be used at quick start step for ssh cmd]

## Expand to other cluster environments:
Like other cloud AWS, ali, existing k8s etc. we could extend dev-box-util.sh to add other CLUSTER_TYPE
 and add related scripts to config related pre-requests.

## Related issue

https://github.com/Microsoft/pai/issues/1716
https://github.com/Microsoft/pai/issues/1715
https://github.com/Microsoft/pai/issues/1714


YanjieGao:
@fanyangCS , i will refactor this pr to a doc which will tell user the prerequest and provide reference documentation whenever possible without maintaining scripts and documents. If you feels ok, I will refactor

PR #1737, opened 2018-11-21
increase node cpu alerting level

xudifsd:
increase to 90%


fanyangCS:
since the system is relatively more robust to high CPU usage, we can increase the alert level of CPU to higher value.

fanyangCS:
Also, please think of how to provide a severity level for each alert.

xudifsd:
We can add label `severity: critical` to the alerts that we think important.

But I think this is only useful when we have other notification methods so that we can decide what notification get pushed. For example, we may add slack notification for alerts not so important, and only send out email when some critical events happened. If no other endpoints included, I see no needs to label severity level.

fanyangCS:
@xudifsd , we can do that after we have severity levels.

PR #1740, opened 2018-11-21
[Cluster Object Model] Add driver parser and refactor the deployment script.

ydye:


src/drivers/config/drivers.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def __init__(self, cluster_configuration, service_configuration, default_service_configuraiton):
+        self.logger = logging.getLogger(__name__)
+
+        self.cluster_configuration = cluster_configuration
+        self.service_configuration = self.merge_service_configuration(service_configuration, default_service_configuraiton)
+
+
+
+    def merge_service_configuration(self, overwrite_srv_cfg, default_srv_cfg):
```
mzmssg: Hi, Yundong, is it better to write a superclass implementing `run` and `merge_service_configuration`, interfacing `validation_pre` and `validation_post`? Then every service config model inherit this superclass and implement the interface. I think this merge function works for most cases, no need to rewrite it in every service.

If some service has customization requirements, it could overwrite `run` function.

xudifsd: agreed

ydye: @mzmssg Can you send pr to implement it?

mzmssg: @ydye Get you, will have a try

PR #1759, opened 2018-11-26
fix-login-redirects

zhkjim:
Fixed a bug in login that was located in pai/src/webportal/src/app/user-login.comment.js, which caused users to log in to template.html and then jump to dashboard.html.
The reason for this bug is the object “query” in user-login.component.js. Its attribute “origin” is not assigned, so query.origin is undefined and executes "window.location.replace(query.origin ? query.origin) : '/template.html');", the page will jump to template.html and then jump to dashboard.html.
Fix this bug: change "/template.html" in "window.location.replace(query.origin ? query.origin : '/template.html');" to ‘/’.


Gerhut:
There are still 2 `.DS_Store`s remain.

Gerhut:
Thank you for contribution.

This is a legacy configuration of marketplace redirection of login.

fanyangCS:
merge for external contribution.

Issue #1557, opened 2018-10-19
[AKS] make driver path configurable


YanjieGao:
## Problem: 
For existing K8s, Such as AKS env, the default driver path maybe not consist with pai driver default path. And driver install schell maybe hanging when hope to uninstall host default driver.  So it is useful to make drive path configurable for different existing K8s environmnet

## Dependency:
- Job launch script
- job-exporter
- Node manager

## Solution & Plan:
- Job launch script
code location:
    https://github.com/Microsoft/pai/blob/master/src/rest-server/src/templates/yarnContainerScript.mustache
    https://github.com/Microsoft/pai/blob/master/src/rest-server/src/models/job.js#L416

? add a configuration items at job / data var
- job-exporter
  Solution: make the driver mount path configurable
- Node manager
  Solution: make the driver mount path configurable

## Dev plan
- yanjie will create a pai/aks branch as master branch for receive all aks related pr
https://github.com/Microsoft/pai/tree/pai/aks
- some knowledge could refer yanjga/aks-test driver related change
- feature dev folk & submit pr to pai/aks branch
- If all tests passed, then cherry pick pai/aks pr to master branch.

hao1939:
Can we skip driver installing if driver path was given?

YanjieGao:
Related PR:
https://github.com/Microsoft/pai/pull/1761

YanjieGao:
NOTE: PR #1761  shoud cherry pick to pai/aks branch for later int testing

PR #1761, opened 2018-11-27
makes drivers pre-installed-driver friendly

### LINKED ISSUE #1747: should using absolute path of nvidia-smi instead of relay on PATH in job-exporter
Since many other services relays on absolute path to reference nvidia library and binary, job-exporter should do this too, in case on-prime nodes have pre-installed gpu driver but didn't configured PATH

xudifsd:
fix #1747 and related to #1557 

Changes this PR made:
* make drivers service check `PRE_INSTALLED_NV_DRIVER_PATH`'s existence, which passed as variable environment, I'll make this configurable with @ydye 's cluster object model
  * if `PRE_INSTALLED_NV_DRIVER_PATH` exist, then drivers will abort driver installation only make a symbolic link from `NV_DRIVER` to `PRE_INSTALLED_NV_DRIVER_PATH` which are `/var/drivers/nvidia/current` and `/usr/local/nvidia` respectively. Otherwise install driver as old way and create symbolic link
* make `hadoop-node-manager` and `node-exporter` mount `/var/drivers/nvidia/current` instead of `/var/drivers`, because `/var/drivers/nvidia/current` is a symbolic link, if we mount `/var/drivers`, the symbolic link will be invalid if it pointing to path not mounted, but if mount point is symbolic link, the kubelet will follow the link and mount target directly.

So after this PR, gpu consumer can assume driver exist under `/var/drivers/nvidia/current` and should mount it into container. It's drivers service's responsibility to make that symbolic link pointing to the right location, no matter it is pre-existed or install by drivers.

@hao1939 I also fixed `nvidia-smi not found` problem in this PR, it turns out, not found not means didn't found nvidia-smi binary, but means didn't found some library which are exist in $DRIVER_PATH/lib64, this can be solved by using `LD_LIBRARY_PATH` or by writing path into `/etc/ld.so.conf.d/`. I choose `LD_LIBRARY_PATH` by making job-exporter configure that itself.


src/drivers/deploy/drivers.yaml.template
```diff
@@ -54,6 +54,13 @@ spec:
           name: etc-path-ld
         - mountPath: /etc/docker
           name: etc-path-docker
+        - mountPath: /usr/local/nvidia # TODO
+          name: pre-install-nv-driver-path
+        env:
+        - name: DRIVER_PATH
+          value: /var/drivers/nvidia
+        - name: PRE_INSTALLED_NV_DRIVER_PATH
```
YanjieGao: This is assume deployer known this config in this file. If he's cluster contains driver in other path, he need to change this path config and deploy?

YanjieGao: Why not make this in service-configuration.yaml or kuberentes-configuration.yaml as a global configuration file config items?

xudifsd: I will make this configurable after @ydye 's cluster object model merged into master. Right now, I think this is enough since we only need to cover aks case.

YanjieGao: ok, pls update doc in this place when it is configurable and tell user how to config
image
https://github.com/Microsoft/pai/blob/master/docs/pai-management/doc/how-to-write-pai-configuration.md#gpu_driver

xudifsd: Sure, I'll take a note.

YanjieGao:
@xudifsd , pls cherry pick this pr to pai/aks branch.  previous agreed process is first pr merge to pai/aks then cherry pick to master. As pai/aks will do test for aks related changes.

xudifsd:
@YanjieGao done

PR #1787, opened 2018-11-29
[rest-server]parser of cluster object model

Gerhut:
- [x] Document

The service port of HDFS, WebHDFS and Yarn is hard-coded in object model of the previous code, but still shared in configurations of different services (rest-server & pylon, because they are written in the same file, call the same methods).

Currently, the parser code of different service is separated, so the services port of dependencies will be hard-coded more than once.

### DIFF
--- a/deployment/quick-start/services-configuration.yaml.template
+++ b/deployment/quick-start/services-configuration.yaml.template
@@ -72,16 +72,11 @@ hadoop:
 frameworklauncher:
   frameworklauncher-port: 9086
 
-restserver:
-  # port for rest api server
-  server-port: 9186
-  # secret for signing authentication tokens, e.g., "Hello PAI!"
+
+rest-server:
   jwt-secret: pai-secret
-  # database admin username
   default-pai-admin-username: admin
-  # database admin password
   default-pai-admin-password: admin-password
-  # rest server would achieve marketplace template from below configed github repository
   github-owner: Microsoft
   github-repository: pai
   github-path: marketplace
--- a/deployment/quick-start/services-configuration.yaml.template
+++ b/deployment/quick-start/services-configuration.yaml.template
@@ -74,12 +74,10 @@ frameworklauncher:
 
 
 rest-server:
-  jwt-secret: pai-secret
+  # database admin username
   default-pai-admin-username: admin
+  # database admin password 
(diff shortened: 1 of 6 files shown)


deployment/quick-start/services-configuration.yaml.template
```diff
@@ -74,16 +74,10 @@ frameworklauncher:
 
 
 
-restserver:
-  # port for rest api server
-  server-port: 9186
-  # secret for signing authentication tokens, e.g., "Hello PAI!"
+rest-server:
   jwt-secret: pai-secret
```
ydye: Maybe in the template, we could only left the necessary configuration. And remove the configuration with the default value.  Because I found that some of the key-value are same with the default.

src/rest-server/deploy/rest-server.yaml.template
```diff
@@ -33,45 +33,45 @@ spec:
...
-        image: {{ clusterinfo['dockerregistryinfo']['prefix'] }}rest-server:{{ clusterinfo['dockerregistryinfo']['docker_tag'] }}
+        image: {{ cluster_cfg['cluster']['docker-registry']['prefix'] }}rest-server:{{ cluster_cfg['cluster']['docker-registry']['tag'] }}
         imagePullPolicy: Always
         env:
         - name: LAUNCHER_WEBSERVICE_URI
-          value: {{ clusterinfo['restserverinfo']['webservice_uri'] }}
+          value: {{ cluster_cfg['yarn-frameworklauncher']['webservice'] }}
         - name: HDFS_URI
-          value: {{ clusterinfo['restserverinfo']['hdfs_uri'] }}
+          value: http://{{ cluster_cfg['hadoop-name-node']['master-ip'] }}:9000
```
ydye: maybe for the port hardcode problems, you could ask the service owner to make it as a new configuration

deployment/quick-start/services-configuration.yaml.template
Gerhut: Sorry for misunderstood both files, now inverted them.

src/rest-server/deploy/rest-server.yaml.template
Gerhut: @mzmssg any ideas?

PR #1795, opened 2018-11-29
[pylon]parser of cluster object model

Gerhut:
2 internal used configurations will be undocumented.

- `webhdfs_legacy_port` is for legacy use
- `deploy_service` is for future use

### DIFF
--- a/deployment/quick-start/services-configuration.yaml.template
+++ b/deployment/quick-start/services-configuration.yaml.template
@@ -127,8 +127,6 @@ drivers:
 #   {% endif %}
 #   {% endif %}
 
-pylon:
-  # port of pylon
-  port: 80
-  # deploy service for cloud load balancer
-  deploy_service: false
+# uncomment following section if you want to customize the port of pylon
+# pylon:
+#  port: 80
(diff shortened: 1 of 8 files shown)


examples/cluster-configuration/services-configuration.yaml
```diff
@@ -132,9 +132,3 @@ prometheus:
   #    smtp_from: foo_bar@gmail.com
   #    smtp_auth_username: user@gmail.com
   #    smtp_auth_password: gmail_password
-
-
-pylon:
-  # port of pylon
-  port: 80
-
```
xudifsd: should also change `deployment/quick-start/services-configuration.yaml.template`. When change it, comment out all pylon default value, and leave a comment, so quick start user can simple uncomment value to change.

Gerhut: Sorry for a bad push, please refresh the PR

deployment/quick-start/services-configuration.yaml.template
```diff
@@ -139,9 +139,3 @@ prometheus:
   ssh-secret-name: ssh-key-secret
   {% endif %}
   {% endif %}
-
-pylon:
```
xudifsd: maybe just comment out all pylon default values, and leave a comments, so quick start users can simple uncomment value to change.

PR #1806, opened 2018-11-30
Closes #1767:Add function to force user login before entering the Submit Job page

### LINKED ISSUE #1767: Ask user to login before editing job, so won't lost input
Now, user could access `Submit Job` page without login, but when `submit` the changes will lost.
So we should force user to login before editing.

zhkjim:
Closes #1767:Add function to force user login before entering the Submit Job page.
If you do not log in when you enter the Submit Job V2 page, you will be redirected to the login page, but this feature is not available in the Submit Job page.
Add this function to /pai/src/webportal/src/app/marketplace/job-submit/job-submit.component.js and call userAuth.checkToken after the DOM is loaded.


zhkjim:
Create a new pull request instead of this one.

PR #1814, opened 2018-12-03
fix drivers bug

xudifsd:
because we set `$PRE_INSTALLED_NV_DRIVER_PATH` as volume, k8s will always have it as directory, so, previous will always think `$PRE_INSTALLED_NV_DRIVER_PATH` is installed, and modify `/var/drivers/nvidia/current` to it.

### DIFF
--- a/src/drivers/build/install-nvidia-drivers
+++ b/src/drivers/build/install-nvidia-drivers
@@ -33,7 +33,8 @@ function nvidiaPresent {
 }
 
 echo ======== If NVIDIA present exit early =========
-if [ nvidiaPresent ] ; then
+nvidiaPresent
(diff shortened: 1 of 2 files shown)


src/drivers/build/install-nvidia-drivers
```diff
@@ -33,7 +33,8 @@ function nvidiaPresent {
 }
 
 echo ======== If NVIDIA present exit early =========
-if [ nvidiaPresent ] ; then
+nvidiaPresent
```
ydye: What's the difference

xudifsd: previous way bash will always take true path, I think bash treat it as no null value. later way bash will check its return value.

PR #1824, opened 2018-12-04
Enable Cluster object model 

ydye:
1) Checkin cluster object model.
2) Remove all dependency of the mock cluster object model
3) Add the forward compatibility for the old cluster configuraiton


deployment/quick-start/services-configuration.yaml.template
```diff
@@ -15,131 +15,110 @@
...
 
   # the docker registry to store docker images that contain system services like frameworklauncher, hadoop, etc.
-  docker-registry-info:
+  docker-registry:
 
     # If public, please fill it the same as your username
-    docker-namespace: openpai
+    namespace: openpai
 
     # E.g., gcr.io. If public, fill docker_registry_domain with word "public"
```
xudifsd: maybe this comment and following comments are outdated?

examples/cluster-configuration/services-configuration.yaml
```diff
@@ -15,120 +15,107 @@
...
+#  # the docker registry to store docker images that contain system services like frameworklauncher, hadoop, etc.
+#  docker-registry:
+#
+#    # domain/namespace/image-name:tag
+#
+#    # If public, please fill it the same as your username
+#    namespace: openpai
+#
+#    # E.g., gcr.io. If public，fill docker_registry_domain with word "public"
+#    # docker_registry_domain: public
```
xudifsd: Same above

deployment/quick-start/services-configuration.yaml.template
```diff
@@ -15,131 +15,110 @@
...
-  clusterid: pai-example
 
-  # HDFS, zookeeper data path on your cluster machine.
-  data-path: "/datastorage"
+cluster:
+  common:
+    clusterid: pai-example
+
+    # HDFS, zookeeper data path on your cluster machine.
+    data-path: "/datastorage"
```
xudifsd: Is this field mandatory? I see it is commented out in example. Should go through all fields in this file, and comment out all optional fields. Also I think we should refine some of comments, so this file can server as document of how to configure pai.

ydye: I will comment the ```common``` field. But the ```docker-registry``` is necessary, because when doing ci/cd, we will fill some value into this key.

examples/cluster-configuration/services-configuration.yaml
ydye: Updated

deployment/quick-start/services-configuration.yaml.template
ydye: Updated

deployment/k8sPaiLibrary/maintainlib/kubectl_install.py
```diff
@@ -57,21 +57,21 @@ def kubectl_install(self):
 
 
     def kubectl_configuration_generate(self):
-
+        com = self.cluster_config
         self.logger.info("Generate the configuation file of kubectl.")
 
-        if self.cluster_config != None:
+        if com != None:
```
hao1939: I'm not sure why we need the assignment?
Could we use cluster_config directly?

deployment/k8sPaiLibrary/maintainlib/update.py
```diff
@@ -84,16 +86,15 @@ def get_node_list_from_k8s_api(self):
     def get_node_config_from_cluster_configuration(self):
 
         self.logger.info("Try to get node confguration from cluster configuration.")
-        cluster_config = self.k8s_configuration
+        com = self.k8s_configuration
```
hao1939: And here, why we need a variable `com`?

hao1939:
My most concern is about the way we implement `forward compatibility`.
Multiple config entries for one purpose would be a nightmare.

I suggest to provide an external configuration migration tool instead of support both the old and new style configuration.

deployment/k8sPaiLibrary/maintainlib/kubectl_install.py
ydye: No，I will left this to make it as another approch

deployment/k8sPaiLibrary/maintainlib/update.py
ydye: Get node configuration to constuct node list. And BTW, this change is introduced before and not in this PR.

ydye:
@hao1939 Do it later.

PR #1825, opened 2018-12-04
optimization: Replace the url module with the querystring module

zhkjim:
This optimization replaces all code in the webprotal that uses the url module with the querystring module and ensures consistent functionality. The result of this optimization is to reduce the total webprotal size from 478.4MB to 477.4MB.
The url module is built using the punycode and querystring modules, which are larger than the query string module. Webprotal uses the "url.parse().query" function of the url module to get the query type parameter of the address bar. However, this feature can be replaced by the "querystring.parse()" function in the smaller querystring module.
The reason is in lines 143 to 151 of url.js `this.query = querystring.parse(this.search.substr(1));`
 used the parse method in querystring.  Both “url.parse().query” and "querystring.parse()" return an object, but "querystring.parse()" does not directly separate the domain name from the query parameter. You need to change it to `querystring.parse(window.location.href.slice(window.location.href.indexOf) ('?') +1)); `will return the same object as "url.parse().query".
The size of url.js, punycode.js, and querystring.js are 23KB, 18KB, and 6KB respectively. After the webpack is packaged, the size of all optimized JavaScript files will be reduced about 100KB. The specific changes are listed below:
login.bundle.js:937KB(before) 826KB(after)
hardwareDetail.bundle.js: 920KB(before) 810KB(after)
templateList.bundle.js:928KB(before) 818KB(after)
templateDetail.bundle.js:932KB(before) 822KB(after)
templateView.bundle.js:3.1MB(before) 3MB(after)
jobSubmit.bundle.js:16.3MB(before) 16.2MB(after)
submit.bundle.js:  1.6MB(before) 1.5MB(after)

### DIFF
--- a/src/webportal/src/app/cluster-view/hardware/hardware-detail.component.js
+++ b/src/webportal/src/app/cluster-view/hardware/hardware-detail.component.js
@@ -20,7 +20,7 @@
 const hardwareDetailComponent = require('./hardware-detail.component.ejs');
 const breadcrumbComponent = require('../../job/breadcrumb/breadcrumb.component.ejs');
 const webportalConfig = require('../../config/webportal.config.js');
-const url = require('url');
+const querystring = require('querystring');
 
 //
 
@@ -35,7 +35,7 @@ $(document).ready(() => {
   $('#sidebar-menu--cluster-view').addClass('active');
   $('#sidebar-menu--cluster-view--hardware').addClass('active');
   let instance = '';
-  const query = url.parse(window.location.href, true).query;
+  const query = querystring.parse(window.location.search.replace(/^\?+/, ''));
   if (query['instance']) {
     instance = query['instance'];
   } else {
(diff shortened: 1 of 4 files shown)


Gerhut:
Nice catch, currently the bundle size is too large to load without blocking, and optimizing needs lots of works.

However, I'd like to use `window.location.search.replace(/^?+/, '')` to grab the query string, this will unify the understanding to "query string" part between browser and our code, or perhaps you could survey how to deal with query string in other routing libraries.

Also, please rebase your code on the newest master branch.

zhkjim:
`window.location.href.slice(window.location.href.indexOf) ('?') +1) `has been modified by a regular expression, `querystring.parse(window.location.href.slice(window.location) .href.indexOf) ('?') +1));` is replaced with `querystring.parse(window.location.search.replace(/^\?+/, ''))`.
In addition, submit.html is not tested because the current "use" operation of the marketplace automatically jumps to the v2 version of the submit. Theoretically feasible, but please help me test it.Thanks!

Gerhut:
Hi @zhkjim 

Could you please rebase your branch on the latest master branch so we can merge your code directly?

PR #1829, opened 2018-12-05
fix driver bug

xudifsd:
it seems rest-server will not pass environment variable into yarn-script. just hardcode in yarn-script since it is hardcoded anyway.

### DIFF
--- a/src/rest-server/deploy/rest-server.yaml.template
+++ b/src/rest-server/deploy/rest-server.yaml.template
@@ -52,8 +52,6 @@ spec:
           value: {{ cluster_cfg['rest-server']['default-pai-admin-password'] }}
         - name: ETCD_URI
           value: {{ cluster_cfg['rest-server']['etcd-uris'] }}
-        - name: NV_DRIVER
-          value: /var/drivers/nvidia/current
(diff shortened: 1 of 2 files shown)


src/rest-server/deploy/rest-server.yaml.template
```diff
@@ -52,8 +52,6 @@ spec:
           value: {{ cluster_cfg['rest-server']['default-pai-admin-password'] }}
         - name: ETCD_URI
           value: {{ cluster_cfg['rest-server']['etcd-uris'] }}
-        - name: NV_DRIVER
-          value: /var/drivers/nvidia/current
```
YanjieGao: for example if on aks env dir is diff with this, need tell user to  change this config?

xudifsd: No, this can be hardcoded. My plan is all gpu-consumers(job-exporter, node-manager) use this symbolic link to find current active driver. And it's driver's responsibility to make this symbolic link pointing to the right place, whether it is pre-installed place or our installation place.

xudifsd:
@Gerhut yes, I know that, but I didn't know when I submit previous PR, since it can be hardcoded, I'll just hardcoding in script instead of passing the env variable.

PR #1831, opened 2018-12-05
[VC update] rest-server

mzmssg:
add 3 vc related api:

1. PUT virtual-clusters/:vcName
add or update a vc quota

2. DELETE virtual-clusters/:vcName
delete a vc

3. PUT virtual-clusters/:vcName/status
change vc status


src/rest-server/src/config/yarn.js
```diff
@@ -45,4 +58,18 @@ if (error) {
...
+  .timeout(2000)
+  .end((res) => {
+    if (res.status === 200) {
+      logger.info('connected to yarn successfully');
+    } else {
+      throw new Error('cannot connect to yarn');
+    }
+  });
+}
+
```
Gerhut: @fanyangCS 

Shall we add a yarn dependency to rest server?

src/rest-server/src/models/vc.js
```diff
@@ -18,44 +18,35 @@
...
+class VirtualCluster {
   getCapacitySchedulerInfo(queueInfo) {
     let queues = {};
+
     function traverse(queueInfo, queueDict) {
       if (queueInfo.type === 'capacitySchedulerLeafQueueInfo') {
         queueDict[queueInfo.queueName] = {
-          capacity: queueInfo.absoluteCapacity,
-          maxCapacity: queueInfo.absoluteMaxCapacity,
+          capacity: parseInt(queueInfo.absoluteCapacity),
```
Gerhut: use `Number()` instead of `parseInt()` to avoid number base issues.

src/rest-server/src/models/vc.js
```diff
@@ -80,6 +71,254 @@ class VirtualCluster {
...
+        if (!vcList.hasOwnProperty('default')) {
+          return callback(createError('Not Found', 'NoVirtualClusterError', `No default vc found, can't free quota`));
+        } else if (!vcList.hasOwnProperty(vcName)) {
+          return callback(createError('Not Found', 'NoVirtualClusterError', `Can't delete a nonexistent vc ${vcName}`));
+        } else {
+          this.stopVc(vcName, (err) => {
+            if (err) {
+              return callback(err);
+            } else {
+              let defaultQuotaIfUpdated = vcList['default']['capacity'] + vcList[vcName]['capacity'];
```
Gerhut: Could you make sure about these 2 values will be Number? `+` will be treated as string concat when one of these is string.

src/rest-server/src/config/yarn.js
fanyangCS: good question. the desired behavior is: if yarn is available, it works. if yarn is unavailable, the system says it doesn't support the feature yet.

---
In reply to: [239478932](https://github.com/Microsoft/pai/pull/1831#discussion_r239478932) [](ancestors = 239478932)

mzmssg: @Gerhut @fanyangCS 
I will move the validation to somewhere, but I still have a question.
Do we consider `vc` as the same concept of `hadoop queue`, or a abstract interface, `hadoop queue` is one implementation? Now I think answer is the former, so moving yarn check under `vc` related API sounds reasonable?

src/rest-server/src/models/vc.js
mzmssg: Resolve it by `math.round`

src/rest-server/src/models/vc.js
mzmssg: Yes, this two value comes from YARN, they are both number.

src/rest-server/src/models/vc.js
Gerhut: Built-in [`Math.round`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round) object is enough

mzmssg:
Checkin firstly to help frontend test, will remove this deps in other pr

PR #1836, opened 2018-12-06
Fix cntk test exit code

mzmssg:


xudifsd:
maybe `set -e`? could `hdfs-mount` also go wrong?

mzmssg:
@xudifsd This tool is not that stable, sometimes it will enter zombie and block exit. I'm afraid that `set -e` will confuse us with this tool's bug.

xudifsd:
if `set -e` could expose this tools' bugs then we can fix them, instead of hide and pretend it's fine.

fanyangCS:
we should remove this test. we do not encourage user to use hdfs-mount

mzmssg:
closed for test removed

Issue #1764, opened 2018-11-27
job-exporter refactor proposal


xudifsd:
# Background
Job exporter is essentially a metrics collector, to full fill its purpose, it will need to call multiple external commands, which makes job exporter vulnerable to commands hang. Following are the commands it calls:
* `docker stats --no-stream --format "table {{.Container}},{{.Name}},{{.CPUPerc}},{{.MemUsage}},{{.NetIO}},{{.BlockIO}},{{.MemPerc}}"`
* `docker inspect $container_id`
* `docker logs --tail all $container_id`
* `systemctl is-active docker | if [ $? -eq 0 ]; then echo "active"; else exit 1; fi`
* `nvidia-smi -q -x`
* `iftop -t -P -s 1 -L 10000 -B -n -N`
* `infilter $pid /usr/bin/lsof -I -n -P`

Except `docker logs` and `infilter` commands, we have experienced cases of all other commands hang.

# Problems
* Hangs on commands, this will causes: [not showing some metrics](https://github.com/Microsoft/pai/issues/1719) or even worse: metrics stay flat and user/admin doesn't know it.
* Single threaded model is sensitive for all commands hang, if one command hangs, all other metrics will not be updated. For example, `docker stats` command has nothing to do with `iftop` command, hanging in `iftop` command should only affect network related metrics not gpu utilization.
* We have no idea of how much time we spent on commands. This info is useful for debugging metrics related problems and alerting.

Previously we have designed [some ways](https://github.com/Microsoft/pai/pull/848) tailored for `nvidia-smi` to mitigate its hanging, but it lacks some features we requires:
* It did not solve the problem of single threaded model.
* It did not record the time spent on command calls.

# Solution
* Use [`subprocess32`](https://pypi.org/project/subprocess32/) which is backport of subprocess in python3 to python2, it provides timeout functionality we required. Or we simply upgrade job-exporter to python3.
* Refactored job-exporter to multi-threaded models.
* Do some alerting on absurd large time (we can change it to reasonable one once we gained enough data).
* Use port to expose metrics instead of current file based one.

fanyangCS:
Sounds good. Since we will be able to figure out the timeout, what will grafana shows in this case? an intermittent line instead of continuous line?

fanyangCS:
relevant issue: https://github.com/Microsoft/pai/issues/1729

xudifsd:
sounds resonable, I'll take a note.

fanyangCS:
[...]

Issue #1719, opened 2018-11-16
OpenPAI page don't show the utilization of the GPU.


lllllwwwwyyyy:
![image](https://user-images.githubusercontent.com/10185324/48626437-cfa43900-e9ec-11e8-87da-fa26abefdc0a.png)
![image](https://user-images.githubusercontent.com/10185324/48626444-d29f2980-e9ec-11e8-9806-3f19f0c10236.png)
I run the example job `tensorflow-cifar10`
The config show below:
`{

  "jobName": "tensorflow-cifar10",

  "image": "openpai/pai.example.tensorflow:stable",

  "dataDir": "/lwy/datadir",

  "outputDir": "/lwy/outputdata",

  "taskRoles": [

    {

      "name": "cifar_train",

      "taskNumber": 1,

      "cpuNumber": 1,

      "memoryMB": 32768,

      "gpuNumber": 2,

      "command": "git clone https://github.com/tensorflow/models && cd models/research/slim && python download_and_convert_data.py --dataset_name=cifar10 --dataset_dir=$PAI_DATA_DIR && python train_image_classifier.py --batch_size=64 --model_name=inception_v3 --dataset_name=cifar10 --dataset_split_name=train --dataset_dir=$PAI_DATA_DIR --train_dir=$PAI_OUTPUT_DIR"

    }

  ]

}`
But when it is running,the page do not show the utilization of GPU.

lllllwwwwyyyy:
below is my dir:(in the hdfs webportal)
![image](https://user-images.githubusercontent.com/10185324/48626641-607b1480-e9ed-11e8-849b-8564f2683c4b.png)
![image](https://user-images.githubusercontent.com/10185324/48626650-653fc880-e9ed-11e8-874a-c3958291d4ad.png)

lllllwwwwyyyy:
Below is my datadir :  in the hdfs webportal 
![image](https://user-images.githubusercontent.com/10185324/48626699-86a0b480-e9ed-11e8-9f5e-3321eb2eff53.png)

![image](https://user-images.githubusercontent.com/10185324/48626673-7b4d8900-e9ed-11e8-92a9-0d6ed305c4f4.png)

lllllwwwwyyyy:
![image](https://user-images.githubusercontent.com/10185324/48655126-563d3280-ea4e-11e8-9f33-fdbf76c4125f.png)
Is there a problem with the configuration?

mzmssg:
[...]

PR #1840, opened 2018-12-06
refactor job-exporter

### LINKED ISSUE #889: [exporter] gpu exporter can not indicate gpu failure
Current gpu exporter use file to export metrics through node-exporter, this file based mechanism can not indicate gpu exporter's failure. For example, if nvidia-smi hanging for 10min, we may expect no gpu data at all, alter may triggered in this case, but current implementation will still reporting last available metrics.

If we continue using file based method, we should maybe remove all metrics relating to gpu if gpu data not available.

xudifsd:
implementation of #1764 , fixed #889 can partially solve #1719 .

I split monolithic job-exporter into 4 threads:
* docker collector: used to collect metric about docker daemon's healthiness
* gpu collector: used to collect metric about nvidia gpu utilization, also provide gpu info to container collector
* container collector: used to collect task/service specific usage metric, consume gpu info provided by gpu collector and provide docker stats info to zombie collector
* zombie collector: used to provide how many zombie task exist in cluster

We utilize python3's timeout parameter of check_output, so hopefully no single external command call will hangs indefinitely. I have already tuned the timeout parameter of each command call by using 99th latency of calling each command plus some buffer time.

Also since we use multi-threaded model, no hangs in single collector will block other collector from emitting metrics.


YanjieGao:
If cmd hangs, current exporter behaviour is continue hanging sending timeout alert Or killed cmd then retry?

xudifsd:
if hangs exceed predefined timeout value, the cmd will be killed and return None, other parts will ignore this None and continue working. It will retry next time.

PR #1841, opened 2018-12-06
make preinstalled driver path configurable

### LINKED ISSUE #1557: [AKS] make driver path configurable

xudifsd:
fix #1557

### DIFF
--- a/src/drivers/config/drivers.md
+++ b/src/drivers/config/drivers.md
@@ -21,6 +21,14 @@ drivers:
     set-nvidia-runtme: true
 ```
 
+Or if your cluster has already installed nvidia-driver, and do not need pai to install it
+again, then you can provide this info in your service-configuration.yaml like:
+
+```yaml
+drivers:
+    pre-installed-nvidia-path: /path/to/your/drivers
+```
+
 #### Generated Configuration <a name="G_Config"></a>
 
 Generated configuration means the object model after parsing. The parsed data will be presented by a yaml format.
@@ -28,10 +36,8 @@ Generated configuration means the object model after parsing. The parsed data wi
 drivers:
     set-nvidia-runtme: false
     version: 384.111
+    pre-installed-nvidia-path: /usr/local/nvidia
 ```
-
-
-
 
 #### Table <a name="T_Config"></a>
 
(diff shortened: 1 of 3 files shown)


src/drivers/config/drivers.md
```diff
@@ -21,18 +21,24 @@ drivers:
     set-nvidia-runtme: true
 ```
 
+Or if your cluster has already installed nvidia-driver, and do not need pai to install it
+again, then you can provide this info in your service-configuration.yaml like:
+
+```yaml
+drivers:
```
YanjieGao: hi @xudifsd  does this pr cherry pick to pai/aks ? i think is useful

xudifsd: why? you need to change the config?

YanjieGao: if not change could work on AKS, no need is ok

xudifsd: yes, the default value /usr/local/driver is where aks's driver exist, so no need to cherry-pick to aks.

YanjieGao: ok thanks

Issue #1857, opened 2018-12-07
[HDFS] enhance HFDS error processing


YanjieGao:
## For potentials cascade error:
![image](https://user-images.githubusercontent.com/5576848/49637555-a2253b00-fa40-11e8-9359-d8e2ae6b7310.png)

fanyangCS:
close this issue and track in #1992 .

PR #1858, opened 2018-12-10
Add hdfs datanode replace policy

mzmssg:
#1857


sterowang:
>   <value>2</value>

Is the default replication set to 2 too small? 3 should be a more frequently selected configuration to balance between better data avaliability and space utilization.

---
Refers to: src/hadoop-data-node/deploy/hadoop-data-node-configuration/hdfs-site.xml:23 in d53c6c9. [](commit_id = d53c6c9200356f323b8c68de71f80e1e40199abb, deletion_comment = False)

mzmssg:
@sterowang 
Do you mean we merge 4 configurations(RM, NM, NN, DN) into 1? 

For the replication number, I noticed we kept it for a long time.  I guess there might be some historical reasons to reduce it?

mzmssg:
@sterowang 
For merge configuration, I think we could create a new PR for it. Then focus on datanode replace policy in current thread.

YanjieGao:
This is client side config, how could user get and use this config? From configmap?

mzmssg:
@YanjieGao 
Node manager will copy its configuration to user's container. Currently, spark example use this way to access YARN.

PR #1860, opened 2018-12-10
Escape injected variables in shell scripts

Gerhut:


src/rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -179,17 +181,19 @@ get_host_list $ENV_LIST
 # Backward compatibility
 cat <<ENV >> $ENV_LIST
 # Backward compatibility
+PAI_CURRENT_CONTAINER_IP=$CONTAINER_IP
+PAI_TASK_ROLE_INDEX=$TASK_INDEX
+ENV
+cat <<'ENV' >> $ENV_LIST
 PAI_USERNAME={{{ jobData.userName }}}
```
sterowang: Any reasons for print ENVs and add to ENV_LIST file multiple times?

Gerhut: The different in two append is one is `cat <<ENV >> $ENV_LIST` and the other one is `cat <<'ENV' >> $ENV_LIST`.

The first heredoc will substitute all `$VARIABLE`s with their values, as the second one will not, so if we inject `$PATH` to `taskData.name`, it will be save.

sterowang:
> 

Here is another ENV command, any reasons to print ENVs multiple times?

---
Refers to: src/rest-server/src/templates/yarnContainerScript.mustache:168 in 8eb5876. [](commit_id = 8eb5876196673a651371fc4764a231f0da5565e5, deletion_comment = False)

Gerhut:
The bottom one is for backward compatibility

sunqinzheng:
https://github.com/kacperszurek/exploits/blob/master/GitList/exploit-bypass-php-escapeshellarg-escapeshellcmd.md

src/rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -159,7 +159,7 @@ env | grep -E "^PAI|PATH|PREFIX|JAVA|HADOOP|NVIDIA|CUDA" > /etc/environment
 function run_user_command()
 {
   printf "%s %s\n\n" "[INFO]" "USER COMMAND START"
-  {{{ taskData.command }}} || exit $?
+  eval {{ taskData.command }} || exit $?
```
sunqinzheng: environment variables won't work if you escape '$'

src/rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -144,6 +144,8 @@ PAI_TASK_INDEX=$TASK_INDEX
 PAI_CONTAINER_HOST_IP=$CONTAINER_IP
 PAI_CONTAINER_HOST_PORT_LIST=$CONTAINER_PORTS
 PAI_CONTAINER_ID=$CONTAINER_ID
+ENV
+cat <<'ENV' >> $ENV_LIST
 PAI_DEFAULT_FS_URI={{{ hdfsUri }}}
 PAI_JOB_NAME={{{ jobData.jobName }}}
 PAI_USER_NAME={{{ jobData.userName }}}
```
sunqinzheng: '\n' should also be escaped.
We can construct a user name like 'noname\nENV\nsome-dangerous-commands'

src/rest-server/src/templates/dockerContainerScript.mustache
Gerhut:
```
bash-3.2$ eval echo\ \$HOME
/Users/gerhut
```

src/rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -144,6 +144,8 @@ PAI_TASK_INDEX=$TASK_INDEX
 PAI_CONTAINER_HOST_IP=$CONTAINER_IP
 PAI_CONTAINER_HOST_PORT_LIST=$CONTAINER_PORTS
 PAI_CONTAINER_ID=$CONTAINER_ID
+ENV
+cat <<'ENV' >> $ENV_LIST
 PAI_DEFAULT_FS_URI={{{ hdfsUri }}}
```
sunqinzheng: I think the job data here should be escaped as well. (double curly braces)

Gerhut: Since the escape rule in heredoc is different from directly in bash, maybe we should drop the heredoc use here

Gerhut: There is another solution, like multipart in chrome, use random string as heredoc boundary.

src/rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -136,35 +136,32 @@ function get_host_list()
...
+echo FRAMEWORK_NAME=$FRAMEWORK_NAME >> $ENV_LIST
+echo HADOOP_USER_NAME=$HADOOP_USER_NAME >> $ENV_LIST
+echo PAI_TASK_INDEX=$TASK_INDEX >> $ENV_LIST
+echo PAI_CONTAINER_HOST_IP=$CONTAINER_IP >> $ENV_LIST
+echo PAI_CONTAINER_HOST_PORT_LIST=$CONTAINER_PORTS >> $ENV_LIST
+echo PAI_CONTAINER_ID=$CONTAINER_ID >> $ENV_LIST
+echo PAI_DEFAULT_FS_URI={{ hdfsUri }} >> $ENV_LIST
+echo PAI_JOB_NAME={{ jobData.jobName }} >> $ENV_LIST
+echo PAI_USER_NAME={{ jobData.userName }} >> $ENV_LIST
+echo PAI_DATA_DIR={{ jobData.dataDir }} >> $ENV_LIST
```
sunqinzheng: User can set/modify arbitrary environment variable here because the job data will be unescaped by echo.
(I don't know how to make use of it...)

Gerhut: Job env will be only enabled in docker containers, so I think any unpredicted modifications will be safe.

PR #1861, opened 2018-12-10
decouple zookeeper image from cluster object module

hao1939:
1. Passing env `ZK_ID`, generate `myid` from bash, no need to substitute `myid` with clusterObjectModule in zookeeper container.
2. Passing env `ZK_SERVERS`, generate the `zoo.cfg` from bash, no need to substitute `zoo.cfg` with clusterObjectModule in container.

@ydye ,
Please help reviewing how it handle multiple zk instances.
Should the `ZK_ID` be unique?


ydye:
This change only support one node zookeeper, so I block this PR. You can keep the content of the two template, then mount them through configmap. So that you can achieve the goal to decouple the content

hao1939:
@ydye I updated the zkid generator.

src/zookeeper/build/run.sh
```diff
@@ -20,9 +20,19 @@
...
+tickTime=2000
+dataDir=/var/lib/zoodata
+clientPort=2181
+initLimit=5
+syncLimit=2
+${ZK_SERVERS}
+" > /etc/zookeeper/conf/zoo.cfg
+
+# create /var/lib/zoodata/myid
+echo $HOST_NAME | sha1sum | grep -Eo "[a-f0-9]{6}"  | head -n1 > /var/lib/zoodata/myid
```
ydye: An example here in comments.

src/zookeeper/deploy/zookeeper.yaml.template
```diff
@@ -49,6 +49,10 @@ spec:
         env:
         - name: JAVA_OPTS
           value: "-server -Xmx8192m"
+        - name: ZK_ID
```
ydye: ZK_ID?

src/zookeeper/build/run.sh
hao1939: It should generate id like below:

> hayua@stcvl-131:~/pai$ echo $HOST_NAME | sha1sum | grep -Eo "[a-f0-9]{6}"  | head -n1
> adc83b

src/zookeeper/deploy/zookeeper.yaml.template
hao1939: It maybe not a suitable name, I fellow the config entry name `zkid`.
Do you have any suggestion?

ydye: You have already generate the zkid through hash and I didn't find there is any place to use this value. Why shouldn't we remove it?

src/zookeeper/build/run.sh
ydye: Maybe you should write these detail information or knowledge in the document. I think zkid is a very important configuration for zk.

ydye:
As far as i can see, you still depends on cluster - object - model.

src/zookeeper/build/run.sh
hao1939: Agree, I put some comments.

src/zookeeper/deploy/zookeeper.yaml.template
hao1939: Yeah, I will remove it.

hao1939:
Yes, you are right.
I'm still working on removing `machine-list` from service deployment.
This PR is the first step, it remove the `machine-list` from zookeeper image.

ydye:
@hao1939 Maybe you should rename the title

hao1939:
Hi @ydye,
The title is about 'zookeeper image'.

src/zookeeper/config/zookeeper.py
```diff
@@ -48,13 +47,15 @@ def run(self):
...
         zookeeper_com["quorum"] = ""
+        zookeeper_com["zk-servers"] = ""
 
         for host_config in self.cluster_configuration["machine-list"]:
             if "pai-master" in host_config and host_config["pai-master"] == "true":
                 zookeeper_com["host-list"].append(host_config["hostname"])
                 if zookeeper_com["quorum"] != "":
                     zookeeper_com["quorum"] = zookeeper_com["quorum"] + ","
                 zookeeper_com["quorum"] = zookeeper_com["quorum"] + host_config["hostip"] + ":2181"
+                zookeeper_com["zk-servers"] = zookeeper_com["zk-servers"] + host_config["hostip"] + ":2888:3888\n"
```
ydye: This new value should be added into the document in this path about zk config. https://github.com/Microsoft/pai/blob/master/src/zookeeper/config/zookeeper.md

ydye:
> Hi @ydye,
> The title is about 'zookeeper image'.

ok

src/zookeeper/config/zookeeper.py
hao1939: Got it, I will update.

src/zookeeper/build/run.sh
```diff
@@ -20,9 +20,19 @@
...
+tickTime=2000
+dataDir=/var/lib/zoodata
+clientPort=2181
+initLimit=5
+syncLimit=2
+${ZK_SERVERS}
+" > /etc/zookeeper/conf/zoo.cfg
+
+# generate a zookeeper server id from ip: remove the dot, and reserve the last 7 digits.
+zkid=${POD_IP//\./} && echo ${zkid: -7} > /var/lib/zoodata/myid
```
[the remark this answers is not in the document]
ydye: Why not just use the Ip digit without dot? I think that value will be unique.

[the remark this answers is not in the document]
hao1939: The assumption based on the rule of ip range.
Can't image zookeeper working crossing such a huge network.

Not using full 9 digital of ipv4 is that it may overflow integer.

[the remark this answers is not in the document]
ydye: > The assumption based on the rule of ip range.
> Can't image zookeeper working crossing such a huge network.
> 
> Not using full 9 digital of ipv4 is that it may overflow integer.

Even you use full 9 digital, the problem still exits. Can we find another solution?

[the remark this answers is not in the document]
hao1939: Well, what about using sha1sum?
It will be better, but won't guarantee the unique of server id.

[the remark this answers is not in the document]
ydye: Little confused. I still remember sha1sum is your first solution. Why did you move to the solution of the last 7 digit of ip address.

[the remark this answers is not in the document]
hao1939: When updating the doc, I realized that I was generating the wrong format `ZK_SERVERS`, the`zkid` was missing.
So I have to generate the zkid in `zookeeper.py` and `bash`, but I didn't want to introduce more dependency into the python script.

ydye:
@hao1939  The myid file consists of a single line containing only the text of that machine's id. So myid of server 1 would contain the text "1" and nothing else. The id must be unique within the ensemble and should have a value between 1 and 255.

REF: https://zookeeper.apache.org/doc/r3.4.13/zookeeperAdmin.html

Issue #1863, opened 2018-12-11
add machine and remove machine issue


YanjieGao:
i follow to add and remove machine 10.0.0.7 to existing p100 cluster  (v0.8.1 pai) but meet problem:
https://github.com/Microsoft/pai/blob/pai-0.8.y/docs/paictl/paictl-manual.md#Machine_Nodelist_Example

For add:

machine-list:
  - hostname: next-a-gpu-0001
    hostip: 10.0.0.7
    machine-type: GENERIC
    k8s-role: worker
    pai-worker: "true"
    sshport: 22
    username: xx
    password: xx
But datanode and namenode IP address is error:
datanode
![image](https://user-images.githubusercontent.com/5576848/49779403-c33da280-fd44-11e8-92ff-02774e74cbe2.png)

namenode log:
![image](https://user-images.githubusercontent.com/5576848/49779422-d0f32800-fd44-11e8-8285-add602f70465.png)

Driver meet rediness probe error, log seems right:
![image](https://user-images.githubusercontent.com/5576848/49779449-f122e700-fd44-11e8-9ad5-ea98f77591ab.png)

For remove: meet error

![image](https://user-images.githubusercontent.com/5576848/49779382-b0c36900-fd44-11e8-90a5-796b9bdaf59f.png)

ydye:
Node-manager issue:
1) refresh the service ```cluster-configuraiton```
2) refresh the service ```hadoop-data-node``` and ```hadoop-node-manager```

remove problems: kind a bug

YanjieGao:
for nm issue:

hostmame is the same with machinelist.yaml hostname:
![image](https://user-images.githubusercontent.com/5576848/49779855-06991080-fd47-11e8-94b2-0d01845b41f9.png)

root@paidevbox:~/PaiDeployment# cd pai-cluster-msr-next-p100/
![image](https://user-images.githubusercontent.com/5576848/49779836-eec18c80-fd46-11e8-94b4-f83778fb605d.png)

Currently i want to remove this node and try to deploy but also meet remove node error.... above

YanjieGao:
For @253 i guess maybe the reason is i directly use config file and not gen config replace this lead to this issue. But this assumption currently not mentioned on our doc. If you have method tell me to remove this node and not throw exception, i could try to redeploy this.

[...]

PR #1864, opened 2018-12-11
[Hotfix] Fix the bug unable to find kubernetes-clean.sh

ydye:
#1863


YanjieGao:
note after this change also contains other error
image

PR #1865, opened 2018-12-11
add refresh for {node,job}-exporter, prometheus and watchdog

### LINKED ISSUE #1748: Webportal display the wrong GPU usage ratio: 179% allocated GPUs
image

Here's the dashboard: http://65.52.108.125:86/dashboard.html.

xudifsd:
fix #1748 , I also fixed a bug when calling `paictl.py service refresh`. Not sure if this is correct. @ydye please have a check.

### DIFF
--- a/deployment/paiLibrary/paiService/service_management_refresh.py
+++ b/deployment/paiLibrary/paiService/service_management_refresh.py
@@ -67,7 +67,7 @@ class service_management_refresh:
 
     def refresh_all_label(self):
         self.logger.info("Begin to refresh all the nodes' labels")
-        machinelist = self.cluster_object_model['machinelist']
+        machinelist = self.cluster_object_model['machine']['machine-list']
         
         labels = ['pai-master', 'pai-worker', 'no-driver', 'no-nodeexporter']
         logging.info("Currently supported labels: " + str(labels))
@@ -102,7 +102,7 @@ class service_management_refresh:
             return
 
         service_conf = file_handler.load_yaml_config("src/{0}/deploy/service.yaml".format(serv))
-        machinelist = self.cluster_object_model['machinelist']
+        machinelist = self.cluster_object_model['machine']['machine-list']
         service_refresher = service_refresh.service_refresh(service_conf, serv, self.label_map)
 
         dependency_list = service_refresher.get_dependency()
--- a/src/prometheus/deploy/refresh.sh
+++ b/src/prometheus/deploy/refresh.sh
@@ -20,9 +20,7 @@
 
 pushd $(dirname "$0") > /dev/null
 
-
-echo "refresh prometheus configuration"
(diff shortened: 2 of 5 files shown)


src/prometheus/deploy/refresh.sh
```diff
@@ -20,9 +20,7 @@
 
 pushd $(dirname "$0") > /dev/null
 
-
-echo "refresh prometheus configuration"
-kubectl apply -f prometheus-configmap.yaml || exit $?
```
hao1939: I'm wandering why it doesn't work. Do you have any idea?

xudifsd: this only works if someone changed some parameters that are allowed to change in description of configmap. It will not restart prometheus service, which might be expected behavior by refresh user.

deployment/paiLibrary/paiService/service_management_refresh.py
```diff
@@ -67,7 +67,7 @@ def get_service_list(self):
 
     def refresh_all_label(self):
         self.logger.info("Begin to refresh all the nodes' labels")
-        machinelist = self.cluster_object_model['machinelist']
+        machinelist = self.cluster_object_model['machine']['machine-list']
```
hao1939: We are going to decouple the statically binding of `machine list`.
Could you query the machine list from k8s api?
Or via `kube get pods`?

xudifsd: you should ask @ydye about this. Anyway, this PR is focus on the refresh of these services not the way of how paictl.py implement it.

fanyangCS: @hao1939, maybe machinelist should be kept for paiservice. it is just for paiservice, machinelist should be retrieved from k8s api, instead of clusterconfig. of course, before deploying paiservice, you will need to validate the k8s cluster state is healthy.

hao1939: We'd better remove statically machinelist, since we are working on a scalable system.
We should base on assume that system is on changing.

hao1939: My currently assume is that:
1. k8s manage the nodes, and label the nodes.
2. PAI service bind to label, for scalability, it should decouple from specific nodes when possible.

fanyangCS: @hao1939 , that makes sense. please present your design once ready.

xudifsd: Ok, please approve this PR if appropriate. Changing to dynamic machine list can be addressed by another PR.

hao1939: ok.

PR #1866, opened 2018-12-11
write timeout parameter explicitly

### LINKED ISSUE #1854: write timeout explicitly in readiness probe
the [default timeout](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#configure-probes) for k8s readiness probe is 1s, this may not be enough in some case, should write timeout value explicitly.

xudifsd:
fix #1854

### DIFF
--- a/src/job-exporter/deploy/job-exporter.yaml.template
+++ b/src/job-exporter/deploy/job-exporter.yaml.template
@@ -41,6 +41,7 @@ spec:
             port: {{ cluster_cfg["job-exporter"]["port"] }}
           initialDelaySeconds: 3
           periodSeconds: 30
+          timeoutSeconds: 10
         command:
         - "python"
         - "/job_exporter/main.py"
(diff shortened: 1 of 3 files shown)


src/job-exporter/deploy/job-exporter.yaml.template
```diff
@@ -41,6 +41,7 @@ spec:
             port: {{ cluster_cfg["job-exporter"]["port"] }}
           initialDelaySeconds: 3
           periodSeconds: 30
+          timeoutSeconds: 3
```
YanjieGao: Does this 3 a little small ? which maybe easy to see readiness probe failed ?

xudifsd: The default is 1s. Our bed shows 1s won't be too small.

fanyangCS: @xudifsd , please consider setting the value to the worst case.

xudifsd: maybe 10?

PR #1870, opened 2018-12-11
decouple data-node image from cluster object module

hao1939:
Passing the POD_IP as data-node hostname.
It should be hostIP when using `docker` host-network mode, and it should also work when using other network modes..

### DIFF
--- a/src/hadoop-data-node/deploy/hadoop-data-node-configuration/datanode-generate-script.sh
+++ b/src/hadoop-data-node/deploy/hadoop-data-node-configuration/datanode-generate-script.sh
@@ -23,7 +23,5 @@ cp  /hadoop-configuration/hadoop-env.sh $HADOOP_CONF_DIR/hadoop-env.sh
 cp  /hadoop-configuration/yarn-env.sh $HADOOP_CONF_DIR/yarn-env.sh
 cp  /hadoop-configuration/mapred-site.xml $HADOOP_CONF_DIR/mapred-site.xml
 
-HOST_NAME=`hostname`
-/usr/local/host-configure.py -c /host-configuration/host-configuration.yaml -f $HADOOP_CONF_DIR/hdfs-site.xml -n $HOST_NAME
-
-sed -i "s/{HDFS_ADDRESS}/${HDFS_ADDRESS}/g" $HADOOP_CONF_DIR/core-site.xml 
+sed -i "s/{POD_IP}/${POD_IP}/g" $HADOOP_CONF_DIR/hdfs-site.xml
+sed -i "s/{HDFS_ADDRESS}/${HDFS_ADDRESS}/g" $HADOOP_CONF_DIR/core-site.xml
--- a/src/hadoop-data-node/deploy/hadoop-data-node.yaml.template
+++ b/src/hadoop-data-node/deploy/hadoop-data-node.yaml.template
@@ -64,6 +64,10 @@ spec:
           value: datanode-generate-script.sh
         - name: START_SERVICE
           value: datanode-start-service.sh
+        - name: POD_IP
+          valueFrom:
+            fieldRef:
+              status.podIP
       imagePullSecrets:
       - name: {{ cluster_cfg["cluster"]["docker-registry"]["secret-name"] }}
       volumes:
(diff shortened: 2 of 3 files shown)


src/hadoop-data-node/deploy/hadoop-data-node-configuration/datanode-generate-script.sh
```diff
@@ -23,7 +23,5 @@ cp  /hadoop-configuration/hadoop-env.sh $HADOOP_CONF_DIR/hadoop-env.sh
 cp  /hadoop-configuration/yarn-env.sh $HADOOP_CONF_DIR/yarn-env.sh
 cp  /hadoop-configuration/mapred-site.xml $HADOOP_CONF_DIR/mapred-site.xml
 
-HOST_NAME=`hostname`
-/usr/local/host-configure.py -c /host-configuration/host-configuration.yaml -f $HADOOP_CONF_DIR/hdfs-site.xml -n $HOST_NAME
-
-sed -i "s/{HDFS_ADDRESS}/${HDFS_ADDRESS}/g" $HADOOP_CONF_DIR/core-site.xml 
+sed -i "s/{POD_IP}/${POD_IP}/g" $HADOOP_CONF_DIR/hdfs-site.xml
+sed -i "s/{HDFS_ADDRESS}/${HDFS_ADDRESS}/g" $HADOOP_CONF_DIR/core-site.xml
```
ydye: Confused about your goal. If you wanna decouple cluster-object-model from the service, why do you still use the value `HDFS_ADDRESS` from cluster-object-model? You should know this value is parsed from `machine-list` in `cluster-configuration.yaml`

ydye: And the PR about zk also confused me a lot.

src/hadoop-data-node/deploy/hadoop-data-node.yaml.template
```diff
@@ -64,6 +64,10 @@ spec:
           value: datanode-generate-script.sh
         - name: START_SERVICE
           value: datanode-start-service.sh
+        - name: POD_IP
+          valueFrom:
+            fieldRef:
```
ydye: Any document about this value?

src/hadoop-data-node/deploy/hadoop-data-node-configuration/datanode-generate-script.sh
hao1939: Hi @ydye ,
I'm refactoring step by step, and I don't expect a breaking change which will block anyone.
So I will decouple the cluster-object-module from the most edge side.

src/hadoop-data-node/deploy/hadoop-data-node.yaml.template
hao1939: Please refer to the k8s doc here: https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-pod-fields-as-values-for-environment-variables.

```
apiVersion: v1
kind: Pod
metadata:
  name: dapi-envars-fieldref
spec:
  containers:
    - name: test-container
      image: k8s.gcr.io/busybox
      command: [ "sh", "-c"]
      args:
      - while true; do
          echo -en '\n';
          printenv MY_NODE_NAME MY_POD_NAME MY_POD_NAMESPACE;
          printenv MY_POD_IP MY_POD_SERVICE_ACCOUNT;
          sleep 10;
        done;
      env:
        - name: MY_NODE_NAME
          valueFrom:
            fieldRef:
              fieldPath: spec.nodeName
        - name: MY_POD_NAME
          valueFrom:
            fieldRef:
              fieldPath: metadata.name
        - name: MY_POD_NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace
        - name: MY_POD_IP
          valueFrom:
            fieldRef:
              fieldPath: status.podIP
        - name: MY_POD_SERVICE_ACCOUNT
          valueFrom:
            fieldRef:
              fieldPath: spec.serviceAccountName
  restartPolicy: Never
```

ydye:
I think you are changing the deployment script rather then the data-node image. And data-node don't have an specific image.

hao1939:
Merged into other branch.

PR #1874, opened 2018-12-12
mount host path root into node-exporter

### LINKED ISSUE #1766: do not alert duplicated file system usage
node-exporter will emit metrics concerning file system usage backed by the same hardware device but have different mount points. Our alerting rule about file system usage did not deduplicate the metrics, so currently we will receive duplicated alerting like
```
FileSystem usage in 10.151.40.156 is   above 80% (current value is: 82.3481878196733) Source
FileSystem usage in 10.151.40.156 is   above 80% (current value is: 82.3481878196733) Source
FileSystem usage in 10.151.40.156 is   above 80% (current value is: 82.3481878196733) Source
FileSystem usage in 10.151.40.156 is   above 80% (current value is: 82.3481878196733) Source
```

Change the alerting rule to remove the duplicated alerting.

xudifsd:
fix #1766 , mount host-root as volume in node-exporter, so that node-exporter can report usage info about all disk mounted in host. Also changed alerting rule to remove duplicated entry in filesystem usage, this will also alert other disks previous ignored by node-exporter.

### DIFF
--- a/src/node-exporter/deploy/node-exporter.yaml.template
+++ b/src/node-exporter/deploy/node-exporter.yaml.template
@@ -80,8 +80,15 @@ spec:
         - containerPort: {{ cluster_cfg["node-exporter"]["port"] }}
           hostPort: {{ cluster_cfg["node-exporter"]["port"] }}
           name: scrape
+        volumeMounts:
+        - name: host-root
+          mountPath: /host-root/
       imagePullSecrets:
       - name: {{ cluster_cfg["cluster"]["docker-registry"]["secret-name"] }}
+      volumes:
+      - name: host-root
+        hostPath:
+          path: /
       hostNetwork: true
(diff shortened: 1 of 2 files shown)


src/node-exporter/deploy/node-exporter.yaml.template
```diff
@@ -80,8 +80,15 @@ spec:
         - containerPort: {{ cluster_cfg["node-exporter"]["port"] }}
           hostPort: {{ cluster_cfg["node-exporter"]["port"] }}
           name: scrape
+        volumeMounts:
+        - name: host-root
```
sterowang: Mount as readonly?

xudifsd: oh, I forgot. Fixed.

PR #1879, opened 2018-12-12
[hadoop-resource-manager] Hard-code ip to ip map in /etc/hosts file.

ydye:


fanyangCS:
for "/etc/hosts", is this in host filesystem or that in RM's docker image?

ydye:
> for "/etc/hosts", is this in host filesystem or that in RM's docker image?

RM's container.

src/hadoop-resource-manager/deploy/hadoop-resource-manager.yaml.template
```diff
@@ -45,6 +45,8 @@ spec:
           name: yarn-resourcemanager-storage
         - mountPath: /var/lib/hadoopdata
           name: hadoop-tmp-storage
+        - mountPath: /host-configuration
```
fanyangCS: why do you this (and the configmap)?

ydye: One machine may have several eth and serveral ip-address. The host-configuration has a yaml containing the machine-list. We can grep the file to find the ip of the machine which configured in our configuration

ydye:
This fix won't work, and i will try to investigate this issue. Try to fix in another way. @sterowang @fanyangCS

PR #1880, opened 2018-12-12
[AKS] Make api-server-url configurable

ydye:


YanjieGao:
pls test affect for k8s part config change

PR #1888, opened 2018-12-14
Pylon: replace WehHDFS URLs in REST server responses

### LINKED ISSUE #1887: SSH info in webportal behind pylon is not replaced.
REST server part in pylon should also replace hdfs ips, now only nm/rm ips.

Gerhut:
Fixes #1887


Gerhut:
@fanyangCS 

50070 is a pure transparent proxy to 5070 without any response replacements, like request the WebHDFS api directly but not through pylon.

So by design it may be redirected to the newer 5070 port, but not pylon's 80 port.

PR #1899, opened 2018-12-17
Fix jenkins failure: /.git/index.lock': File exists.

hao1939:
clean up untracked files, prevent git cmd from failing


Gerhut:
Why `git reset` to another branch? Does `git checkout --force` work?

hao1939:
`get reset` to exactly the same branch, in order to avoid conflict.
`reset` mean regardless any local changes, so it would work. Still don't know why `git pull` fails.

PR #1902, opened 2018-12-18
[Fix] [hadoop-resource-manager] Fix name resolve behavior for rm.

ydye:
Add the ip to ip mapper in the /etc/hosts


ydye:
This fix won't work, and i will try to investigate this issue. Try to fix in another way. @sterowang @fanyangCS

PR #1905, opened 2018-12-18
add TaintBasedEvictions

xudifsd:
if not enable this feature gate, toleration for disk/memory pressure will not take effect. See [this](https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/#taint-based-evictions)


YanjieGao:
here is prerequisites but related with os level
https://github.com/Microsoft/pai#1-prerequisites

As k8s is deployed by pai (not by user), i am not sure which is the place it should mentioned for now. 

But future decouple service deployment with k8s deployment, it is better to add prerequest k8s version for service deployment.

PR #1934, opened 2018-12-21
[QoS] qos switch to disable the resource requirement.

ydye:
- k8s level

Add a configuration named ```kubernetes.qos-switch```. And its default value will be ```"true"```.

If ```kubernetes.qos-switch``` is ```"true"```, the ```resources.requests or resource.limits``` part will be enabled.
If ```kubernetes.qos-switch``` is ```"false"```, the ```resources.requests or resource.limits``` part will be removed from the component.

See the example code following.
```
{% if cluster_cfg['kuberentes']['qos-switch] == "true" -%}
    resources:
      requests:
        memory: "1Gi"
        cpu: "1000m"
{%- endif %}
``` 

- service level

Add a configuration named ```cluster.common.qos-switch```. And its default value will be ```"true"```.

If ```cluster.common.qos-switch``` is ```"true"```, the ```resources.requests or resource.limits``` part will be enabled.
If ```cluster.common.qos-switch``` is ```"false"```, the ```resources.requests or resource.limits``` part will be removed from the service.

```
{% if cluster_cfg['cluster']['common']['qos-switch] == "true" -%}
        resources:
          limits:
            memory: "36Gi"
          requests:
            memory: "8Gi"
{%- endif %}
```

### DIFF
--- a/deployment/clusterObjectModel/mainParser/kubernetes.py
+++ b/deployment/clusterObjectModel/mainParser/kubernetes.py
@@ -97,6 +97,8 @@ class Kubernetes:
         else:
             com_kubernetes["etcd-data-path"] = k8s_cfg["etcd-data-path"]
 
+        com_kubernetes["qos-switch"] = k8s_cfg["qos-switch"] if ("qos-switch" in k8s_cfg) else "true"
+
         k8s_master_list = self.get_k8s_master_machine()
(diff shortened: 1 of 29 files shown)


ydye:
I think we only need to modify them this time. And combine it with template and handle it by cluster-configuration, it's friendly to user.

PR #1947, opened 2018-12-26
Break down Kubernetes and PAI deployment

hao1939:
TODO: update the description later.

### DIFF
--- a/src/hadoop-node-manager/deploy/hadoop-node-manager-configuration/yarn-site.xml
+++ b/src/hadoop-node-manager/deploy/hadoop-node-manager-configuration/yarn-site.xml
@@ -92,7 +92,7 @@
 
     <property>
       <name>yarn.nodemanager.hostname</name>
-      <value>{{ host_config['ip'] }}</value>
+      <value>{POD_IP}</value>
     </property>  
 
     <property>  
@@ -137,7 +137,7 @@
 
     <property>
       <name>yarn.nodemanager.address</name>
-      <value>{{ host_config['ip'] }}:8041</value>
+      <value>{POD_IP}:8041</value>
     </property>
 
     <property>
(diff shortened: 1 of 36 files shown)


src/hadoop-node-manager/deploy/hadoop-node-manager-configuration/yarn-site.xml
```diff
@@ -92,7 +92,7 @@
 
     <property>
       <name>yarn.nodemanager.hostname</name>
-      <value>{{ host_config['ip'] }}</value>
+      <value>{POD_IP}</value>
```
ydye: Though, it's good for me. But you'd better decouple it from thhis PR,

src/prometheus/deploy/prometheus-configmap.yaml.template
```diff
@@ -32,7 +32,7 @@ data:
     - job_name: 'pai_serivce_exporter'
       scrape_interval: {{ prom_info["scrape_interval"] }}s
       kubernetes_sd_configs:
-      - api_server: "{{ cluster_cfg["kubernetes"]["api-servers-url"] }}"
+      - api_server: {{ cluster_cfg["cluster"]["kubernetes"]["api-servers-url"] }}
```
ydye: I think kubernetes should have a independent key, rather than a child key of cluster.

hao1939: Hi @ydye,
I haven't decided where to place it.
Now, put it under 'cluster' because don't need to create a separately config parser.
Will discuss it later.

src/hadoop-node-manager/deploy/hadoop-node-manager-configuration/yarn-site.xml
hao1939: En, Ideally we need to separate it from this PR.
Put it here because I want to avoid a complex fixing of the start script.

Issue #1946, opened 2018-12-26
zombie CNTK test job


mzmssg:
**Organization Name**: 
ms

**Short summary about the issue/question**: 
Occassionally, CNTK test can't exit normally( highly suspect it's hdfs_mount tool reason). Becasue it's a periodical job, might leave many zombie containers on hosts. These zombie containers will further lead docker daemon hangs. Need to replace with a new test example or rewrite hdfs operations. 

**Brief what process you are following**: 

**How to reproduce it**: 

**OpenPAI Environment**:
- OpenPAI version:
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Hardware (e.g. core number, memory size, storage size, GPU type etc.):
- Others:

**Anything else we need to know**:

PR #1965, opened 2019-01-02
[test] replace e2e cntk to tensorflow job to avoid "mount HDFS case zoombie container of CNTK job"

YanjieGao:
related to this issue:
https://github.com/Microsoft/pai/issues/1946

Test results:
image


mzmssg:
could the test run on singlebox?

YanjieGao:
> could the test succeed on singlebox?

it passed ci singlebox test

YanjieGao:
@mzmssg , for singlebox deployment has tested and successed. To reduce the resource cost, optimized
 resource and training config

PR #2008, opened 2019-01-15
add more node related alerts

xudifsd:


fanyangCS:
I remember we already have alert for memory pressure?

xudifsd:
Oh, yes, it subtract total memory with used from node-exporter, newly added memory_pressure is kubernetes defined. May duplicated. Which one to use?

fanyangCS:
we use our own. k8s cannot modify job containers.

xudifsd:
ok, removed k8s one

xudifsd:
@fanyangCS please approve this if appropriate.

PR #2018, opened 2019-01-16
hot fix for job no random number replace problem

YanjieGao:


YanjieGao:
need cherry pick to v0.9.y

PR #2038, opened 2019-01-18
[Restserver] Append ip hostanme pairs into /etc/hosts in rdma workload.

ydye:
1) Append ip hostanme pairs into /etc/hosts
2) Keep the line starting with 127.0.0.1 as ```127.0.0.1 localhost```


src/rest-server/src/models/job.js
```diff
@@ -434,6 +434,10 @@ class Job {
   }
 
   generateDockerContainerScript(data, idx) {
+    let paiMachineList = '';
+    if (fs.existsSync('/pai-cluster-config/cluster-configuration.yaml')) {
```
Gerhut: In fact, this is a pretty slow operation, at least we could move it into an azRDMA condition block.

src/rest-server/src/models/job.js
```diff
@@ -434,6 +434,10 @@ class Job {
...
+      paiMachineList = yaml.safeLoad(fs.readFileSync('/pai-cluster-config/cluster-configuration.yaml', 'utf8'))['machine-list']
```
Gerhut: Use try-catch block around read instead of exists-read to prevent race conditions.

src/rest-server/src/models/job.js
ydye: Yes, agree with you that the solution is solw. But the cluster-configuration file is not a static file. And it will be updated, if admin perform the config update command. Are there any solution to update the config for a specify time period.

PR #2041, opened 2019-01-21
[DO NOT MERGE] implement kube-runtime & rest-lit

xudifsd:
commit kube-runtime & rest-lite to dev branch. PR only for review purpose, do not merge


src/kube-runtime/build/kube-runtime.dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM golang:alpine as builder
+
+ENV PROJECT_DIR=${GOPATH}/src/github.com/microsoft/frameworkcontroller
```
yqwang-ms: suggest to do not build frameworkbarrier, just copy the binary from official image?
Such as:
1.	https://docs.docker.com/develop/develop-images/multistage-build/
2.	https://stackoverflow.com/questions/48265608/is-it-possible-to-copy-a-file-out-of-a-docker-image-without-actually-running-the
3.	Run barrier as docker run, instead of raw process run?

yqwang-ms: suggest to do not build frameworkbarrier, just copy the binary from official image?
Such as:
1.	https://docs.docker.com/develop/develop-images/multistage-build/
2.	https://stackoverflow.com/questions/48265608/is-it-possible-to-copy-a-file-out-of-a-docker-image-without-actually-running-the
3.	Run barrier as docker run, instead of raw process run?

xudifsd: What's the name of official image? frameworkbarrier?

I do not think call `docker run` in container is a good practice.

yqwang-ms: Pls check: https://hub.docker.com/r/frameworkcontroller/frameworkbarrier
So, I list it in the last option, is it possible to use previous 1 and 2?

src/kube-runtime/build/kube-runtime.dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM frameworkcontroller/frameworkbarrier:v0.3.0 as barrier
```
yqwang-ms: >v0.3.0 [](start = 42, length = 6)

So, we need to tell user to use v0.3.0 frameworkcontroller?

PR #2047, opened 2019-01-22
increase job-exporter memory limit

xudifsd:
`k8s_job-exporter_job-exporter-l4ptn_default_d804d9e9-1d74-11e9-b630-000d3ab7e6d6_0 126.6MiB / 128MiB`.

increase memory limit for job-exporter and fix some code error.


fanyangCS:
by the way, I assume this is the limit, not minimum requirement of the memory usage, right?

xudifsd:
@fanyangCS yes, requirement is represented by field `request` which is not specified by job-exporter.

PR #2050, opened 2019-01-22
Prevent schedule to stuck GPUs

mzmssg:
Mark a node unhealthy will:
1. Exclude its resource in total resource 
2. Kill running container on this node


mzmssg:
Some details of unhealthy status
http://johnjianfang.blogspot.com/2014/10/unhealthy-node-in-hadoop-two.html

src/hadoop-node-manager/deploy/hadoop-node-manager-configuration/health_checker.sh
```diff
@@ -0,0 +1,35 @@
...
+# DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+
+
+function check_gpu_hang()
+{
+  stuck_processes=`ps axo pid,state,command | awk '$0~"nvidia-smi" && $2~"D" {print $1}' | xargs`
+  if [ -n "$stuck_processes" ]; then
+    echo "ERROR: found stuck processes: ${stuck_processes}"
```
fanyangCS: does it need to return something to report is not ok?

mzmssg:
@fanyangCS 
Check YARN scheduler logic, it doesn't support exclude resource but keep container. If we want to black a node but don't kill node containers, YARN resource view might become messy.

PR #2058, opened 2019-01-23
[Paictl] SSH tool for admin to maintain cluster.

ydye:


paictl.py
```diff
@@ -502,6 +504,49 @@ def update_external_config(self, args):
         external_conf_update.update_latest_external_configuration()
 
 
+class utility(SubCmd):
```
xudifsd: use `Utility`

paictl.py
```diff
@@ -502,6 +504,49 @@ def update_external_config(self, args):
...
+    def register(self, parser):
+        utility_parser = parser.add_subparsers(help="utility for maintaining in a easy way.")
+
+        ssh_parser = SubCmd.add_handler(utility_parser, self.cluster_ssh, "ssh")
+        ssh_parser.add_argument("-p", "--config-path", dest="config_path", required=True, help="path of cluster configuration file")
```
xudifsd: Do we need -p? Since the config is in configMap.

ydye: In this case, just a tool to help admin execute a command in a cluster level. For example, first copy the az-rdma.sh to the machine which has been selected, and then execute the script on all the machine.

ydye:
> What's the purpose of this PR? Will merge into master?

help admin enable az-rdma in cluster.

PR #2072, opened 2019-01-24
Reduce am resource

mzmssg:
1. reduce AM memory to 1G
2. increase YARN virtual memory ratio to 6(1g physical memory with 6g virtual memory), previous value is 2.1. Launcher AM might consume 3G+ virtual memory.

Test case:
A framework with 9 taskroles, 5 tasks per taskrole, 2-node cluster, for one hour.


Gerhut:
Shall we make it as default config in launcher?

yqwang-ms:
By setting this, it is already default for all jobs.

mzmssg:
@Gerhut 
Seems it would be better if we put it in job config instead hard-code it here?

mzmssg:
@Gerhut 
Updated

Issue #1986, opened 2019-01-09
Add jsonc support for job submission page (V1)


zhouhaoyu:
**What would you like to be added**: JSONC (JSON with comments) support on the job submission page (V1) of PAI web portal.

**Why is this needed**: We are currently working on vscode extension for OpenPAI, and would like to improve readability for editing job config file by adding relevant comments in json. To streamline the job submission experience, it could be better if user can upload the commented json straightly to PAI web portal.

**Without this feature, how does the current module work**：User need to strip out the comments from json in order to submit job in PAI web portal.

**Components that may involve changes**: PAI Web portal

fanyangCS:
@scarlett2018 , yes. Yaml has built-in support for comments.

fanyangCS:
closed.

PR #2084, opened 2019-01-25
webportal: allow jsonc in job submission

zhouhaoyu:
Allowing user to import json with comment (JSONC) on job submission page.
Addressing #1986.

### DIFF
--- a/src/webportal/package.json
+++ b/src/webportal/package.json
@@ -67,6 +67,7 @@
     "sass-loader": "~6.0.6",
     "serve-favicon": "~2.5.0",
     "slick-carousel": "~1.8.1",
+    "strip-json-comments": "^2.0.1",
     "style-loader": "~0.19.0",
     "uglifyjs-webpack-plugin": "~1.1.4",
     "util": "~0.10.3",
--- a/src/webportal/src/app/job/job-submit/job-submit.component.ejs
+++ b/src/webportal/src/app/job/job-submit/job-submit.component.ejs
@@ -6,7 +6,7 @@
       <div class="row">
         <div class="form-group col-md-4">
           <label for="fileUpload">Import Json</label>
-          <input class="btn btn-default" id="fileUpload" type="file" accept=".json">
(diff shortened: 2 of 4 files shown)


Gerhut:
Hi @zhouhaoyu , thanks for your contribution.

I just read #1986, and have not understand your expected experience yet.

To submit JSONC to PAI in vscode extension, there are two solutions:

1. Convert JSONC to JSON (strip comments) in client side, and submit JSON to PAI. 
2. Make REST server able to receive JSONC as job config, also store JSONC for latter retrieving.

I assume you choose solution 1, I think these steps should be done 1 by 1:

1. Make the extension support JSONC, strip comments in extension side.
2. Propose that web portal should align to the vscode extension, to fill the expected experience:
      > Export JSONC job config from vscode extension (beacuse JSONC will not stored in REST server) and import it to webportal for submission

(Is this your expected experience?)

As I know, vscode extension have not implement the JSONC support now. So directly making the step 2 done seems useless currently.

@fanyangCS @scarlett2018 for discussion.

zhouhaoyu:
> To submit JSONC to PAI in vscode extension, there are two solutions:
> 
> 1. Convert JSONC to JSON (strip comments) in client side, and submit JSON to PAI.
> 2. Make REST server able to receive JSONC as job config, also store JSONC for latter retrieving.
> 
In fact, what I'm trying to achieve is allowing user to edit jsonc in vscode directly as the commented configuration file will be easier to read.

When user finishes their editing and wants to submit a job, they can either

1. ...submit the file directly to web portal, without an extra step to strip the comments, which is currently not available due to web portal not allowing jsonc
2. ...or use the extension to submit the configuration to PAI via the REST api directly, in which case the extension will take care of the conversion from jsonc to json

By allowing jsonc in web portal job submission page, the first option will be available.

Gerhut:
Offline synced with @fanyangCS , this change could be checked-in as long as it introduces no extra dependencies. So approved, feel free to :shipit:

PR #2087, opened 2019-01-28
[doc fix] tell user notice dev-box version

YanjieGao:


YanjieGao:
ok, thanks , has merged

PR #2091, opened 2019-01-28
[Azure RDMA] Merge Azure RDMA change into master branch

ydye:
- [x] Cluster Level Configuration Switch for Admin to enable the Az RDMA feature
- [x] User Level Job Parameter to get the Az RDMA Capable container
- [x] Necessary code change in restserver to enable az-RDMA environment to the job container.
- [x] Some useful tool of ssh and sftp-copy to help admin to maintain the cluster machines.
- [x] An example job of intel MPI benchmark based on azure RDMA. And Guide user to run the mpi task.
- [x] A tutorial for admin to enable azure rdma for the cluster.

- existing issue
1. If the machine in the cluster is removed or added, admin should restart restserver to refresh the machine list.  Pending it as a feature and do it in the feature.


src/rest-server/src/models/job.js
```diff
@@ -425,6 +427,8 @@ class Job {
           'jobData': data,
           'inspectFormat': '{{.State.Pid}}',
           'jobEnvs': jobEnvs,
+          'azRDMA': azureEnv.azRDMA == 'false' ? false : true,
```
mzmssg: `===` might be better?

deployment/utility/sftp_copy.py
```diff
@@ -0,0 +1,95 @@
...
+import logging.config
+
+
+class OpenPaiSftpCopy:
+
+    def __init__(self, filename, source, dest, cluster_config_path, filter):
+        self.filename = filename
+        self.source = source
+        self.dest = dest
+        objectModelFactoryHandler = cluster_object_model.cluster_object_model(configuration_path=cluster_config_path)
```
hao1939: What about do not pass the whole object_model? 
For example, here you may only want a `machine-list`.
So we can narrow down which is need by the function.

deployment/utility/sftp_copy.py
```diff
@@ -0,0 +1,95 @@
...
+            elif user_input == "Y":
+                break
+            else:
+                print(" Please type Y or N.")
+            count_input = count_input + 1
+            if count_input == 3:
+                logger.warning("3 Times.........  Sorry,  we will force stopping your operation.")
+                sys.exit(1)
+
+    def run(self):
```
hao1939: We'd better choose some name more readable.
For example, it could be: uploade_xxx_ and so on.

deployment/utility/ssh.py
```diff
@@ -0,0 +1,97 @@
...
+        self.logger.info("\n")
+        self.logger.info("\n")
+
+        self.logger.info("=============================================")
+        self.logger.info("======= Machine List After filtered =========")
+        self.logger.info("=============================================")
+        for hostname in self.cluster_object_model['machine']["machine-list"]:
+            host = self.cluster_object_model['machine']["machine-list"][hostname]
+            flag = True
+            for rule in rule_list:
```
mzmssg: `for else` could get rid of the flag

paictl.py
```diff
@@ -502,6 +505,51 @@ def update_external_config(self, args):
         external_conf_update.update_latest_external_configuration()
 
 
+class Utility(SubCmd):
```
hao1939: Should we put a new command to paictl?
It looks like won't work for the platform we support. What about leave it as a separately script?

deployment/utility/sftp_copy.py
ydye: I don't think so. I think the unified entry point could make more sense.

src/rest-server/src/models/job.js
ydye: Waht's the difference in the string compare between ```===``` and ```==```

paictl.py
ydye: I prefer to manage them in one tool like paictl. And why it won't work for the platform we support?

deployment/utility/sftp_copy.py
hao1939: `run` is a too common name, which should be avoided.

src/rest-server/src/models/job.js
mzmssg: It works in this case since we have validation before, but never use `==` is a better spec

ydye: Yes. So I'm really curious about in which input without validation will cause an issue. Do you have an example

deployment/utility/sftp_copy.py
ydye: This should be an optimization of cluster object model. And maybe you could fire an new issue to track this.

paictl.py
hao1939: My concern is that: 
1. paictl should focus on the major features of PAI.
2. We could provide associated tools for convenient, and release it as a separately script. In case there are qualified thirdparty tools, we can replace it.

There are plenty of tool for copying or uploading files, and we are not completing with them.

deployment/utility/sftp_copy.py
hao1939: I mean, the function don't need to accept a 'cluster_object_model'.

It's just a pseudo:

machine_list = ...  # how to generate machine list is out of the scope of the class OpenPaiSftp
openPaiSftp = OpenPaiSftp(machine_list) # remove copy from the class name. Class name should better be a `noun`.
result = openPaiSftp.copy(from_dir, to_dir) # So we don't need to call `run`, we call the `verb` copy.

src/rest-server/src/models/job.js
ydye: @Gerhut What's your opinion?

paictl.py
ydye: 1. And you also should find a tool which could leverage our machine-list. So that, user could avoid to write another machine-list to do it.  
2. Our solution is based on paramiko.

deployment/utility/sftp_copy.py
ydye: Thank you for your suggestion. And the machine-list part will be optimized. But I will keep the naming style as before.

paictl.py
ydye: Maybe we could rise an issue to track this idea.

ydye:
Optimization of machine-list is done.

ydye:
Depending on ```for...else...``` done.

src/rest-server/src/models/job.js
mzmssg: BTW, seems we don't need `? :` here?

ydye:
```
> true=="false"
false
> undefined=="false"
false
> 0=="false"
false
> false=="false"
false
>
```
@mzmssg

ydye: Replace "==" with "===" .

mzmssg: test `[false] == 'false'`

ydye:
Replace ```==``` with ```===```

deployment/utility/ssh.py
```diff
@@ -26,7 +27,7 @@ class OpenPaiSSH:
     def __init__(self, command, cluster_config_path, filter):
         self.cmd = command
         objectModelFactoryHandler = cluster_object_model.cluster_object_model(configuration_path=cluster_config_path)
-        self.cluster_object_model = objectModelFactoryHandler.run()
+        self.origin_machine_list = objectModelFactoryHandler.run()["machine"]["machine-list"]
```
hao1939: Hi @ydye,

Noticed the changes, and there could be one more step move forward:
passing the `machine_list` instead of `cluster_config_path`.

Consideration: here only need a `machine_list`, don't have to know it generate from `cluster_config_path` using a `objectModelFactoryHandler`.

paictl.py
hao1939: Hi @ydye,

It's about we should be carefully when we are adding a new command.
We frequently update commands, user will be annoyed .
I'm against adding a new 'utils' sub command, because we are not sure it's really needed.

In the same time, I recommend to provide it in a more flex format, for example, a standalone script.

deployment/utility/ssh.py
ydye: I think put this logic to paictl.py is not a good idea. Because we should make paictl more easy to read. And the logic should be separated into the class implementation module. And paictl only pass the raw parameters from the input.

ydye: Done

ydye:
Move machine list generate logic to paictl. Done.

PR #2094, opened 2019-01-29
Test plan outlines

hao1939:
It's the outlines of test plan.

Here's some assumption:
1. Benchmark task run on PAI as jobs
2. Benchmark task could be run repeatedly, separately and incrementally
3. Benchmark task generate report and aggregate to a shared directory

Hi @fanyangCS , could you review and put some feedback? For example:
1. Are we on the same page about the big picture?
2. Which part should be high priority?
3. Strategies.

### DIFF
--- a/docs/benchmark.md
+++ b/docs/benchmark.md
@@ -8,17 +8,32 @@ The benchmark runs on PAI as a normal job, consistents of many metric jobs. All 
 
 Fully test and demostrate API & Protocol, report the supported protocols.
 
-- User APIs
-- Job APIs
-- Metric APIs
+### User APIs
+
+### Job APIs
+
+- verify GPU drivers works
+- verify port works
+
+### Metric APIs
+
+- verify metrics works
 
 ## Capability and Limitations
 
 Consistent of bunch of extrame jobs which would push PAI into crash, try to figure out the limitation.
 
-- Memory stress
-- Disk stress
-- Massive count of jobs, test the scheduler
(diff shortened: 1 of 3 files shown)


fanyangCS:
@hao1939 , suggest to firstly test on the tricky issues we have found during the k8s deployment. 
1. Specific Docker configuration in our deployment (e.g., log rotation, path customization, driver path config)
2. Tolerance of node memory/disk pressure. we disable k8s memory/disk pressure, will need to figure out if it is possible to do that in AKS, and if can't how to provide an alternative to customers.
3. Raw Docker instance vs. AKS pod. We need to verify whether Yarn/Prometheus works correctly for job containers. If job containers can get what they need (e.g., port, host IP).
4. pai-ctl. if pai-ctl works correctly. if some features are missing, we need to document them clearly.
5. please work with team to collect those tricky issues to see if I miss anything

hao1939:
Hi @fanyangCS ,

I breakdown some test cases for the situation your mentioned.

For the paictl.py testing, it's very tricky, I don't put it here.
And here's some suggestion about the paictl.py testing:
1. We have already covered the most frequently used cmd in our CI. We could add more if necessary.
2. Remove the cmds that are not covering in CI. Those cmd may be rarely used or for specific environments.
3. We provide some tools without officially support. For example those cmds we removed from paictl.py.

scarlett2018:
closing as out of date

PR #2119, opened 2019-02-01
Diawang/dockercleaner

wangdian:
Clean logic V0.1
1. Monitor disk usage on docker root, do not need to mount docker disk as volume manually
2. Stop container that uses largest disk space, use a white list to avoid killing system containers.
3. Send SIGUSR1(10) to container as termination signal


src/cleaner/cleaner_main.py
```diff
@@ -76,31 +76,20 @@ def sync(self):
...
-    else:
-        raise ValueError("arguments %s is not supported.", arg)
-    return worker
-
-
-liveness_files = {
-    "docker_cache": "docker-cache-cleaner-healthy",
-    "deleted_files": "deleted-files-cleaner-healthy"
-}
+def get_worker(threshold):
```
mzmssg: Doesn't plan support multiworker?

deployment/quick-start/services-configuration.yaml.template
```diff
@@ -128,3 +128,7 @@ rest-server:
 # uncomment following section if you want to customize the port of pylon
 # pylon:
 #  port: 80
+
+# uncomment following section if you want to customize the docker path of cleaner
```
fanyangCS: >uncomment [](start = 2, length = 9)

remove these three lines if you have not implemented it yet.

src/cleaner/config/cleaner.py
```diff
@@ -18,23 +18,29 @@
...
-class Cleaner:
-
-    def __init__(self, cluster_configuration, service_configuration, default_service_configuraiton):
+    def __init__(self, cluster_conf, service_conf, default_service_conf):
         self.logger = logging.getLogger(__name__)
-
-        self.cluster_configuration = cluster_configuration
+        self.cluster_conf = cluster_conf
+        self.service_conf = service_conf
+        self.default_service_conf = default_service_conf
```
fanyangCS: >default_service_conf [](start = 36, length = 20)

there is a cluster-object-model you should follow. Please ping Yundong for help.

src/cleaner/cleaner_main.py
```diff
@@ -19,7 +19,7 @@
 import argparse
 import os
 from datetime import timedelta
-from cleaner.scripts import clean_docker_cache, check_deleted_files
+from cleaner.scripts import clean_docker_cache, check_deleted_files, clean_docker
```
mzmssg: you could remove unnecessary import if whithout mutltiworker support

src/cleaner/cleaner_main.py
```diff
@@ -76,31 +76,20 @@ def sync(self):
...
     parser = argparse.ArgumentParser()
-    parser.add_argument("option", help="the functions currently supported: [docker_cache | deleted_files]")
+    parser.add_argument("-t", "--threshold", help="the disk usage precent to start cleaner")
     args = parser.parse_args()
 
     common.setup_logging()
 
-    cleaner = Cleaner(liveness_files[args.option])
-    cleaner.add_worker(args.option, get_worker(args.option))
+    cleaner = Cleaner("docker-cleaner-healthy")
```
mzmssg: Could we move this liveness file path to `clean_docker`, less hard-code here seems to be better

src/cleaner/config/cleaner.py
```diff
@@ -18,23 +18,29 @@
...
-        return com
-
-    def validation_post(self, cluster_object_model):
+        result = copy.deepcopy(self.default_service_conf)
+        result.update(self.service_conf)
+        return result
+
+    def validation_post(self, conf):
+        threshold = conf["cleaner"].get("threshold")
+        if type(threshold) != int:
```
mzmssg: might convert it to int in a `try ... catch`, then check whether it's in `(0, 100)` is better.

src/cleaner/scripts/clean_docker.py
```diff
@@ -0,0 +1,85 @@
...
+                size = int(splitline[4][:-1])
+    except ValueError:
+        logger.error("cannot get disk size, reset size to 0")
+        size = 0
+    logger.info("Checking disk, disk usage = {0}%".format(size))
+    return size
+
+
+def check_and_clean(threshold):
+    if check_disk_usage("/") > int(threshold):
```
mzmssg: So you will only check root filesystem?

src/cleaner/scripts/clean_docker.py
```diff
@@ -0,0 +1,85 @@
...
+        logger.info("Disk usage is above {0}%, Try to remove containers".format(threshold))
+        kill_largest_container()
+
+
+# Clean logic v1: kill largest container
+white_list = ["k8s_kube", "k8s_pylon", "k8s_zookeeper", "k8s_rest-server", "k8s_yarn", "k8s_hadoop", "k8s_job-exporter", "k8s_watchdog", "k8s_grafana", "k8s_node-exporter", "k8s_webportal", "k8s_prometheus", "k8s_nvidia-drivers", "k8s_etcd-container", "k8s_apiserver-container", "k8s_docker-cleaner", "kubelet"]
```
mzmssg: @fanyangCS 
Seems we should add `dev-box` here before we complete dev-box management

mzmssg: I don't remember we have some container starting with `k8s_yarn`. 

Whatever, I think you could simply regard `k8s_` and `kubelet` is our service.

src/cleaner/scripts/clean_docker.py
```diff
@@ -0,0 +1,85 @@
...
+def kill_largest_container():
+    containers = []
+    # Only try to stop PAI jobs and user created containers
+    containers_source = subprocess.Popen(["docker", "ps", "-a", "--format", r'{{.ID}}\t{{.Image}}\t{{.Size}}\t{{.Names}}'], stdout=subprocess.PIPE)
+    for line in containers_source.stdout:
+        splitline = line.split("\t")
+        ignore = False
+        for prefix in white_list:
```
mzmssg: Leverage `for ... else` to remove the flag is more pythonic.

src/cleaner/scripts/clean_docker.py
```diff
@@ -0,0 +1,85 @@
...
+        logger.info("Disk usage is above {0}%, Try to remove containers".format(threshold))
+        kill_largest_container()
+
+
+# Clean logic v1: kill largest container
+white_list = ["k8s_kube", "k8s_pylon", "k8s_zookeeper", "k8s_rest-server", "k8s_yarn", "k8s_hadoop", "k8s_job-exporter", "k8s_watchdog", "k8s_grafana", "k8s_node-exporter", "k8s_webportal", "k8s_prometheus", "k8s_nvidia-drivers", "k8s_etcd-container", "k8s_apiserver-container", "k8s_docker-cleaner", "kubelet"]
+def kill_largest_container():
+    containers = []
+    # Only try to stop PAI jobs and user created containers
+    containers_source = subprocess.Popen(["docker", "ps", "-a", "--format", r'{{.ID}}\t{{.Image}}\t{{.Size}}\t{{.Names}}'], stdout=subprocess.PIPE)
```
mzmssg: We might need to record what size is considered in `docker ps`.
Currently following items is excluded:
- disk space used for log-files (if you use the json-file logging driver) - which can be quite a bit if your container generates a lot of logs, and log-rotation (max-file / max-size logging options) is not configured
- volumes used by the container
- disk space used for the container's configuration files (hostconfig.json, config.v2.json, hosts, hostname, resolv.conf) - although these files are small
- memory written to disk (if swapping is enabled)
- checkpoints (if you're using the experimental checkpoint/restore feature)

Comes from https://github.com/docker/docker.github.io/issues/1520

src/cleaner/scripts/clean_docker.py
```diff
@@ -0,0 +1,85 @@
...
+
+size_defs={'B':1, 'K':1024, 'M':1024**2, 'G':1024**3, 'T':1024**4, 'b':1, 'k':1024, 'm':1024**2, 'g':1024**3, 't':1024**4}
+def calculate_size(size_str):
+    size_search = re.search(r"[BbKkMmGgTt]", size_str)
+    return float(size_str[0:size_search.start()]) * size_defs[size_search.group()]
+
+
+if __name__ == "__main__":
+    common.setup_logging()
+    check_and_clean(60)
```
mzmssg: I think expose the monitor interval to configuration is necessary, otherwise we need rebuild image every time

src/rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -29,10 +29,17 @@ function exit_handler()
...
+  printf "%s %s\n" \
+    "[DEBUG]" "Docker container kill handler: SIGTERM signal received in docker container, exiting ..."
+    exit 0
+}
+
 set -x
 PS4="+[\t] "
 trap exit_handler EXIT
-
+trap kill_handler SIGTERM
```
mzmssg: `docker kill` support `-s` option to choose signal, please trap other signal instead of SIGTERM, which is used in many places, such as https://github.com/Microsoft/pai/blob/acb982d1272924dffa49d5cc220d5f1bb6a40594/src/hadoop-node-manager/deploy/hadoop-node-manager-delete/delete.sh#L27

src/rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -29,10 +29,17 @@ function exit_handler()
     "[DEBUG]" "Docker container exit handler: EXIT signal received in docker container, exiting ..."
 }
 
+function kill_handler()
+{
+  printf "%s %s\n" \
+    "[DEBUG]" "Docker container kill handler: SIGTERM signal received in docker container, exiting ..."
+    exit 0
```
mzmssg: Don't `exit 0`, it will be considered as a succeeded job. 
Also, you could dump some log like "consume too much space" and suggest to use hdfs or nfs.

src/cleaner/scripts/clean_docker.py
mzmssg: Sorry, fogget `yarn exporter`

src/cleaner/cleaner_main.py
fanyangCS: agree. If you plan to support multiworker in the future. you can bring it back when you need it. the current implementation is a bit obscure.


---
In reply to: [252962426](https://github.com/Microsoft/pai/pull/2119#discussion_r252962426) [](ancestors = 252962426)

wangdian: Sure in current version, we can remove the worker wrap. Zhaoyu's original design is separate the cleaner into two parts: docker cleaner and file cleaner, and now the clean process is all based on docker. Later if we want to manage diskspace in different dimensions, we will bring it back.

deployment/quick-start/services-configuration.yaml.template
wangdian: Already implemented

src/cleaner/config/cleaner.py
wangdian: I followed other module's config code to do this. Confirmed with Yundong and the format is right. Will add Yundong to review when I finish other fixes.

src/cleaner/scripts/clean_docker.py
wangdian: Not include in this version.

src/cleaner/scripts/clean_docker.py
```diff
@@ -0,0 +1,85 @@
...
+                break
+        if ignore == False:
+            size = calculate_size(splitline[2].split()[0])
+            containers.append([size, splitline[0], splitline[1]])
+
+    containers.sort(key=lambda x:x[0], reverse=True)
+
+    if containers.count > 0 and containers[0][0] > 1024**3:
+        logger.warning("Kill container {0} due to disk pressure. Container size: {1}".format(containers[0][1], containers[0][0]))
+        subprocess.Popen(["docker", "container", "stop", containers[0][1]])
```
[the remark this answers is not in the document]
wangdian: Good point. That depends on how we define the scope of the cleaner. I suggest we follow the docker's way to kill the container, and  discuss other cases when we really get it.

src/rest-server/src/templates/dockerContainerScript.mustache
wangdian: Accepted

src/rest-server/src/templates/dockerContainerScript.mustache
wangdian: Accepted

src/cleaner/scripts/clean_docker.py
wangdian: Thanks

src/cleaner/config/cleaner.yaml
```diff
@@ -15,3 +15,5 @@
 # DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
+threshold: 94
```
ydye: A markdown document is needed to describe this service's configuration. Like this: [https://github.com/Microsoft/pai/blob/pai-0.9.y/src/cluster/config/cluster.md](https://github.com/Microsoft/pai/blob/pai-0.9.y/src/cluster/config/cluster.md)

src/cleaner/scripts/clean_docker.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def check_and_clean(self):
+        if self.check_disk_usage("/") >= self.__threshold:
+            self.logger.info("Disk usage is above {0}%, Try to remove containers".format(self.__threshold))
+            self.kill_largest_container()
+
+
+    # Clean logic v1: kill largest container
+    white_list = ["k8s_kube", "k8s_pylon", "k8s_zookeeper", "k8s_rest-server", "k8s_yarn", "k8s_hadoop", "k8s_job-exporter", "k8s_watchdog", "k8s_grafana", "k8s_node-exporter", "k8s_webportal", "k8s_prometheus", "k8s_nvidia-drivers", "k8s_etcd-container", "k8s_apiserver-container", "k8s_docker-cleaner", "kubelet", "dev-box"]
+    def kill_largest_container(self):
```
ydye: seems controller and scheduler are not in the white list.

src/cleaner/config/cleaner.yaml
wangdian: Sure, will add it

src/cleaner/scripts/clean_docker.py
wangdian: Cleaner will only clean the containers larger then 1G.
Also will add k8s pods to ignore list.

src/rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -29,10 +29,17 @@ function exit_handler()
     "[DEBUG]" "Docker container exit handler: EXIT signal received in docker container, exiting ..."
 }
 
+function kill_handler()
+{
+  printf "%s %s\n" \
+    "[INFO]" "Docker container killed due to disk pressure. If your job needs large disk space, please use HDFS or NFS to store your data."
```
fanyangCS: Container killed probably due to disk or memory pressure.

src/cleaner/scripts/clean_docker.py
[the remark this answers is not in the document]
mzmssg: np

src/rest-server/src/templates/dockerContainerScript.mustache
wangdian: After experiment, I changed the signal to SIGUSR1(10) and the container can handle the signal as expected. So if we trapped signal 10, means the container is killed due to disk pressure.

PR #2124, opened 2019-02-02
[Pod Eviction] Disable kubernetes's hard enicvtion

ydye:
1: Disable pod eviction behavior
2: Set the image gc threshold to 100%. And when image gc is triggered, kubelet will continue do image gc until the file system's space is below 95%.


fanyangCS:
have you tested it?

ydye:
> have you tested it?

Haven't test it yet.

fanyangCS:
please have it test

PR #2143, opened 2019-02-13
Diawang/cleaner test

wangdian:
Add cleaner test job. Fill disk to <threshold> with in <time> seconds. Threshold and time are variables in job command line.


src/cleaner/test/job/cleaner-test-job.json
```diff
@@ -1,6 +1,6 @@
 {
   "jobName": "cleaner-test-job",
-  "image": "ubuntu",
+  "image": "openpai/testcleaner",
```
mzmssg: add a tag might be better

wangdian: Then I will use openpai/testcleaner:stable

PR #2145, opened 2019-02-14
Tools for migrate config and validate config

hao1939:
This PR includes tools for:
1. Migrating config from old style(v0.8) to the latest (v0.10)
(Usage: ./deployment/configMigration.py from_dir/ to_dir/)
2. Validating config
(Usage: ./paictl.py check -p /config_dir)


deployment/configMigration.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env python
+
+from clusterObjectModel import forward_compatibility
+import shutil
+import yaml
+import os
+import sys
+print("This script is used for migrating config from v0.8 to v0.10!")
```
ydye: logger

deployment/configMigration.py
```diff
@@ -0,0 +1,61 @@
...
+# got dashboard ip
+dashboard_ip = ""
+for node in old_cluster_config["machine-list"]:
+    if ("dashboard" in node and (node["dashboard"] == "true")):
+        dashboard_ip = node["hostip"]
+        break
+# assign api-servers-url and dashboard-url
+layout["kubernetes"] = {
+    "api-servers-url": "http://"+kubernetes_config["kubernetes"]["load-balance-ip"]+":443",
+    "dashboard-url": "http://"+dashboard_ip+":80"}
```
ydye: 443?

deployment/configMigration.py
```diff
@@ -0,0 +1,61 @@
...
+
+# write out the file
+layout_file = os.path.join(output_dir, "layout.yaml")
+if(os.path.exists(layout_file)):
+    print"!!! Error, file already existing:", layout_file
+    exit(-1)
```
ydye: sys.exit(-1).
[https://stackoverflow.com/questions/6501121/difference-between-exit-and-sys-exit-in-python](https://stackoverflow.com/questions/6501121/difference-between-exit-and-sys-exit-in-python)

deployment/configMigration.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env python
```
ydye: Don't put the tool in our module. Please put it in another place.

deployment/configMigration.py
hao1939: Hi ydye,
It's a one off script, so I try to make it simple.

deployment/configMigration.py
hao1939: Cool, I will fix.

deployment/configMigration.py
hao1939: Great.

deployment/configMigration.py
hao1939: Agree. It's a standalone script, we can move it else place.
Where's the best place in you opinion?

deployment/configMigration.py
hao1939: Hi @ydye ,

One more question, one more disadvantage is that:

> If I use exit() in a code and run it in the shell, it shows a message asking whether I want to kill the program or not. It's really disturbing.

But I haven't see it. Maybe I have been using the sys.exit()? Because I had imported sys.
Any suggestion?

deployment/configMigration.py
hao1939: Hi @ydye,

Do we support deploy a https api-server now?

ydye: No, it isn't supported yet.

deployment/configMigration.py
ydye: I have no idea about this, maybe only happened in a desktop?

deployment/configMigration.py
ydye: ```src/tool``` ?

deployment/configMigration.py
ydye: ok

deployment/configMigration.py
hao1939: maybe in deployment/tool?
It's more about deployment, other than any component.

Gerhut:
Hi @hao1939 

After you move the `./deployment/configMigration.py` to `./deployment/tools/configMigration.py`, the `from clusterObjectModel import forward_compatibility` cannot find the module.

hao1939:
Hi @Gerhut ,
Fix it in #2163.

PR #2149, opened 2019-02-15
Unify PAI runtime environment variable names

abuccts:
Unify PAI runtime environment variable names in docs and envfile.
Here's the exposed environment variable list:
- PAI_JOB_NAME
- PAI_JOB_VC_NAME
- PAI_USER_NAME
- PAI_DEFAULT_FS_URI
- PAI_TASK_ROLE_COUNT
- PAI_TASK_ROLE_LIST
- PAI_TASK_ROLE_TASK_COUNT\_`$taskrole`
- PAI_HOST_IP\_`$taskRole`\_`$taskIndex`
- PAI_PORT_LIST\_`$taskRole`\_`$taskIndex`\_`$portType`
- PAI_RESOURCE\_`$taskRole` 
- PAI_MIN_FAILED_TASK_COUNT\_`$taskRole`
- PAI_MIN_SUCCEEDED_TASK_COUNT\_`$taskRole`
- PAI_CURRENT_TASK_ROLE_NAME
- PAI_CURRENT_TASK_ROLE_CURRENT_TASK_INDEX

### DIFF
--- a/docs/job_tutorial.md
+++ b/docs/job_tutorial.md
@@ -153,33 +153,22 @@ Those environment variables can also be used in the job config file.
 
 Below we show a complete list of environment variables accessible in a Docker container:
 
-| Environment Variable Name          | Description                              |
-| :--------------------------------- | :--------------------------------------- |
-| PAI_WORK_DIR                       | Working directory in Docker container    |
-| PAI_DEFAULT_FS_URI                 | Default file system uri in PAI           |
-| PAI_JOB_NAME                       | `jobName` in config file                 |
-| PAI_JOB_VC_NAME                    | The virtual cluster in which the job is running     |
-| PAI_USER_NAME                      | User who submit the job                  |
-| PAI_DATA_DIR                       | `dataDir` in config file                 |
-| PAI_OUTPUT_DIR                     | `outputDir`in config file or the generated path if `outputDir` is not specified |
-| PAI_CODE_DIR                       | `codeDir` in config file. **Note**: we plan to obsolete PAI_CODE_DIR soon. Users are encouraged to keep their code in Github or any privately managed code repo if the code is not public. When running the Docker image,user can insert some script in the taskRole.command to “git clone …” the code down to the Docker instance.                 |
-| PAI_CURRENT_TASK_ROLE_NAME         | `taskRole.name` of current task role     |
-| PAI_CURRENT_TASK_ROLE_TASK_COUNT   | `taskRole.taskNumber` of current task role |
-| PAI_CURRENT_TASK_ROLE_CPU_COUNT    | `taskRole.cpuNumber` of current task role  |
-| PAI_CURRENT_TASK_ROLE_MEM_MB       | `taskRole.memoryMB` of current task role   |
-| PAI_CURRENT_TASK_ROLE_SHM_MB       | `taskRole.shmMB` of current task role      |
-| PAI_CURRENT_TASK_ROLE_GPU_COUNT    | `taskRole.gpuNumber` of current task role  |
-| PAI_CURRENT_TASK_ROLE_MIN_FAILED_TASK_COUNT    | `taskRole.minFailedTaskCount` of current task role    |
-| PAI_CURRENT_TASK_ROLE_MIN_SUCCEEDED_TASK_COUNT | `taskRole.minSucceededTaskCount` of current task role |
-| PAI_CURRENT_TASK_ROLE_CURRENT_TASK_INDEX | Index of current task in current task role, starting from 0 |
-| PAI_JOB_TASK_COUNT                 | Total tasks' number in config file       |
-| PAI_JOB_TASK_ROLE_COUNT            | Total task roles' number in config file  |
-| PAI_JOB_TASK_ROLE_LIST             | Comma separated all task role names in config file |
-| PAI_CONTAINER_HOST_IP              | Allocated ip for current docker container |
-| PAI_CONTAINER_HOST_PORT_LIST       | Allocated port list for current docker container, in `portLabel0:port0,port1,port2;portLabel1:port3,port4` format |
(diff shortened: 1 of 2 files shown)


hao1939:
Hi @abuccts,

When I was writing script, I wonder whether we could get all those info as a 'json' or 'yaml', so my python script could easily handle it.
For example, we put a `job.json` or `job.yaml` file in the container which including all those info.

What'd your suggestion?

docs/job_tutorial.md
```diff
@@ -153,33 +153,22 @@ Those environment variables can also be used in the job config file.
...
+|                   | PAI_HOST_IP\_`$taskRole`\_`$taskIndex`    | The host IP for `taskIndex` task in `taskRole`              |
+|                   | PAI_PORT_LIST\_`$taskRole`\_`$taskIndex`\_`$portType` | The `$portType` port list for `taskIndex` task in `taskRole`     |
+|                   | PAI_RESOURCE\_`$taskRole`                 | Resource requirement for the task role in "gpuNumber,cpuNumber,memMB,shmMB" format |
+|                   | PAI_MIN_FAILED_TASK_COUNT\_`$taskRole`    | `taskRole.minFailedTaskCount` of the task role              |
+|                   | PAI_MIN_SUCCEEDED_TASK_COUNT\_`$taskRole` | `taskRole.minSucceededTaskCount` of the task role           |
+| Current task role | PAI_CURRENT_TASK_ROLE_NAME                | `taskRole.name` of current task role                        |
+| Current task      | PAI_CURRENT_TASK_ROLE_CURRENT_TASK_INDEX  | Index of current task in current task role, starting from 0 |
 
 
 ### A deep learning job example <a name="example"></a>
```
xudifsd: Should also update example below. And examples in example dir.

@fanyangCS Maybe all examples should not using dataDir/outputDir/codeDir and environment any more, right?

fanyangCS: that can wait in a separate PR. that is a lot of work.

abuccts:
@hao1939 
In python, you can use `os.environ["PAI_ENV"]` to get the environment value. Just treat the `os.environ` as a dictionary loaded by json of environment values.

docs/job_tutorial.md
```diff
@@ -153,33 +153,22 @@ Those environment variables can also be used in the job config file.
...
-| PAI_CONTAINER_HOST\_`$type`\_PORT_LIST | Allocated port list for `portList.label == $type`, comma separated `port` string |
-| PAI_TASK_ROLE\_`$name`\_HOST_LIST  | Host list for `PAI_TASK_ROLE_NAME == $name`, comma separated `ip:port` string, sorted by current task index in task role. Each task role has a host list environment variable with the corresponding task role name |
-| PAI\_`$taskRole`\_`$taskIndex`\_`$type`_PORT | The port for `portList.label == $type, task role name == $taskRole and task index of current task role == $taskIndex`. Each port has corresponding environment variable and is exposed to all tasks. |
+| Category          | Environment Variable Name                 | Description                                                 |
+| :---------------- | :---------------------------------------- | :---------------------------------------------------------- |
+| Job level         | PAI_JOB_NAME                              | `jobName` in config file                                    |
+|                   | PAI_JOB_VC_NAME                           | The virtual cluster in which the job is running             |
```
mzmssg: Do we need to unify k8s_launcher and yarn_launcher runtime. 
i.e. 
A common base env sets such as JOB_NAME, USER_NAME. 
Then platform-specific env sets such as YARN_VC_NAME in yarn_launcher, or other necessary envs in k8s_launcher.

fanyangCS: we should remove PAI_JOB_VC_NAME

abuccts: Updated.

hao1939:
Hi @abuccts ,

Here's the case: user want to collect all the ssh port of all the taskroles. Now user will have to concat the keys using: taskroles, taskindexs, and porttypes. It's at least 3 iterations.



The point is that: `concat the ENV key is annoying`.
A job define in json/yaml would be more API friendly. User could call restful API, but it will need some additional info, for example: auth.
So for convenience, we can put the job define as txt file in job container (as json or yaml):

```
{
JobName: "job-1",
User: "username",
...
taskroles:[
  master: {portList: {ssh: 22}, elseAttributes: "ToBe Filled"},
  slave: {portList: {ssh: 22}, elseAttributes: "ToBe Filled"}
],
jobEnvs{
...
}
}
```

abuccts:
@hao1939 I think there's not much difference if getting values in scripts, except that string will be shorter. Here's your case:

```py
# from env
for taskrole in os.environ["PAI_TASK_ROLE_LIST"]:
    for i in range(os.environ["PAI_TASK_ROLE_TASK_COUNT_{}".format(taskrole)]):
        sshports = os.environ["PAI_PORT_LIST_{}_{}_ssh".format(taskrole, i)]
```

```py
# from json
pai_variables = json.load(filename)
for taskrole in pai_variables["taskroles"]:
    for task in pai_variables["taskroles"][taskrole]:
        sshports = task["portlist"]["ssh"]
```

However, if user wants to use those variables in one line command as entrypoint, then the first method will be more convenient.

xudifsd:
@hao1939 I think you're talking about the runtime definition, this will affect both pai & pai-lite. Maybe you're proposing drop job config file in some location? We can discuss this in another thread, and leave this PR as unifying environment variables.

hao1939:
Hi @abuccts,

Your example works, but still not cover the pain: for example, user may always want the `ip` as the same time of `port`.
ENVs are flat key-values pairs, and object is hierarchically which is better for APIs. I'm not against one another, want both.

Hi @xudifsd ,

Agree with you, it's helpful is user could get runtime context. We can discuss in other thread.
By the way, I seem to lost track of the runtime discussion, please loop me in.

PR #2154, opened 2019-02-18
REST server: classify two states to WAITING

Gerhut:


Gerhut:
Coveralls may meet an [issue](https://github.com/lemurheavy/coveralls-public/issues/1264)

Issue #2126, opened 2019-02-02
etcd cannot auto recover in case of data corruption


fanyangCS:
**Organization Name**: OpenPAI

**Short summary about the issue/question**: one etcd instance fails and is trapped in infinite crash recovery loop.

**Brief what process you are following**: 

**How to reproduce it**: 
corruption of etcd data

**OpenPAI Environment**: 
- OpenPAI version: 0.9.0
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Hardware (e.g. core number, memory size, storage size, GPU type etc.):
- Others:

**Anything else we need to know**:

fanyangCS:
here is the log of etcd.

2019-02-02 07:32:32.972383 I | etcdmain: etcd Version: 3.2.17 
2019-02-02 07:32:32.972434 I | etcdmain: Git SHA: 28c47bb2f 
2019-02-02 07:32:32.972438 I | etcdmain: Go Version: go1.8.7 
2019-02-02 07:32:32.972441 I | etcdmain: Go OS/Arch: linux/amd64 
2019-02-02 07:32:32.972444 I | etcdmain: setting maximum number of CPUs to 16, total number of available CPUs is 16 
2019-02-02 07:32:32.972477 N | etcdmain: the server is already initialized as member before, starting as etcd member... 
2019-02-02 07:32:32.972523 I | embed: listening for peers on http://0.0.0.0:2380 
2019-02-02 07:32:32.972551 I | embed: listening for client requests on 0.0.0.0:4001 
2019-02-02 07:32:32.973554 I | etcdserver: name = etcdid3 
2019-02-02 07:32:32.973565 I | etcdserver: data dir = /var/etcd/data 
2019-02-02 07:32:32.973584 I | etcdserver: member dir = /var/etcd/data/member 
2019-02-02 07:32:32.973588 I | etcdserver: heartbeat = 100ms 
2019-02-02 07:32:32.973590 I | etcdserver: election = 1000ms 
2019-02-02 07:32:32.973593 I | etcdserver: snapshot count = 100000 
2019-02-02 07:32:32.973603 I | etcdserver: advertise client URLs = http://10.1.0.6:4001 
2019-02-02 07:32:32.973784 I | etcdserver: restarting member 981d16e74d5b93b0 in cluster 2a4e4334d3f30cc3 at commit index 0 
[...]

PR #2159, opened 2019-02-18
add script to upgrade etcd aiding monitoring etcd

xudifsd:
Use readiness probe to monitor etcd service.

As long as pod with labels: app=name, monitored=pai(previous watchdog only requires app=name labels), the watchdog will monitor its readiness status, and generate corresponding metrics, and will sent out email if service is not ready for a long time.

This PR added all service with these two labels and add readiness probe to etcd service. For etcd service deployed by previous pai code, the PR has a script for user to upgrade. If users want to monitor old version etcd, they can refer the doc and do some operation.

I assume this PR can go into v0.11.0, and can modify upgrade doc according if this PR can not make it.


hao1939:
Hi @xudifsd ,

Why we need two labels? `labels: app=name, monitored=pai`.

fanyangCS:
what if the k8s is not deployed by ourselved. does the design still work?

docs/monitor-etcd.md
```diff
@@ -0,0 +1,29 @@
+# Background
+
+In Pai version before 0.10.0, pai did not have a monitor solution for etcd deployed
+by OpenPai, so admin will not get notified when something wrong with etcd and can not
+running correctly.
+
+To monitor etcd, we have watchdog service, this service monitor pods with special
+labels and check its running status via readiness probe. If you have deployed
```
fanyangCS: what is the value of the special label?

docs/monitor-etcd.md
```diff
@@ -0,0 +1,29 @@
...
+before 0.10.0, you should do following upgrade to configure etcd correctly so it can
+be monitored.
+
+# Upgrade steps
+
+Please note following steps involves restart etcd pods, the etcd process will stop for around 5 minutes, and **kubernetes can not work during the time**, please plan upgrade in advance to minimize the possible impact.
+
+1. Please use a [dev-box](pai-management/doc/how-to-setup-dev-box.md) to do following operations, as dev-box can provide a clean environment.
+2. Checkout the 0.10.0 version branch of pai: `cd $PAI_ROOT && git checkout v0.10.0`.
+3. Apply upgrade script: `cd $PAI_ROOT/src/etcd-upgrade/build && kubectl apply -f ds.yaml`.
```
fanyangCS: >ds.yaml [](start = 82, length = 7)

use a daemonSet to upgrade etcd is way too heavy. by the way, can we just label the pod? k8s does support attaching a label to a node on the fly, right?

docs/monitor-etcd.md
xudifsd: app=name, monitored=pai

docs/monitor-etcd.md
xudifsd: Yes, we are updating the etcd.yaml, by using DaemonSet to deploy update script in every node and update etcd.yaml.

xudifsd:
> Hi @xudifsd ,
> 
> Why we need two labels? `labels: app=name, monitored=pai`.

Previous, I just use app=name, but this is too simple, and if user deploy other pods using app labels, their pods will be monitored by pai, which not what we expected. To use monitored=pai, this marks pod we want to monitored and is obvious for users.

xudifsd:
related to #2126

hao1939:
> > Hi @xudifsd ,
> > Why we need two labels? `labels: app=name, monitored=pai`.
> 
> Previous, I just use app=name, but this is too simple, and if user deploy other pods using app labels, their pods will be monitored by pai, which not what we expected. To use monitored=pai, this marks pod we want to monitored and is obvious for users.

Hi Di,

We should alter if user deploy other pods on the k8s, shouldn't we?
Running un-managed pod on the k8s will affect our stability.
At default, we should monitor everything, so we'd better add an exclusive rules instead of inclusive rules.

xudifsd:
@hao1939 I am fine with it, just leave it as app=name. I thought since editing etc label is not easy, we should do it right, and do it explicitly with monitored=pai.

What's your opinion? @fanyangCS.

hao1939:
Hi @xudifsd ,

As you said, edits labels on pod is not easy, could we put the rules inside the alter manager?
So we can avoid to touch the pod be monitored.

xudifsd:
> Hi @xudifsd ,
> 
> As you said, edits labels on pod is not easy, could we put the rules inside the alter manager?
> So we can avoid to touch the pod be monitored.

I'm not sure what you mean, the pod needs to attached special labels to be found by watchdog, and the discovery rule is defined in watchdog, not in pod. And this PR is propose using rule `app=name && monitored=pai` to select pods.

hao1939:
> > Hi @xudifsd ,
> > As you said, edits labels on pod is not easy, could we put the rules inside the alter manager?
> > So we can avoid to touch the pod be monitored.
> 
> I'm not sure what you mean, the pod needs to attached special labels to be found by watchdog, and the discovery rule is defined in watchdog, not in pod. And this PR is propose using rule `app=name && monitored=pai` to select pods.
I mean that we only need the minimum tags, for example `app=name`. Once we got the app name, we can easily decision whether to include/exclude it from monitoring.

xudifsd:
@hao1939  removed monitored=pai rule

fanyangCS:
@hao1939, make sure you can handle this new alert in AKS case.

xudifsd:
AKS will not need to monitor etcd

PR #2171, opened 2019-02-20
generate gpu memory util as used/total

### LINKED ISSUE #2121: show the true GPU memory utilization
checked the manual of nvidia-smi, seems memory utilization is
   Percent of time over the past sample period during which global (device) memory was being read or written. The sample period may be between 1 second and 1/6 second depending on the product.

Not how much memory application used.

xudifsd:
fixed #2121

### DIFF
--- a/src/grafana/src/dashboards/gpu.js
+++ b/src/grafana/src/dashboards/gpu.js
@@ -125,7 +125,7 @@ for (var i = 0; i < gpuNum; i++) {
           },
           "yaxes": [
             {
-              "format": "short",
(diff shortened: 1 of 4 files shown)


YanjieGao:
how about attach an image to view the results directly on PR

xudifsd:
this PR has no graphic change, only changed the meaning of `nvidiasmi_utilization_memory` from `Percent of time over the past sample period during which global (device) memory was being read or written` to `gpu memory used / gp memory total`

Issue #2169, opened 2019-02-20
Virtual Cluster with empty space can be created but can not be used


scarlett2018:
**Organization Name**: PAI PROD-1

**How to reproduce it**: 
- Created a VC named "A B" in web UI, the VC has been created successfully
- Add a user to this VC in User "Edit Info" page, it failed.
![image](https://user-images.githubusercontent.com/39592018/53071674-d2e41f00-351d-11e9-9635-cff0043d19fa.png)

**OpenPAI Environment**:
- OpenPAI version: 0.9.0

mzmssg:
Fixed in #2180

PR #2180, opened 2019-02-20
Restrict vcname character in VC related API

mzmssg:
#2169 
YARN actually supports more kinds of characters such as space, but PAI doesn't.
This pr is to keep consistent with other PAI API.


src/rest-server/src/controllers/vc.js
```diff
@@ -23,7 +23,9 @@ const createError = require('../util/error');
  * Validation, not allow operation to "default" vc.
  */
 const validate = (req, res, next, vcName) => {
-  if (vcName === 'default' && req.method !== 'GET') {
+  if (! /^[A-Za-z0-9_,]+$/.test(vcName)) {
```
abuccts: Can `vcName` contains comma `,`?
```suggestion
  if (! /^[A-Za-z0-9_]+$/.test(vcName)) {
```

mzmssg: Fixed

PR #2194, opened 2019-02-22
fix missing gpu metrics

xudifsd:
Current prometheus will failed to get task gpu metrics as shown below, this is due to GpuCollector(producer) and ContainerCollector(consumer) is in different thread, and consumer running in shorter interval, so consumer may failed to get data needed.

This PR make GpuCollector interval shorter to fix this problem. This should be a temporary fix, should change structure for long term fix.

image


fanyangCS:
can this go to 0.10.y?

Gerhut:
@fanyangCS OK

@xudifsd Please cherry-pick as you will, I will do revalidate next Monday

PR #2197, opened 2019-02-22
add a dashboard in grafana to list all tasks in node

xudifsd:
Add a page to show all tasks util in certain node

image

add a link to tasks in node page in node util page like

image


fanyangCS:
Looks good. In the first page, can you add a link directed back to the second (root) page?

xudifsd:
@fanyangCS Ok

xudifsd:
@fanyangCS added, so the link looks like

image

PR #2209, opened 2019-02-26
alert on unhealthy gpu

### LINKED ISSUE #2192: Need alarms for unhealthy GPU cases
**Organization Name**:  Microsoft

There are some GPU exceptions PAI can't perfectly handle. For these case, as the first phase, we will alert and introunce manual operations.
1. job-exporter collect gpu status
2. alert admin
3. admin manually decommission nodes.

Currently, we summarized some unhealthy cases, including:
* Nvidia-smi hangs
   1.	In kernal mode
        Detection: Filter processes list of command 'nvidia-smi' and status 'D', to reduce false positive alarms, we could add a timeout: 
* Nvidia-smi works
   1. GPU memory leak
   Detection: Call nvidia-smi to get memory-utilized GPUs and processes running GPUs,
Subtraction results is the memory leak GPU.
  2. GPU used by external processes
  Detection: Call nvidia-smi, get running processes, then get their cgroup in /proc/$pid/cgroup, check whether it’s under job container
  3. GPU used by zombie container
  Detection: Based on 2, further check whether corresponding yarn container survives.
  4. GPU ECC error
  Detection: Call nvidia-smi to get ecc error code.

**OpenPAI Environment**:
- OpenPAI version: master

Alarms and manual operations are the first step. 
As the next step, PAI itself should tolerate such cases.

xudifsd:
fixed #2192 , working in progress

- [x] zombie process
- [x] gpu memory leak
- [x] gpu used by external process
- [x] gpu ecc error

### DIFF
--- a/src/job-exporter/src/collector.py
+++ b/src/job-exporter/src/collector.py
@@ -272,7 +272,7 @@ class GpuCollector(Collector):
     cmd_histogram = Histogram("cmd_nvidia_smi_latency_seconds",
             "Command call latency for nvidia-smi (seconds)")
 
-    cmd_timeout = 5 # 99th latency is 0.97s
+    cmd_timeout = 60 # 99th latency is 0.97s
 
     def __init__(self, name, sleep_time, atomic_ref, iteration_counter, gpu_info_ref):
         Collector.__init__(self, name, sleep_time, atomic_ref, iteration_counter)
--- a/src/job-exporter/src/collector.py
+++ b/src/job-exporter/src/collector.py
@@ -57,6 +57,11 @@ def gen_gpu_mem_util_gauge():
             "gpu memory utilization of card",
             labels=["minor_number"])
 
+def gen_gpu_ecc_counter():
+    return GaugeMetricFamily("nvidiasmi_ecc_error_count",
+            "count of nvidia ecc error",
+            labels=["minor_number", "type"])
+
 
 class ResourceGauges(object):
     def __init__(self):
@@ -284,12 +289,15 @@ class GpuCollector(Collector):
         if gpu_info is not None:
             core_utils = gen_gpu_util_gauge()
             mem_utils = gen_gpu_mem_util_gauge()
+            ecc_errors = gen_gpu_ecc_counter()
 
             for minor, info in gpu_info.items():
-                core_utils.add_metric([minor], info["gpu_util"])
-                mem_utils.add_metric([minor], info["gpu_mem_util"])
+                core_utils.add_metric([minor], info.gpu_util)
+                mem_utils.add_metric([minor], info.gpu_mem_util)
+                ecc_errors.add_metric([minor, "single"], info.ecc_errors.single)
+                ecc_errors.add_metric([minor, "double"], info.ecc_errors.double)
 
-            return [core_utils, mem_utils]
+            return [core_utils, mem_utils, ecc_errors]
 
         return None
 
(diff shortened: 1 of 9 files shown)


xudifsd:
@fanyangCS @mzmssg ready for review

src/job-exporter/src/collector.py
```diff
@@ -267,13 +294,76 @@ class GpuCollector(Collector):
...
+        if not os.path.isfile(path):
+            return False, ""
+
+        with open(path) as f:
+            content = f.read()
+
+        for line in content.split("\n"):
+            line = line.strip()
+            if "pids" in line and "/docker/" in line:
+                parts = line.split("/docker/")
```
mzmssg: Could we also consider kubernetes pod? Then if we switch to static pod, we won't need to change here.

xudifsd: How to consider k8s pod?

mzmssg: @xudifsd 
Not find a clean definition of k8s pod cgroup, we could add a TODO here to remind us.

src/job-exporter/src/collector.py
```diff
@@ -267,13 +294,76 @@ class GpuCollector(Collector):
...
+        zombie_container = gen_gpu_used_by_zombie_container_counter()
+
+        pids_use_gpu = {} # key is gpu minor, value is an array of pid
+
+        for minor, info in gpu_info.items():
+            core_utils.add_metric([minor], info.gpu_util)
+            mem_utils.add_metric([minor], info.gpu_mem_util)
+            ecc_errors.add_metric([minor, "single"], info.ecc_errors.single)
+            ecc_errors.add_metric([minor, "double"], info.ecc_errors.double)
+            if info.gpu_mem_util > 20 * 1024 * 1024 and len(info.pids) == 0:
```
mzmssg: Could 20M be a configurion?

xudifsd: This 20M was mentioned by @fanyangCS , I'm not sure if we want it to be configurable or hardcode should be enough.

mzmssg: @xudifsd 
In our yarn patch, this is a configuration, so an item with default value seems reasonable.

src/prometheus/deploy/alerting/gpu.rules
```diff
@@ -0,0 +1,58 @@
...
+    - name: gpu_related
+      rules:
+      - alert: NvidiaSmiLatencyTooLarge
+        expr: histogram_quantile(0.95, sum(rate(cmd_nvidia_smi_latency_seconds_bucket[5m])) by (le, instance)) > 40
+        for: 5m
+        annotations:
+          summary: "95th nvidia-smi call latency is larger than 40s in {{$labels.instance}}, should check the gpu status manually"
+
+      - alert: NvidiaSmiEccError
+        expr: nvidiasmi_ecc_error_count{type="double"} > 0
```
mzmssg: > However, occassionaly there will be multiple such single-bit errors or a double-bit error on the same memory page. When this happens, the job will exit with segfault and burn a retry.

Seems we should consider multiple single-bit errors as well

src/job-exporter/src/collector.py
xudifsd: ok

src/prometheus/deploy/alerting/gpu.rules
xudifsd: Then, how to alert? Also alert on single error?

PR #2244, opened 2019-03-01
YARN and HDFS log persistence 

mzmssg:
Currently PAI only output service log to console.
This PR is to persist yarn and hdfs log to files, to avoid log lost accross service restart.

After this pr, admin could find yarn and hdfs log under `{{ cluster_cfg["cluster"]["common"][ "data-path" ] }}/pai-service-log`


src/hadoop-resource-manager/deploy/hadoop-resource-manager.yaml.template
```diff
@@ -115,6 +122,9 @@ spec:
       - name: hadoop-tmp-storage
         hostPath:
           path: {{ cluster_cfg["cluster"]["common"][ "data-path" ] }}/hadooptmp/resourcemanager
+      - name: log-dir
+        hostPath:
+          path: /var/log/resource-manager
```
ydye: ```{{ cluster_cfg["cluster"]["common"][ "data-path" ] }}/pai-service-log/resourcemanager```?

mzmssg: Done

Issue #2238, opened 2019-02-28
Simplified Job Templates for NFS configured OpenPAI


scarlett2018:
**What would you like to be added**:
A simplified job submission page for simple job types.

**Why is this needed**:
Current Job submission page is too complex to understand for most of the users.

**Goal**
- 2 TF Job Types we listed for jobs running on OpenPAI Instance that has NFS configured.

**No Goal**
- Jobs that does not have NFS configured
- Other Job Types

**UX Mock Up** **(need to update with latest discussed changes)**

_**Sample Job Type - Tensorflow-SingleNode**_
![image](https://user-images.githubusercontent.com/39592018/54176714-e2c2a380-44ca-11e9-9543-3ee0b887a503.png)

_**Sample Job Type - Tensorflow-Distributed**_
![image](https://user-images.githubusercontent.com/39592018/54176703-d63e4b00-44ca-11e9-925e-512aadc3b7dd.png)

-----------------
**Exit Criteria**
- [x]  3 Job Type Examples **that can run end 2 end** @abuccts Anbang

-----------------
**Dependency**
Plugin Changes Need: issue to be created by @Gerhut 
- [x]  server side change to make endpoint per plugin: plugin configuration changes to pass plugin dictionary 
- [x] plugin versioning
- [x] submit v1 json to initialize form

Job Type Need:
- [x] * Job Type Examples **that can run end 2 end** @abuccts Anbang

fanyangCS:
Looks great! Please make the UI on mount path configurable, I.e., we can config to make it disappear. Some deployment may not have a storage system that is suitable to be mounted on (e.g., hdfs)

squirrelsc:
For JobType, is there any details about it? And there could be a clone button to clone from self jobs, or even search more jobs.

squirrelsc:
For extensibility, resource part should be able to customize, and support different resource type for clusters with different server model.

squirrelsc:
For directory paths in docker, it should be able configurable. So that user don't need to update their code to align with OpenPAI.

scarlett2018:
> For JobType, is there any details about it? And there could be a clone button to clone from self jobs, or even search more jobs.

created #2241 for this input, will not cover this in current issue.

scarlett2018:
> For extensibility, resource part should be able to customize, and support different resource type for clusters with different server model.

yes, this will be part of the current end 2 end story. Admin is the one to customize the Resource options.

scarlett2018:
[...]

PR #2253, opened 2019-03-01
Web portal submit job: support init json from sessionStorage.

Gerhut:
Dependency of #2238

Since the maximum length in address bar of browser will be (usually) 2000 chars, transfer the job JSON through url might be risky, so we choose sessionStorage

## Usage

1. Save the initial job JSON string by `window.sessionStorage.setItem('job-init', job)`
2. Navigate to `/submit.html?op=init`
3. The saved job JSON string will be automatically loaded from sessionStorage, than cleared.

### DIFF
--- a/src/webportal/src/app/job/job-submit/job-submit.component.js
+++ b/src/webportal/src/app/job/job-submit/job-submit.component.js
@@ -160,31 +160,41 @@ $(document).ready(() => {
       resize();
     };
     const query = url.parse(window.location.href, true).query;
+    const op = query.op;
     const type = query.type;
     const username = query.user;
     const jobname = query.jobname;
-    if (type != null && username != null && jobname != null) {
-      const url = username==''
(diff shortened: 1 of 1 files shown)


src/webportal/src/app/job/job-submit/job-submit.component.js
```diff
@@ -160,31 +160,41 @@ $(document).ready(() => {
...
-          const res = JSON.parse(xhr.responseText);
-          if (res.message === 'ConfigFileNotFound') {
-            alert('This job\'s config file has not been stored.');
-          } else {
-            alert('Error: ' + res.message);
-          }
-        },
-      });
+    if (op === 'resubmit') {
+      if (type != null && username != null && jobname != null) {
```
abuccts: use `!==`?

Gerhut: In fact `== null` & `!= null` in JavaScript have a clearily meaning, which is `=== undefined || === null` or `!== undefined && !== null`

Issue #2259, opened 2019-03-04
[Job Debugging] Tutorial for user to enable job debugging 


ydye:
**What would you like to be added**:

```
Update doc to introduce user to enable debugging reservation.
```

**Why is this needed**:

```
User may don't know how to enable it.
```

**Without this feature, how does the current module work**：

**Components that may involve changes**:

PR #2257, opened 2019-03-04
Make debugging reservation time configurable.

ydye:

Issue #2252, opened 2019-03-01
[Job Debugging] Make reservation time configurable to admin in cluster-configuration


ydye:
**What would you like to be added**:

```
Make the debugging reservation time configurable to cluster admin. And cluster admin could configure it 
in cluster configuration.
```

**Why is this needed**:

```
More flexible.
```

**Without this feature, how does the current module work**：

```
Unable to configure it.
```

**Components that may involve changes**:

```
Restserver
```

Issue #2218, opened 2019-02-26
[Job Debugging] Provide detail information when the job container exits.


ydye:
**What would you like to be added**:
```
Sometimes, users' job may be kill by OpenPAI due to several reasons. It's different from
the users' error. We can class these error as system error. And the container can't be reserved 
if it's failed due to system error. So more detailed log should be provided. So that users could know the reason why their job container exits and can't be reserverd for job debugging. 

An example of system error which will make your job failed:
    - Disk pressure

```
**Why is this needed**:

```
If the job container can't be reserved if it failed due to system error. Users may feel confused. 
```
**Without this feature, how does the current module work**：
```
User should investigate the job log.
```
**Components that may involve changes**:
```
TBD
```

Binyang2014:
refer to kubernetes/kubernetes#140

Binyang2014:
This item will work on K8S version. Several tasks:
- [ ] Map the docker exit info mechanism from dockerScript to kube runtime
- [ ] Reorganize the error spec. Remove yarn error, add new error type for k8s version

Binyang2014:
Add failurePattern for runtime:
```yaml
- errorType: user/system
  patterns:
    pattenExitCode:
      exitCode: 132
      userLog: ""
      runtimeLog: ""
      # more can be added here
    pattenUserLog:
      userLog: "This is an error"
  matches: ${pattens.pattenExitCode} or ${pattens.pattenUserLog}
  reason: 'User program terminated by SIGILL'
  solution: 'Please check the log and retry again'
  containerExitCode: 132
```

PR #2272, opened 2019-03-05
[Job Debugging] Basic Implement Of Job Debugging.

ydye:
### This PR will achieve

The ability to hold the failed job which is casued by user's error. User error means there is something wrong happened in user's command and the command exit with a none-zero code. 

### This PR won't solve 

We can't hold the failed job which is caused by OpenPAI's system error. System error means the job is killed by system alerts or system issue. For example, user's job will be killed if the container write too much data into local path.

### Relative Issues:
- Design Overview: #2210
    - Reserve job container whose user command exits with none zero code. #2213
    - An option for user to enable job debugging feature for their job. #2214
    - Tutorial for user to use job debugging. #2259 
    - Configurable reservation time. #2252 

### Sub-PR detail information

- ``` Tutorial for user ``` #2264 

Provide the document for user to use job debugging feature

- ```Configurable job debugging reservation time for admin ``` #2257

Admin could configure the debugging reservation time in ```services-configuration.yaml -> rest-server -> debugging-reservation-seconds```.  The default value will be 604800 seconds.

- ``` Debug option for user and the ability to reserved a container due to user error ``` #2225

When user submit the job, set the following property in the jobEnv of jobConfig. If the job's user command fails, the container will be kept for 1 weeks. And user could debug the container after ssh to it. After debugging, user should manually stop it to recycle the system resources.

##### When submitting job from webportal:
image

##### when submitting job from json file:
```
  "jobEnvs": {
    "isDebug": true
  }
```

##### When you job failed, click into the ```Go to Tracking Page``` and looking into the ```stdout```. You could find following log.
```
[INFO] USER COMMAND START

job has finished with exit code 2
=============================================================================
======   The job container failed, so it will be reserved for 1 week   ======
======          After debugging, please stop the job manually.         ======
=============================================================================
```

### Future Plan

- Provide more detailed failure info, if the job is killed due to system error. #2218 
- Policy to collect the job status such as job debugging on each node. #2215  
- Show the job which is reserved in job debugging status for user. #2216 
- The ability to email user when the job is reserved for job debugging. #2217


src/rest-server/src/config/paiConfig.js
```diff
@@ -26,18 +26,19 @@ try {
     paiMachineList = yaml.safeLoad(fs.readFileSync('/pai-cluster-config/layout.yaml', 'utf8'))['machine-list'];
 } catch (err) {
     paiMachineList = [];
-    logger.info('Unable to load machine list from cluster-configuration.');
-    logger.info(err.stack);
+    logger.error('Unable to load machine list from cluster-configuration.');
```
Gerhut: Since it's not a fatal error in this process, I think `warn` is better, also add details of what it'll cause.

ydye: Done

PR #2273, opened 2019-03-05
use twisted in yarn-exporter

xudifsd:
similar to #2093 , use twisted in yarn-exporter for better performance and disable gc collector to avoid deadlock


src/yarn-exporter/src/yarn_exporter.py
```diff
@@ -16,128 +16,116 @@
...
-        # yield nodes_lost
 
-        # nodes_unhealthy = YarnMetric('nodes_unhealthy', YarnMetric.GAUGE,
-        #                              'The number of unhealthy nodes', ['cluster']).create_metric()
-        # nodes_unhealthy.add_metric([self.cluster_name], metric['unhealthyNodes'])
-        # yield nodes_unhealthy
+class HealthResource(Resource):
+    def render_GET(self, request):
+        request.setHeader("Content-Type", "text/html; charset=utf-8")
+        return "<html>Ok</html>".encode("utf-8")
```
mzmssg: It's a python3 code, why need encode

xudifsd: Twisted requires application returning byte, so encode is needed.

PR #2260, opened 2019-03-04
Job submission v2

abuccts:
Back-end part for job submission v2 #2232.

Work Items
-------------
- [x] [Docs] Finalize protocol spec: merged [pai-proto](https://github.com/Microsoft/pai/tree/pai-proto) branch.
- [x] [Docs] Add new yaml files that follow new protocol: #2292.
- [x] [Rest Server] Protocol yaml parsing, validation and rendering: #2274.
- [x] [Rest Server] Convert protocol to framework launcher description file: #2291.
- [x] [Rest Server] Refactor and update rest API v2: #2329.

Exit Criteria
-------------
- [x] BERT end to end demo.
- [x] Distributed end to end example.

PR #2274, opened 2019-03-05
[Rest Server] Protocol yaml parsing, validation, and rendering

abuccts:
Part of #2260. Protocol yaml parsing, validation, and rendering:
- Use `ajv` for protocol schema and validation.
- Use `mustache` for protocol rendering.

### DIFF
--- a/src/rest-server/src/middlewares/v2/protocol.js
+++ b/src/rest-server/src/middlewares/v2/protocol.js
@@ -117,7 +117,7 @@ const protocolRender = async (req, res, next) => {
       }
     }
     let entrypoint = '';
-    const tokens = mustache.parse(commands.join(' ; '), ['<%', '%>']);
+    const tokens = mustache.parse(commands.join(' && '), ['<%', '%>']);
     const context = new mustache.Context({
       '$parameters': protocolJSON.parameters,
       '$script': protocolJSON.prerequisites['script'][protocolJSON.taskRoles[taskRole].script],
--- a/src/rest-server/src/middlewares/v2/protocol.js
+++ b/src/rest-server/src/middlewares/v2/protocol.js
@@ -19,6 +19,7 @@
 // module dependencies
 const yaml = require('js-yaml');
 const mustache = require('mustache');
+const createError = require('../../util/error');
 const protocolSchema = require('../../config/v2/protocol');
 
 const prerequisiteTypes = [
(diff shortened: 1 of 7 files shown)


src/rest-server/src/middlewares/v2/protocol.js
```diff
@@ -16,6 +16,9 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 
+// skip guard-for-in becasue protocolValidate has already filtered keys
```
Gerhut: Why not `for (key of Object.keys(obj))` ?

src/rest-server/src/middlewares/v2/protocol.js
```diff
@@ -0,0 +1,117 @@
...
+        if (taskRole in protocolJSON.deployments) {
+            if ('preCommands' in protocolJSON.deployments[taskRole]) {
+                commands = protocolJSON.deployments[taskRole].preCommands.concat(commands);
+            }
+            if ('postCommands' in protocolJSON.deployments[taskRole]) {
+                commands = commands.concat(protocolJSON.deployments[taskRole].postCommands);
+            }
+        }
+        let entrypoint = commands.join(' ; ').replace('<%', '<%=');
+        const compiled = template(entrypoint);
```
Gerhut: EJS-style template will introduce lots of vulnerabilities and could not be used for user's input.

For example: if user input `<% process.exit(1) %>`, it will kill the rest-server itself.

abuccts: Because the input string has been replaced `<%` with `<%=`, all strings inside "evaluate" delimiter won't be executed and will be treated as "interpolate" delimiter. In your case, nothing will happen.

src/rest-server/src/middlewares/v2/protocol.js
abuccts: Updated.

src/rest-server/src/middlewares/v2/protocol.js
abuccts: Change to `mustache`.

fanyangCS:
unit test pleas.

src/rest-server/src/middlewares/v2/protocol.js
```diff
@@ -0,0 +1,162 @@
...
+    if (deployment != null && taskRole in deployment.taskRoles) {
+      if ('preCommands' in deployment.taskRoles[taskRole]) {
+        commands = deployment.taskRoles[taskRole].preCommands.concat(commands);
+      }
+      if ('postCommands' in deployment.taskRoles[taskRole]) {
+        commands = commands.concat(deployment.taskRoles[taskRole].postCommands);
+      }
+    }
+    let entrypoint = '';
+    const tokens = mustache.parse(commands.join(' ; '), ['<%', '%>']);
```
mydmdm: A simple question, whether we should use **;** or **&&** to combine commands? In **&&** case, the succeeding commands will be executed only if previous commands return an exit status of zero, this may help to find early bugs

abuccts: Update to `&&`.

PR #2290, opened 2019-03-08
Diawang/extras

wangdian:
Add extras field to JobConfig.
The extras field is used to save any job information.
the extras.submitFrom field is used to save the plugin id if the job was submitted by plugin.

### DIFF
--- a/docs/job_tutorial.md
+++ b/docs/job_tutorial.md
@@ -117,7 +117,6 @@ Below please find the detailed explanation for each of the parameters in the con
 | `jobEnvs.paiAzRDMA`              | Boolean, optional          | If you cluster is azure rdma capable, you could specify the parameter to make your container azure rdma capable. How to use azure rdma? Please follow this [job example](../examples/azure-rdma-inte-mpi-benchmark-with-horovod-image) |
 | `jobEnvs.isDebug`                | Boolean, optional          | after this flag is set as ```true```, if user's command exits with a none-zero value, the failed container will be reserved for job debugging.  [More detail](./job_debugging.md)|
 | `extras`                         | Object, optional           | Extra parameters, key-value pairs, save any information that job may use |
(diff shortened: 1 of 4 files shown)


docs/job_tutorial.md
```diff
@@ -109,8 +114,11 @@ Below please find the detailed explanation for each of the parameters in the con
...
-| `jobEnvs.isDebug`                        | Boolean, optional           | after this flag is set as ```true```, if user's command exits with a none-zero value, the failed container will be reserved for job debugging.  [More detail](./job_debugging.md)|
+| `jobEnvs.paiAzRDMA`              | Boolean, optional          | If you cluster is azure rdma capable, you could specify the parameter to make your container azure rdma capable. How to use azure rdma? Please follow this [job example](../examples/azure-rdma-inte-mpi-benchmark-with-horovod-image) |
+| `jobEnvs.isDebug`                | Boolean, optional          | after this flag is set as ```true```, if user's command exits with a none-zero value, the failed container will be reserved for job debugging.  [More detail](./job_debugging.md)|
+| `extras`                         | Object, optional           | Extra parameters, key-value pairs, save any information that job may use |
+| `extras.submitFrom`              | String, optional           | The plugin id if a job was submitted by plugin |
```
Gerhut: I think this line is better to be along with "clone to plugin" feature PR

src/rest-server/src/config/job.js
```diff
@@ -105,6 +105,11 @@ const jobConfigSchema = Joi.object().keys({
         /\w+/,
         Joi.required()
     ),
+  extras: Joi.object()
+    .pattern(
+        /\w+/,
```
Gerhut: `/\w+/` means every keys should be made up of `[A-Za-z0-9_]`, it's better to add dot(.) at least for namespaced keys.

Gerhut:
add @sunqinzheng for vscode extension code review.

PR #2292, opened 2019-03-08
[Docs] add new yaml files that follow pai protocol

abuccts:
Part of #2260. Add new protocol marketplace for protocol yaml files.


protocol-marketplace/tensorflow-cifar10.yaml
```diff
@@ -0,0 +1,149 @@
...
+        --variable_update=parameter_server
+        --ps_hosts=$PAI_TASK_ROLE_ps_server_HOST_LIST
+        --worker_hosts=$PAI_TASK_ROLE_worker_HOST_LIST
+        --task_index=$PAI_CURRENT_TASK_ROLE_CURRENT_TASK_INDEX
+        --data_name=<% $data.name %>
+        --data_dir=$PAI_WORK_DIR/data_<% $data.name %>
+        --train_dir=$PAI_WORK_DIR/output_<% $output.name %>
+        --model=<% $parameters.model %>
+        --batch_size=<% $parameters.batchsize %>
+  ps_server:
```
fanyangCS: I assume by now there are few people using ps mode to train a distributed TF work? maybe we should remove this example and add MPI example (Horovod) later?

abuccts: ok

protocol-marketplace/caffe-mnist.yaml
```diff
@@ -0,0 +1,58 @@
+protocolVersion: 2
```
sterowang: Isn't "marketplace/v2" a more aligned folder path to echo with "*/v2" in PR #2260?

abuccts: Because the original marketplace serve the whole dir, so I added a new separate directory.

abuccts: renamed to `marketplace-v2`

abuccts:
For `commands` field in `taskRoles`, it's an array type, so user can write multiple command items or concatenate several commands using `&&` as one item. It depends on user himself.

For `preCommands` and `postCommands` fields in `deployments`, they are dynamic and related to runtime clusters instead of the job itself, so we separate a `deployments` field to achieve this by design.

Issue #2294, opened 2019-03-10
Why the containers always show all of the GPUs?


ICEORY:
- OpenPAI 0.9.1

After upgrading the OpenPAI from 0.8.2 to 0.9.1, I find that the containers always allocate all of GPUs of the host machine. 

e.g. ,
![image](https://user-images.githubusercontent.com/11911903/54080878-b117d400-4334-11e9-9d8a-4b83764e8d56.png)

when I run "nvidia-smi", I got

![image](https://user-images.githubusercontent.com/11911903/54080837-fb4c8580-4333-11e9-8391-6c5009b03f3c.png)

- More information: 

![image](https://user-images.githubusercontent.com/11911903/54080849-2931ca00-4334-11e9-8aff-29b447320aa2.png)

- docker version: 
![image](https://user-images.githubusercontent.com/11911903/54080884-1370d480-4335-11e9-88b8-852278783174.png)

- nvidia-container-runtime --version
runc version 1.0.0-rc6+dev
commit: 12f6a991201fdb8f82579582d5e00e28fba06d0a-dirty
spec: 1.0.1-dev

ICEORY:
@abuccts thanks, adding `-e NVIDIA_VISIBLE_DEVICES=$gpu_id` to the file `yarnContainerScript.mustache` solve the issue

xudifsd:
@ICEORY what's your config? Did you enabled the [`set-nvidia-runtime`](https://github.com/Microsoft/pai/blob/pai-0.9.y/deployment/quick-start/services-configuration.yaml.template#L82)?

ICEORY:
Yes, I have enabled the `set-nvidia-runtime`. Should I disable it?

xudifsd:
@ICEORY Yes, how did you know that config?

ICEORY:
well, when upgrading the OpenPAI, I got the issues "fork/exec /usr/bin/nvidia-container-runtime: no such file or directory: unknown" and "cannot locate software nvidia-container-runtime". Then, I added the source of `nvidia-container-runtime` and enabled the `set-nvidia-runtime`(this option is provided in `service-configuration.yaml` ). I guessed the issues above are caused by `nvidia-container-runtime` :-)

By the way,  I also need to delete the old replica set of `prometheus`, `watchdog`, `grafana` by hand, in order to deploy the new replica set.

xudifsd:
[...]

PR #2300, opened 2019-03-11
Support nvidia-container-runtime for gpu isolation

### LINKED ISSUE #1667: provide NVIDIA_VISIBLE_DEVICES environment variable to nvidia-container-runtime
Currently, we use `--device=/dev/nvidia0` to tell docker isolate gpu for jobs. But this parameter seems is ignored by `nvidia-container-runtime`. If we want to achieve same effect of gpu isolation, we should pass NVIDIA_VISIBLE_DEVICES to `docker run` like `-e NVIDIA_VISIBLE_DEVICES=0,1`.

abuccts:
Support nvidia-container-runtime for gpu isolation.
Works for w/ and w/o `--runtime=nvidia`.

Fixes #1667, #2294.

### DIFF
--- a/src/rest-server/src/templates/yarnContainerScript.mustache
+++ b/src/rest-server/src/templates/yarnContainerScript.mustache
@@ -324,6 +324,7 @@ docker run --name $docker_name \
   --label PAI_JOB_VC_NAME={{ jobData.virtualCluster }} \
   --label PAI_USER_NAME={{ jobData.userName }} \
   --label PAI_CURRENT_TASK_ROLE_NAME={{ taskData.name }} \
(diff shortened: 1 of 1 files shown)


xudifsd:
Have you deploy in your cluster and run some example jobs? With and without the default runtime set to nvidia-container-runtime?

abuccts:
1. GPU ids specified by `NVIDIA_VISIBLE_DEVICES` env var aren't mapped to `/dev/nvidia*` correspondingly, e.g. `NVIDIA_VISIBLE_DEVICES=0` may be mapped to `/dev/nvidia3`.
2.  `--runtime=nvidia` is compatible with `--device=/dev/nvidia*`. As long as privileged mode is not enabled, not all gpus will be mounted into container.

Issue #2208, opened 2019-02-26
[Close as Dup] display resource utility per vc/queue metrics


xudifsd:
**What would you like to be added**:

A graph/list to show resource util per vc/queue

**Why is this needed**:

User sometime found their job is always in waiting state, this may be due to resource not enough in VC or resource fragment in node, need a graph to help them debug the issue

**Without this feature, how does the current module work**：

User check yarn's page manually.

**Components that may involve changes**:

* yarn-exporter
* grafana

Anbang-Hu:
Is it possible to have metrics showing current resource fragmentation status (per VC)? We should manage users' expectation properly.

xudifsd:
I'm not sure how to show resource fragmentation status, showing node count with certain free card number? How about memory? Like what hadoop showing disk?

![image](https://user-images.githubusercontent.com/665253/53539945-8e283b80-3b4e-11e9-90ef-98f84f934d3e.png)

But this is the whole picture, I think no vc view about this. Am I right? @mzmssg

fanyangCS:
related. https://github.com/Microsoft/pai/issues/2013

fanyangCS:
@xudifsd , a high priority issue is to show how many jobs are in waiting state in a VC.

scarlett2018:
This is fit for the "Virtual Cluster" Page refactoring. Currently, when click on the VC instance name it goes to the single VC page. The current single VC page only has the generic VC info. We can add this info above the current table.

**VC Table**
![image](https://user-images.githubusercontent.com/39592018/53548340-37c9f580-3b6c-11e9-9afe-ec494cc21170.png)

**Single VC Page**
![image](https://user-images.githubusercontent.com/39592018/53548377-4f08e300-3b6c-11e9-89d0-aac2dcae0fe7.png)

xudifsd:
@fanyangCS yes, yarn-exporter should get that metric. We will need to display that and node count with certain free gpu card, memory, cpu graph in some page, so user can get a sense why their job is waiting.

mzmssg:
@fanyangCS 
> a high priority issue is to show how many jobs are in waiting state in a VC.
What's the definition of waiting status? AM launcher but no job container is regarded as  RUNNING or WAITING?

yqwang-ms:
Job Waiting: Job Not Complete and Not Exist Container is Running

xudifsd:
[...]

PR #2302, opened 2019-03-11
Add Troubleshooting guide for jobs

squirrelsc:


hao1939:
It's great to have the feature 'Simulate environment locally'!
Before that, I usually run a job 'sleep infinity' for debugging.

docs/user/troubleshooting_job.md
```diff
@@ -17,51 +17,197 @@
...
+
+## Diagnostic issues
+
+### Job is waiting for hours
+
+In general, jobs of OpenPAI stays in waiting status about 1 minute. But if there is not enough resource, a job may stay in waiting status longer.
+
+![waiting](imgs/web_job_list_waiting.png)
+
+In this case, the job is in queue to wait resource. When some jobs complete, this job will get resource to run. Another way is to reduce requested resource, so this job may get enough resource easier.
```
xudifsd: Can direct user to virtual cluster page to let user checks if requesting resource can not be satisfied by this VC.

xudifsd: Also #2208 will be useful for user to debug.

docs/user/troubleshooting_job.md
```diff
@@ -17,51 +17,197 @@
...
+
+- Click *Go to Job Metrics Page* as below in job details page, if some tasks are assigned IP, and ports.
+
+![job link](imgs/web_job_details_metrics.png)
+
+- A new page is opened and show metrics of this job.
+
+![job link](imgs/web_job_metrics.png)
+
+- The *memory usage*, *network traffic* and *disk bandwidth* uses absolute value. It's easy to understand.
```
xudifsd: Should notice: PAI sacrifices network metric accuracy by using host network in docker container for better performance, so, PAI can not relay on docker virtual bridge to get accurate network traffic consumption.

PAI uses another way: get all connections and traffic used by each connection, then break down into each container. This logic is implemented in [job-exporter](https://github.com/Microsoft/pai/blob/master/src/job-exporter/src/network.py). So, if your connection is short-lived or not happened in container namespace(for example, NFS mount), traffic in this connection will not captured by job-exporter and hence will show 0 in panel.

docs/user/troubleshooting_job.md
squirrelsc: VC is an advanced topic for admin. A cluster may enable or disable preemption, it's hard to let user to judge it, and even in the virtual cluster, fragments resource still blocks allocation. The goal is to let user know fragmentation may block resource allocation, that's enough to a user.

docs/user/troubleshooting_job.md
squirrelsc: Thanks, I will update it accordingly.

docs/user/troubleshooting_job.md
xudifsd: Ok, maybe we can change the doc later when #2208 has finished.

PR #2306, opened 2019-03-11
[Rest API] Add task state; Add job's retry details; Refine job config

sunqinzheng:
- add task state field
- add job's retry details field
- refine job config api
    - use request's Accept header to determine whether json or yaml will be sent.
    - BREAK CHANGE: fix the bug that old job config api may return an escaped json string instead of a json object

### DIFF
--- a/docs/rest-server/API.md
+++ b/docs/rest-server/API.md
@@ -440,10 +440,16 @@ Status: 200
     createdTime: "createdTimestamp",
     completedTime: "completedTimestamp",
     executionType: "executionType",
-    // Sum of succeededRetriedCount, transientNormalRetriedCount,
-    // transientConflictRetriedCount, nonTransientRetriedCount,
-    // and unKnownRetriedCount
-    retries: retriedCount,
+    // sum of retries
+    retries: retries,
+    retryDetails: {
+      // Failed caused by user's code
+      user: userRetries,
+      // System failed, and it can ensure that it will success within a finite retry times
+      system: systemRetries,
+      // System cannot get required resource to run.
+      resource: resourceRetries,
+    },
     appId: "applicationId",
     appProgress: "applicationProgress",
     appTrackingUrl: "applicationTrackingUrl",
@@ -461,6 +467,7 @@ Status: 200
       },
       taskStatuses: {
         taskIndex: taskIndex,
+        taskState: taskState,
         containerId: "containerId",
         containerIp: "containerIp",
         containerPorts: {
--- a/docs/rest-server/API.md
+++ b/docs/rest-server/API.md
@@ -445,8 +445,8 @@ Status: 200
     retryDetails: {
       // Job failed due to user or unknown error
       user: userRetries,
-      // Job failed due to system/plaform error
-      system: systemRetries,
+      // Job failed due to platform error
+      platform: platformRetries,
       // Job cannot get required resource to run within timeout
       resource: resourceRetries,
     },
--- a/docs/rest-server/API.md
+++ b/docs/rest-server/API.md
@@ -443,11 +443,11 @@ Status: 200
     // sum of retries
     retries: retries,
     retryDetails: {
-      // Failed caused by user's code
+      // Job cannot get required resource to run within timeout
       user: userRetries,
-      // System failed, and it can ensure that it will success within a finite retry times
+      // Job failed due to system/plaform error
       system: systemRetries,
-      // System cannot get required resource to run.
+      // Job failed due to user or unknown error
       resource: resourceRetries,
     },
     appId: "applicationId",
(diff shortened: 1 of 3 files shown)


docs/rest-server/API.md
```diff
@@ -440,10 +440,16 @@ Status: 200
     createdTime: "createdTimestamp",
     completedTime: "completedTimestamp",
     executionType: "executionType",
-    // Sum of succeededRetriedCount, transientNormalRetriedCount,
-    // transientConflictRetriedCount, nonTransientRetriedCount,
-    // and unKnownRetriedCount
-    retries: retriedCount,
+    // sum of retries
+    retries: retries,
+    retryDetails: {
```
yqwang-ms: better name to be retryBreakdown?

docs/rest-server/API.md
```diff
@@ -440,10 +440,16 @@ Status: 200
...
+      // Failed caused by user's code
+      user: userRetries,
+      // System failed, and it can ensure that it will success within a finite retry times
+      system: systemRetries,
+      // System cannot get required resource to run.
```
yqwang-ms: "Job cannot get required resource to run within timeout"

docs/rest-server/API.md
```diff
@@ -440,10 +440,16 @@ Status: 200
...
-    // Sum of succeededRetriedCount, transientNormalRetriedCount,
-    // transientConflictRetriedCount, nonTransientRetriedCount,
-    // and unKnownRetriedCount
-    retries: retriedCount,
+    // sum of retries
+    retries: retries,
+    retryDetails: {
+      // Failed caused by user's code
+      user: userRetries,
+      // System failed, and it can ensure that it will success within a finite retry times
```
yqwang-ms: "Job failed due to system/plaform error"

docs/rest-server/API.md
```diff
@@ -440,10 +440,16 @@ Status: 200
...
     completedTime: "completedTimestamp",
     executionType: "executionType",
-    // Sum of succeededRetriedCount, transientNormalRetriedCount,
-    // transientConflictRetriedCount, nonTransientRetriedCount,
-    // and unKnownRetriedCount
-    retries: retriedCount,
+    // sum of retries
+    retries: retries,
+    retryDetails: {
+      // Failed caused by user's code
```
yqwang-ms: "Job failed due to user error"

docs/rest-server/API.md
sunqinzheng: fixed

docs/rest-server/API.md
sunqinzheng: fixed

docs/rest-server/API.md
sunqinzheng: fixed

docs/rest-server/API.md
sunqinzheng: I think "details" is better...

docs/rest-server/API.md
```diff
@@ -440,10 +440,16 @@ Status: 200
...
     completedTime: "completedTimestamp",
     executionType: "executionType",
-    // Sum of succeededRetriedCount, transientNormalRetriedCount,
-    // transientConflictRetriedCount, nonTransientRetriedCount,
-    // and unKnownRetriedCount
-    retries: retriedCount,
+    // sum of retries
+    retries: retries,
+    retryDetails: {
+      // Job cannot get required resource to run within timeout
```
yqwang-ms: wrong comment location?

sunqinzheng: fixed

docs/rest-server/API.md
```diff
@@ -440,10 +440,16 @@ Status: 200
...
-    // transientConflictRetriedCount, nonTransientRetriedCount,
-    // and unKnownRetriedCount
-    retries: retriedCount,
+    // sum of retries
+    retries: retries,
+    retryDetails: {
+      // Job failed due to user or unknown error
+      user: userRetries,
+      // Job failed due to system/plaform error
+      system: systemRetries,
```
yqwang-ms: "platform" is better than "system" due to:
1. More accurate, user job may also be a "system"
2. Our PAI is "platform" for AI, not "system" for AI
3. There are PaaS, but no "system" as a Service

so can you replace all "system" words with "platform" in this PR?

sunqinzheng: I find that "system" is used in rest api document (https://github.com/Microsoft/pai/blob/master/docs/rest-server/API.md#post-token) multiple times. Shall we change them as well?

yqwang-ms: Seems there is no need to distingush with user in that doc, so no ambiguity in the doc (the "system" in the doc implies PAI).
So, it is better to replace "in the system" to "in PAI", but it is not necessary, it depends on you :)

sunqinzheng: fixed

sunqinzheng:
#2382

PR #2310, opened 2019-03-12
Add upgrading doc

hao1939:
Add upgrading doc, could upgrade from v0.8.


hao1939:
PR to release branch instead.

PR #2314, opened 2019-03-12
[HOT-FIX] Set the qucik-start template image tag to v0.10.1

hao1939:
Here's the case:

1. When user install, the `latest` was v0.9.
2. Somehow one service, for example rest-server, restart. Kubernetes would try to using the `latest` image, which is v0.10.

So we should fix the quick-start template tag to the currently release.
In this case, it should be v0.10.1.


hao1939:
@mzmssg Should update it in the v0.11 release too.

mzmssg:
@hao1939 
Is this in our release process?

hao1939:
Yes, it should.

PR #2319, opened 2019-03-13
Query unhealthy GPU from prometheus

mzmssg:
A part of #2192

After #2209, unhealthy gpu info will be collected by prometheus. This PR provides following operation guide after alerting.

In our design, admin need to do:
1. Query prometheus, collect unhealthy nodes.
2. Add unhealthy nodes to blacklist.
3. Enforce service to reload the blacklist.

More details in https://github.com/Microsoft/pai/blob/8b186958f9ed55c4c0fac157977f58f9b56adc8c/docs/tools/node_maintain.md

### DIFF
--- a/docs/tools/node_maintain.md
+++ b/docs/tools/node_maintain.md
@@ -4,7 +4,8 @@
 
 This [tool](../../src/tools/node_maintain.py) help you to gracefully decommission unhealthy nodes in OpenPAI. 
 
-**NOTES: This tool should be invoked in the dev-box, under `{PAI_ROOT_DIR}/src/tools`. For now, it only supports decommissioning YARN nodes.**
+**NOTES: This tool should be invoked in the dev-box, under `{PAI_ROOT_DIR}/src/tools`. For now, it only supports decommissioning YARN nodes.
+HDFS decommissioning might be added in the future.**
 
 ## Commands
 
@@ -15,6 +16,8 @@ python node_maintain.py {object} {action} {arguments}
 ```
 
 ### Get current unhealthy GPU nodes
+
+PAI monitors gpu healthiness and aggregate them in prometheus. You could query the bad GPUs from it in some ways:
 
 ###### 1. Query prometheus by this tool
 ```bash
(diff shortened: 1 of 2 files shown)


docs/tools/node_maintain.md
```diff
@@ -2,45 +2,60 @@
...
 
-Graceful decommissioning could minimize the impact on existing job when removing unhealthy nodes. 
-This is a tool to help you achieve it in OpenPAI. 
+This [tool](../../src/tools/node_maintain.py) help you to gracefully decommission unhealthy nodes in OpenPAI. 
 
-In OpenPAI, an unhealthy node-list records the nodes to be decommissioned. 
-Components could sync the list and decommission them as needed.
-
-**NOTES: For now, only YARN nodes are decommissioned, HDFS doesn't.**
+**NOTES: This tool should be invoked in the dev-box, under `{PAI_ROOT_DIR}/src/tools`. For now, it only supports decommissioning YARN nodes.**
```
hao1939: So HDFS node was not decommissioned, suggest to point it out.

docs/tools/node_maintain.md
```diff
@@ -2,45 +2,60 @@
...
+python node_maintain.py {object} {action} {arguments}
+```
+
+### Get current unhealthy GPU nodes
 
-So, a common invoking looks like:
+###### 1. Query prometheus by this tool
 ```bash
-python node_maintain.py {master_ip} sub-command sub-command-arguments
+python node_maintain.py badgpus get -m {prometheus_ip} [--prometheus-port {prometheus-port}]
```
hao1939: We'd better add some background about the why we 'query prometheus'. For example: We monitor some gpu metrics using prometheus, we can query the bad GPUs from prometheus.

docs/tools/node_maintain.md
mzmssg: Added

docs/tools/node_maintain.md
mzmssg: Added

PR #2327, opened 2019-03-15
Add convert code to add id for webportal plugins

wangdian:
Add converter to add id for webportal plugins. Use this when migrating from v0.10 to v0.11


hao1939:
For the v0.11 release, there still need code to call the script during the `service start`.

PR #2330, opened 2019-03-15
support get task usage of kube tasks

xudifsd:
Remove `container_label_PAI_HOSTNAME` from task resource labels, and support get task usage from kube tasks


src/job-exporter/src/docker_inspect.py
```diff
@@ -25,7 +25,7 @@
 
 logger = logging.getLogger(__name__)
 
-target_label = {"PAI_HOSTNAME", "PAI_JOB_NAME", "PAI_USER_NAME", "PAI_CURRENT_TASK_ROLE_NAME", "GPU_ID"}
+target_label = {"PAI_JOB_NAME", "PAI_USER_NAME", "PAI_CURRENT_TASK_ROLE_NAME", "GPU_ID"}
```
yqwang-ms: Can we use these as env to find k8s docker container?

xudifsd: what do you mean?

PR #2334, opened 2019-03-15
[Job Debugging] Add --cap-add=SYS_PTRACE to use gdb

ydye:


src/rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -301,6 +301,9 @@ docker run --name $docker_name \
   --shm-size={{ taskData.shmMB }}m \
   $nvidia_devices \
   --device=/dev/fuse \
+{{# isDebug }}
+  --cap-add=SYS_PTRACE
```
abuccts: * syntax error, add `\` at the end
* add `--security-opt seccomp:unconfined` flag as well

```suggestion
  --cap-add=SYS_PTRACE \
  --security-opt seccomp:unconfined \
```

ydye: Done

PR #2338, opened 2019-03-18
map gpu in kube tasks

xudifsd:
Tasks run by nvidia-container-runtime will have ENV of `NVIDIA_VISIBLE_DEVICES` and value is GPU uuids, these uuid can mapped into minor number by info get from nvidia-smi.

This PR will make job-exporter able to get resource usage of tasks run by either yarn-launcher and kube-launcher, including GPU usage.

### DIFF
--- a/src/job-exporter/src/collector.py
+++ b/src/job-exporter/src/collector.py
@@ -485,13 +485,26 @@ class ContainerCollector(Collector):
         if inspect_info.gpu_ids:
             ids = inspect_info.gpu_ids.replace("\"", "").split(",")
             for id in ids:
+                # If the container was scheduled by yarn, we get its GPU usage
+                # info from label GPU_ID, value of the label is minor_number, and
+                # will be digits.
+                # If the container was scheduled by kube launcher, we get its GPU
+                # usage info from environment NVIDIA_VISIBLE_DEVICES, the value
+                # is like GPU-dc0671b0-61a4-443e-f456-f8fa6359b788. The mapping
+                # from uuid to minor_number is get via nvidia-smi, and gpu_infos
+                # should have key of this uuid.
                 if id.isdigit():
                     gpu_ids.append(id)
                 elif id and gpu_infos is not None:
                     # id is in form of UUID like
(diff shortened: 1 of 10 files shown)


src/job-exporter/src/collector.py
```diff
@@ -470,20 +473,27 @@ def collect_impl(self):
...
+
+        if inspect_info.gpu_ids:
+            ids = inspect_info.gpu_ids.replace("\"", "").split(",")
+            for id in ids:
+                if id.isdigit():
+                    gpu_ids.append(id)
+                elif id and gpu_infos is not None:
+                    # id is in form of UUID like
+                    # GPU-dc0671b0-61a4-443e-f456-f8fa6359b788
+                    if gpu_infos.get(id) is not None:
```
yqwang-ms: Do we need a warning log if it is None?

src/job-exporter/src/collector.py
```diff
@@ -470,20 +473,27 @@ def collect_impl(self):
...
+
+        result_labels["username"] = inspect_info.username or "unknown"
+        result_labels["job_name"] = inspect_info.job_name or "unknown"
+        result_labels["role_name"] = inspect_info.role_name or "unknown"
+        result_labels["task_index"] = inspect_info.task_index or "unknown"
+
+        if inspect_info.gpu_ids:
+            ids = inspect_info.gpu_ids.replace("\"", "").split(",")
+            for id in ids:
+                if id.isdigit():
```
yqwang-ms: In which case, it is directlt the gpu id, and in which case it is just a UUID?

scarlett2018:
TODO: help check regression during release testing process
@xudifsd

src/job-exporter/src/collector.py
xudifsd: added

src/job-exporter/src/collector.py
xudifsd: added comment in code.

Containers from yarn will have minor in gpu id, which are digits.

Containers from kube launcher will have UUID in gpu id.

xudifsd:
test pass in release process

Issue #2335, opened 2019-03-17
GPU used by kube-launcher's task will be reported as external process


xudifsd:
**Short summary about the issue/question**:
External processes are detected by using `/proc/$pid/cgroup` and check if they match /docker/$container_id

But tasks scheduled by kube-launcher will have content in `/proc/$pid/cgroup` of

```
12:cpuset:/kubepods/pod69c72bad-46fa-11e9-8937-000d3ab38724/29fe36ad918457348a6429aa74283efc2beceb9d807722410e256d0968cd1d47
11:rdma:/
10:pids:/kubepods/pod69c72bad-46fa-11e9-8937-000d3ab38724/29fe36ad918457348a6429aa74283efc2beceb9d807722410e256d0968cd1d47
9:blkio:/kubepods/pod69c72bad-46fa-11e9-8937-000d3ab38724/29fe36ad918457348a6429aa74283efc2beceb9d807722410e256d0968cd1d47
8:perf_event:/kubepods/pod69c72bad-46fa-11e9-8937-000d3ab38724/29fe36ad918457348a6429aa74283efc2beceb9d807722410e256d0968cd1d47
7:freezer:/kubepods/pod69c72bad-46fa-11e9-8937-000d3ab38724/29fe36ad918457348a6429aa74283efc2beceb9d807722410e256d0968cd1d47
6:net_cls,net_prio:/kubepods/pod69c72bad-46fa-11e9-8937-000d3ab38724/29fe36ad918457348a6429aa74283efc2beceb9d807722410e256d0968cd1d47
5:memory:/kubepods/pod69c72bad-46fa-11e9-8937-000d3ab38724/29fe36ad918457348a6429aa74283efc2beceb9d807722410e256d0968cd1d47
4:hugetlb:/kubepods/pod69c72bad-46fa-11e9-8937-000d3ab38724/29fe36ad918457348a6429aa74283efc2beceb9d807722410e256d0968cd1d47
3:devices:/kubepods/pod69c72bad-46fa-11e9-8937-000d3ab38724/29fe36ad918457348a6429aa74283efc2beceb9d807722410e256d0968cd1d47
2:cpu,cpuacct:/kubepods/pod69c72bad-46fa-11e9-8937-000d3ab38724/29fe36ad918457348a6429aa74283efc2beceb9d807722410e256d0968cd1d47
[...]

PR #2341, opened 2019-03-18
prevent tasks scheduled by kube launcher be deemed as external user

xudifsd:
We have detection of external gpu users, previous this is implemented as check /proc/$pid/cgroup content, if it not starts with /docker/, it will be treat as external process.

But tasks started by kube launcher will have content started with `/kubepods/`.

This PR will treat these tasks also as normal users.


xudifsd:
fixed #2335

Issue #2146, opened 2019-02-15
hadoop-ai can't detect ECC error


mzmssg:
**Organization Name**: 

Microsoft

**Short summary about the issue/question**:

Job scheduled to ECC error GPUs

**How to reproduce it**: 

Current ECC checker will fail if ECC error code greater than 1.

mzmssg:
fixed in #2343

PR #2343, opened 2019-03-18
Fix hadoop-ai ecc eccor detection

mzmssg:
Origin code only consider ecc code 0/1, but it's actually an int.

Readable patch is here:
https://github.com/mzmssg/hadoop/pull/3


mzmssg:
#2146

PR #2352, opened 2019-03-19
Support nvidia-container-runtime for gpu isolation

abuccts:
Support nvidia-container-runtime for gpu isolation:

1. *Support nvidia-container-runtime for gpu isolation*
Set `NVIDIA_VISIBLE_DEVICES` to `void` to avoid conflict with runc, reference:
https://github.com/NVIDIA/nvidia-container-runtime#nvidia_visible_devices.
Works for w/ and w/o `--runtime=nvidia`.

2. *Set default runtime to runc*
Set `--runtime` to `runc` explicitly to overwrite default runtime.

3. *Drop `MKNOD` capability*
Drop `MKNOD` capability when starting container.
`nvidia-smi` will make all gpu devices show up under /dev, drop [mknod](https://linux.die.net/man/2/mknod) capability to avoid this.
Reference: NVIDIA/nvidia-docker#170


xudifsd:
fixed #1667

Issue #2336, opened 2019-03-18
Should count docker errors as transient system errors


mzmssg:
**Organization Name**: 
Microsoft

**Short summary about the issue/question**:
Job failure with dockerd error code  125, not sure the root case. Anyway, this job should get a retry.

**Anything else we need to know**:
Docker error code list: https://github.com/moby/moby/pull/14012
Job error log:
```
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
time="2019-03-04T05:45:26Z" level=error msg="error waiting for container: unexpected EOF" 
```
```
+[05:45:26] echo '[DEBUG] Yarn container exit handler: yarn container exit code: 125'
```

mzmssg:
fixed in #2355

PR #2355, opened 2019-03-19
yarn container script exit handler fix and refactor

mzmssg:
This PR includes:
1. Revert #1108, remove unnecessary logs.
2. Improve log readability. Killed job and finished job will have different logs. 
Killed job:
image
Finished job:
image
3. #2336 
We assume only docker errors raise 125 errorcode. This pr covert it to 205, which is launcher USER_APP_TRANSIENT_ERROR. Then launcher will retry it without burning retry count.

### DIFF
--- a/src/rest-server/src/templates/yarnContainerScript.mustache
+++ b/src/rest-server/src/templates/yarnContainerScript.mustache
@@ -44,7 +44,7 @@ function exit_handler()
 {
   rc=$?
   printf "[DEBUG] EXIT signal received in yarn container, performing clean up action...\n"
-  if [ $rc -eq 125]; then
+  if [ $rc -eq 125 ]; then
     printf "[DEBUG] Exit code 125, it may be docker daemon error, convert it to a transient error\n"
     rc=205
   fi
--- a/src/rest-server/src/templates/yarnContainerScript.mustache
+++ b/src/rest-server/src/templates/yarnContainerScript.mustache
@@ -26,44 +26,49 @@ export PAI_DEFAULT_FS_URI={{ hdfsUri }}
 exec 13>$LAUNCHER_LOG_DIR/YarnContainerDebug.log
 BASH_XTRACEFD=13
 
+function sigterm_handler()
+{
+  printf "[DEBUG] SIGTERM received, job may be killed by YARN\n"
+
+  pid=$(docker inspect --format={{{ inspectPidFormat }}} $docker_name 2>/dev/null)
+  if [ $pid -gt 0 ]; then
+    kill -15 $pid &&\
+      printf "[DEBUG] Docker container $docker_name killed successfully." ||\
+      printf "[DEBUG] Try to kill the container $docker_name but failed. Maybe it has already exited."
+  fi
+  exit 143
+}
+
 
 function exit_handler()
 {
   rc=$?
-  echo "Exited with $rc"
-  local handler="Yarn container exit handler"
-  debug_log "$handler"  "EXIT signal received in yarn container, performing clean up action..."
+  printf "[DEBUG] EXIT signal received in yarn container, performing clean up action...\n"
+  if [ $rc -eq 125]; then
+    printf "[DEBUG] Exit code 125, it may be docker daemon error, convert it to a transient error\n"
+    rc=205
+  fi
 
-  debug_log "$handler" "trying to kill docker container $docker_name"
-  docker logs $docker_name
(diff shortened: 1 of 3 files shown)


mzmssg:
> What if yarn container killed by sigkill?

It will return with 137 immediately, no any clean action. Then container is left on the host.

fanyangCS:
> > What if yarn container killed by sigkill?
> 
> It will return with 137 immediately, no any clean action. Then container is left on the host.

I assume this will be a rare case. we can ask diskcleaner to remove such dead containers periodically.

src/rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -26,44 +26,53 @@ export PAI_DEFAULT_FS_URI={{ hdfsUri }}
...
-  pkill --parent $$
-
-  debug_log "$handler" "write exit code to file"
-  debug_log "$handler" "yarn container exit code: $rc"
-  debug_log "$handler" "exit code file path: /var/lib/hadoopdata/nm-local-dir/nmPrivate/$APP_ID/$CONTAINER_ID/$CONTAINER_ID.pid.exitcode"
+  printf "[DEBUG] Write exit code $rc to file /var/lib/hadoopdata/nm-local-dir/nmPrivate/$APP_ID/$CONTAINER_ID/$CONTAINER_ID.pid.exitcode.\n"
   echo $rc > "/var/lib/hadoopdata/nm-local-dir/nmPrivate/$APP_ID/$CONTAINER_ID/$CONTAINER_ID.pid.exitcode"
 
+  # Any time-consuming operations please put here, otherwise NM couldn't get exitcode as restart
+  docker container rm $docker_name
```
fanyangCS: docker rm -f $docker_name ?

src/rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -26,44 +26,53 @@ export PAI_DEFAULT_FS_URI={{ hdfsUri }}
...
+  if [ $rc -eq 125 ]; then
+    # We assume 125 is a docker error, and retry it without burning user's retry count
+    printf "[DEBUG] Exit code 125, it may be docker error, convert it to a transient error\n"
+    rc=205
+  fi
 
-  debug_log "$handler" "trying to kill docker container $docker_name"
-  docker logs $docker_name
-  docker inspect $docker_name
   pid=$(docker inspect --format={{{ inspectPidFormat }}} $docker_name 2>/dev/null)
```
fanyangCS: docker inspect only when exit code is 137 (SIGKILL) and 143 (SIGTERM)?

fanyangCS:
instead of using docker inspect. can we conside using "journalctl -k | grep -i -e memory -e oom" to detect the OOM event? 
Maybe we don't have to use pid, using some the label to retrieve the container from journalctl?

src/rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -26,44 +26,47 @@ export PAI_DEFAULT_FS_URI={{ hdfsUri }}
...
+    rc=55
+  fi
 
-  debug_log "$handler" "write exit code to file"
-  debug_log "$handler" "yarn container exit code: $rc"
-  debug_log "$handler" "exit code file path: /var/lib/hadoopdata/nm-local-dir/nmPrivate/$APP_ID/$CONTAINER_ID/$CONTAINER_ID.pid.exitcode"
+  printf "[DEBUG] Write exit code $rc to file /var/lib/hadoopdata/nm-local-dir/nmPrivate/$APP_ID/$CONTAINER_ID/$CONTAINER_ID.pid.exitcode.\n"
   echo $rc > "/var/lib/hadoopdata/nm-local-dir/nmPrivate/$APP_ID/$CONTAINER_ID/$CONTAINER_ID.pid.exitcode"
 
+  # After all necessary steps, do a best effort kill here
```
fanyangCS: what is the behavior when we stuck in (docker inspect), and yarn sends a SIGKILL to yarncontainer?
in this case, "/var/lib/hadoopdata/nm-local-dir/nmPrivate/$APP_ID/$CONTAINER_ID/$CONTAINER_ID.pid.exitcode" might be different from the actual exit code?

mzmssg: That's fine, the container statemachine might be a little different, but RM won't care that.

PR #2365, opened 2019-03-21
add yarn max allocation per heatbeat

### LINKED ISSUE #2364: Resource manager get stuck
**Organization Name**:
Microsoft

RM get stuck, can't respond to any request.

**How to reproduce it**:
Not sure

**OpenPAI Environment**: 
- OpenPAI version: 0.10, but it may be always there.

These is a similar issue: https://issues.apache.org/jira/browse/YARN-8513?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel
And this is the workaround: https://issues.apache.org/jira/browse/YARN-8896

mzmssg:
adapt this patch to 2.9.0 to fix #2364 : https://issues.apache.org/jira/browse/YARN-8896


mzmssg:
@fanyangCS 
Can't reproduce, but our scheduler get stuck in the same loop, so this workaround should work.

src/hadoop-ai/build/build.sh
```diff
@@ -32,6 +32,8 @@ git checkout branch-2.9.0
 git apply /hadoop-2.9.0.gpu-port.patch
 git apply /HDFS-13773.patch
 git apply /docker-executor.patch
+# to avoid potential endless loop, refer to https://issues.apache.org/jira/browse/YARN-8513?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel
+git apply /12944563.patch
```
yqwang-ms: why not use the orignal Patch name, i.e. YARN-8896.patch

mzmssg: It's not 2.9 patch

PR #2370, opened 2019-03-21
fix delete script of rollback nvidia runtime

xudifsd:
fix bug in rollback nvidia container runtime


fanyangCS:
need a hotfix?

xudifsd:
> need a hotfix?

No, no one currently used this, this will only triggered by setting `set-nvidia-runtime=true`.

Issue #2387, opened 2019-03-25
The code_dir storage error log is not consistent with size


debuggy:
**Organization Name**:
microsoft

**Short summary about the issue/question**:

When User code_dir_size is 0. There will be an error log says that dir is larger than 200MB.
```
[INFO] User code_dir_size is 0
hdfs://10.0.0.4:9000/qinsu/qinsu-test2_04320b0b is larger than 200MB, exit.
```
The related bug location:
https://github.com/Microsoft/pai/blob/0f80208482eade027af88c61b4d58b78bf39b57a/src/rest-server/src/templates/yarnContainerScript.mustache#L262

fanyangCS:
https://github.com/Microsoft/pai/pull/2388

PR #2388, opened 2019-03-25
 fix: code_dir_size report incorrect error message

debuggy:
- Add a condition when $code_dir_size is less than 0.


src/rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -262,6 +262,9 @@ if [[ -n $PAI_CODE_DIR ]]; then
   if [ $code_dir_size -gt 0 ] && [ $code_dir_size -le 200000000 ]; then
```
debuggy: Since the code_dir will be removed in the future. Should we pass the size==0 condition instead of report error?

debuggy:
#2387

src/rest-server/src/templates/yarnContainerScript.mustache
fanyangCS: yes size==0 should be allowed.

debuggy: I am confused about when to use single bracket and when to use double in the bash. Is there any convention in this project?

https://github.com/Microsoft/pai/blob/0f80208482eade027af88c61b4d58b78bf39b57a/src/rest-server/src/templates/yarnContainerScript.mustache#L256

debuggy: ok, I will make some changes corresspondingly.

src/rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -262,6 +262,12 @@ if [[ -n $PAI_CODE_DIR ]]; then
...
     hdfs dfs -get $PAI_CODE_DIR $code_dir \
       || { echo "Can not get code from HDFS. HDFS may be down."; exit 1; }
+  elif [ $code_dir_size -eq 0]; then
+    # pass when code_dir_size is 0
+    : 
+  elif [ $code_dir_size -lt 0 ]; then
```
fanyangCS: >$code_dir_size [](start = 9, length = 14)

in which case code_dir_size will be less than 0? should be a bug?

debuggy: It should be a bug if less than 0. So the point is whether to print an error log here or just pass and let other parts catch this bug.

fanyangCS: exit here seems ok.

---
In reply to: [268521074](https://github.com/Microsoft/pai/pull/2388#discussion_r268521074) [](ancestors = 268521074)

PR #2393, opened 2019-03-25
[Webportal] Add active sidebar menu in job list/detail pages

abuccts:
Add active sidebar menu in job list/detail pages.


fanyangCS:
should this go to 0.11 too?

abuccts:
Yep, will add in #2394.

PR #2406, opened 2019-03-26
Add support for multiline user commands

zhiltsov-max:
- Added support for multiline user commands
    - User commands gets written to a separate script file. By default, the script fails on the first error and prints executed commands
    - Command has some reasonable restrictions on line count and line length
- Proxy environment variables forwarded to `/etc/environment` (can be improved by using more portable  `/etc/profile`, forwarding all the environment, and exposing proxy settings somewhere)
- Removed unused code from Docker entry point
- Slightly improved log messages
- Fixed `mv` file access issue for `/pai/ssh_config` read-only mount

### DIFF
--- a/src/rest-server/src/templates/dockerContainerScript.mustache
+++ b/src/rest-server/src/templates/dockerContainerScript.mustache
@@ -143,7 +143,7 @@ function generate_ssh_connect_info()
   fi
 }
 
-# Check is SSH exists, if not - ignore ssh preparation and start part
+# Check if SSH exists, if not - ignore ssh preparation and start part
 # Start sshd in docker container
 if service --status-all 2>&1 | grep -q ssh; then
   prepare_ssh
@@ -180,40 +180,16 @@ azure_rdma_preparation
 {{/ azRDMA }}
 {{/ reqAzRDMA }}
 
-function write_user_command()
-{
-  local scriptPath="$1"
-
-  # Read an escaped string to an array and unescape
-  local lines=$'{{ taskData.command }}'
-  readarray -n 50 -t lines <<< "${lines}"
-  read -n 1000 -a lines <<< "${lines[@]}"
-
-  echo "#!/bin/bash" > "${scriptPath}"
-
-  # By default, exit on the first error
-  echo 'set -e' >> "${scriptPath}"
-  # Enable logging for the user command
-  echo 'set -x' >> "${scriptPath}"
-
-  printf '%s\n' "${lines[@]}" >> "${scriptPath}"
-
-  chmod a+x "${scriptPath}"
-}
-
 function run_user_command()
 {
-  local scriptPath="$1"
+  set -a
   printf "%s %s\n\n" "[INFO]" "USER COMMAND START"
-  exec "${scriptPath}" || exit $?
+  $1 || exit $?
   printf "\n%s %s\n\n" "[INFO]" "USER COMMAND END" # job_exporter relay on this exact output, if you are going to change this, remember to change job_exporter also
   exit 0
 }
 
-PAI_USER_COMMAND_PATH="/pai/command.sh"
-write_user_command "${PAI_USER_COMMAND_PATH}"
-
-run_user_command "${PAI_USER_COMMAND_PATH}" &
+run_user_command {{ userCommandScriptContainerPath }} &
 user_command_pid=$!
 
 while [ $(( $(date +%s) - $(stat -c %Y /alive/yarn_$PAI_CONTAINER_ID) )) -lt 120 ] && \
--- a/src/rest-server/src/templates/dockerContainerScript.mustache
+++ b/src/rest-server/src/templates/dockerContainerScript.mustache
@@ -47,13 +47,16 @@ touch "/alive/docker_$PAI_CONTAINER_ID"
 
 export PAI_WORK_DIR="$(pwd)"
 PAI_WEB_HDFS_PREFIX={{ webHdfsUri }}/webhdfs/v1/Container
-HDFS_LAUNCHER_PREFIX=$PAI_DEFAULT_FS_URI/Container
-export CLASSPATH="$(hadoop classpath --glob)"
+
 # Make all GPUs allocated to the container accessible
 export NVIDIA_VISIBLE_DEVICES=all
 
 task_role_no={{ idx }}
+
 printf "%s %s\n%s\n\n" "[INFO]" "ENV" "$(printenv | sort)"
+
+# Write env to system-wide environment
+env | grep -E "^PAI|PATH|PREFIX|JAVA|HADOOP|NVIDIA|CUDA|(.*_PROXY)" > /etc/environment
 
 cp -r /pai/code/* ./
 
@@ -109,6 +112,7 @@ function get_ssh_key_files()
   info_source="webhdfs"
   localKeyPath=/root/.ssh/{{ jobData.jobName }}.pub
   localPrivateKeyPath=/root/.ssh/{{ jobData.jobName }}
+
   if [[ -f $localKeyPath ]]; then
     rm -f $localKeyPath
   fi
@@ -139,7 +143,7 @@ function generate_ssh_connect_info()
   fi
 }
 
-# Check whether hdfs bianry and ssh exists, if not ignore ssh preparation and start part
+# Check is SSH exists, if not - ignore ssh preparation and start part
 # Start sshd in docker container
 if service --status-all 2>&1 | grep -q ssh; then
   prepare_ssh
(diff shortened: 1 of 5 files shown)


src/rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -177,17 +179,39 @@ function  azure_rdma_preparation()
 azure_rdma_preparation
 {{/ azRDMA }}
 {{/ reqAzRDMA }}
-# Write env to system-wide environment
-env | grep -E "^PAI|PATH|PREFIX|JAVA|HADOOP|NVIDIA|CUDA" > /etc/environment
+
+function write_user_command()
+{
```
abuccts: We do not need to support multiple line commands here:
* For current version, users can set `command` to `command1 && command2 && ...` or `command1 ; command2 ; ...` to execute multiple commands. If you really think "\n" is useful in `command`, you can add a wrapper to replace `\n` with `;` for `taskData.command` in [mustache renderer](https://github.com/Microsoft/pai/blob/master/src/rest-server/src/models/job.js#L442).
* For future job configuration file, we do support multiple line commands with pre-commands/post-commands, you can refer to [protocol docs](https://github.com/Microsoft/pai/blob/feature/submission-v2/docs/pai-job-protocol.yaml#L183-L196) for example.

zhiltsov-max: Thanks for the comment. Yes, complex commands can be chained by `&&` and others, but the use of `\n` makes them much more readable - this is the main reason for this change. Could you name any downsides of multiline commands? It looks like PAI tries to avoid them, is there any reason for that?

abuccts: `\n` is ok but I prefer to support it when rendering mustache (one or two line js code) instead of runtime shell scripts (20+ lines sh code) which will introduce more complexity.

`\n` does make the command more readable, but it's on web page or config file. How we support it inside Docker container wouldn't affect user experience.

zhiltsov-max: Doing this during the mustache processing is a good idea and I'm, actually, stand for it. To do this, I think, we have to remove command escaping and directly insert the user command to the target script, possibly forwarding its path. I have no idea how to do this, could you give me some pointers? 

Simple replacement of all the `\n`s should be wrong as they can be a part of the user command.

abuccts: Hi, sorry for the late response.
Replacing all `\n` is not a good idea indeed, but writing command to a target script then execute is not good neither. User may rely on relative path (e.g. `python ./code/run.py`) inside Docker so writing to another path will break existing jobs.

zhiltsov-max: Hm, relative paths should be unaffected as long as we don't change current directory. BTW, I thought you are going to deprecate `PAI_CODE_DIR`, aren't you?

abuccts: You're right.
Yes, we are going to deprecate `PAI_CODE_DIR` and let users download code/data themselves. To be more specific, it will be implemented in [`deployments` section](https://github.com/Microsoft/pai/blob/feature/submission-v2/docs/pai-job-protocol.yaml#L198-L220) in the new config file.

src/rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -177,17 +179,36 @@ function  azure_rdma_preparation()
...
+  PAI_USER_COMMAND_DIR="/pai/taskData"
+  PAI_USER_COMMAND_PATH="${PAI_USER_COMMAND_DIR}/command.sh"
```
abuccts: you can use the path `/pai/command.sh`, maybe make it an argument to functions `write_user_command` and `run_user_command`?

src/rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -177,17 +179,36 @@ function  azure_rdma_preparation()
...
+  mkdir -p "${PAI_USER_COMMAND_DIR}"
+
+  echo "#!/bin/bash" > "${PAI_USER_COMMAND_PATH}"
+
+  # By default, exit on the first error
+  echo 'set -e' >> "${PAI_USER_COMMAND_PATH}"
+  # Enable logging for the user command
+  echo 'set -x' >> "${PAI_USER_COMMAND_PATH}"
```
abuccts: move these `echo` to the beginning of `cat`?

mzmssg:
> /cc @mzmssg
> After writing commands to a file, it will be easy to separate user's stdout/stderr from container's.

Yes, actually I agree this pr, writing command to a new file seems more unified with pai-lite.

zhiltsov-max:
@mzmssg, if you want to separate container output, you can drop `--tty` flag from `docker run`. The dowside is that the output of some programs can be modified:

``` bash
# With tty:
$ ls
file1     file2
file3     file4

# Without tty:
$ ls
file1
file2
file3
file4
```

The problem with pseudo-terminals is general in Linux - they always combine stdout and stderr.

src/rest-server/src/templates/dockerContainerScript.mustache
zhiltsov-max: The idea was to hide the previous setup messages from the log - they should not be interesting to the user.

src/rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -177,17 +179,36 @@ function  azure_rdma_preparation()
...
+
+  cat <<EOF >> "${PAI_USER_COMMAND_PATH}"
+{{{ taskData.command }}}
+EOF
```
zhiltsov-max: I've realized it should not be done like this - if the user command contains heredocs it can fail. So, I'll revert to the previous escaped variant.

src/rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -139,7 +143,7 @@ function generate_ssh_connect_info()
   fi
 }
 
-# Check whether hdfs bianry and ssh exists, if not ignore ssh preparation and start part
+# Check is SSH exists, if not - ignore ssh preparation and start part
```
abuccts: ”whether SSH exists“ ?

src/rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -177,18 +179,41 @@ function  azure_rdma_preparation()
...
 {{/ reqAzRDMA }}
-# Write env to system-wide environment
-env | grep -E "^PAI|PATH|PREFIX|JAVA|HADOOP|NVIDIA|CUDA" > /etc/environment
+
+function write_user_command()
+{
+  local scriptPath="$1"
+
+  # Read an escaped string to an array and unescape
+  local lines=$'{{ taskData.command }}'
```
abuccts: what if there's single quotation mark `'` in command?

src/rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -177,18 +179,41 @@ function  azure_rdma_preparation()
...
+  readarray -n 50 -t lines <<< "${lines}"
+  read -n 1000 -a lines <<< "${lines[@]}"
```
abuccts: `read` uses space as delimiter? it will split `lines` by space in my test.

zhiltsov-max: Yes, it does, but all whitespaces should come escaped at this point. Only unescaped should be from `\n`s. What `{{ }}` mustache braces do is a substitution of html-escaped string, which is not the same as bash-escaped. I think it's better just to escape all the characters in the command, substitute it with `{{{ }}}`, and then use `read`.

src/rest-server/src/templates/dockerContainerScript.mustache
zhiltsov-max: It should come escaped from `{{ }}`.

abuccts: The `{{ }}` is used in mustache syntax, instead of bash, nodejs will replace the `{{ taskData.command }}` part with real command. [Here's](http://mustache.github.io/mustache.5.html) how it works.

For example, if the real command is `echo "Hello"`, nodejs will generate this line from dockerContainerScript.mustache to `local lines=$'echo "Hello"'` in dockerContainerScript.sh. The `{{ }}` won't be kept in runtime bash.

src/rest-server/src/templates/dockerContainerScript.mustache
abuccts: the escaped characters have been reconfigured [here](https://github.com/Microsoft/pai/blob/master/src/rest-server/src/config/index.js#L32-L48), so `{{{ }}}` isn't needed.

src/rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -177,18 +179,41 @@ function  azure_rdma_preparation()
...
+
+  chmod a+x "${scriptPath}"
+}
 
 function run_user_command()
 {
+  local scriptPath="$1"
   printf "%s %s\n\n" "[INFO]" "USER COMMAND START"
-  eval {{ taskData.command }} || exit $?
+  exec "${scriptPath}" || exit $?
```
mzmssg: exec will overwrite parent context, so `printf "\n%s %s\n\n" "[INFO]" "USER COMMAND END"` will never be executed?

zhiltsov-max: You were right, fixed.

PR #2407, opened 2019-03-26
User pretending for a task

zhiltsov-max:
- Enabled user changing in a container
    - Now, it's possible to change user parameters of the user in a task container. This is similar to using `docker run --user <user/uid>:<gid> --groups-add <gid> ...` parameters. The main reason for this is LDAP auth support and access to protected file shares (NFS, Hadoop, etc.) inside the container.
    - Currently, user can specify the user parameters if it is needed, or decide not to leave from `root` (questionable option)
    - `sudo` rights are added to the selected container user if `sudo` is available in a container. This is useful for interactive work to experiment with container setup (enables `apt` calls etc.) and if something like `mount` command is required.
- Task command is extracted to a separate file
- No new requirements for a task container

Depends on #2406

### DIFF
--- a/src/rest-server/src/templates/dockerContainerScript.mustache
+++ b/src/rest-server/src/templates/dockerContainerScript.mustache
@@ -47,13 +47,16 @@ touch "/alive/docker_$PAI_CONTAINER_ID"
 
 export PAI_WORK_DIR="$(pwd)"
 PAI_WEB_HDFS_PREFIX={{ webHdfsUri }}/webhdfs/v1/Container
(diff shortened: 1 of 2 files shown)


fanyangCS:
@zhiltsov-max , you will need to sign the CLA for us to proceed. Meanwhile, we have some ongoing efforts that also cover some of your PRs, let's see how much we can leverage your efforts. By the way, may I ask if this change comes from your internal PAI deployment? If yes, what is the organization?

fanyangCS:
we've moved to k8s version.

PR #2408, opened 2019-03-26
Graceful stop for tasks and OOM conditions handling

zhiltsov-max:
- Added graceful stop support for user tasks
    - Now, when the user clicks on "Stop" button in the web UI or stops a task by a REST request, the task receives a signal. After a period, the task receives a `KILL` signal and gets stopped. 
    - This enables, for example, to save a training checkpoint before the exit.
- Added Out-Of-Memory conditions handling
    - Now, when one of the container processes gets killed due to OOM, the user command receives a signal. The user command decides if it is needed to stop.
- Implementation details: 
    - Signals are sent to all running container processes in user session. Thus, if some do not have a registered handler, they will stop.
    - Signals and timeouts are customizable by setting environment variables for a task. By default, the OOM kill signal is `INT`, the task stop signal is `TERM`, task stop timeout is 30s.
    - No signal specified = no notification.
    - User timeouts have non-modifiable "hard" timeout - 65 seconds (set in cluster configuration)
    - Hard timeouts and signal sequence during task stopping:
        ``` bash
        kill -TERM <yarn container> # <timeout 60s + 65s from YARN config> (needs some clarification)
        kill -KILL <yarn container>
        ```
    - OOM kill checks are performed periodically during the task run and after by `dmesg` output checking. This is not very reliable in some cases, but for most situations it should be enough. The reason is restricted `dmesg` log length. In can probably be resolved by listening cgroup events, but it is not portable way.
    - If a task has failed and there were some OOM notifications, the task exit code is `55 (unknown error)`
- Examples
    To execute a command, which expects some child can fail, do:
    In the task description set:
    ```
    {
      "PAI_GRACEFUL_EXIT_TIMEOUT": 30, # set some custom exit timeout
      "PAI_GRACEFUL_EXIT_SIGNAL": "USR1", # some custom signal to send on stop
      "PAI_OOM_KILLED_SIGNAL": "USR2" # and on OOM kill
    }
    ```

    And launch the following user command:
    ``` bash
    set +e # the most important line, script exits on children kill by default (with -e)
    trap "echo 'got USR1!, exiting'; exit 0" USR1
    trap "echo 'got USR2!'" USR2

    # launch some big task in a separate shell periodically
    (while :; sleep 7; do /oom_test/alloc_test 10000000000; done) &

    # and just wait in the main shell
    while :; do sleep 5; done
    ```
    `stdout` output:
    ```
    [INFO] USER COMMAND START
    
    + set +e
    + trap 'echo '\''got USR1!, exiting'\''; exit 0' USR1
    + trap 'echo '\''got USR2!'\''' USR2
    + :
    + sleep 5
    + :
    + sleep 7
    + :
    + sleep 5
    + /oom_test/alloc_test 10000000000
    Trying to allocate 10000000000 bytes...
    /pai/taskData/command.sh: line 9:   163 Killed                  /oom_test/alloc_test 10000000000
    + :
    + sleep 7
    + :
    + sleep 5
    + /oom_test/alloc_test 10000000000
    Trying to allocate 10000000000 bytes...
    /pai/taskData/command.sh: line 9:   166 Killed                  /oom_test/alloc_test 10000000000
                                   ^--- (comment: from the kernel)
    + :
    + sleep 7
    + :
    + sleep 5
    + :
    + sleep 5
    [WARNING] One of the child processes has been killed by the OOM killer. Notifying the others with the 'USR2' signal
    User defined signal 2 <-- (comment: from Bash, can not be suppressed)
    ++ echo 'got USR2!'
    got USR2!
              ^---- (comment: user command received an OOM kill notification signal and decided to continue)
    + :
    + sleep 5
    + :
    ...
    ```
- Fixed AM hang with infinite spin waiting on an error
- AM now supports graceful stop

### DIFF
--- a/src/rest-server/src/templates/yarnContainerScript.mustache
+++ b/src/rest-server/src/templates/yarnContainerScript.mustache
@@ -344,5 +344,6 @@ docker run --name $docker_name \
   --env NVIDIA_VISIBLE_DEVICES=void \
   --env-file $ENV_LIST \
   --entrypoint="" \
+  --user "root" \
    {{ jobData.image }} \
   /bin/bash '/pai/bootstrap/docker_bootstrap.sh'
--- a/src/rest-server/src/templates/yarnContainerScript.mustache
+++ b/src/rest-server/src/templates/yarnContainerScript.mustache
@@ -26,31 +26,55 @@ export PAI_DEFAULT_FS_URI={{ hdfsUri }}
 exec 13>$LAUNCHER_LOG_DIR/YarnContainerDebug.log
 BASH_XTRACEFD=13
 
-# YARN kill container will send SIGTERM(15) and SIGKILL(9)
-function sigterm_handler()
+
+function get_docker_container_id()
 {
-  printf "[DEBUG] SIGTERM received, job may be killed by YARN\n"
-  killed_by_yarn=true
-  exit 143
+  {{=<% %>=}}
+  docker inspect --format {{.Id}} $1
+  <%={{ }}=%>
 }
 
+function check_docker_oom_killed()
+{
+  {{=<% %>=}}
+  docker inspect --format {{.State.OOMKilled}} $1
+  <%={{ }}=%>
+}
+
+function check_kernel_oom_killer()
+{
+  container_id="$1"
+  [ -e "$container_id" ] && return 1
+  dmesg | grep -i "kill" | grep "/docker/$container_id"
+}
 
 function exit_handler()
 {
-  rc=$?
-  printf "[DEBUG] Performing clean up action...\n"
+  local rc=$?
+  local finish_time="$(date +'%s')"
+
   if [[ $rc -eq 125 ]]; then
     # We assume 125 is a docker error, and retry it without burning user's retry count
-    printf "[DEBUG] Exit code 125, it may be docker error, convert it to a transient error\n"
+    printf "[DEBUG] Exit code 125, it may be Docker error, convert it to a transient error\n"
     rc=205
   fi
 
-  if [[ ${killed_by_yarn} != true && $rc -eq 143 || $rc -eq 137 ]]; then
-    printf "[DEBUG] Exit code $rc, it may be due to out of memory (OOM) or other critical errors.\n"
-    rc=55
+  printf "[DEBUG] EXIT signal received in yarn container, performing clean up action...\n"
+
+  docker stop --time="${PAI_GRACEFUL_EXIT_TIMEOUT:-65}" $docker_name >/dev/null
+
+  # 137, 143 - not handled KILL and TERM signals
+  if [[ "$rc" = "137" || "$rc" = "143" || "$(check_docker_oom_killed $docker_name)" ]]; then
+    if [ "$(check_docker_oom_killed $docker_name)" = "true" ] || \
+         check_kernel_oom_killer $(get_docker_container_id $docker_name) >/dev/null; then
+      echo "[ERROR] One of the container processes has been killed by the OOM killer recently"
+      rc=55 # Some undefined erroneous exit code
+    fi
   fi
 
-  printf "[DEBUG] Write exit code $rc to file /var/lib/hadoopdata/nm-local-dir/nmPrivate/$APP_ID/$CONTAINER_ID/$CONTAINER_ID.pid.exitcode.\n"
+  docker logs --since "$finish_time" $docker_name
+  docker rm -f $docker_name >/dev/null
+
   echo $rc > "/var/lib/hadoopdata/nm-local-dir/nmPrivate/$APP_ID/$CONTAINER_ID/$CONTAINER_ID.pid.exitcode"
 
   # After all necessary steps, do a best effort kill here
@@ -239,9 +263,13 @@ echo PAI_CURRENT_CONTAINER_PORT="$(cut -f 1 -d"," <<< $PAI_CONTAINER_HOST_http_P
 
 export $(grep -v '^#' $ENV_LIST | xargs)
 # put it after the 'export', so we won't source it into yarn container.
-# we want job environment parameters to be available only in job contianer.
+# we want job environment parameters to be available only in job contianer
+# PAI defaults can be overridden this way.
 {{# jobEnvs }}
 PAI_TEMP={{ key }}={{ value }} && echo $PAI_TEMP >> $ENV_LIST
+if [ "{{ key }}" = "PAI_GRACEFUL_EXIT_TIMEOUT" ]; then
+  PAI_GRACEFUL_EXIT_TIMEOUT="{{ value }}"
+fi
 {{/ jobEnvs }}
 
 # Prepare docker bootstrap script
@@ -264,7 +292,7 @@ if [[ -n $PAI_CODE_DIR ]]; then
       || { echo "Can not get code from HDFS. HDFS may be down."; exit 1; }
   elif [ $code_dir_size -eq 0]; then
     # pass when code_dir_size is 0
-    : 
+    :
   elif [ $code_dir_size -lt 0 ]; then
     echo "$PAI_CODE_DIR is less than 0MB, exit."
     exit 1
@@ -299,6 +327,7 @@ docker pull {{ jobData.image }} \
 ## network consumption
 docker run --name $docker_name \
   --init \
+  --detach \
   --tty \
   --rm \
   --runtime=runc \
@@ -347,3 +376,6 @@ docker run --name $docker_name \
   --user "root" \
    {{ jobData.image }} \
   /bin/bash '/pai/bootstrap/docker_bootstrap.sh'
+
+docker attach --no-stdin $docker_name
+
--- a/subprojects/frameworklauncher/yarn/src/main/java/com/microsoft/frameworklauncher/applicationmaster/ApplicationMaster.java
+++ b/subprojects/frameworklauncher/yarn/src/main/java/com/microsoft/frameworklauncher/applicationmaster/ApplicationMaster.java
@@ -1,19 +1,19 @@
 // Copyright (c) Microsoft Corporation
-// All rights reserved.
+// All rights reserved. 
 //
 // MIT License
 //
-// Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
-// documentation files (the "Software"), to deal in the Software without restriction, including without limitation
-// the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated 
+// documentation files (the "Software"), to deal in the Software without restriction, including without limitation 
+// the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and 
 // to permit persons to whom the Software is furnished to do so, subject to the following conditions:
 // The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
 //
-// THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING
-// BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
-// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
-// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING 
+// BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND 
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, 
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, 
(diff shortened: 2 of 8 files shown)


zhiltsov-max:
Depends on #2407

subprojects/frameworklauncher/yarn/src/main/java/com/microsoft/frameworklauncher/applicationmaster/ApplicationMaster.java
```diff
@@ -201,7 +207,19 @@ public synchronized void stop(StopStatus stopStatus) {
     AggregateException ae = new AggregateException();
 
     // Stop AM's SubServices
+
     // No need to stop nmClient, since it may be time consuming to stop all Containers, leave it for RM.
+    try {
+      if (stopStatus.getNeedUnregister()) {
+        if (nmClient != null) {
+          LOGGER.logInfo("Trying to perform graceful shutdown and stop all the containers");
+          nmClient.stop();
```
yqwang-ms: As comment said, it may be time consuming. We at least need a flag to skip the graceful shutdown.

subprojects/frameworklauncher/yarn/src/main/java/com/microsoft/frameworklauncher/applicationmaster/FrameworkInfoPublisher.java
```diff
@@ -82,9 +82,9 @@ protected void run() throws Exception {
     new Thread(() -> {
       while (true) {
         try {
-          publishFrameworkInfo();
-
           Thread.sleep(conf.getLauncherConfig().getAmFrameworkInfoPublishIntervalSec() * 1000);
+
+          publishFrameworkInfo();
```
yqwang-ms: Why move it?

subprojects/frameworklauncher/yarn/src/main/java/com/microsoft/frameworklauncher/applicationmaster/StatusManager.java
```diff
@@ -795,6 +795,12 @@ public synchronized int getOutstandingTaskAppearedRound() {
     return outstandingTaskAppearedRound;
   }
 
+  public synchronized ExecutionType getFrameworkExecutionType() throws Exception {
+    final FrameworkRequest frameworkRequest =
+        zkStore.getFrameworkRequest(conf.getFrameworkName());
+    return frameworkRequest.getFrameworkDescriptor().getExecutionType();
+  }
```
yqwang-ms: Pls do not directly get ExecutionType from ZK, reuse the in-memory one.

subprojects/frameworklauncher/yarn/src/main/java/com/microsoft/frameworklauncher/service/Service.java
```diff
@@ -303,6 +303,7 @@ private ContainerLaunchContext setupContainerLaunchContextForDefaultAM(
 
     // SetupEntryPoint
     Vector<CharSequence> vargs = new Vector<>(30);
+    vargs.add("trap ' ' TERM;"); // prevents intermediate "bash -c" interpreter from immediate exit after receiving TERM signal
     vargs.add(ApplicationConstants.Environment.JAVA_HOME.$$() + "/bin/java");
```
yqwang-ms: Launcher is general purposed, do we have to hard code it here?

subprojects/frameworklauncher/yarn/src/main/java/com/microsoft/frameworklauncher/applicationmaster/ApplicationMaster.java
```diff
@@ -348,6 +366,14 @@ private void stopForInternalUnKnownError(String customizedDiagnostics) {
     stop(new StopStatus(ExitStatusKey.AM_INTERNAL_UNKNOWN_ERROR.toInt(), false, diagnostics));
   }
 
+  private void stopForApplicationStopRequest(String customizedDiagnostics) {
+    String diagnostics = ExitDiagnostics.generateDiagnostics(
+        ExitStatusKey.LAUNCHER_STOP_FRAMEWORK_REQUESTED, customizedDiagnostics);
+
+    // Do not unregister to treat it conservatively.
+    stop(new StopStatus(ExitStatusKey.LAUNCHER_STOP_FRAMEWORK_REQUESTED.toInt(), true, diagnostics));
+  }
```
yqwang-ms: LAUNCHER_STOP_FRAMEWORK_REQUESTED is defined to be NON_TRANSIENT, so we should unregister

subprojects/frameworklauncher/yarn/src/main/java/com/microsoft/frameworklauncher/applicationmaster/ApplicationMaster.java
```diff
@@ -1361,8 +1387,21 @@ public void onError(Throwable e) {
   }
 
   public void onShutdownRequest() {
-    stopForInternalTransientNormalError(
-        "onShutdownRequest called into AM from RM, maybe this Attempt does not exist in RM.");
+    Boolean isApplicationStopped = false;
+    try {
+      isApplicationStopped = statusManager.getFrameworkExecutionType() == ExecutionType.STOP;
+    } catch (Throwable e) {
+      LOGGER.logWarning(e, "Failed to check framework execution type, AM will not unregister");
```
yqwang-ms: You catch all things, it may be a breaking change if RMClient calls onShutdownRequest

subprojects/frameworklauncher/yarn/src/main/java/com/microsoft/frameworklauncher/applicationmaster/FrameworkInfoPublisher.java
zhiltsov-max: @yqwang-ms, this is the bug with infinite spin-waiting you were asking about. If `publishFrameworkInfo()` has thrown, the timeout was never started. This function throws, at least, if `zkStore` is unavailable. It becomes unavailable, for example, after its stop in the stop signal handling thread.

subprojects/frameworklauncher/yarn/src/main/java/com/microsoft/frameworklauncher/applicationmaster/StatusManager.java
zhiltsov-max: Please, suggest better place for this function. Actually, in AM at least 3 services do similar calls, as I've noticed. BTW, will the in-memory entry be actual?

subprojects/frameworklauncher/yarn/src/main/java/com/microsoft/frameworklauncher/service/Service.java
zhiltsov-max: Check the comment in [dockerContainerScript](src/rest-server/src/templates/dockerContainerScript.mustache) for explanations - the same reason we add empty handlers in a user command script. It should not break the generality, as it makes the difference between sending a signal only to the main container process or to its process group (YARN sends to a foreground group like bash does in the interactive mode). But yes, I'd rather used `exec "..."` in auto-generated Docker entry point in YARN without of `bash -c "..."` as it does not spawn an intermediate shell:

``` bash
exec /bin/bash -c "trap ' ' TERM; $JAVA_HOME/bin/java -DLAUNCHER_LOG_DIR=$LAUNCHER_LOG_DIR -Xmx256m com.microsoft.frameworklauncher.applicationmaster.Bootstrap 1>$LAUNCHER_LOG_DIR/stdout 2>$LAUNCHER_LOG_DIR/stderr "
```

So I'm open for better solutions. Without this handler, YARN task container just exits immediately.

subprojects/frameworklauncher/yarn/src/main/java/com/microsoft/frameworklauncher/applicationmaster/ApplicationMaster.java
zhiltsov-max: No, in the case of any errors, old behaviour is saved.

subprojects/frameworklauncher/yarn/src/main/java/com/microsoft/frameworklauncher/applicationmaster/ApplicationMaster.java
zhiltsov-max: We do.

subprojects/frameworklauncher/yarn/src/main/java/com/microsoft/frameworklauncher/applicationmaster/FrameworkInfoPublisher.java
yqwang-ms: But why zkStore can be unavailable? When we stop AM, we stop the zkStore at last.

subprojects/frameworklauncher/yarn/src/main/java/com/microsoft/frameworklauncher/applicationmaster/ApplicationMaster.java
yqwang-ms: OK but you forget to change the comment.

subprojects/frameworklauncher/yarn/src/main/java/com/microsoft/frameworklauncher/service/Service.java
yqwang-ms: let's keep the default YARN behavior, otherwise, it is better to contribute "trap ' ' TERM" to YARN code and make a flag there.

subprojects/frameworklauncher/yarn/src/main/java/com/microsoft/frameworklauncher/applicationmaster/FrameworkInfoPublisher.java
zhiltsov-max: Yes, it happened during the stopping as I described above. The impact is ~1-10 Gb of useless error spam in AM logs during the stop.

yqwang-ms: If so, maybe we can skip to call zkStore.stop() (it is the old behaiour and works fine)

zhiltsov-max:
@yqwang-ms, please read my comments above and here. 

> Do we have to change Launcher code to achieve your goal?

Probably yes, since Launcher is responsible for building the command to be executed in Docker container. YARN is also responsible as it adds an extra shell to the one Docker starts with. To resolve the issue, we have either:
- to remove `bash -c` and install a TERM handler in YARN Docker command config
- to remove `bash -c` and install a TERM handler in Launcher command

Currently, both traps are required. If you think it is simpler to modify YARN - I have nothing against.

> What is the benefit?

Users can opt in to save their work (long running training) before the container stop by a signal. `Caffe` framework does this, for instance. Python training scripts also can do this.

yqwang-ms:
I mean "What is the benefit to change Launcher code". 

After app is killed (Launcher Framework Stopped), YARN can already graceful kill containers by sending signal first then force kill. So all container can catch this signal and graceful shutdown. 

It is reasonable to let task container to graceful shutdown, such as saving checkpoint etc, but we do not need AM container to graceful shutdown, because LauncherAM can tolerate unexpected shutdown, it does not need graceful shutdown.

zhiltsov-max:
@yqwang-ms, 

> YARN can already graceful kill containers

Yes, and this is used here, actually. I've just checked one more time - the only reason for the change in Launcher is graceful stop of AM. If AM does not need to unregister in this case, I think there are no reasons left to modify AM or Launcher, like you are suggesting. The only reason to stop gracefully for AM is just a correct stop then. 

What I'm not sure about is the difference between `yarn.nodemanager.sleep-delay-before-sigkill.ms` and `yarn.nodemanager.process-kill-wait.ms`.

yqwang-ms:
NM try to kill container pipeline:
1. getContainerPid (This can fail, so retry for max yarn.nodemanager.process-kill-wait.ms, but usually this can succeed immediately)
2. TERM the container identified by the pid
3. Wait yarn.nodemanager.sleep-delay-before-sigkill.ms
4. KILL the container identified by the pid

yqwang-ms:
So I think your PR only need to take care of yarn.nodemanager.sleep-delay-before-sigkill.ms

fanyangCS:
Might be a good idea to increase the value of sleep before sigkill

zhiltsov-max:
@yqwang-ms, removed AM graceful stop as it is not very important.

@fanyangCS, what do you think the default value should be? This value is a cluster administrator responsibility, it could probably have a more natural way for its modification - through the cluster model, for example.

yqwang-ms:
@zhiltsov-max For Launcher, it would be better if you could remove all callers of zkStore.stop(); instead of just move around publishFrameworkInfo, i.e. remove all
image

Thanks!

yqwang-ms:
Hi @zhiltsov-max, to unblock the launcher fix, I have moved it to another PR:
https://github.com/Microsoft/pai/pull/2435
Thanks for reporting and the proposal solution!

src/rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -299,6 +327,7 @@ docker pull {{ jobData.image }} \
 ## network consumption
 docker run --name $docker_name \
   --init \
+  --detach \
```
mzmssg: We still think removing `--rm` to achieve oom checker is too costly.

mzmssg: Other change is fine.

subprojects/frameworklauncher/yarn/src/main/java/com/microsoft/frameworklauncher/applicationmaster/FrameworkInfoPublisher.java
zhiltsov-max: Should be done in another PR.

yqwang-ms: @zhiltsov-max, to unblock the launcher fix, I have moved it to another PR:
#2435
Thanks for reporting and the proposal solution!

src/rest-server/src/templates/yarnContainerScript.mustache
zhiltsov-max: In this PR it is only used to determine what is (or was) the container `cgroup` to scan `dmesg` log for that value. I'm open for better solutions preserving `--rm`. Technically, we can write `cgroup` name to a file from the task container and the read it in exit handler.

mzmssg: Agree, I think we need to implement it in this PR as well, given that currently PAI has no service to reap trash.

mzmssg:
We add yarn container graceful stop in #2536. Big thanks for your contribution.

fanyangCS:
Implemented in k8s version

fanyangCS:
Implemented in k8s version

PR #2412, opened 2019-03-26
 v0.11 upgrade process

mzmssg:
Including:
- change template tag to v0.11.0
- update configMigration.py to support 0.10 -> 0.11
- add plugin migration docs.


deployment/tools/pluginIdMigration.py
```diff
@@ -31,14 +31,11 @@
             plugin["id"] = slugify(plugin["title"])
 else:
     print "No webportal plugins to convert, exit..."
-    sys.exit(0)
+    sys.exit(-1)
```
wangdian: I suggest to overwrite whether the file will change or not.

mzmssg: Done

Issue #2385, opened 2019-03-25
add resource usage monitor for external processes


xudifsd:
**What would you like to be added**:
Job-exporter currently monitored both task and service resource usage, but it ignores other process's resource usage, some bugs in these programs can lead to OOM or consuming too much CPU, we need to monitor all external processes to be able to find the root causes of node problem and trigger alerts.

**Why is this needed**:
External processes may also jeopardize the stability of the system. And with alert, the admin can do something before it's too late.

**Without this feature, how does the current module work**:
Wait for other alerts like service not up or node not up to know something is wrong.

**Components that may involve changes**:
job-exporter

PR #2419, opened 2019-03-27
monitor process memory consumption and alert for om[is]agent

xudifsd:
Get process memory consumption by ps rss, report those who use more than 500M to save space in prometheus. Also defined an alert rule for process `omiagent` and `omsagent`, these two processes are frequently causing OOM in azure VM, which DRI should take care of.

### DIFF
--- a/src/job-exporter/src/collector.py
+++ b/src/job-exporter/src/collector.py
@@ -24,6 +24,7 @@ import subprocess
 import time
 import copy
 import os
+import collections
 
 from prometheus_client import make_wsgi_app, Counter, Gauge, Histogram
 from prometheus_client.core import GaugeMetricFamily
@@ -84,6 +85,10 @@ def gen_gpu_used_by_zombie_container_counter():
             "count of gpu used by zombie container",
             labels=["minor_number", "container_id"])
 
+def gen_process_mem_usage_gauge():
+    return GaugeMetricFamily("process_mem_usage_byte",
+            "memory usage of process, to save space in prometheus, we only expose those who consume more than 500Mb of memory",
+            labels=["pid", "cmd"])
 
 class ResourceGauges(object):
     def __init__(self):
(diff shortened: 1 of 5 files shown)


xudifsd:
fixed #2385

src/job-exporter/src/ps.py
```diff
@@ -18,41 +18,46 @@
...
 
         return parse_result(ps_output)
     except subprocess.CalledProcessError as e:
         logger.exception("command '%s' return with error (code %d): %s",
                 e.cmd, e.returncode, e.output)
     except subprocess.TimeoutExpired:
-        logger.warning("ps aux timeout")
+        logger.warning("ps ax timeout")
     except Exception:
         logger.exception("exec ps aux error")
```
mzmssg: ps ax

src/job-exporter/src/ps.py
```diff
@@ -18,42 +18,47 @@
...
         logger.exception("command '%s' return with error (code %d): %s",
                 e.cmd, e.returncode, e.output)
     except subprocess.TimeoutExpired:
-        logger.warning("ps aux timeout")
+        logger.warning("ps ax timeout")
     except Exception:
-        logger.exception("exec ps aux error")
+        logger.exception("exec ps ax error")
 
     return None
```
mzmssg: `return []`?

src/job-exporter/src/collector.py
```diff
@@ -768,15 +773,31 @@ def __init__(self, name, sleep_time, atomic_ref, iteration_counter):
         Collector.__init__(self, name, sleep_time, atomic_ref, iteration_counter)
 
     def collect_impl(self):
-        process = ps.get_zombie_process(ProcessCollector.cmd_histogram,
+        process_info = ps.get_process_info(ProcessCollector.cmd_histogram,
                 ProcessCollector.cmd_timeout)
 
-        if len(process) > 0:
+        if len(process_info) > 0:
```
mzmssg: Might get error if `get_process_info` return `None`.

src/job-exporter/src/collector.py
```diff
@@ -768,15 +773,31 @@ def __init__(self, name, sleep_time, atomic_ref, iteration_counter):
...
+                    if "nvidia-smi" in info.cmd:
+                        # override command name to make alert rule easier
+                        zombie_count["nvidia-smi"] += 1
+                    else:
+                        cmd = info.cmd.split()[0] # remove args
+                        zombie_count[cmd] += 1
+
+                if info.rss > 500 * 1024 * 1024:
+                    # only record large memory consumption to save space in prometheus
+                    cmd = info.cmd.split()[0] # remove args
```
mzmssg: Here we strip the command details, so most commands will be `bash` or `python`, count them seems meaningless.
How about excluding such processes

xudifsd: I removed args because omiagent will also have args like `/opt/omi/bin/omiagent 9 10 --destdir / --providerdir /opt/omi/lib --loglevel WARNING`, I think the args is useless.

mzmssg: Why not filter omiagent here, instead of in alert rules.

xudifsd: that will be a special case in code

Issue #2228, opened 2019-02-28
Add TotalGpuCount and TotalTaskCount into job list


yqwang-ms:
**What would you like to be added**:
Add more info into job list, such as TotalGpuCount, TotalTaskCount, etc.

**Why is this needed**:
User can find jobs easiliy with these more info in list.

**Without this feature, how does the current module work**：
Check into job detail info one by one.

**Components that may involve changes**:
web portal, launcher, rest server.

--------------
I'm deferring the "TotalTaskRoleCount" ask from this item, as it seems not that critical to be known in a glance. let's hold this ask to later.

yqwang-ms:
Please help to propose these additional info that you want to add.
Current plan is to add:
TotalGpuCount in Request, TotalTaskCount in Request.

fanyangCS:
maybe to show a meta data field.

yqwang-ms:
@fanyangCS 
What meta data field do you mean? 
Is frameworkDescription below (it can be any string) enough?
![image](https://user-images.githubusercontent.com/32826762/53543433-a5225a00-3b5d-11e9-9ccd-4095f5b61af1.png)

fanyangCS:
Yes. what is the max length? 512?

yqwang-ms:
Yes, 512, it should be enough at least for job list.

yqwang-ms:
@fanyangCS do you need more? Or could you please help to involve PAIExp team?

yqwang-ms:
@fanyangCS I increased the description len to 1024

scarlett2018:
Rest is ready for UX.

PR #2435, opened 2019-03-28
[Launcher]: Add more info into SummarizedFrameworkInfo

yqwang-ms:
1. Add more info into SummarizedFrameworkInfo
https://github.com/Microsoft/pai/issues/2228

2. Skip to stop ZKStore in case it is still using
https://github.com/Microsoft/pai/pull/2408


yqwang-ms:
Example:
image

mzmssg:
We only provide total GPU number, without cpu or memory?

yqwang-ms:
Good point, let me added it.

mzmssg:
@yqwang-ms 
Another question, will it count AM resource

yqwang-ms:
It seems not easy to judge this. So, let's only add current fields, the memory and cpu is TBD in future release.

PR #2438, opened 2019-03-28
send resolved email and make user can config repeat interval

xudifsd:
@squirrelsc think there are too many same alert email, so maybe should enlarge the interval of same alerts, I make it as configurable and default to our old value.

At the same time, it would be useful to send resolved email to notify problem solved.

### DIFF
--- a/src/alert-manager/config/alert-manager.md
+++ b/src/alert-manager/config/alert-manager.md
@@ -25,6 +25,8 @@ alert-manager:
 
 In addition, if you deployed pai behind firewall, you should configure alert-manager with `use-pylon: True`, to make url from alert email public available.
 
+Also you can control the interval of sending same alert email if problem unsolved, the default interval is 4h, you can shorten it to 30m or enlarge it to 24h, by adding `repeat-interval: 30m` or `repeat-interval: 24h` undert alert-manager config.
+
 #### Generated Configuration <a name="G_Config"></a>
 
 After parsing, if you configured the alert-manager the model will be like:
@@ -40,6 +42,7 @@ alert-manager:
     host: master_ip
     url: "http://master_ip:9093"
     use-pylon: False
+    repeat-interval: 4h
 ```
 
 if you didn't configured alert-manager, it will be like:
@@ -50,4 +53,5 @@ alert-manager:
     host: master_ip
     url: "http://master_ip:9093"
     use-pylon: False
+    repeat-interval: 4h
 ```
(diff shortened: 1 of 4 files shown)


src/alert-manager/config/alert-manager.md
```diff
@@ -50,4 +53,5 @@ alert-manager:
     host: master_ip
     url: "http://master_ip:9093"
     use-pylon: False
+    repeat-interval: 4h
```
squirrelsc: we can extend it to 24h, and depends on resolved mails.

src/alert-manager/deploy/alert-configmap.yaml.template
```diff
@@ -18,12 +18,13 @@ data:
...
       group_interval: 5m
-      repeat_interval: 4h
+      repeat_interval: {{ cluster_cfg["alert-manager"]["repeat-interval"] }}
       group_by: [alertname]
     receivers:
     - name: "pai-alert"
       email_configs:
         - to: {{ alert_info["receiver"] }}
+          send_resolved: true
           html: '{{ "{{" }} template "email.pai.html" . {{ "}}" }}'
```
squirrelsc: Let's make sure resolve mail have a green title.

src/alert-manager/config/alert-manager.md
xudifsd: yes, I provide this as a config, user who want to extend can extend when deploying.

src/alert-manager/deploy/alert-configmap.yaml.template
xudifsd: I do not know how to pass this into template to generate a different email.

src/alert-manager/deploy/alert-templates/pai.tmpl
```diff
@@ -43,7 +43,11 @@ SOFTWARE.
       <div style="font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif; box-sizing: border-box; font-size: 14px; max-width: 600px; display: block; margin: 0 auto; padding: 0;">
         <table width="100%" cellpadding="0" cellspacing="0" style="font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif; box-sizing: border-box; font-size: 14px; border-radius: 3px; background-color: #fff; margin: 0; border: 1px solid #e9e9e9;" bgcolor="#fff">
           <tr>
+            {{ if gt (len .Alerts.Firing) 0 }}
```
squirrelsc: please double check firing and resolving won't be in the same mail.

xudifsd: Yes, I've already checked.

Issue #2361, opened 2019-03-20
Clone jobs' name is too long to be used


scarlett2018:
When job is clonned, the default job name append with a timestamp currently, when you clone a job several times, the name is too long to be used.

Proposed change: 
option 1: only change the timestamp
option 2: any other options that can make the name more normal

![image](https://user-images.githubusercontent.com/39592018/54665080-f56b5700-4b20-11e9-9959-4c284329eee1.png)

PR #2451, opened 2019-03-29
[Web Portal] Replace the suffix if a cloned job is resubmited

sunqinzheng:
#2361 
- add checksum to cloned job's name suffix
- prepare for job config api change (fix)

### DIFF
--- a/src/webportal/src/app/job/job-submit/job-submit.component.js
+++ b/src/webportal/src/app/job/job-submit/job-submit.component.js
@@ -28,7 +28,6 @@ const userAuth = require('../../user/user-auth/user-auth.component');
 const jobSchema = require('./job-submit.schema.js');
 const querystring = require('querystring');
 const stripJsonComments = require('strip-json-comments');
-const uuid = require('uuid');
 
 const jobSubmitHtml = jobSubmitComponent({
   breadcrumb: breadcrumbComponent,
@@ -39,11 +38,10 @@ let editor;
 let jobDefaultConfig;
 
 const getChecksum = (str) => {
-  const buffer = Buffer.from(str);
   let res = 0;
-  buffer.forEach((x) => {
-    res = res ^ x;
-  });
+  for (const c of str) {
+    res^= c.charCodeAt(0) & 0xff;
+  }
   return res.toString(16);
 };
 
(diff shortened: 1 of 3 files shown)


Gerhut:
Hi @sunqinzheng 

Since [Buffer](https://nodejs.org/api/buffer.html) is a node.js only feature, and ES2015 standard have [TypedArray](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray) which provides similar functionalities. Shall we use TypedArray instead, to follow more to standard?

src/webportal/src/app/job/job-submit/job-submit.component.js
```diff
@@ -28,6 +28,7 @@ const userAuth = require('../../user/user-auth/user-auth.component');
 const jobSchema = require('./job-submit.schema.js');
 const querystring = require('querystring');
 const stripJsonComments = require('strip-json-comments');
+const uuid = require('uuid');
```
Gerhut: Have not seen the `uuid` included in `package.json`, and I don't think uuid is required here. In browser side, uuid just uses crypto level random number instead of `Math.random()` to avoid more collisions, since we have a directly response when user is submitting with the same job name, so I don't think avoid more collisions here is important.

sunqinzheng:
Webpack uses [feross/buffer](https://github.com/feross/buffer) v4 as nodejs Buffer's browser implementation if node.Buffer is set to true. (https://github.com/webpack/node-libs-browser)

feross/buffer uses Uint8Array by default and object as shim.

Gerhut:
> Webpack uses feross/buffer v4 as nodejs Buffer's browser implementation if node.Buffer is set to true. (https://github.com/webpack/node-libs-browser)
> feross/buffer uses Uint8Array by default and object as shim.
> TextEncoder is another solution if we just want to convert a string to uint8array.

I see and think that it is not necessary to introduce such a dependency.

sunqinzheng:
OK, I'll use TextEncoder instead

Gerhut:
> OK, I'll use TextEncoder instead

Please pay attention to MS Edge compatibility since encoding standard is still [in development](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/encodingstandard/)

src/webportal/src/app/job/job-submit/job-submit.component.js
sunqinzheng: fixed

sunqinzheng:
Since we only need to calc a checksum, I'll use the last byte of the charcter code

PR #2461, opened 2019-03-29
Add protocol submission plugin

abuccts:
Add protocol submission plugin:
* import protocol yaml file
* view/edit protocol yaml file with monaco editor

Known issues:
* monaco editor cannot be bundled by webpack for plugin (fixed)
* editor worker has the wrong base url (fixed)
* language highlight and some editor styles do not work in plugin (fixed)


abuccts:
image
image

abuccts:
Monaco editor webpack related issues can be found at [monaco-editor-webpack-plugin issue](https://github.com/Microsoft/monaco-editor-webpack-plugin/issues). It seems the monaco-editor-webpack-plugin is not supported anymore.

If we cannot solve those issues, here're some options:
* Implement protocol submission inside webportal instead of a plugin.
* Change to another module bundler to replace webpack.
* Use another light weight editor.

@Gerhut @sunqinzheng Could you please help to have a look?

contrib/submit-protocol/webpack.config.ts
```diff
@@ -0,0 +1,66 @@
...
+
+import { resolve } from "path";
+import { Configuration } from "webpack";
+import MonacoWebpackPlugin from "monaco-editor-webpack-plugin";
+
+const configuration: Configuration = {
+  context: resolve(__dirname, "src"),
+  entry: {
+    "plugin": "./index.ts",
+    "editor.worker": "monaco-editor/esm/vs/editor/editor.worker.js",
```
sunqinzheng: I think `editor.worker` is unnecessary if monaco-webpack-plugin is used. https://github.com/Microsoft/monaco-editor/blob/master/docs/integrate-esm.md

contrib/submit-protocol/src/App/ProtocolForm.tsx
```diff
@@ -0,0 +1,227 @@
...
+              </div>
+            </form>
+          </div>
+        </Fabric>
+      </>
+    );
+  }
+
+  private setJobName = (jobName: string) => {
+    const protocol = this.state.protocol;
```
sunqinzheng: React state is immutable. Please use `lodash.cloneDeep` or something else to get a clone.
https://medium.freecodecamp.org/handling-state-in-react-four-immutable-approaches-to-consider-d1f5c00249d5

contrib/submit-protocol/src/App/ProtocolForm.tsx
```diff
@@ -0,0 +1,227 @@
...
+    const source = this.props.source;
+    if (source && source.jobName && source.user) {
+      window.fetch(
+        `${this.props.api}/api/v1/user/${source.user}/jobs/${source.jobName}/config`
+      ).then((res) => {
+        return res.json();
+      }).then((body) => {
+        const protocol = jsyaml.safeLoad(body);
+        this.setState({ protocol });
+        this.setJobName(`${source.jobName}_clone_${crypto.randomBytes(4).toString("hex")}`);
```
sunqinzheng: React `setState` is an async operation. So `setJobName` may not get the `protocol` set here. You can use a setState callback to get the correct `protocol`.
https://reactjs.org/docs/react-component.html#setstate

contrib/submit-protocol/src/App/ProtocolForm.tsx
```diff
@@ -0,0 +1,227 @@
...
+    this.state = {
+      jobName: "",
+      protocol: Object.create(null),
+      protocolYAML: "",
+      showEditor: false,
+    };
+
+    const source = this.props.source;
+    if (source && source.jobName && source.user) {
+      window.fetch(
```
sunqinzheng: Side effects (async http request) should be put in componentDidMount

contrib/submit-protocol/src/App/ProtocolForm.tsx
```diff
@@ -0,0 +1,227 @@
...
+              theme="vs-dark"
+              options={{ wordWrap: "on", readOnly: false }}
+            />
+            <div style={{ marginTop: "15px" }}>
+              <PrimaryButton text="Save" onClick={this.saveEditor} style={{ marginRight: "10px" }}/>
+              <DefaultButton text="Discard" onClick={this.discardEditor} />
+            </div>
+          </Panel>
+          <div className="content">
+            <form className="modal-dialog">
```
sunqinzheng: I think these class names are from bootstrap. But it seems the plugin doesn't require bootstrap now.

contrib/submit-protocol/src/index.ts
```diff
@@ -0,0 +1,56 @@
...
+  public connectedCallback() {
+    const api = this.getAttribute("pai-rest-server-uri") as string;
+    const user = this.getAttribute("pai-user");
+    const token = this.getAttribute("pai-rest-server-token");
+    if (user === null || token === null) {
+      window.location.href = "/login.html";
+      return;
+    }
+
+    const query = parse(window.location.search.replace(/^\?/, ""));
```
sunqinzheng: I prefer using https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams here

contrib/submit-protocol/src/App/ProtocolForm.tsx
```diff
@@ -0,0 +1,227 @@
...
+    }
+    window.fetch(`${this.props.api}/api/v2/jobs`, {
+      body: this.state.protocolYAML,
+      headers: {
+        "Authorization": `Bearer ${this.props.token}`,
+        "Content-Type": "text/yaml",
+      },
+      method: "POST",
+    }).then((res) => {
+      return res.json();
```
sunqinzheng: Maybe you need to parse the result manually if yaml is returned.

contrib/submit-protocol/webpack.config.ts
abuccts: Removed.

contrib/submit-protocol/src/App/ProtocolForm.tsx
abuccts: Updated, use [react `update`](https://reactjs.org/docs/update.html).

contrib/submit-protocol/src/App/ProtocolForm.tsx
abuccts: Updated.

contrib/submit-protocol/src/index.ts
abuccts: Updated.

contrib/submit-protocol/src/App/ProtocolForm.tsx
abuccts: The result is a json response like:
```js
{
  "status": "202",
  "messgae": "..."
}
```
Only the config file from user is yaml.

contrib/submit-protocol/src/App/ProtocolForm.tsx
abuccts: Updated.

contrib/submit-protocol/src/App/ProtocolForm.tsx
abuccts: Right, the form uses some styles from webportal bootstrap css.

The goal of this PR is an end-to-end protocol submission pipeline. The form UI is just a prototype without experience design.

contrib/submit-protocol/src/App/ProtocolForm.tsx
```diff
@@ -0,0 +1,246 @@
...
+          ),
+        );
+      }).catch((err) => {
+        window.alert(err.message);
+      });
+    }
+  }
+
+  componentDidMount() {
+    this.submitProtocol;
```
sunqinzheng: I think `submitProtocol` here should be a function call...

contrib/submit-protocol/src/App/ProtocolForm.tsx
sunqinzheng: I prefer using bootstrap css modules here.
We can't assume that bootstrap will be loaded in web portal in the future.

contrib/submit-protocol/src/App/ProtocolForm.tsx
sunqinzheng: Fetch in constructor should be moved to componentDidMount as well...

abuccts: only move `fetch` from constructor to `componentDidMount`, which is called before `render`.

`fetch` in `submitProtocol` is called when submit button is clicked, so I removed from `componentDidMount`.

contrib/submit-protocol/src/App/ProtocolForm.tsx
abuccts: Added.

contrib/submit-protocol/src/App/ProtocolForm.tsx
abuccts: Updated.

contrib/submit-protocol/src/App/ProtocolForm.tsx
abuccts: added loading spinner.

contrib/submit-protocol/src/App/ProtocolForm.tsx
abuccts: changed to `react-bootstrap`.

abuccts: updated to css modules.

contrib/submit-protocol/src/App/ProtocolForm.tsx
```diff
@@ -0,0 +1,373 @@
...
+        `${this.props.api}/api/v1/user/${source.user}/jobs/${source.jobName}/config`,
+      ).then((res) => {
+        return res.json();
+      }).then((body) => {
+        const protocol = jsyaml.safeLoad(body);
+        this.setState(
+          { protocol },
+          () => this.setJobName(
+            null as any,
+            `${source.jobName}_clone_${crypto.randomBytes(4).toString("hex")}`,
```
Gerhut: Please use Math.random directly since the crypto level random number is not required here, and reduce the bundle size.

contrib/submit-protocol/src/App/ProtocolForm.tsx
```diff
@@ -0,0 +1,373 @@
...
+  token: string;
+  source ?: {
+    jobName: string;
+    user: string;
+  };
+}
+
+interface IProtocolState {
+  jobName: string;
+  protocol: {[key: string]: string | object};
```
Gerhut: If you don't want to manage the protocol schema here, just use `any`

contrib/submit-protocol/src/App/ProtocolForm.tsx
```diff
@@ -0,0 +1,373 @@
...
+  }
+
+  public componentDidMount() {
+    this.fetchConfig();
+  }
+
+  public render() {
+    if (this.state.loading) {
+      return (
+        <>
```
Gerhut: Why an extra Fragment?

contrib/submit-protocol/src/App/ProtocolForm.tsx
```diff
@@ -0,0 +1,373 @@
...
+  loading: boolean;
+  showParameters: boolean;
+  showEditor: boolean;
+}
+
+export default class ProtocolForm extends React.Component<IProtocolProps, IProtocolState> {
+  constructor(props: IProtocolProps) {
+    super(props);
+
+    this.state = {
```
Gerhut: Use `public state = {}` out of the constructor instead.

contrib/submit-protocol/src/App/ProtocolForm.tsx
Gerhut: If there are 2 independent condition branch, use 2 (private) render methods instead like `renderLoading` and `renderContent`

contrib/submit-protocol/src/App/ProtocolForm.tsx
```diff
@@ -0,0 +1,373 @@
...
+          <Fabric>
+            <div className={bootstrapStyles.container}>
+              <div className={bootstrapStyles.modalDialog}>
+                <div className={bootstrapStyles.modalContent}>
+                  <div className={bootstrapStyles.modalHeader}>
+                    <h3 className={bootstrapStyles.modalTitle}>
+                      Submit Job v2 <small>Protocol Preview</small>
+                    </h3>
+                  </div>
+                  <div className={`${bootstrapStyles.modalBody} ${bootstrapStyles.row}`}>
```
Gerhut: Use classnames package instead

contrib/submit-protocol/src/App/ProtocolForm.tsx
abuccts: updated

contrib/submit-protocol/src/App/ProtocolForm.tsx
abuccts: updated

contrib/submit-protocol/src/App/ProtocolForm.tsx
abuccts: updated

contrib/submit-protocol/src/App/ProtocolForm.tsx
abuccts: updated

contrib/submit-protocol/src/App/ProtocolForm.tsx
abuccts: updated

contrib/submit-protocol/src/App/ProtocolForm.tsx
```diff
@@ -0,0 +1,367 @@
...
+  token: string;
+  source ?: {
+    jobName: string;
+    user: string;
+  };
+}
+
+interface IProtocolState {
+  jobName: string;
+  protocol: {[key: string]: any};
```
Gerhut:
```suggestion
  protocol: any;
```

contrib/submit-protocol/src/App/ProtocolForm.tsx
```diff
@@ -0,0 +1,367 @@
...
+    jobName: "",
+    protocol: Object.create(null),
+    protocolYAML: "",
+    loading: true,
+    showParameters: false,
+    showEditor: false,
+  };
+
+  constructor(props: IProtocolProps) {
+    super(props);
```
Gerhut: No constructor need here

contrib/submit-protocol/package.json
```diff
@@ -0,0 +1,55 @@
...
+    "immutability-helper": "^3.0.0",
+    "js-yaml": "^3.13.0",
+    "monaco-editor": "^0.16.2",
+    "office-ui-fabric-react": "^6.162.1",
+    "react": "^16.8.4",
+    "react-dom": "^16.8.4",
+    "react-monaco-editor": "^0.25.1"
+  },
+  "devDependencies": {
+    "@types/classnames": "^2.2.7",
```
Gerhut: I've found typing files in classnames package itself: https://github.com/JedWatson/classnames/blob/master/index.d.ts

contrib/submit-protocol/package.json
```diff
@@ -0,0 +1,55 @@
...
+    "@types/immutability-helper": "^2.6.3",
```
Gerhut: Also immutability-helper https://github.com/kolodny/immutability-helper/blob/master/index.d.ts

contrib/submit-protocol/package.json
abuccts: the latest [classnames version](https://www.npmjs.com/package/classnames) (10 months ago) does not contain typing files.

contrib/submit-protocol/src/App/ProtocolForm.tsx
abuccts: removed

contrib/submit-protocol/package.json
abuccts: removed

Issue #2127, opened 2019-02-02
Need usage report


scarlett2018:
Job Summary, User Usage Summary, VC Usage Summary.

By time, by status

xudifsd:
These info can be get from launcher from `:9086/v1/Frameworks`, if combined info from #2073 , we can calculated resource that wasted(the job finally failed).

scarlett2018:
> These info can be get from launcher from `:9086/v1/Frameworks`, if combined info from #2073 , we can calculated resource that wasted(the job finally failed/killed).

Would you like to merge #2073 with this item? what's the estimation to have both #2127 and #2073 in place? Let's combine them if it makes tracking easier.

xudifsd:
No, I think #2073 is relatively easy to implement, but this requires more efforts. Let's track them in different issues.

scarlett2018:
There are some experiment for job dashboard done in PowerBI, it's time to revisit, whether these things work well for v1.x. And whether there are any new needs to better understand the overall job utilization.

Issue #2449, opened 2019-03-29
place prometheus data in host path to avoid data loss during prometheus upgrade/restart


xudifsd:
Current prometheus data are place in container space, so whenever prometheus get restarted, all time series data will lost.

Issue #2073, opened 2019-01-24
VC usage (activity) reports


scarlett2018:
As we are suggesting to use VC to organize functional group teams, the operation team needs to know how well users/jobs in each VC is.
i.e. need a daily/weekly/monthly report for each team's VC usage, including jobs, gpu, memory.
i.e. need to know the individual's activity in a VC. who use the most resources in the vc for example
[To be evaluated the value ] ~i.e. need alert and email notification for VC loads~
i.e. would like to have report summary send to admin through email

xudifsd:
Do we need to alert on VC loads? I think this is kind of abuse the alert email, since this should a normal situation, and admin can do nothing about it.

Maybe we can use power BI to implement the report, it's not possible to send report via alert rule, and if we want to generate a report we can write another services, but I think BI maybe more appropriate.

xudifsd:
Two problems with monthly report:
* prometheus currently only retain 15 days data
* we do not save prometheus data in host path, so all data will be lost in prometheus redeploy

We can prolong the retention date but may incur too much disk consumption, will need to investigate.

scarlett2018:
> Do we need to alert on VC loads? I think this is kind of abuse the alert email, since this should a normal situation, and admin can do nothing about it.
> 
> Maybe we can use power BI to implement the report, it's not possible to send report via alert rule, and if we want to generate a report we can write another services, but I think BI maybe more appropriate.

Sounds reasonable, just updated the original request and mark that item as to be evaluated. Let's pending that one, and gather more feedbacks about whether it is needed.

scarlett2018:
> Two problems with monthly report:
> 
> * prometheus currently only retain 15 days data
> * we do not save prometheus data in host path, so all data will be lost in prometheus redeployment
> 
> We can prolong the retention date but may incur too much disk consumption, will need to investigate.

All good pionts, my quick thinking is we should figure out a way to persist usage and log related histories. Please investigate and get @fanyangCS and @sterowang 's technical suggestions accordingly.

xudifsd:
user/vc's resource usage can be scrapped from yarn's api `/ws/v1/cluster/apps`, it has application's resource usage info and final status. But this API will only retain specific number of entries, like 1000 entries, so we need another service to periodically get this info and persist somewhere for later report use.

xudifsd:
[...]

PR #2507, opened 2019-04-04
add script to generate reports for OpenPai cluster

xudifsd:
generate csv reports for OpenPai cluster.

Document is in the header of the script.

### DIFF
--- a/docs/tools/how-to-setup-report-script.md
+++ b/docs/tools/how-to-setup-report-script.md
@@ -6,7 +6,7 @@ But since not everyone is interested in this report, we do not maintain such a s
 
 ## What the script will report
 
-The reports will be 3 csv files `job.csv`, `alert.csv` and `raw_job.csv`.
+The reports will be 4 csv files `job.csv`, `alert.csv`, `raw_job.csv` and `gpu.csv`.
 
 ### job.csv
 
@@ -62,6 +62,16 @@ The file have following columns:
 * memory allocated: how much memory(GB) allocated to the job, this include the memory allocated to app master
 * max memory usage: maximum memory(GB) usage of this job
 * gpu allocated: how many gpu card allocated to the job
+
+### gpu.csv
+
+This file is about all gpu util info in cluster.
+
+The file have following columns:
+
+* host_ip: where this gpu installed
+* gpu_id: gpu minor number in the node
+* avg: avg utils during the time frame
 
 ## Prerequisite
 
@@ -153,6 +163,7 @@ With http server setup, you can now get those reports with the same name of csv 
 http://$IP:10240/job
 http://$IP:10240/raw_job
 http://$IP:10240/alert
+http://$IP:10240/gpu
 ```
 
 These end point all accept `span` argument, you can provide with value: `day`, `week` or `month`, which will generate report in that time span. The default span is week.
--- a/docs/tools/how-to-setup-report-script.md
+++ b/docs/tools/how-to-setup-report-script.md
@@ -71,14 +71,14 @@ The file have following columns:
 
 ## Prerequisite
 
-You should prepare a node that have access to OpenPai services, the script will need to access hadoop-resource-manager, framework-launcher and Prometheus deployed by OpenPai. This node do not need to have much memory resource and do not need to have GPU cards. You only need to make sure this node will not restart frequently. Usualy the master node of the OpenPai cluster is a good choice.
+You should prepare a node that have access to OpenPai services, the script will need to access hadoop-resource-manager, framework-launcher and Prometheus deployed by OpenPai. This node do not need to have much memory resource and do not need to have GPU cards. You only need to make sure this node will not restart frequently. Usually the master node of the OpenPai cluster is a good choice.
 
 After you choose a node, please make sure you have following software installed:
 
 * python3
 * requests library
 
-If your node is ubuntu node you can install these software using following commands:
+If your node is ubuntu node, you can install these software using following commands:
 
 ``` sh
 sudo apt-get install -y python3 python3-pip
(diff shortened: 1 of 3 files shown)


src/tools/reports.py
```diff
@@ -0,0 +1,672 @@
...
+    """ for example a=[{"a": {"b": 2}}]
+    walk_json_field_safe(a, 0, "a", "b") will get 2
+    walk_json_field_safe(a, 0, "not_exist") will get None
+    """
+    try:
+        for f in fields:
+            obj = obj[f]
+        return obj
+    except:
+        return None
```
mzmssg: suggest move "return None" out of "except", or in "finally"

src/tools/reports.py
```diff
@@ -0,0 +1,672 @@
...
+import sys
+
+import sqlite3
+import requests
+
+logger = logging.getLogger(__name__)
+
+
+def walk_json_field_safe(obj, *fields):
+    """ for example a=[{"a": {"b": 2}}]
```
mzmssg: If `a=a=[{"a": {"b": "asdf"}}]`, should `walk_json_field_safe(a, 0, "a", "b", 1)` return `None` or `s`?

src/tools/reports.py
```diff
@@ -0,0 +1,672 @@
...
+        for f in fields:
+            obj = obj[f]
+        return obj
+    except:
+        return None
+
+
+def request_with_error_handling(url):
+    try:
+        response = requests.get(url, allow_redirects=True)
```
mzmssg: `response.raise_for_status()` seems better, no need to check status_code.

src/tools/reports.py
```diff
@@ -0,0 +1,672 @@
...
+                                (framework["finished_time"], framework["content"], framework["name"]))
+            else:
+                cursor.execute("""INSERT INTO frameworks(name,finished_time,content)
+                                VALUES(?,?,?)""",
+                                (framework["name"], framework["finished_time"], framework["content"]))
+
+        db.conn.commit()
+
+
+def get_vc_report(days):
```
mzmssg: I think passing "begin_date" and "end_date" is more flexible. Because more common request might be monthly report such as March report, not the past 30 days report.

mzmssg:
Should we add a document?

src/tools/reports.py
xudifsd: `s`

src/tools/reports.py
xudifsd: I'm not sure, we will delete data older than 31 days, can not generate that report. Will need to check if this is required.

mzmssg: Yes, +@scarlett2018 for details.

xudifsd:
The intended usage will be
```
python3 reports.py refresh -y $YARN_URL -p $PROMETHEUS_URL -l $LAUNCHER_URL -d cluster.sqlite
```

and

```
python3 reports.py report -y $YARN_URL -p $PROMETHEUS_URL -l $LAUNCHER_URL -d cluster.sqlite
```

The refresh command should be called periodically, maybe every 10 minutes.

Will write a doc on how to setup the cron job and more fancy usage.

xudifsd:
fixed #2127 and #2073

xudifsd:
may relay on #2449

scarlett2018:
@xudifsd Just made some inline changes for first 2 paragraphs: https://github.com/Microsoft/pai/blob/07c464838e7db343045a96dcacdb2d5c10532ec9/docs/tools/how-to-setup-report-script.md. Will review the rest next Monday.

xudifsd:
@scarlett2018 

1. you mean something like `2019/04/15-14:50:39`?
2. They are not duplicate in a sense that not all labels are the same, for example, JobExporter will have `docker_collector` and `container_collector`, if they all hanging, two alerts will triggered, although they are in same host, and appears to be duplicated in reports if we only extract IP. I did not find a better way to do this, because alerts have different labels and even different number of labels. Maybe we can only keep alerts what we want in the first place: `PaiServicePodNotReady`.
3. This is `disk-cleaner` pod, the origin requirement is from @fanyangCS to get a rough sense of what pai service is venerable to not ready, so no IP is required.

docs/tools/how-to-setup-report-script.md
```diff
@@ -0,0 +1,169 @@
...
+The `report` action will query data about vc usage and job statistic from sqlite3 DB and generate vc/job/raw_job/gpu csv files, it will also get data from Prometheus to generate alert reports. You can execute this action whenever you want the reports.
+
+The `serve` action will start a http server so outside world can query report through web server instead of using files.
+
+Both `serve` and `report` will need `refresh` being called periodically to fetch data
+from underlaying source.
+
+First, log into the node you choose, put the [script](../../src/tools/reports.py) somewhere, for example, I put it in directory `/home/core/report`, edit the crontab using
+
+``` sh
```
mzmssg: Currently it's more like services than tools, why not make it an optional service, like alert-manager.

xudifsd: Bin object this, if it is a service, we have to maintain its status, and if it is a tool, the users are responsible for that.

src/tools/reports.py
```diff
@@ -0,0 +1,959 @@
...
+        self.conn.commit()
+
+
+def get_yarn_apps(yarn_url):
+    apps_url = urllib.parse.urljoin(yarn_url, "/ws/v1/cluster/apps")
+    result = []
+
+    obj = request_with_error_handling(apps_url)
+
+    if obj.get("apps") is None or obj["apps"].get("app") is None:
```
mzmssg: More checker?
```suggestion
    if obj is None or obj.get("apps") is None or obj["apps"].get("app") is None:
```

src/tools/reports.py
```diff
@@ -0,0 +1,959 @@
...
+    return result
+
+
+def get_frameworks(launcher_url):
+    launcher_url = urllib.parse.urljoin(launcher_url, "/v1/Frameworks")
+    result = []
+
+    obj = request_with_error_handling(launcher_url)
+
+    if obj.get("summarizedFrameworkInfos") is None:
```
mzmssg:
```suggestion
    if obj is None or obj.get("summarizedFrameworkInfos") is None:
```

docs/tools/how-to-setup-report-script.md
mzmssg: I see, but it's not that user friendly, and might mess host envs. Is it necessary to tell user how to uninstall this tool?

xudifsd: Yes, maybe we should change it to service in the future.

Issue #2515, opened 2019-04-08
[v0.11.0] Job list missing on IE


scarlett2018:
**Short summary about the issue/question**:
Job List is missing on IE.
Work around: use Edge, Chrome, Firefox or other browsers.

**How to reproduce it**:

**OpenPAI Environment**:
- OpenPAI version: 0.11.0

Gerhut:
Closed by #2516

PR #2516, opened 2019-04-09
add ie11 support

yxwithu:
fix IE11 job missing bug #2515
https://github.com/Microsoft/pai/issues/2515


Gerhut:
Please introduce URL & URLSearchParams polyfill together to make the browser runtime more stable.

Gerhut:
I choose to upgrade to babel@7.4 which supports core-js@3, that includes polyfill of `URL` & `URLSearchParams`. core-js@2(or @babel/polyfill) do not have these 2.

@sunqinzheng please have a look

Gerhut:
The reason for using the "entry" mode is that we still have some CommonJS typed modules, and "usage" mode will prepend ES2015 typed `import "core-js/foo/bar"`s to each file, which confused webpack.

Issue #1893, opened 2018-12-14
P00 - part of 4022 - Prerequisites/node check


hao1939:
Before installation, check:
1. Hardware requirements (cpu/memory/gpu, etc)
2. System configuration (os, docker, ntp, user, etc)

After installation, collect those info for debug purpose.

hao1939:
#2520 would check os version.

fanyangCS:
done in https://github.com/microsoft/pai/blob/master/contrib/kubespray/doc/requirement.md

PR #2520, opened 2019-04-09
check if the nodes are ubuntu 16.04

### LINKED ISSUE #1657: systemctl docker restart fail on some machine
On some machine this command failed with error;
Failed to issue method call: Unit docker.service failed to load: No such file or directory. See system logs and 'systemctl status docker.service' for details.

But the command "service docker restart" can be used. We may need to fix our script to execute this command if systemctl fails.

-----------------
**Resolution**:
+ add a system validation

hao1939:
Add a prerequisites check for os version, as part of #1893. Fixes #1657 .

### DIFF
--- a/deployment/checkCmd.py
+++ b/deployment/checkCmd.py
@@ -13,7 +13,7 @@ def validate_config(config_path):
     layout_file = os.path.join(config_path, "layout.yaml")
     old_cluster_config_file = os.path.join(config_path, "cluster-configuration.yaml")
     if(os.path.exists(old_cluster_config_file) and not os.path.exists(layout_file)):
-        logger.error("[Error], please upgrade config files!")
+        logger.error("[Error] - Please upgrade config files!")
         exit(-1)
 
     # kubernetes config
@@ -44,7 +44,7 @@ def validata_node_os(config_path):
             host_not_match.append(host["hostip"])
 
     if len(host_not_match) > 0:
-        logger.error("The nodes be listed are not ubunut 16.04: %s", host_not_match)
+        logger.error("[Error - unsupported OS] - OpenPAI only supports ubuntu 16.04, the following node(s) are not ubuntu 16.04: %s", host_not_match)
         exit(-1)
 
 
--- a/deployment/checkCmd.py
+++ b/deployment/checkCmd.py
@@ -2,8 +2,50 @@ from clusterObjectModel.cluster_object_model import cluster_object_model
 import os
 import logging
 import logging.config
+import yaml
+import re
+from k8sPaiLibrary.maintainlib import common as pai_common
 
 logger = logging.getLogger(__name__)
+
+
+def validate_config(config_path):
+    layout_file = os.path.join(config_path, "layout.yaml")
+    old_cluster_config_file = os.path.join(config_path, "cluster-configuration.yaml")
+    if(os.path.exists(old_cluster_config_file) and not os.path.exists(layout_file)):
+        logger.error("[Error], please upgrade config files!")
+        exit(-1)
+
+    # kubernetes config
+    my_cluster_object_model = cluster_object_model(config_path)
+    kubernetes_config = my_cluster_object_model.kubernetes_config()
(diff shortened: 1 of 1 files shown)


deployment/checkCmd.py
```diff
@@ -2,10 +2,52 @@
...
+    regex = r"Description:\s+Ubuntu\s+16.04(.\d)?\s+LTS"
+    for host in hosts:
+        logger.info("Check os for %s ...", host["hostip"])
+        result_stdout, result_stderr = pai_common.ssh_shell_paramiko_with_result(
+            host,
+            "sudo lsb_release -d")
+        logger.info(result_stdout)
+        # should match  Ubuntu 16.04.5 LTS
+        if re.match(regex, result_stdout) is None:
+            host_not_match.append(host["hostip"])
```
scarlett2018: do you want to append a seperation charactor beween hosts?

deployment/checkCmd.py
```diff
@@ -2,10 +2,52 @@
...
+
+    if len(host_not_match) > 0:
+        logger.error("The nodes be listed are not ubunut 16.04: %s", host_not_match)
```
scarlett2018: > "The nodes be listed are not ubunut 16.04: %s"
Maybe change to:
[Error - unsupported OS] - OpenPAI only supports ubuntu 16.04 and above OS, the following node(s) are not ubuntu 16.04: %s.

deployment/checkCmd.py
hao1939: It's a list, so no need.

deployment/checkCmd.py
hao1939: Will update.
One notes: we haven't tested `16.04 above`.

deployment/checkCmd.py
```diff
@@ -2,10 +2,52 @@
...
+from k8sPaiLibrary.maintainlib import common as pai_common
 
 logger = logging.getLogger(__name__)
 
 
+def validate_config(config_path):
+    layout_file = os.path.join(config_path, "layout.yaml")
+    old_cluster_config_file = os.path.join(config_path, "cluster-configuration.yaml")
+    if(os.path.exists(old_cluster_config_file) and not os.path.exists(layout_file)):
+        logger.error("[Error], please upgrade config files!")
```
abuccts: do not need `,`

deployment/checkCmd.py
```diff
@@ -2,10 +2,52 @@
...
+def validata_node_os(config_path):
+    layout_file = os.path.join(config_path, "layout.yaml")
+    hosts = yaml.load(open(layout_file))["machine-list"]
+    print hosts
+
+    host_not_match = []
+    regex = r"Description:\s+Ubuntu\s+16.04(.\d)?\s+LTS"
+    for host in hosts:
+        logger.info("Check os for %s ...", host["hostip"])
+        result_stdout, result_stderr = pai_common.ssh_shell_paramiko_with_result(
```
abuccts: do not need to use regex here.
you can check file `/etc/os-release`, grep `ID=ubuntu` and `VERSION_ID="16.04"`.

hao1939: Hi @abuccts, 

Agree, using grep also works. The two method are equivalent, so either is fine.

deployment/checkCmd.py
hao1939: Yeah, done.

deployment/checkCmd.py
```diff
@@ -2,10 +2,52 @@
...
 import logging.config
+import yaml
+import re
+from k8sPaiLibrary.maintainlib import common as pai_common
 
 logger = logging.getLogger(__name__)
 
 
+def validate_config(config_path):
+    layout_file = os.path.join(config_path, "layout.yaml")
```
ydye: Directly read from yaml file? Can you read the machine list or the so called layout from cluster object model? Other wise, if the name of ```machine-list``` is changed, you will have to change 2 places.

PR #2525, opened 2019-04-10
Add environment and handle corner cases

wangdian:
1. Add dev-box environment for storage plugin.
2. Handle case when setup storage server with no storage user.


wangdian:
Test case:
1. Use a clean PAI instance or check you k8s configmap in **[your ip]/kubernetes-dashboard/#!/configmap?namespace=default**, backup **storage-user** and remove it. 
2. In dev box. Go to **/pai/contrib/storage_plugin**, run  
**python storagectl.py server set [teamname] nfs [nfs ip] [nfs root path] --sharedfolders data --privatefolders users**
3. The test case passed if no exception occurs.
4. Recover **storage-user** configmap

PR #2527, opened 2019-04-10
[hadoop] remove duplicate diagnostics

mzmssg:
More readable patch: https://github.com/mzmssg/hadoop/pull/4/files
There will no duplicate stderr in diagnostics:
Original
image

Fixed
image


yqwang-ms:
Can we just move line 1575 to 1579 and 1596?
It seems more nature.
image

mzmssg:
It's the same. But we need three `if (exitEvent.getDiagnosticInfo() != null)` in different block, and split diagnostic and its prefix `Diagnostic message from attempt` in different blocks. 
I don't think it's more readable.

PR #2528, opened 2019-04-10
add cluster version info

### LINKED ISSUE #2498: Add cluster version info
A simple mechanism for Store/Query cluster version info.

Goal:
1. Store kubernetes/PAI version separately in the cluster.
2. Provide a interface(paictl or other) for query the kubernetes/PAI version.

hao1939:
resolve #2498 .

Implementation. Store the ‘version info’ as raw string in the files:
/PAI_CODE_ROOT/
                /version
                                PAI.VERSION
                                K8S.VERSION

The format would be ‘v0.12.0’ and ‘V0.12.0-1.9.9’.

There are 4 places where would need ‘version info’:
1.	[TODO] It will in another PR. Webportal. It will need to display the ‘version info’ on the page, generate from the ‘version info’ during the image building process.
2.	Quick-start. It will need to generate the ‘image tag’, reading the ‘version info’ on the fly.
3.	Dev-box, reading the ‘version info’ on the fly. When dev-box version is not the same as the cluster, would print a warning message.
4.	Build. NOTICE, not using the ‘version info’ , it will read from user config.


src/cluster-configuration/deploy/stop.sh.template
```diff
@@ -31,8 +31,17 @@ if kubectl get configmap | grep -q "gpu-configuration"; then
     kubectl delete configmap gpu-configuration || exit $?
 fi
 
+if kubectl get configmap | grep -q "pai-version"; then
+    kubectl delete configmap pai-version || exit $?
+fi
+
+if kubectl get configmap | grep -q "k8s-version"; then
+    kubectl delete configmap k8s-version || exit $?
```
ydye: kubernetes-configuration.yaml has this information.

version/K8S.VERSION
```diff
@@ -0,0 +1 @@
+v0.11.0-1.9.9
```
ydye: Remove K8S version

src/cluster-configuration/deploy/stop.sh.template
hao1939: There is slight difference: the k8s version is our 'customized k8s' version.
For example: v0.11.0-1.9.9 may different from v0.12.0-1.9.9, although they both basing on the k8s 1.9.9.

version/K8S.VERSION
hao1939: My consideration is to make it different from `PAI version`, they would be decoupled later.

PR #2534, opened 2019-04-10
Fix gpu N/A in hardware page.

GeoffreyChen777:
In ver.0.11.0, the GPU and GPU Mem of some nodes are N/A on hardware page.

These nodes both consist of many Nvidia 2080ti GPU.

I checked the json dict from `http://x.x.x.x/prometheus/api/v1/query_range?query=avg(nvidiasmi_utilization_gpu%7Binstance%3D~%22x(%3A%5B0-x.x.x.x9%5D*)%3F%24%22%7D)&start=1554901559&end=1554905159&step=30`. And, there is no data in the result.

The `job_explore` of these nodes report an `out of range` error:
```
  File "/job_exporter/nvidia.py", line 165, in nvidia_smi
    return parse_smi_xml_result(smi_output)
  File "/job_exporter/nvidia.py", line 134, in parse_smi_xml_result
    single = volatile.getElementsByTagName("single_bit")[0].getElementsByTagName("total")[0]
IndexError: list index out of range
```

In the `nvidia_smi_log` of 2080ti, there is no `single_bit` tag in it. But I find some other tags such as `sram_uncorrectable`:

```
		<ecc_errors>
			<volatile>
				<sram_correctable>N/A</sram_correctable>
				<sram_uncorrectable>N/A</sram_uncorrectable>
				<dram_correctable>N/A</dram_correctable>
				<dram_uncorrectable>N/A</dram_uncorrectable>
			</volatile>
			<aggregate>
				<sram_correctable>N/A</sram_correctable>
				<sram_uncorrectable>N/A</sram_uncorrectable>
				<dram_correctable>N/A</dram_correctable>
				<dram_uncorrectable>N/A</dram_uncorrectable>
			</aggregate>
		</ecc_errors>
```

After rewrite the `/job_exporter/nvidia.py`, all the nodes work well.


src/job-exporter/src/nvidia.py
```diff
@@ -131,8 +131,12 @@ def parse_smi_xml_result(smi):
             volatile = ecc_errors[0].getElementsByTagName("volatile")
             if len(volatile) > 0:
                 volatile = volatile[0]
-                single = volatile.getElementsByTagName("single_bit")[0].getElementsByTagName("total")[0]
-                double = volatile.getElementsByTagName("double_bit")[0].getElementsByTagName("total")[0]
+                try:
+                    single = volatile.getElementsByTagName("single_bit")[0].getElementsByTagName("total")[0]
+                    double = volatile.getElementsByTagName("double_bit")[0].getElementsByTagName("total")[0]
+                except:
```
xudifsd: Do you know what exact exception this may cause? If so, please specify it here.

src/job-exporter/src/nvidia.py
```diff
@@ -131,8 +131,12 @@ def parse_smi_xml_result(smi):
...
+                    single = volatile.getElementsByTagName("sram_uncorrectable")[0]
```
xudifsd: Could you please add some comments here so reader will know what's going on here.
I also suggest you add some unit test like [here](https://github.com/Microsoft/pai/blob/master/src/job-exporter/test/test_nvidia.py#L47). You can add sample nvidia-smi output in data directory.

xudifsd:
@GeoffreyChen777 Thank you for your contribution, I added some comments, would you help change a little bit?

GeoffreyChen777:
@xudifsd Sure, I will change it soon.

GeoffreyChen777:
@xudifsd Hi, I pushed the new code now. 

I searched a lot of information about the ecc error in `nvidia-smi -q -x` report. In this doc([https://developer.download.nvidia.com/compute/DCGM/docs/nvidia-smi-367.38.pdf](https://developer.download.nvidia.com/compute/DCGM/docs/nvidia-smi-367.38.pdf)), Nvidia mentioned that the ECC errors are either single or double bit, where single bit errors are corrected and double bit errors are uncorrectable. 

So, I summarize the value of the `sram_correctable` and `dram_correctable` tags as the single_bit error number. The double_bit error number are calculated by `sram_uncorrectable` and `dram_uncorrectable`.
BTW, I'm not a professional Nvidia driver developer. I'm not sure that's the right way to calculate it. :)

mzmssg:
@GeoffreyChen777 
Thanks for your fixing.
Seems that 2080 is not for scientific computing and have no ecc correction. That's the difference.
I check the field `ecc_object` is `<ecc_object>3.0</ecc_object>` for us  and `<ecc_object>N/A</ecc_object>` for you. 

Maybe we should disable ecc checker for this gpus.

mzmssg:
I think `ecc_object` seems also not enough, given that ecc could be manually turn off in support GPU.
My suggestion is putting this part in a `try...catch...` block.

@xudifsd 
How about your opinion?

xudifsd:
@mzmssg will turn off ecc make `ecc_object` as "N/A"? If so I think this is enough.

xudifsd:
I have confirmed that the disabled ecc wil still has `ecc_object` as previous value, `3.0` in my case.

and the `ecc_errors` will have:

``` xml
<ecc_errors>
                        <volatile>
                                <single_bit>
                                        <device_memory>N/A</device_memory>
                                        <register_file>N/A</register_file>
                                        <l1_cache>N/A</l1_cache>
                                        <l2_cache>N/A</l2_cache>
                                        <texture_memory>N/A</texture_memory>
                                        <texture_shm>N/A</texture_shm>
                                        <cbu>N/A</cbu>
                                        <total>N/A</total>
                                </single_bit>
                                <double_bit>
                                        <device_memory>N/A</device_memory>
                                        <register_file>N/A</register_file>
                                        <l1_cache>N/A</l1_cache>
                                        <l2_cache>N/A</l2_cache>
                                        <texture_memory>N/A</texture_memory>
                                        <texture_shm>N/A</texture_shm>
                                        <cbu>N/A</cbu>
                                        <total>N/A</total>
                                </double_bit>
                        </volatile>
                        <aggregate>
                                <single_bit>
                                        <device_memory>N/A</device_memory>
                                        <register_file>N/A</register_file>
                                        <l1_cache>N/A</l1_cache>
                                        <l2_cache>N/A</l2_cache>
                                        <texture_memory>N/A</texture_memory>
                                        <texture_shm>N/A</texture_shm>
                                        <cbu>N/A</cbu>
                                        <total>N/A</total>
                                </single_bit>
                                <double_bit>
                                        <device_memory>N/A</device_memory>
                                        <register_file>N/A</register_file>
                                        <l1_cache>N/A</l1_cache>
                                        <l2_cache>N/A</l2_cache>
                                        <texture_memory>N/A</texture_memory>
                                        <texture_shm>N/A</texture_shm>
                                        <cbu>N/A</cbu>
                                        <total>N/A</total>
                                </double_bit>
                        </aggregate>
                </ecc_errors>
```

Which will also make that logic failed.

@GeoffreyChen777 Could you wrap that logic into try..catch as @mzmssg said?

Maybe we should think about a safer way to access xml, like [`walk_json_field_safe`](https://github.com/Microsoft/pai/blob/master/src/job-exporter/src/utils.py#L42).

GeoffreyChen777:
@xudifsd ok, I will do it tonight.

GeoffreyChen777:
Hi, @xudifsd, According to your xml output after disabling ecc, when the ecc_error is supported by the GPU but we manually disable it, there is still a `single_bit tag` in the xml output. So the `ecc_single` will be set to 0 at here:
https://github.com/Microsoft/pai/blob/dc15181edfdc7ad8377a4a292da95b877303e630/src/job-exporter/src/nvidia.py#L128

When the ecc_object is N/A which means that the GPU don't support ecc_error checking. We will skip the ecc_error checking code here:
https://github.com/Microsoft/pai/blob/dc15181edfdc7ad8377a4a292da95b877303e630/src/job-exporter/src/nvidia.py#L134

So, @mzmssg, I think we don't need to wrap that logic into try..catch block.

mzmssg:
@GeoffreyChen777 
Thanks for your diving.
Because we havn't found a detailed nvidia explaination about these fields, we can't guarantee all cases are coverd . My suggestion is for a safe purpose.

GeoffreyChen777:
@mzmssg ok, I will wrap it by try and catch.

xudifsd:
Thank you for your contribution.

PR #2546, opened 2019-04-12
Authn configuration for login module

ydye:


src/authentication/config/authentication.py
```diff
@@ -0,0 +1,58 @@
...
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+# DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+class Authentication:
+
+    def __init__(self, cluster_configuration, service_configuration, default_service_configuration):
+        self.cluster_configuration = cluster_configuration
+        self.service_configuration = dict(default_service_configuration, **service_configuration)
+
```
abuccts: too many blank lines.
I suggest to follow [pep8](https://www.python.org/dev/peps/pep-0008/) style for all Python code, at least for newly added Python code currently. [Pylint](https://www.pylint.org/) can be used to do static code analysis.

src/rest-server/deploy/auth-configmap/oidc.yaml.template
```diff
@@ -0,0 +1,32 @@
+{% if cluster_cfg['authentication']['OIDC'] %}
+
+{% if cluster_cfg['authentication']['OIDC-type'] == 'AAD' %}
+
+OIDC-type: aad
```
abuccts: do not need more config files, you can add those items as env variables in pod yaml (deploy/rest-server.yaml.template) directly.

src/authentication/config/authentication.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def validation_pre(self):
+        print self.service_configuration["OIDC"]
```
abuccts: there's also `python3` in this change.
I suggest to make Python code compatible for both python 2 and 3, for newly added code currently, e.g.,
```py
from __future__ import print_function
print(self.service_configuration["OIDC"])
```

ydye: I will remove this line. Its's for debugging.

src/rest-server/deploy/auth-configmap/oidc.yaml.template
ydye: Whole file content will be a environment？

src/authentication/config/authentication.py
ydye: Done

src/rest-server/deploy/auth-configmap/oidc.yaml.template
abuccts: Are all those values needed in rest server? Or this file is just a config file like `pai-configuration`?

It seems that this yaml is just a k8s configmap, so why do we need to place it here? Maybe other services will also need it?

ydye: https://github.com/Microsoft/pai/blob/43a7f8e520775e515418cf68fb6ace0062bad535/src/rest-server/src/config/express.js#L62
See the example, all the value is needed.

ydye: Other service won't have the permission to communicate with AAD.

abuccts: I see.

PR #2564, opened 2019-04-15
Remove useless log

Binyang2014:
Remove useless log such as:
% Total    % Received % Xferd  Average Speed ...

### DIFF
--- a/src/rest-server/src/templates/dockerContainerScript.mustache
+++ b/src/rest-server/src/templates/dockerContainerScript.mustache
@@ -95,16 +95,12 @@ cp -r /pai/code/* ./
 function webhdfs_create_file()
 {
   webHdfsRequestPath=${1}"?user.name="{{ jobData.userName }}"&op=CREATE"
-  redirectResponse=$(curl -i -X PUT ${webHdfsRequestPath} -o /dev/null -w %{redirect_url}' '%{http_code} 2>/dev/null)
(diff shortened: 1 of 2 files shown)


Binyang2014:
> What if the curl fail? We need the log to diagnose. I suggest to keep the log somewhere .

When the curl is failed, we will log some errors when checking the return result, such as HttpCode and response. But the error generated by curl will be missed.

While if we keep such message here, the aggregated runtime error info will contain such log, and mess the runtime error info. 

So If the httpCode/resp can be used to diagnose, I suggest remove such curl error message here. After we rewrite the runtime, we can make this component cleaner and easy to diagnose.

yqwang-ms:
Do we need to check curl exit code?
And its stderr should be redirect to runtime.pai.error file.

Binyang2014:
> Do we need to check curl exit code?
> And its stderr should be redirect to runtime.pai.error file.

We already redirect the stderr to runtime.yarn.pai.error by using
`1>> $RUNTIME_YARN_LOG 2>> $RUNTIME_YARN_ERR_LOG`

Add curl response message check, log error message if curl failed.
Since I don't want to message "% Total % Received % Xferd Average Speed"  to mess the error log, put the stderr from curl to /dev/null

Anyway I think we need to refactor the runtime code to dismiss some tricky codes.

PR #2595, opened 2019-04-18
[Deployment] Disable IB driver installation by default

abuccts:
Disable IB driver installation by default.

Azure VM builtin IB kernel modules into vmlinux image,
IB driver installation will fail in this case.

If IB installation is needed during deployment,
set `enable-ib-installation` field to `true` in config.

### DIFF
--- a/src/drivers/config/drivers.yaml
+++ b/src/drivers/config/drivers.yaml
@@ -27,8 +27,6 @@ version: "384.111"
 
 pre-installed-nvidia-path: /usr/local/nvidia
 
-# Azure VM builtin IB kernel modules into vmlinux image,
-# IB driver installation will fail in this case.
-# If IB installation is needed during deployment,
-# you can set `enable-ib-installation` field to true.
+# IB driver installation will fail when VM builtin IB kernel modules into vmlinux image.
+# If IB installation is needed during deployment, you can set the following field to true.
 enable-ib-installation: false
--- a/src/drivers/config/drivers.yaml
+++ b/src/drivers/config/drivers.yaml
@@ -28,4 +28,4 @@ version: "384.111"
 pre-installed-nvidia-path: /usr/local/nvidia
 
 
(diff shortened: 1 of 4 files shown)


src/drivers/config/drivers.yaml
```diff
@@ -28,4 +28,4 @@ version: "384.111"
 pre-installed-nvidia-path: /usr/local/nvidia
 
 
-skip-ib-installation: false
+skip-ib-installation: true
```
squirrelsc: 1. Can the name of configuration is positive (enable-IB-driver-installation), instead of negative(skip, disable, turn-off)?
2. add some comments and update document to explain this setting, and potential issue once it's enabled.

abuccts: updated

src/drivers/config/drivers.yaml
```diff
@@ -27,5 +27,8 @@ version: "384.111"
 
 pre-installed-nvidia-path: /usr/local/nvidia
 
-
-skip-ib-installation: false
+# Azure VM builtin IB kernel modules into vmlinux image,
```
squirrelsc: It may not apply to Azure only. how about as below?

Some servers has already installed IB drivers, so this flag is disabled by default. If this flag is enabled, OpenPAI will try best to install the correct IB driver, but it may be failed due to compatibility.

abuccts: The installation works if ib drivers have already been installed, and we need to re-install drivers in `/var/drivers` path. The flag should be enabled in this case.
The only issue is linux kernel on Azure has builtin ib kernel modules, which fails the re-installation.

squirrelsc: We haven't tested on AWS, aliyun, and other various hardware configurations. So don't specify Azure here only, it may be a common case in cloud providers.

Issue #2539, opened 2019-04-12
Refine Home UI Part II - User (not Admin user) home page


scarlett2018:
Homepage for user 
- Job status
- My virtual clusters
- Available GPU nodes
- My recent jobs: Create / Tutorial (first time login)

![PAI-Home-1](https://user-images.githubusercontent.com/44491713/55927539-fd15ab80-5c47-11e9-9010-bb1df3bd9824.png)
![PAI-Home-2](https://user-images.githubusercontent.com/44491713/55927543-01da5f80-5c48-11e9-9f76-9b9630c62e49.png)

abuccts:
Merged in #2614,

PR #2614, opened 2019-04-18
[Web Portal] Refine Home UI Part II - User (not Admin user) home page 

sunqinzheng:
#2539


abuccts:
Could you attach some snapshots?

sunqinzheng:
图片

PR #2620, opened 2019-04-19
Fix runtime errorspec issues

Binyang2014:
1. make sure user error output will redirect to stderr
2. Remove some useless error message
3. Correct error info

Testing Done:
Tested in singlebox

### DIFF
--- a/src/rest-server/src/templates/dockerContainerScript.mustache
+++ b/src/rest-server/src/templates/dockerContainerScript.mustache
@@ -20,13 +20,13 @@
 
 # Bootstrap script for docker container.
 
-exec 17>/pai/log/DockerContainerDebug.log
+exec 17>/pai/log/runtime.docker.pai.log
 BASH_XTRACEFD=17
 
 RUNTIME_DOCKER_LOG=/pai/log/runtime.docker.pai.log
 RUNTIME_DOCKER_ERR_LOG=/pai/log/runtime.docker.pai.error
-USER_LOG=/pai/log/user.pai.log
-USER_ERROR_LOG=/pai/log/user.pai.error
+USER_LOG=/pai/log/user.pai.stdout
+USER_ERROR_LOG=/pai/log/user.pai.stderr
 DISK_CLEANER_ERROR_FILE=/pai/log/diskCleaner.pai.error
 
 # create runtime docker log here. Since multi shell IO block will write to the files,
@@ -83,7 +83,6 @@ touch "/alive/docker_$PAI_CONTAINER_ID"
 export PAI_WORK_DIR="$(pwd)"
 PAI_WEB_HDFS_PREFIX={{ webHdfsUri }}/webhdfs/v1/Container
 HDFS_LAUNCHER_PREFIX=$PAI_DEFAULT_FS_URI/Container
-export CLASSPATH="$(hadoop classpath --glob)"
 # Make all GPUs allocated to the container accessible
 export NVIDIA_VISIBLE_DEVICES=all
 
@@ -219,8 +218,6 @@ azure_rdma_preparation
 # Write env to system-wide environment
 env | grep -E "^PAI|PATH|PREFIX|JAVA|HADOOP|NVIDIA|CUDA" > /etc/environment
 
-} 1>> $RUNTIME_DOCKER_LOG 2>> $RUNTIME_DOCKER_ERR_LOG
-
 function run_user_command()
 {
   printf "%s %s\n\n" "[INFO]" "USER COMMAND START"
@@ -229,9 +226,8 @@ function run_user_command()
   exit 0
 }
 
-run_user_command > >(tee $USER_LOG) 2> >(tee $USER_ERROR_LOG >&2) &
+run_user_command >$USER_LOG 2>$USER_ERROR_LOG &
 
-{
 user_command_pid=$!
 
 function grep_kill_command_from_file()
--- a/src/rest-server/src/templates/dockerContainerScript.mustache
+++ b/src/rest-server/src/templates/dockerContainerScript.mustache
@@ -65,7 +65,7 @@ function exit_handler()
     exit 0
   fi
 
-  log_runtime_error "ERROR" "REASON" "\"Exiting because of user error, exit code: $rc\""
+  log_runtime_error "ERROR" "REASON" "\"User process exited with code: $rc\""
   exit $rc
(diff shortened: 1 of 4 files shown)


src/rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -505,7 +505,8 @@ docker pull {{ jobData.image }} \
 ## network consumption
 docker run --name $docker_name \
   --init \
-  --tty \
+  --attach STDOUT \
+  --attach STDERR \
```
mzmssg: > While a client is connected to container’s stdio using docker attach, Docker uses a ~1MB memory buffer to maximize the throughput of the application. If this buffer is filled, the speed of the API connection will start to have an effect on the process output writing speed. This is similar to other applications like SSH. Because of this, it is not recommended to run performance critical applications that generate a lot of output in the foreground over a slow client connection. Instead, users should use the docker logs command to get access to the logs.

So it might degrade performance?

mzmssg: Anyway, I think it's not critical

yqwang-ms: Should be caused by this https://github.com/moby/moby/issues/25542.
Dose tty also have performance degradation?
If not, we may need to think about to de-multiplex STDOUT and STDERR with -tty by ourselves

yqwang-ms: Or could we let the docker container directly write stdout and stderr to yarn log dir?

Binyang2014: I am not sure if it will degrade the performance. This description comes from "docker attach" command, here we use "docker run --attach". I'm not sure if the implementation are same for these two commands.

BTW, before this change, all container stdout and stderr will redirect to current stdout. This change just separate it. It not causes more message output to the current tty.

yqwang-ms: could we let the docker container directly write stdout and stderr to yarn log dir? This is the safest way to be high perf

Binyang2014: Currently, we already redirected user command stdout to /pai/log/user.pai.log and stderr to /pai/log/user.pai.error. We can use these two files as stdout and stderr directly.

src/rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -20,13 +20,13 @@
...
-exec 17>/pai/log/DockerContainerDebug.log
+exec 17>/pai/log/runtime.docker.pai.log
 BASH_XTRACEFD=17
 
 RUNTIME_DOCKER_LOG=/pai/log/runtime.docker.pai.log
 RUNTIME_DOCKER_ERR_LOG=/pai/log/runtime.docker.pai.error
-USER_LOG=/pai/log/user.pai.log
-USER_ERROR_LOG=/pai/log/user.pai.error
+USER_LOG=/pai/log/user.pai.stdout
+USER_ERROR_LOG=/pai/log/user.pai.stderr
```
yqwang-ms: >USER_ERROR_LOG [](start = 0, length = 14)

USER_STDERR

src/rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -20,13 +20,13 @@
...
 
-exec 17>/pai/log/DockerContainerDebug.log
+exec 17>/pai/log/runtime.docker.pai.log
 BASH_XTRACEFD=17
 
 RUNTIME_DOCKER_LOG=/pai/log/runtime.docker.pai.log
 RUNTIME_DOCKER_ERR_LOG=/pai/log/runtime.docker.pai.error
-USER_LOG=/pai/log/user.pai.log
-USER_ERROR_LOG=/pai/log/user.pai.error
+USER_LOG=/pai/log/user.pai.stdout
```
yqwang-ms: >USER_LOG [](start = 0, length = 8)

USER_STDOUT

src/rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -52,6 +53,12 @@ CONTAINER_OOM_ERROR=196
 CONTAINER_OOD_ERROR=198
 RUNTIME_UNRECOGNIZED_ERROR=255
 
+SIG_2="SIGINT"
+SIG_6="SIGABRT"
+SIG_9="SIGKILL"
+SIG_11="SIGSEGV"
+SIG_15="SIGTERM"
```
yqwang-ms: Seems you only need to exit with the exitcode, and no need to care about the string represent if the exitcode?
So, remove them for easy to maintain?

src/rest-server/src/templates/yarnContainerScript.mustache
```diff
@@ -222,23 +231,16 @@ function exit_handler()
   rc=$?
 
   log_runtime_error "INFO" "USER_EXIT_CODE" "$rc"
-
-  if [[ $rc -eq 125 ]]; then
-    # We assume 125 is a docker error, and retry it without burning user's retry count
-    printf "[DEBUG] Exit code 125, it may be docker error, convert it to a transient error\n"
-    log_runtime_error "ERROR" "REASON" "\"Current container is killed, may be due to the docker deamon error\""
+  if [[ $rc -eq 130 || $rc -eq 134 || $rc -eq 137 || $rc -eq 139 || $rc -eq 143 ]]; then
+    log_runtime_error "ERROR" "MESSAGE" "Container exited because of the OS signal"
```
yqwang-ms: Miss some exitcodes, please check:
[https://github.com/Microsoft/pai/blob/errorspec/src/job-exit-spec/config/job-exit-spec.yaml](https://github.com/Microsoft/pai/blob/errorspec/src/job-exit-spec/config/job-exit-spec.yaml)

yqwang-ms:
>   echo "  disk: |" >> $RUNTIME_AGG_ERROR_FILE

change this to diskCleaner

---
Refers to: src/rest-server/src/templates/yarnContainerScript.mustache:220 in b9c7c1c. [](commit_id = b9c7c1cb56278927db76f5ccca403d48b185a045, deletion_comment = False)

src/rest-server/src/templates/yarnContainerScript.mustache
Binyang2014: I have removed hard code signal number and just maintain the signal name.

PR #2628, opened 2019-04-22
[Protocol Plugin] Support job clone

### LINKED ISSUE #2589: Protocol V2: Some different types of job could not be cloned
Need to make old yaml job not clonable, add submitForm field to new yaml job.

abuccts:
Closes #2589.

* Add plugin id in submit protocol plugin
* Add protocol job clone in webportal, add plugin id to env
* Add extra schema in rest server api v2

### DIFF
--- a/contrib/submit-protocol/src/App/ProtocolForm.tsx
+++ b/contrib/submit-protocol/src/App/ProtocolForm.tsx
@@ -134,10 +134,11 @@ interface IProtocolProps {
   api: string;
   user: string;
   token: string;
-  source ?: {
+  source?: {
     jobName: string;
     user: string;
   };
+  pluginId?: string;
 }
 
 interface IProtocolState {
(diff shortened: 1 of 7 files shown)


src/rest-server/src/config/v2/protocol.js
```diff
@@ -317,6 +317,14 @@ const protocolSchema = {
         },
       },
     },
+    extras: {
+      type: 'object',
+      properties: {
```
Gerhut: `extras` is an uncontrolled and free for adding any sub fields by any plugins, so it could not be limited with only `submitForm` here.

Refer: https://github.com/Microsoft/pai/blob/f52d98ec34da4d6d192f5f715b391e60d0b8adb3/src/rest-server/src/config/job.js#L108-L112

contrib/submit-protocol/src/App/ProtocolForm.tsx
```diff
@@ -318,30 +319,37 @@ export default class ProtocolForm extends React.Component<IProtocolProps, IProto
...
         const protocol = yaml.parse(body);
+        if (protocol.extras.submitFrom !== pluginId) {
+          throw new Error(`Unknown plugin id ${protocol.extras.submitFrom}`);
+        }
         this.setState(
           { protocol },
           () => this.setJobName(
             null as any,
-            `${source.jobName}_clone_${Math.random().toString(36).slice(2, 10)}`,
+            this.getCloneJobName(source.jobName),
```
Gerhut: Why split the state update into two here?

src/rest-server/src/config/v2/protocol.js
abuccts: removed properties

contrib/submit-protocol/src/App/ProtocolForm.tsx
abuccts: to reuse `setJobName`, merged

Gerhut: Actually you can call 2 `setState`s in one event loop. React will merge them into one update.

PR #2635, opened 2019-04-22
[Launcher]: Infer TaskStatus after App completed

yqwang-ms:
By infer TaskStatus to be completed after App completed, we have not downgraded Launcher's Task/Container Consistent Semantic. 
I.e. the Container completed state is not confirmed, such as we already have not got confirmed that a container is fenced/killed totally if its node is lost.

In this way, before we support more accurate state than YARN (such as by Launcher Agent), we choose to totally believe YARN. 
So YARN consider the app/container is completed, Launcher also consider it is completed.


src/job-exit-spec/config/job-exit-spec.yaml
```diff
@@ -973,6 +973,24 @@ spec:
...
+################################
+# Range: [-7499, -7400]
+# Owner: PAI_LAUNCHER WebServer
+# Description: Task Failure
+################################
+- code: -7400
+  phrase: TASK_STOPPED_ON_APP_COMPLETION
+  issuer: PAI_LAUNCHER
+  causer: USER_STOP
+  type: USER_STOP
```
yqwang-ms: @sunqinzheng you may need to use code: -7400 or type: USER_STOP to judge whether the TaskState is Stop/Success/Failure...

Issue #2492, opened 2019-04-02
Allow edit user info before submit or submit failed.


mslichao:

mslichao:
The PR is #2636

Issue #2491, opened 2019-04-02
Preview the new users' info before submit.


mslichao:

mslichao:
The PR is #2636

Issue #2078, opened 2019-01-25
As a teacher, I need able to Batch create/update/delete user accounts for my class


scarlett2018:
This is an EDU scenario, before the class started, teacher needs to create accounts for users. after the class the user might need to batch reset the user psw, or directly delete all the users.

mslichao:
Split this issue to some sub issues and finish them step by step.

* Import users from CSV file directly and show the final result.
* Preview the new users' info before submit.
* Allow edit user info before submit or submit failed.
* Refine UI with the design.
* Refine User management page.

Below is ths design V2:
![1](https://user-images.githubusercontent.com/40487035/55391377-d7174980-556b-11e9-91d1-a96d1d8a0a19.png)
![2](https://user-images.githubusercontent.com/40487035/55391392-de3e5780-556b-11e9-95d5-e9223a7cbb3b.png)
![3](https://user-images.githubusercontent.com/40487035/55391395-ded6ee00-556b-11e9-85da-d2e468b3dfc0.png)
![4](https://user-images.githubusercontent.com/40487035/55391396-df6f8480-556b-11e9-850e-638659d9d74c.png)
![5](https://user-images.githubusercontent.com/40487035/55391399-df6f8480-556b-11e9-87d8-b6cf60229e51.png)
![6](https://user-images.githubusercontent.com/40487035/55391401-e0081b00-556b-11e9-9564-923411f7f6d7.png)
![7](https://user-images.githubusercontent.com/40487035/55391402-e0081b00-556b-11e9-896b-ec0620d9014e.png)
![8](https://user-images.githubusercontent.com/40487035/55391403-e0a0b180-556b-11e9-997f-568f66e8d941.png)
![9](https://user-images.githubusercontent.com/40487035/55391405-e1394800-556b-11e9-84cc-41d3bcdef81b.png)
![10](https://user-images.githubusercontent.com/40487035/55391407-e1394800-556b-11e9-888f-49fdc5611da6.png)

mslichao:
This is the design for add/edit user

![image](https://user-images.githubusercontent.com/40487035/57833139-f4d30200-77eb-11e9-93cb-3a2f7684655d.png)

fanyangCS:
close as it has been resolved.

PR #2636, opened 2019-04-23
[Web Portal] Preview and edit user info before batch create users.

mslichao:
#2491 and #2492 in #2078 

Teacher want to batch create/update/delete user accounts for his class.

This PR is preview and edit user info before batch create users, the whole steps is in #2078 .

image

image


ydye:
Is it possible to add a configuration switch in service-configuration.yaml for admin

mslichao:
Close and re-send another PR to master branch.

Issue #2195, opened 2019-02-22
move job container into k8s pod


fanyangCS:
**What would you like to be added**:

When NM launches a container, move job container into k8s pod. (presumably as a static pod to reduce impacts to api server).

**Why is this needed**:

The approach can let k8s takes over local resource coordination efforts, without PAI to worry about it. 
For example, 
i)The QoS mechanism and Pod eviction mechanism of K8s
Right now Docker and k8s may make conflict decision. Our current approach is to disable k8s pod eviction and let PAI takes over the control. Thus we cannot benefit from k8s' more mature and rich local resource management feature.

ii) InitContainer
We can leverage InitContainer to implement global barrier, etc. DockerContainerScript can also be implemented here, paving the way to a full PAI runtime implementation.

Potential problem of this approach.
Currently k8s pod might not be able to support device mount method like [GPU](https://github.com/Microsoft/pai/blob/pai-0.10.y/src/rest-server/src/templates/yarnContainerScript.mustache#L266), [RDMA](https://github.com/Microsoft/pai/blob/pai-0.10.y/src/rest-server/src/templates/yarnContainerScript.mustache#L269). Need further investigation.

**Without this feature, how does the current module work**：

run as a native Docker instance

**Components that may involve changes**:

YarnContainerScript, DockerContainerScript, k8s device plug-in.

fanyangCS:
also relate to other docker flags. https://github.com/Microsoft/pai/issues/2227

yqwang-ms:
**Seems we can only use these docker flags which are explicitly exposed by K8S.**
Otherwise, for example, user do not need to ask memory-swap flag to be exposed by K8S, see https://github.com/kubernetes/kubernetes/issues/7294
https://github.com/kubernetes/kubernetes/issues/53533#issuecomment-355526636

**To support Azure File Share through SMB**
Maybe we leverage below K8S exposed flags to achieve it:
[Pod Security Context Spec](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/)

**To support device mount method like GPU and RDMA**
We can develop or leverage existing device-plugins to achieve it:
[Examples](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/#examples)
[RDMA device plugin](https://github.com/hustcat/k8s-rdma-device-plugin)
[Topology Aware GPU device plugin](https://github.com/Microsoft/KubeGPU)

fanyangCS:
[...]

PR #2657, opened 2019-04-24
Add capability/device for jobs to use InfiniBand

abuccts:
Add IPC_LOCK capability `--cap-add IPC_LOCK` and /dev/infiniband devices `--device=/dev/infiniband` for jobs to use InfiniBand in non privileged Docker. Will check the IB device automatically.

Before this PR, host can install IB OFED drivers but jobs cannot use infiniband network inside container. With this PR, infiniband network will be enabled automatically for jobs if IB device is available.

Reference: https://community.mellanox.com/s/article/how-to-create-a-docker-container-with-rdma-accelerated-applications-over-100gb-infiniband-network#jive_content_id_Option_2


abuccts:
To use InfiniBand in k8s pod #2195, we may need [k8s-rdma-sriov-dev-plugin](https://github.com/Mellanox/k8s-rdma-sriov-dev-plugin) from Mellanox, here's an [example](https://www.alibabacloud.com/blog/using-rdma-on-container-service-for-kubernetes_594462).

PR #2666, opened 2019-04-25
Remove duplicate FROM in job exporter Dockerfile

abuccts:
Remove duplicate FROM in job exporter Dockerfile.

### DIFF
--- a/src/job-exporter/build/job-exporter.dockerfile
+++ b/src/job-exporter/build/job-exporter.dockerfile
@@ -21,8 +21,6 @@ RUN apt-get update && apt-get install --no-install-recommends -y build-essential
     git clone https://github.com/yadutaf/infilter --depth 1 && \
     cd infilter && make
 
-FROM python:3.7
-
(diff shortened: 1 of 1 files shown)


src/job-exporter/build/job-exporter.dockerfile
```diff
@@ -21,8 +21,6 @@ RUN apt-get update && apt-get install --no-install-recommends -y build-essential
     git clone https://github.com/yadutaf/infilter --depth 1 && \
     cd infilter && make
 
-FROM python:3.7
```
xudifsd: This is not a duplicated FROM, line 35 requires this.

See [this](https://docs.docker.com/develop/develop-images/multistage-build/)

xudifsd: Maybe you can rewrite the first FROM into `FROM python:3.7 as builder` and change line 35 to `COPY --from=builder` to avoid ambiguity.

PR #2679, opened 2019-04-26
Add jobhistory and resource manager cross origin

mzmssg:


mzmssg:
Actually we could aggressively keep only 1 copy of hadoop configuraion. I think it's fine

PR #2729, opened 2019-05-08
[Protocol] Support secrets in protocol

abuccts:
Support secrets in protocol
------------------------------

Users may need sensitive information including password or API key in their jobs, `secrets` section is used to help user hide those sensitive information in plain text and avoid sharing.

Work items
------------
- [x] Agree on an approach to add secrets in protocol.
- [x] Update the protocol implementation to support read secret from yaml and mask with `*` when storing.
- [x] Hide the secrets in webportal, including submission, config view, and job clone.

### DIFF
--- a/docs/pai-job-protocol.yaml
+++ b/docs/pai-job-protocol.yaml
@@ -31,7 +31,8 @@ parameters: # Optional, can be omitted.
 
 # If sensitive information including password or API key is needed in the protocol,
 # it should be specified here in secrets section and referenced as `$secrets`.
-# All secrets will be hidden in the plain text and won't be shared.
+# A system that supports PAI protocol should not keep information marked as secretes.
+# For example, the yaml file used for job cloning, the stdout/stderr should remove all information marked as secrets.
 secrets: # Optional, can be omitted.
   <secret1>: password # Specify name and value of all secrets that will be used in the whole job template.
   <secret2>: key # Can be referenced by `<% $secrets.secret1 %>`, `<% $secrets.secret2 %>`.
--- a/docs/pai-job-protocol.yaml
+++ b/docs/pai-job-protocol.yaml
@@ -232,5 +232,5 @@ deployments:
           # Assume the model is output locally, and this command copies the local output to hdfs. One can output to hdfs directly.
           # In this case, you will have to change "--train_dir=$PAI_WORK_DIR/output_<% $output.name %>".
 
-default:
+defaults:
   deployment: prod # Use prod deployment in job submission.
--- a/docs/pai-job-protocol.yaml
+++ b/docs/pai-job-protocol.yaml
@@ -31,6 +31,7 @@ parameters: # Optional, can be omitted.
 
 # If sensitive information including password or API key is needed in the protocol,
 # it should be specified here in secrets section and referenced as `$secrets`.
+# Scope of reference `$secrets`: the reference is global, shared among all task roles and docker image's `auth` field.
 # A system that supports PAI protocol should keep the secret information away from
 # unauthorized users (how to define unauthorized user is out of the scope of this protocol).
 # For example, the yaml file used for job cloning, the stdout/stderr should protect all information marked as secrets.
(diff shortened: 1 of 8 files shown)


docs/pai-job-protocol.yaml
```diff
@@ -19,16 +19,23 @@ prerequisites: # Optional
...
 # If specified, the whole parameters object can be referenced as `$parameters`.
-# Scope of reference `$parameters`: the refercen is global and shared among all task roles.
+# Scope of reference `$parameters`: the reference is global and shared among all task roles.
 parameters: # Optional, can be omitted.
   <param1>: value1 # Specify name and value of all the referencable parameters that will be used in the whole job template.
   <param2>: value2 # Can be referenced by `<% $parameters.param1 %>`, `<% $parameters.param2 %>`.
 
+# If sensitive information including password or API key is needed in the protocol,
+# it should be specified here in secrets section and referenced as `$secrets`.
+# All secrets will be hidden in the plain text and won't be shared.
```
fanyangCS: >All secrets will be hidden in the plain text and won't be shared [](start = 2, length = 64)

A system that supports PAI protocol should not keep information marked as secretes. For example, the yaml file used for job cloning, the stdout/stderr should remove all information marked as secrets.

docs/pai-job-protocol.yaml
```diff
@@ -126,6 +133,10 @@ parameters:
   model: resnet20
   batchsize: 32
 
+secrets:
+  docker_password: password
+  api_token: cGFzc3dvcmQ=
```
fanyangCS: >api_token [](start = 2, length = 9)

where do you use api_token?

abuccts: it hasn't been used in this file, just shows that the secret can be either password or token.

docs/pai-job-protocol.yaml
abuccts: updated

docs/pai-job-protocol.yaml
```diff
@@ -19,16 +19,24 @@ prerequisites: # Optional
...
 # If specified, the whole parameters object can be referenced as `$parameters`.
-# Scope of reference `$parameters`: the refercen is global and shared among all task roles.
+# Scope of reference `$parameters`: the reference is global and shared among all task roles.
 parameters: # Optional, can be omitted.
   <param1>: value1 # Specify name and value of all the referencable parameters that will be used in the whole job template.
   <param2>: value2 # Can be referenced by `<% $parameters.param1 %>`, `<% $parameters.param2 %>`.
 
+# If sensitive information including password or API key is needed in the protocol,
+# it should be specified here in secrets section and referenced as `$secrets`.
+# A system that supports PAI protocol should not keep information marked as secretes.
```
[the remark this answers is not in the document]
abuccts: updated

fanyangCS:
> default:

should be "defaults"?

---
Refers to: docs/pai-job-protocol.yaml:235 in 9e65442. [](commit_id = 9e654422c09dff193594b7f9eb313d46227799f7, deletion_comment = False)

abuccts:
> > default:
> 
> should be "defaults"?

right, fixed

docs/pai-job-protocol.yaml
```diff
@@ -19,16 +19,26 @@ prerequisites: # Optional
...
 # If specified, the whole parameters object can be referenced as `$parameters`.
-# Scope of reference `$parameters`: the refercen is global and shared among all task roles.
+# Scope of reference `$parameters`: the reference is global and shared among all task roles.
 parameters: # Optional, can be omitted.
   <param1>: value1 # Specify name and value of all the referencable parameters that will be used in the whole job template.
   <param2>: value2 # Can be referenced by `<% $parameters.param1 %>`, `<% $parameters.param2 %>`.
 
+# If sensitive information including password or API key is needed in the protocol,
+# it should be specified here in secrets section and referenced as `$secrets`.
+# Scope of reference `$secrets`: the reference is global, shared among all task roles and docker image's `auth` field.
```
fanyangCS: "global" means all sections. I think the reference scope is taskRole and auth field.

abuccts: updated

PR #2754, opened 2019-05-14
[Rest Server] Check default runtime before starting Docker

### LINKED ISSUE #2739: [Rest Server] Cannot start job container by nvidia-docker
Currently, rest server cannot start job container by nvidia-docker. If nvidia-docker is installed on the host, job container cannot use gpu correctly.

abuccts:
Check default runtime before starting Docker, fix #2739.

If nvidia-docker is installed and nvidia runtime is enabled by default, start nvidia-docker and isolate GPU through `NVIDIA_VISIBLE_DEVICES` provided by nvidia runtime.

### DIFF
--- a/src/rest-server/src/models/v2/job.js
+++ b/src/rest-server/src/models/v2/job.js
@@ -142,7 +142,7 @@ const generateYarnContainerScript = (frameworkName, userName, config, frameworkD
     azRDMA: azureEnv.azRDMA === 'true' ? true : false,
     reqAzRDMA: false,
     inspectPidFormat: '{{.State.Pid}}',
-    inspectOOMKilledFormat: '{{.State.OOMKilled}}',
+    infoDefaultRuntimeFormat: '{{json .DefaultRuntime}}',
   });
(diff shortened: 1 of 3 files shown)


src/rest-server/src/models/v2/job.js
```diff
@@ -142,7 +142,7 @@ const generateYarnContainerScript = (frameworkName, userName, config, frameworkD
     azRDMA: azureEnv.azRDMA === 'true' ? true : false,
     reqAzRDMA: false,
     inspectPidFormat: '{{.State.Pid}}',
-    inspectOOMKilledFormat: '{{.State.OOMKilled}}',
+    infoDefaultRuntimeFormat: '{{json .DefaultRuntime}}',
```
Binyang2014: Seems this field is static and we can put it into the shell script.

abuccts: the `{{ ... }}` will be replaced by `mustache` in shell script template, so we have to put it here as a workaround.

PR #2755, opened 2019-05-14
Samba server with AAD integration in Microsoft.

wangdian:
A samba server with AAD integration in Microsoft.
usage: 
```
./build.sh
./start.sh <DOMAIN> <DOMAINUSER> <DOMAINPWD> <PAISMBUSER> <PAISMBPWD>
```


contrib/samba-aad-server/build/krb5.conf
```diff
@@ -0,0 +1,62 @@
...
+    default_domain = CORP.MICROSOFT.COM                         
+  }                                                             
+                                                                
+  EUROPE.CORP.MICROSOFT.COM = {                                 
+    kdc = europe.corp.microsoft.com                             
+    default_domain = CORP.MICROSOFT.COM                         
+  }                                                             
+                                                                
+  FAREAST.CORP.MICROSOFT.COM = {                                
+    kdc = fareast.corp.microsoft.com
```
ydye: It seems that this is only a service could be deployed in the internal env, right?

wangdian: Yes, this is config is same as DLTS's winbind server

PR #2757, opened 2019-05-14
add gpu temperature detection

xudifsd:
required by DLWS, may also be useful to pai.

### DIFF
--- a/src/job-exporter/src/nvidia.py
+++ b/src/job-exporter/src/nvidia.py
@@ -69,26 +69,29 @@ class NvidiaGpuStatus(object):
         gpu_util the gpu util of this gpu, float number, range 0~100
         gpu_mem_util the gpu memory usage/total of this gpu, float number, range 0~100
         pids an array of pid numbers that uses this card
-        ecc_errors instance of EccError class """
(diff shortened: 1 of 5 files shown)


src/job-exporter/src/nvidia.py
```diff
@@ -69,26 +69,29 @@ class NvidiaGpuStatus(object):
...
                 self.gpu_mem_util == o.gpu_mem_util and \
                 self.pids == o.pids and \
                 self.ecc_errors == o.ecc_errors and \
                 self.minor == o.minor and \
-                self.uuid == o.uuid
+                self.uuid == o.uuid and \
+                self.temperature == o.temperature
 
 
 def parse_smi_xml_result(smi):
```
abuccts: It's better to use [NVIDIA System Management Interface (NVML)](https://developer.nvidia.com/nvidia-system-management-interface), or its [Python bindings]( http://pypi.python.org/pypi/nvidia-ml-py/) to get GPU status, which is backwards compatible, while `nvidia-smi`'s output is not guaranteed to be backwards compatible according to its docs.

xudifsd: Ok, will try to investigate and use them in future PR.

PR #2771, opened 2019-05-16
[Web Portal] Remove dup components, move ajax to one file when batch add user

mslichao:
Refine code for batch add user page:

* Remove dup components
* move ajax to one file
* using colorClassNames and spacing


src/webportal/src/app/user/fabric/batchRegister/TopBar.jsx
```diff
@@ -37,7 +37,7 @@ function TopBar() {
     iconProps: {
       iconName: 'OpenFolderHorizontal',
     },
-    buttonStyles: {root: buttonPrimaryStyles},
+    buttonStyles: {root: [buttonMarginStyle, {marginLeft: '-8px'}]},
```
sunqinzheng: I think here should be `-spacing.s1`.

src/webportal/src/app/user/fabric/batchRegister/Table.jsx
```diff
@@ -92,6 +95,14 @@ export default function Table() {
     },
   };
 
+  const DropdownDisabledDiv = styled.div`
+    background-color: #ffffff;
+    border: 1px solid #a6a6a6;
+    margin-left: -12px;
```
sunqinzheng: I think `styles={{title: {padding: 0}}}` could be used to clear dropdown's style.

sunqinzheng: https://developer.microsoft.com/en-us/fabric#/controls/web/dropdown#IDropdownStyles

src/webportal/src/app/user/fabric/batchRegister/TopBar.jsx
sunqinzheng: or maybe you can use IButtonStyle to clear the style.
https://developer.microsoft.com/en-us/fabric#/controls/web/button#IbuttonStyles

mslichao: Changed to use `-spacing.s1`.

src/webportal/src/app/user/fabric/batchRegister/Table.jsx
mslichao: Thanks for your suggestion. Fixed.

src/webportal/src/app/user/fabric/batchRegister/TopBar.jsx
```diff
@@ -37,7 +37,7 @@ function TopBar() {
     iconProps: {
       iconName: 'OpenFolderHorizontal',
     },
-    buttonStyles: {root: buttonPrimaryStyles},
+    buttonStyles: {root: [buttonMarginStyle, {marginLeft: `-${spacing.s1}`}]},
```
sunqinzheng: I think we'd better use CommandBar's styles -> root to set the horizontal padding to spacing.s1 (8px) and use margin right to set the spacing between buttons.
Another option is to use a basic Stack instead of CommandBar.

mslichao: Good suggestion. Fixed.

Issue #2772, opened 2019-05-16
[Web Portal] [User management] Remove Github PAT


mslichao:
Synced with Yifan that we don't save github PAT, so we need remove related UI.

PR #2775, opened 2019-05-16
[Web Portal] Remove Github PAT related code from web portal

mslichao:
#2772 [Web Portal] [User management] Remove Github PAT

Synced with Yifan that we don't save github PAT, so we need remove related UI.


mslichao:
@abuccts The link to change PAT in `github-throttled.js` should be removed too as we discussed.

PR #2782, opened 2019-05-17
[Web Portal] Remove entry for outdated link

### LINKED ISSUE #2356: Marketplace need update because the octokit function have changed
the rest api have changed, like `pageno -> page`,`query->q`, please update it.
the file like below:
[https://github.com/octokit/rest.js/blob/master/plugins/rest-api-endpoints/routes.json](url)

` "search": {
    "code": {
      "method": "GET",
      "params": {
        "order": {
          "enum": [
            "desc",
            "asc"
          ],
          "type": "string"
        },
        "page": {
          "type": "integer"
        },
        "per_page": {
          "type": "integer"
        },
        "q": {
          "required": true,
          "type": "string"
        },
        "sort": {
          "enum": [
            "indexed"
          ],
          "type": "string"
        }
      },`

abuccts:
Remove entry for preview submission and old marketplace.

Close #2356.

### DIFF
--- a/src/webportal/config/webpack.common.js
+++ b/src/webportal/config/webpack.common.js
@@ -70,11 +70,9 @@ const config = (env, argv) => ({
     'hardwareDetail': './src/app/cluster-view/hardware/hardware-detail.component.js',
     'k8s': './src/app/cluster-view/k8s/k8s.component.js',
     'docs': './src/app/job/job-docs/job-docs.component.js',
-    'jobSubmit': './src/app/marketplace/job-submit/job-submit.component.js',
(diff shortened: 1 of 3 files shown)


mslichao:
Seems `src\app\plugin\plugin.component.js` used `<pai-plugin>` which is `MarketplaceElement`, and `MarketplaceElement` required `template-view.component` which required `githubThrottled`. 
Should we fix this too?

abuccts:
`plugin.component.js` is just the plugin instead of `MarketplaceElement`, it doesn't require marketplace.

PR #2783, opened 2019-05-17
add response time & response code monitoring for services

xudifsd:
pods with label key of `app` and annotation `monitor.watchdog/port`, `monitor.watchdog/path` will be requested by watchdog, and watchdog will generate response related metrics like return code and latency.

With these metrics, we can write following alert rules:

* `rate(service_response_code_total{code="200"}[5m]) == 0` to alert service return not returning 200
* `histogram_quantile(0.9, sum(rate(service_response_latency_seconds_bucket[5m])) by (le, service_name, service_ip)) > 5` to alert 90% latency is larger than 5s.

### DIFF
--- a/src/watchdog/src/watchdog.py
+++ b/src/watchdog/src/watchdog.py
@@ -30,6 +30,7 @@ import faulthandler
 import gc
 import re
 import collections
+import datetime
 
 import yaml
 import prometheus_client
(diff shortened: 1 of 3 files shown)


xudifsd:
@mzmssg Some yarn services can take advantage of this to monitor service load?

Issue #2780, opened 2019-05-17
Protocol V2 release readiness check


scarlett2018:
The item is to record all the area that need to check before formally release protocol V2 in End May release. Any issue founded should be recorded here so that we can easily make go or no-go decisions.

**Check list**
- [x] UX polish and refine
- [x] Job submission experience in web ui, rest api, client (if any)
- [x] Job browsing in job list view, job detail page
- [x] Job information (error status, etc.) experience
- [x] Job V1 and V2 co-exist experience
- [x] Test Cases and CICD
- [x] Documentation for the Protocol, sample, and release note
@abuccts @fanyangCS @sterowang @sunqinzheng @qfyin - please add more.

**Issues**
- [x] Rest server and web portal Dockerfiles didn't use `yarn.lock` (#2768)
- [x] Cannot use <kbd>Backspace</kbd> in paictl (#2769)
- [x] Cannot load vc and login issue in home page (#2774)
- [x] Outdated links for previous submission v2 and marketplace (#2775, #2782)
- [x] Linked job pages from home page didn't work correctly (#2787)
- [x] User management page style issues (#2786)
- [x] Submission v2 and marketplace style issues (#2789)
- [x] Cannot copy and paste url in marketplace directly (#2789)
- [x] Jobs with the same name can be submitted through api v2 (#2837)
- [x] Protocol files in marketplace cannot be used directly (#2827)
- [x] Add extras field in protocol (#2827)
- [x] Do not support responsive mode (#2798, #2805)

fanyangCS:
When submit a job from marketplace, the submit Job v2 should show the selected job from marketplace, instead of setting on the "Upload from local disk"
![image](https://user-images.githubusercontent.com/5820832/57912269-77c38d80-78bc-11e9-9cdd-aff76316c8d3.png)

fanyangCS:
Set the default marketplace view to grid view. Remove "Protocol Preview". Remove the "start" icon for each job. Change the "submit" icon to a better one.
![image](https://user-images.githubusercontent.com/5820832/57912943-1997aa00-78be-11e9-9172-d199ad6eecff.png)

fanyangCS:
when submitting a tensorflow job, I get the error below.
![image](https://user-images.githubusercontent.com/5820832/57914075-8f047a00-78c0-11e9-82ee-edad38a54637.png)

fanyangCS:
the page to change the url of marketplace need a "ok" and a "cancel" button.
[...]

PR #2789, opened 2019-05-20
[Plugin] Update protocol submission form and marketplace

abuccts:
Resolve comments in #2780.

* Update default marketplace option in submission form
* Set default view to grid view in marketplace
* Remove subtitle in marketplace
* Remove star icon
* Add apply and discard buttons for uri configuration
* Support copy url from GitHub / DevOps to marketplace directly.

### DIFF
--- a/contrib/submit-protocol/src/App/MarketplaceForm.tsx
+++ b/contrib/submit-protocol/src/App/MarketplaceForm.tsx
@@ -171,13 +171,61 @@ export default class MarketplaceForm extends React.Component<IMarketplaceProps, 
     }
   }
 
+  private formatMarketplaceURI = async (uri: string) => {
+    // regex for https://github.com/{owner}/{repo}/tree/{branch}/{path}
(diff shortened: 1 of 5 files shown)


contrib/submit-protocol/src/App/MarketplaceForm.tsx
```diff
@@ -126,6 +131,10 @@ export default class MarketplaceForm extends React.Component<IMarketplaceProps,
                   value={this.state.uriToken}
                   onChange={this.setMarketplaceURIToken}
                 />
+                <Stack gap={20} padding="15px auto 0" horizontalAlign="center" horizontal={true}>
```
sunqinzheng: If marketplace will be a built-in feature and need to use similar style as other components, you can try to use the following css constants:
- Font: FontClassNames
- Color: ColorClassNames
- Spacing
  - Stack: `l1 l2 m s1 s2`
  - Constant value: `const theme = getTheme(); marginLeft = theme.spacing.l1; blahblahblah`

abuccts: Let's refine css styles later in a separate PR.

PR #2801, opened 2019-05-21
[User & Group Manager] Change the API to get grouplist from v2 api.

ydye:


src/rest-server/src/models/v2/job.js
```diff
@@ -23,7 +23,7 @@ const status = require('statuses');
 const keygen = require('ssh-keygen');
 const mustache = require('mustache');
 const yaml = require('js-yaml');
-const userModel = require('../user');
+const userModelV2 = require('.user');
```
abuccts: use `./user`?

ydye: Changed to './user'

PR #2817, opened 2019-05-23
[Web Portal] extract monaco editor components

sunqinzheng:
## layout
- header
  - dom node, between title and monaco editor
- footer
  - dom node, previous primary button area
## feature
- monacoProps
  - will be appended to [react-monaco-editor](https://github.com/react-monaco-editor/react-monaco-editor) component
- schemas
  - json schema validation, only support `language=json`. format: https://microsoft.github.io/monaco-editor/playground.html#extending-language-services-configure-json-defaults
- completionItems
  - string array


src/webportal/src/app/components/monaco-panel.jsx
```diff
@@ -0,0 +1,163 @@
...
+                // completion provider
+                for (const lang of ['json', 'yaml', 'plaintext']) {
+                  monaco.current.languages.registerCompletionItemProvider(lang, {
+                    provideCompletionItems() {
+                      return completionList.current;
+                    },
+                  });
+                }
+                // json schema
+                setSchemas(monaco.current, schemas);
```
Binyang2014: I'm not sure how to use this schema. I only can use it when language is Json? If I set the language to yaml, how can I validate the content? By customize the onDismiss function?

sunqinzheng: MonacoEditor's built in schema check only support `language=json`.

If you want to use your own validation logic (yaml/joi)
-  use [react-monaco-editor](https://github.com/react-monaco-editor/react-monaco-editor)'s `onChange` event to get the content.
    - use `monacoProps` to pass your callback func
-  display the validation result in header area.

PR #2851, opened 2019-05-29
[Rest Server] Add get job api v2

abuccts:
Add get job api v2.

### DIFF
--- a/src/rest-server/src/controllers/v2/job.js
+++ b/src/rest-server/src/controllers/v2/job.js
@@ -35,7 +35,7 @@ const update = asyncHandler(async (req, res) => {
       throw createError('Conflict', 'ConflictJobError', `Job ${frameworkName} already exists.`);
     }
   } catch (error) {
-    if (error.code === 'UnknownError') {
+    if (error.code !== 'NoJobError') {
       throw error;
     }
   }
--- a/src/rest-server/src/controllers/v2/job.js
+++ b/src/rest-server/src/controllers/v2/job.js
@@ -20,7 +20,7 @@
 const yaml = require('js-yaml');
 const status = require('statuses');
 const asyncHandler = require('../../middlewares/v2/asyncHandler');
-const {get, put, getJobConfig} = require('../../models/v2/job');
+const job = require('../../models/v2/job');
 const createError = require('../../util/error');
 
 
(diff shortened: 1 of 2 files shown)


src/rest-server/src/controllers/v2/job.js
```diff
@@ -17,29 +17,29 @@
...
   try {
-    const getJobError = await util.promisify(v1Job.prototype.getJob)(jobName, userName);
-    if (getJobError.code !== 'NoJobError') {
+    const data = await job.get(frameworkName);
+    if (data != null) {
       throw createError('Conflict', 'ConflictJobError', `Job ${frameworkName} already exists.`);
     }
   } catch (error) {
-    throw createError('Conflict', 'ConflictJobError', `Job ${frameworkName} already exists.`);
+    if (error.code === 'UnknownError') {
```
Binyang2014: Only throw error when error code is UnknownError? It's an expect behavior?

abuccts: yes

abuccts: updated

src/rest-server/src/models/v2/job.js
```diff
@@ -271,10 +271,13 @@ async function get(frameworkName) {
     url: launcherConfig.frameworkPath(frameworkName),
     headers: launcherConfig.webserviceRequestHeaders(userName),
   });
-  if (response.status !== status('OK')) {
+  if (response.status === status('OK')) {
+    return response.data;
+  } else if (response.status === status('Not Found')) {
```
Binyang2014: remove else
```js
if (status === statue('OK')) {
  return ...;
}

if (...) {
  throw ...
}
```

src/rest-server/src/controllers/v2/job.js
```diff
@@ -35,7 +35,7 @@ const update = asyncHandler(async (req, res) => {
       throw createError('Conflict', 'ConflictJobError', `Job ${frameworkName} already exists.`);
     }
   } catch (error) {
-    if (error.code === 'UnknownError') {
+    if (error.code !== 'NoJobError') {
```
Binyang2014: Can log this?
```js
if (error.code === 'NoJobError') {
  log.Debug('...');
  return;
}

throw error;
```

src/rest-server/src/models/v2/job.js
abuccts: updated

src/rest-server/src/controllers/v2/job.js
abuccts: it's not an error if job not found here, no need to log.

PR #2882, opened 2019-06-04
[Plugin] Upgrade fstream to fix vulnerability

abuccts:
Versions of `fstream` prior to `1.0.12` are vulnerable to Arbitrary File Overwrite.
Upgrade` fstream` to `1.0.12` to fix vulnerability.


abuccts:
Tracked in #2886.

PR #2885, opened 2019-06-05
Binyli/job submit

Binyang2014:
This PR contains following changes；
1. Add converter to convert input to protocol format
2. Enhance some widgets
3. Performance improve for textField, only render when onBlur happends


src/webportal/src/plugins/job-submission/components/App.jsx
```diff
@@ -54,38 +58,36 @@ export class App extends React.Component {
...
               </Stack>
-            </Stack>
-            <Stack styles={{root: {width: '30%'}}}>
+            </StackItem>
+            <StackItem disableShrink styles={{root: [t.w30]}}>
               <Stack className={topForm}>
-                <Parameter/>
+                <Parameters defaultValue={parameters}
+                            environment={[]}
+                            onChange={(parameters) => this.setState({parameters: parameters})}/>
```
sunqinzheng: I think the `Parmeters` component is actually a controlled component since the `defaultValue` prop is mutable.

debuggy:
image
the command field not aligned with title, and the blank might be a little shorter

debuggy:
image

Maybe each row should align with top

debuggy:
The ```Ports, Task retry count, Completion and Deployment``` should be put into a more trivial place

debuggy:
> 
> 
> As the convention, the file name should be in lower cases

Binyang2014:
> [#2885 (review)](https://github.com/microsoft/pai/pull/2885#pullrequestreview-247458991)
> 
> As the file name comment: in this project, the name of files are all like `xword-yword.suffix`, maybe we should follow this convention.

Fine, will fix this issue

PR #2887, opened 2019-06-05
[Web Portal] AutoComplete component

sunqinzheng:
## What's this?

> The AutoComplete is an extension of a regular text-field that will auto-complete the input dynamically.

The difference between AutoComplete and ComboBox is that it is mainly focused on text input.

## Props
- `items`
  - string[], required
  - Array of suggestions. Only suggestions prefixed by current text field `value` are displayed by default.
- `value`
  - string, required
  - TextField's value
- `onChange`
  - callback function, required
    - (newValue: string) => {}
  - TextField's onChange function
- `showAllSuggestions`
  - boolean, required
  - Whether to disable suggestions' default prefix filter.

## Example

```
const app = () => {
  const [val, setVal] = useState('');
  return (
    <AutoComplete
      value={val}
      onChange={setVal}
      items={['pen', 'apple', 'applepen']}
    />
  );
}
```

### DIFF
--- a/src/webportal/src/app/components/auto-complete.jsx
+++ b/src/webportal/src/app/components/auto-complete.jsx
@@ -1,0 +1,176 @@
+// Copyright (c) Microsoft Corporation
+// All rights reserved.
+//
+// MIT License
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
+// documentation files (the "Software"), to deal in the Software without restriction, including without limitation
+// the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and
+// to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+// The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING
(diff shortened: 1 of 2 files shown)


src/webportal/src/app/components/auto-complete.jsx
```diff
@@ -0,0 +1,176 @@
...
+  const root = useRef();
+  const input = useRef();
+  // get callout's gap space
+  const {spacing} = getTheme();
+  const gap = useMemo(
+    () => getSpacingValue(spacing.s1),
+    [spacing],
+  );
+  // suggestions
+  useEffect(
```
Binyang2014: Can you give a simple example about how to use it?
And the auotComplete can support only autocomplete the last word? For example I have a sentence: "May the force be with you", and I input "May the fo", the suggestion should be "May the force".

sunqinzheng: The AutoComplete component behaves like a combo box. So you need to provide the whole sentence "May the force".
Since the value and suggestionItems are controlled by its parent, you can add your own suggestion logic.
e.g. add "May the force" option only when user has input "May the f"

sunqinzheng: If you want a multiline word/phrase level AutoComplete component, I recommend you use Monaco Editor.

Binyang2014: Got it, thanks for the explanation.

PR #2936, opened 2019-06-13
[webportal]remove the vc page information icon

fenghuajia:


src/webportal/src/app/vc/vc.component.js
```diff
@@ -4,8 +4,6 @@ require('datatables.net-bs/js/dataTables.bootstrap.js');
 require('datatables.net-bs/css/dataTables.bootstrap.css');
 require('datatables.net-plugins/sorting/natural.js');
 require('datatables.net-plugins/sorting/title-numeric.js');
-require('office-ui-fabric-js/dist/css/fabric.min.css');
```
Gerhut: Will we still need `office-ui-fabric-js` in other parts? If no, we could remove it in `package.json` and `yarn.lock`

Issue #2788, opened 2019-05-17
user's vc permission is not updated when a vc is removed.


sunqinzheng:

fanyangCS:
can you close the issue?

PR #2939, opened 2019-06-13
[Rest Server] update user's vc permission when create/remove vc

sunqinzheng:
#2788


mzmssg:
@ydye 
After AAD interagetion, I think this change is unnecessary?

PR #2947, opened 2019-06-14
Trap SIGTERM in entrypoint to avoid yarn container early stop

### LINKED ISSUE #2910: Job failure due to two jobs on same GPU
**Organization Name**:
Microsoft

Two jobs might have an overlap in using GPUs.
Then it will report `CUDA error: out of memory`

**How to reproduce it**:
Unknown

mzmssg:
fix #2910


fanyangCS:
is it a fix or just a mitigation?

mzmssg:
@fanyangCS 
A fix

PR #2952, opened 2019-06-18
update dev-box Dockerfile

NothingLZ:
Path problem error


abuccts:
Hi @NothingLZ,

Thanks for your contribution. The build's context for dev-box is `src/dev-box` instead of `src/dev-box/build`, so the original path is correct. The path of script is relative to `src/dev-box`, which is `build/start-script.sh`.

You can also refer to [dev-box build in CI](http://40.76.53.138:33080/blue/organizations/jenkins/pai/detail/PR-2952/1/pipeline#step-47-log-215).

PR #2955, opened 2019-06-18
Rest server job model refactor

abuccts:
Rest server job model refactor.


src/rest-server/yarn.lock
```diff
@@ -2420,6 +2420,11 @@ mocha@~5.0.0:
     mkdirp "0.5.1"
     supports-color "4.4.0"
 
+module-alias@^2.2.0:
```
SparkSnail: does this file need to be changed?

src/rest-server/package.json
```diff
@@ -60,6 +61,15 @@
     "winston": "~2.4.0",
     "xml2js": "~0.4.19"
   },
+  "_moduleAliases": {
+    "@config": "src/config",
```
MatsukawaTadashi: why not create a alias of root dir `src/`

sunqinzheng: Why not create an alias of root directory `src/`?

sunqinzheng: P.S. webpack also has a built-in module alias feature https://webpack.js.org/configuration/resolve/#resolvealias

src/rest-server/yarn.lock
abuccts: yes, added a new package `module-alias` in `package.json`

src/rest-server/package.json
abuccts: it's OK to create an alias for `src` but the path will be longer?

we don't use `webpack` in rest server

Gerhut: Have you tried `NODE_PATH` environment variable?

abuccts: > Have you tried `NODE_PATH` environment variable?

seems it does not work on windows using dotenv or command inline env

src/rest-server/src/models/user.js
```diff
@@ -259,7 +259,7 @@ if (config.env !== 'test') {
         if (errMsg) {
           logger.warn('get user list failed', errMsg);
         } else {
-          logger.warn('users:', userInfoList);
+          // logger.warn('users:', userInfoList);
```
SparkSnail: why annotate this?

abuccts: do not need to list all users in log, seems it was a debug log

PR #2964, opened 2019-06-19
collect DLWS_USER_EMAIL labels if available

xudifsd:
scrape user_email from docker inspect if available. For feature of sending email to users if job idle for a long time.


xudifsd:
related to https://github.com/microsoft/DLWorkspace/pull/366

PR #2978, opened 2019-06-21
[VS Code] Add schema and snippets for YAML text editor.

yiyione:
Add "redhat.vscode-yaml" extension dependency to support YAML job config autocomplete.
Add some snippets for YAML text editor.


contrib/pai_vscode/src/pai/yamlJobConfigCompletionProvider.ts
```diff
@@ -0,0 +1,48 @@
...
+ */
+export class YamlJobConfigCompletionProvider implements vscode.CompletionItemProvider {
+    private snippets: IYamlJobConfigSnippet[] = [];
+
+    public constructor() {
+        this.loadSnippets();
+    }
+
+    public provideCompletionItems(document: vscode.TextDocument, position: vscode.Position): vscode.CompletionItem[] {
+        const word: string = document.getText(document.getWordRangeAtPosition(position));
```
sunqinzheng: I think vscode's completion provider already uses fuzzy search to filter the result.
Maybe you can just return all snippets and let vscode do the filtering.

PR #2979, opened 2019-06-23
[Rest Server] Support framework controller

abuccts:
Support framework controller in rest server, migrate api v2:
* POST /api/v2/jobs
* GET /api/v2/jobs


abuccts:
Move to #2987.

Issue #2117, opened 2019-02-01
node exporter not ready for a few hours in D states


fanyangCS:
unknown out of memory![mem](https://user-images.githubusercontent.com/5820832/52099294-95226380-260d-11e9-826c-cc39203e2c3b.png), losing data 
points for several hours ![data](https://user-images.githubusercontent.com/5820832/52099281-89cf3800-260d-11e9-8d7a-f76d51971648.png)

xudifsd:
Happened again, the process is in D state:

```
core@paigcr-b-infra-10:~$ sudo ps aux | grep node_exporter
nobody    1141  0.0  0.1 4360512 90660 ?       Dsl  10:51   0:01 /bin/node_exporter --no-collector.arp --no-collector.bcache --no-collector.bonding --no-collector.conntrack --no-collector.edac --no-collector.entropy --no-collector.hwmon --no-collector.infiniband --no-collector.ipvs --no-collector.mdadm --no-collector.nfs --no-collector.nfsd --no-collector.sockstat --no-collector.stat --no-collector.time --no-collector.timex --no-collector.vmstat --no-collector.wifi --no-collector.xfs --no-collector.zfs
```

and pstack cannot retrieve its stack.

xudifsd:
Should try some other [methods](https://serverfault.com/a/942525) to see what the process is doing if D state happened again.

xudifsd:
the stack of node_exporter during D state:

```
core@paigcr-b-infra-10:~$ sudo cat /proc/32363/stack
[<0>] rpc_wait_bit_killable+0x24/0xa0 [sunrpc]
[<0>] __rpc_execute+0x15e/0x420 [sunrpc]
[<0>] rpc_execute+0x5d/0xc0 [sunrpc]
[<0>] rpc_run_task+0xf8/0x140 [sunrpc]
[<0>] rpc_call_sync+0x50/0x90 [sunrpc]
[<0>] nfs3_rpc_wrapper.constprop.12+0x39/0xb0 [nfsv3]
[<0>] nfs3_proc_statfs+0x56/0xb0 [nfsv3]
[<0>] nfs_statfs+0x6e/0x190 [nfs]
[<0>] statfs_by_dentry+0x72/0x90
[<0>] vfs_statfs+0x1b/0xb0
[<0>] user_statfs+0x58/0xa0
[<0>] SYSC_statfs+0x27/0x60
[<0>] SyS_statfs+0xe/0x10
[...]

PR #2988, opened 2019-06-24
ignore network based file systems in node-exporter

xudifsd:
related to #2117 , ignore those network based filesystems in node-exporter. So the source down will not affect node-exporter.

We append `cifs|nfs4?|lxcfs` to the [`defIgnoredFSTypes`](https://github.com/prometheus/node_exporter/blob/v0.18.0/collector/filesystem_linux.go#L33).

### DIFF
--- a/src/node-exporter/deploy/node-exporter.yaml.template
+++ b/src/node-exporter/deploy/node-exporter.yaml.template
@@ -78,6 +78,8 @@ spec:
           - '--no-collector.wifi'
           - '--no-collector.xfs'
           - '--no-collector.zfs'
+          - '--collector.filesystem.ignored-fs-types'
+          - 'autofs|binfmt_misc|bpf|cgroup2?|configfs|debugfs|devpts|devtmpfs|fusectl|hugetlbfs|iso9660|mqueue|nsfs|overlay|proc|procfs|pstore|rpc_pipefs|securityfs|selinuxfs|squashfs|sysfs|tracefs|tmpfs|cifs|nfs4?|lxcfs'
         ports:
         - containerPort: {{ cluster_cfg["node-exporter"]["port"] }}
           hostPort: {{ cluster_cfg["node-exporter"]["port"] }}
--- a/src/node-exporter/deploy/node-exporter.yaml.template
+++ b/src/node-exporter/deploy/node-exporter.yaml.template
@@ -79,7 +79,7 @@ spec:
           - '--no-collector.xfs'
           - '--no-collector.zfs'
           - '--collector.filesystem.ignored-fs-types'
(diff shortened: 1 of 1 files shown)


src/node-exporter/deploy/node-exporter.yaml.template
```diff
@@ -78,6 +78,8 @@ spec:
           - '--no-collector.wifi'
           - '--no-collector.xfs'
           - '--no-collector.zfs'
+          - '--collector.filesystem.ignored-fs-types'
+          - 'autofs|binfmt_misc|bpf|cgroup2?|configfs|debugfs|devpts|devtmpfs|fusectl|hugetlbfs|iso9660|mqueue|nsfs|overlay|proc|procfs|pstore|rpc_pipefs|securityfs|selinuxfs|squashfs|sysfs|tracefs|tmpfs|cifs|nfs4?|lxcfs'
```
abuccts: use `'^(autofs|...|lxcfs)$'` ?

xudifsd: I'm not sure, in that case lxcfs should be `fuse.lxcfs`, not sure if others are same. Since they are from default value, I think it's ok.

abuccts: I found the [default values](https://github.com/prometheus/node_exporter/blob/v0.18.0/collector/filesystem_linux.go#L33) are limited by `^(...)$`, so maybe keep the origin?

xudifsd: ah, did not see that.

src/node-exporter/deploy/node-exporter.yaml.template
```diff
@@ -79,7 +79,7 @@ spec:
           - '--no-collector.xfs'
           - '--no-collector.zfs'
           - '--collector.filesystem.ignored-fs-types'
-          - 'autofs|binfmt_misc|bpf|cgroup2?|configfs|debugfs|devpts|devtmpfs|fusectl|hugetlbfs|iso9660|mqueue|nsfs|overlay|proc|procfs|pstore|rpc_pipefs|securityfs|selinuxfs|squashfs|sysfs|tracefs|tmpfs|cifs|nfs4?|lxcfs'
+          - '^(autofs|binfmt_misc|bpf|cgroup2?|configfs|debugfs|devpts|devtmpfs|fusectl|hugetlbfs|iso9660|mqueue|nsfs|overlay|proc|procfs|pstore|rpc_pipefs|securityfs|selinuxfs|squashfs|sysfs|tracefs|tmpfs|cifs|nfs4?|fuse.lxcfs)$'
```
abuccts: need escape here `fuse\.lxcfs`

PR #3001, opened 2019-06-25
[Rest Server] Convert framework response in job list and detail

abuccts:
Convert framework response to formats used in job list and detail, keep interface with webportal.

### DIFF
--- a/src/rest-server/src/models/v2/job/k8s.js
+++ b/src/rest-server/src/models/v2/job/k8s.js
@@ -42,9 +42,9 @@ const convertFrameworkSummary = (framework) => {
     },
     createdTime: new Date(framework.status.startTime).getTime(),
     completedTime: new Date(framework.status.completionTime).getTime(),
-    appExitCode: taskStatus.attemptStatus.completionStatus ? taskStatus.attemptStatus.completionStatus.code : null,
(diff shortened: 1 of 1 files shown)


src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -27,6 +27,93 @@ const convertName = (name) => {
...
+      appExitSpec: {}, // TODO
+      appExitDiagnostics: completionStatus ? completionStatus.diagnostics : null,
+      appExitMessages: {
+        container: null,
+        runtime: null,
+        launcher: null,
+      },
+      appExitTriggerMessage: completionStatus ? completionStatus.diagnostics : null,
+      appExitTriggerTaskRoleName: null, // TODO
+      appExitTriggerTaskIndex: null, // TODO
```
abuccts: @yqwang-ms Could framework controller return some structured exit information along with exit diagnostics,
e.g. `"TaskRole master: FailedTaskCount 1 has reached MinFailedTaskCount 1: Triggered by Task: [Index: 0, Diagnostics: Pod failed with non-zero container exit code: [Container main, ExitCode: 127, Reason: ContainerCannotRun, Message: ...]]"`?

yqwang-ms: Yep, we can do it in future. (Not first stage)

PR #3011, opened 2019-06-27
[webportal]dedicated vc table add NodeList cloumn

fenghuajia:


src/webportal/src/app/vc/vc.component.js
```diff
@@ -126,29 +159,21 @@ const virtualClustersAdd = () => {
 };
 
 //
-const addGroup = () => {
+const groupAdd = (groupmane) => {
```
mzmssg: Could you seperate group changes to another PR.

PR #3018, opened 2019-06-27
Sdk release v0.4.00

mydmdm:
python sdk and command line release branch


contrib/python-sdk/openpaisdk/cluster.py
```diff
@@ -0,0 +1,212 @@
...
+        Returns:
+            [type]: [description]
+        """
+
+        job_list = self.rest_api_jobs(job_name)
+        return [j['name'] for j in job_list] if name_only else job_list
+
+    def rest_api_jobs(self, job_name: str=None, info: str=None, user: str=None):
+        user = self.user if user is None else user
+        pth = '{}/rest-server/api/v1/user/{}/jobs'.format(self.pai_uri, user)
```
sunqinzheng: You should check if pylon is enabled before using link like `/rest-server`.
If pylon is disabled, user needs to specify the rest server's port/url itself.

contrib/python-sdk/openpaisdk/job.py
```diff
@@ -0,0 +1,274 @@
...
+            while True:
+                try:
+                    status = client.jobs(self.name)
+                    ip = status["taskRoles"]["main"]["taskStatuses"][0]["containerIp"]
+                except:
+                    ip = None
+                    time.sleep(10)
+                if ip:
+                    break
+            browser_open("http://%s:8888/?token=%s" % (ip, token))
```
sunqinzheng: I think the port 8888 should be added to job config's port list.

contrib/python-sdk/openpaisdk/cluster.py
mydmdm: add dependencies description in the README as below

> Pylon is required to parse the REST api path like /reset-server/.

contrib/python-sdk/openpaisdk/job.py
mydmdm: update it to the port list
```python
resources.setdefault("ports", {})["jupyter"] = 1
cmds = [
    "jupyter notebook --no-browser --ip 0.0.0.0 --port $PAI_CONTAINER_HOST_jupyter_PORT_LIST  --NotebookApp.token={} --allow-root --NotebookApp.file_to_run={}".format(token, nb_file),
]
```
and 
```python
browser_open("http://%s:%s/notebooks/%s" % (status["containerIp"], status["containerPorts"]["jupyter"], nb_file))
```

PR #3021, opened 2019-06-28
fix  bug: clone job can not submit

Binyang2014:


src/webportal/src/app/job-submission/components/submission-section.jsx
```diff
@@ -98,6 +98,10 @@ export const SubmissionSection = (props) => {
     const newValidationMessage = JobProtocol.validateFromObject(protocol);
     if (newValidationMessage !== validationMsg) {
       setValidationMsg(newValidationMessage);
+    }
+
+    const storedErrorMessage = errorMessages[VALIDATION_ERROR_MESSAGE_ID] || '';
+    if (storedErrorMessage !== newValidationMessage) {
```
sunqinzheng: I think we'd better change the `setErrorMessage` function in job-submission.jsx if needed

PR #3024, opened 2019-06-28
[Rest Server] Add annotations in framework config

abuccts:
Add annotations for `userName`, `virtualCluster`, `config` in framework.

### DIFF
--- a/src/rest-server/src/models/v2/job/k8s.js
+++ b/src/rest-server/src/models/v2/job/k8s.js
@@ -30,7 +30,7 @@ const convertName = (name) => {
 const convertFrameworkSummary = (framework) => {
   return {
     name: framework.metadata.name,
-    username: 'unknown',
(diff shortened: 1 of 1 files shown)


src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -227,12 +227,17 @@ const generateTaskRole = (taskRole, config) => {
...
 };
 
-const generateFrameworkDescription = (frameworkName, config) => {
+const generateFrameworkDescription = (frameworkName, userName, config, rawConfig) => {
   const frameworkDescription = {
     apiVersion: launcherConfig.apiVersion,
     kind: 'Framework',
     metadata: {
       name: frameworkName,
+      annotations: {
```
yqwang-ms: Make the userName, virtualCluster also be the label for framework K8s object ?

PR #3040, opened 2019-07-02
[Web Portal] remove pylon address dependency

sunqinzheng:


src/webportal/src/app/home/index.jsx
```diff
@@ -79,7 +80,10 @@ const Index = () => {
       if (config.authnMethod === 'basic') {
         setLoginModal(true);
       } else {
-        location.href = config.restServerUri + '/api/v1/authn/oidc/login?redirect_uri=' + encodeURIComponent(config.pylonAddress+'/index.html')+ '&callback=' + encodeURIComponent(location.href);
+        location.href = config.restServerUri + `/api/v1/authn/oidc/login?${querystring.stringify({
+          redirect_uri: new URL('/index.html', window.location.href).href,
```
ydye: Anything is needed to be updated here? 
https://github.com/microsoft/pai/blob/0f44c46f9e61d6601b1ce19694bee9b5741ca529/src/rest-server/src/controllers/v2/azureAD.js#L34

sunqinzheng: I think express's `req.query` is already parsed by `querystring` or `qs`. You don't need to decode it twice.

ydye: you are right,

PR #3048, opened 2019-07-02
 fix bug when pylonAddress not set in .env 

debuggy:


src/webportal/src/app/job-submission/components/data/data-component.jsx
```diff
@@ -46,8 +46,14 @@ function reducer(state, action) {
 }
 
 export const DataComponent = React.memo((props) => {
-  // add pylonAddress to .env for local debug
-  const hdfsHost = config.hasOwnProperty('pylonAddress') ? getHostNameFromUrl(config.pylonAddress) : window.location.hostname;
+  const envsubRegex = /^\${.*}$/;
+  let hdfsHost;
+  if (!config.pylonAddress || envsubRegex.test(config.pylonAddress)) {
```
Binyang2014: Here we use pylonAddress  to decide the hdfsHost address. Why not use WebHdfsUri directly.
Change the env config, if it contains WebHdfsUri, then use it. Otherwise use window.location.hostname.

.env file can be wrote like:
{
...
WEB_HDFS_URI=http://10.151.40.252:50700
}

src/webportal/src/app/job-submission/components/data/data-component.jsx
```diff
@@ -46,8 +46,14 @@ function reducer(state, action) {
...
+    hdfsHost = window.location.hostname;
+  } else {
+    // add pylonAddress to .env for local debug
+    hdfsHost = getHostNameFromUrl(config.pylonAddress);
+  }
   // TODO: add a judgement whether pylon is ready
   const hdfsClient = new WebHDFSClient(hdfsHost);
```
Binyang2014: Should put port and path to the parameters?
If we use customized hdfs url. The port may different with the default value.

src/webportal/src/app/job-submission/components/data/data-component.jsx
debuggy: the env.js.template file does not contain WebHdfsUri key, unless we change the template file, which will change the window.ENV too

src/webportal/src/app/job-submission/components/data/data-component.jsx
debuggy: yes, I will put the parameters outside

src/webportal/src/app/job-submission/components/data/data-component.jsx
Binyang2014: Actually, I prefer change the template file and change the windows.ENV.
I will make code more reasonable and don't need to care about the logical between pylon and webhdfs
And this filed is optional.

If add a new filed in template file will cause any side-effect, please correct me.

src/webportal/src/app/job-submission/components/data/data-component.jsx
```diff
@@ -46,10 +46,16 @@ function reducer(state, action) {
...
+  let hdfsHost;
+  if (!config.pylonAddress || envsubRegex.test(config.pylonAddress)) {
+    hdfsHost = window.location.hostname;
+  } else {
+    // add pylonAddress to .env for local debug
+    hdfsHost = getHostNameFromUrl(config.pylonAddress);
+  }
   // TODO: add a judgement whether pylon is ready
-  const hdfsClient = new WebHDFSClient(hdfsHost);
+  const hdfsClient = new WebHDFSClient(hdfsHost, undefined, undefined, 80, '/webhdfs/api/v1');
```
Binyang2014: Hmm, here I mean if we set customized hdfsHost for local debug, the hdfsClient may need to be created like:
```js
new WebHDFSClient(hdfsHost, undefined, undefined, 50700, '/webhdfs/v1');
```
or
```js
new WebHDFSClient(hdfsHost, undefined, undefined, port, '/webhdfs/v1');
```

debuggy: you mean not use pylon?

PR #3066, opened 2019-07-04
Return empty command if no teamwise mounts

wangdian:


debuggy:
the ci was failed in this pr, please check it out

wangdian:
CI failed due to network error when building image. Restarting CI

PR #3068, opened 2019-07-04
fix clone job bug

Binyang2014:
Add workaround for comments in commands
fix clone job job


src/webportal/src/app/job-submission/job-submission.jsx
```diff
@@ -215,7 +215,10 @@ const JobSubmission = () => {
           .then((jobConfig) => {
             const [jobInfo, taskRoles, parameters] = getJobComponentsFromConfig(
               jobConfig,
+              {vcNames}
             );
+            const timestamp = Date.now().toString();
+            jobInfo.name = jobInfo.name + '_' + timestamp.substring(timestamp.length - 4);
```
wangdian: This will append job name, I remember we used to replace encoded string after last "_".

PR #3072, opened 2019-07-04
[Web Portal] display red border when a task role is invalid

sunqinzheng:


src/webportal/src/app/job-submission/components/tab-form.jsx
```diff
@@ -59,28 +61,44 @@ export class TabForm extends React.Component {
...
       return null;
     }
 
+    const {spacing, palette} = getTheme();
+    const {items} = this.props;
+    const {selectedIndex} = this.state;
+    const idx = this._getItemIndexByKey(itemProps.itemKey);
+    const item = items[idx];
+    const taskRolesObject = item.content.convertToProtocolFormat();
+    const {error} = Joi.validate(taskRolesObject, taskRolesSchema);
```
Binyang2014: May be we should also validate the prerequisites which associate with this taskRole here.

sunqinzheng: I think you mean docker info. Is there anything else could be edited in task-role section?

Binyang2014: I think that's enough, other fields can be validated by taskRole JOI. Thanks for this change.

PR #3080, opened 2019-07-05
disable submit if command is empty

Binyang2014:


Binyang2014:
> We need to tell user the reason in UI explicitly

Currently, we don't have good solution for this. I think keep as it is fine since we have a default value(the comments) in command filed. Customer will know what happens when he/she delete the all commands.

PR #3091, opened 2019-07-05
fix bug: redict v2 job to default submission papge if plugin not installed

Binyang2014:


src/webportal/src/app/job-submission/job-submission.jsx
```diff
@@ -244,6 +244,9 @@ const JobSubmission = () => {
               {vcNames}
             );
             jobInfo.name = generateJobName(jobInfo.name);
+            if (get(jobConfig, 'extras.submitFrom')) {
+              delete jobConfig.extras.submitFrom;
```
abuccts: should not delete

Binyang2014: The config is cloned from a plugin. But if resubmit, the config should indicate it submitted from the default submission page. So I think remove is the correct behavior

src/webportal/src/app/job/job-view/fabric/job-detail/conn.js
```diff
@@ -134,6 +134,11 @@ export async function cloneJob(rawJobConfig) {
   const plugins = window.PAI_PLUGINS;
   const pluginIndex = plugins.findIndex((x) => x.id === pluginId);
   if (pluginIndex === -1) {
+    // redirect v2 job to default submission page
+    if (isJobV2(rawJobConfig)) {
+      window.location.href = `/submit.html?${qs.stringify(query)}`;
```
sunqinzheng: I think we should still pop up a warning message to user.

Binyang2014: sure, will fix it

PR #3097, opened 2019-07-05
Binyli/docker

Binyang2014:


src/webportal/src/app/job-submission/components/docker-section.jsx
```diff
@@ -73,7 +78,7 @@ AuthTextFiled.propTypes = {
 const options = [
   {
     key: 'all',
-    text: 'all-in-one (image: ufoym/deepo:all)',
+    text: 'all-in-one+python3.6+gpu (image: ufoym/deepo:all-py36-cu100)',
     image: 'ufoym/deepo:all',
```
sunqinzheng: Image should also be `ufoym/deepo:all-py36-cu100`

Binyang2014: Thanks, will fix

PR #3100, opened 2019-07-07
sync webportal version with pai

debuggy:


src/webportal/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "pai-webportal",
-  "version": "0.0.0",
+  "version": "0.14.0",
```
abuccts: the version is labeled [here](https://github.com/microsoft/pai/blob/master/src/webportal/build/build-pre.sh#L25) before building webportal

debuggy:
The pr is deprecated and this issue will be fixed in another pr.

PR #3112, opened 2019-07-08
[Web Portal] fix job submission page's `jobRetryCount` and `taskRetryCount` field

sunqinzheng:


sunqinzheng:
> 
> 
> image
> 
> there are still bugs in `Min Failed/Success Instances` field with yaml editor

fixed

PR #3134, opened 2019-07-09
Binyli/bug fix

Binyang2014:


src/webportal/src/app/job-submission/components/docker-section.jsx
```diff
@@ -187,14 +187,14 @@ export const DockerSection = ({onValueChange, value}) => {
     if (optionKey !== 'customize-image' && isCutomizedImageEnabled) {
       setCutomizedImageEnabled(false);
     }
-  }, []);
+  }, [uri]);
```
sunqinzheng: I think the `isCustomizedImageEnabled` state should be determined by 'uri' props.
You can use a `useMemo` hook to save the value.

Binyang2014: `isCustomizedImageEnabled ` is not just depends on the `uri` pop. Customers can also change its' value. So this situation could happens: `uri` not changed but `isCustomizedImageEnabled ` is changed by customer.
Maybe use state is better for this

sunqinzheng: `useEffect` will be triggered each time page is rerendered. So if user input an existing docker option in the custom docker url text field like `ufoym/deepo:tensorflow-py36-cu100`, this hook will be triggered and `isCutomizedImageEnabled` will be set to `false`.

sunqinzheng:
By the way, the `ufoym/deepo:all` key in previous mapping function `getDockerImageOptionKey` is not matched with `options` object.

Binyang2014:
> By the way, the `ufoym/deepo:all` key in previous mapping function `getDockerImageOptionKey` is not matched with `options` object.

Thanks for this, notice that

PR #3135, opened 2019-07-09
Add launcher-type to service config. Use kube-runtime build in project.

wangdian:
Add launcher-type to service config
Use kube-runtime build in project.

### DIFF
--- a/deployment/quick-start/services-configuration.yaml.template
+++ b/deployment/quick-start/services-configuration.yaml.template
@@ -65,6 +65,8 @@ cluster:
 
 
 rest-server:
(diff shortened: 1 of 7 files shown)


deployment/quick-start/services-configuration.yaml.template
```diff
@@ -65,6 +65,8 @@ cluster:
 
 
 rest-server:
+#  # launcher type. k8s or yarn
```
abuccts: indent

wangdian: It's following the indent rule of comment's comment.

PR #3152, opened 2019-07-10
Fix the issue of failed to install the kubectl on single box.

RaphaelSun:
When I deployed the OpenPAI in a single box, I faced a permission denied problem. And it releases the error "deployment.k8sPaiLibrary.maintainlib.common : Failed to install kubectl on your dev-box".
After I checked the file kubectl-install.sh, I found that in line 40 the request mv does not combine with sudo. Therefore, it does not have permission when deploying kubernetes cluster with OpenPAI.
After I add sudo to "mv ./kubectl /usr/local/bin/kubectl", this problem solved.

### DIFF
--- a/deployment/k8sPaiLibrary/maintaintool/kubectl-install.sh
+++ b/deployment/k8sPaiLibrary/maintaintool/kubectl-install.sh
@@ -37,4 +37,4 @@ pathofusllocalbin="/usr/local/bin"
     mkdir -p /usr/local/bin
 }
 
-mv ./kubectl /usr/local/bin/kubectl
+sudo mv ./kubectl /usr/local/bin/kubectl


ydye:
Hi @RaphaelSun , only one question, what's the command , through which you deploy k8s?

RaphaelSun:
I deploy k8s by using "python paictl.py cluster k8s-bootup \
  -p ~/pai-config"

ydye:
> I deploy k8s by using "python paictl.py cluster k8s-bootup 
> -p ~/pai-config"

I think you could try the command to deploy OpenPAI. And I think adding sudo in front of the paictl will solve this issue too.
```
sudo python paictl.py cluster k8s-bootup -p ~/pai-config
```

abuccts:
> @abuccts What's you opinion?

Agree. And user should already run as root user in dev box.

abuccts:
@RaphaelSun Thanks for your contribution.
Could you run `sudo python paictl.py cluster k8s-bootup -p ~/pai-config` instead?

PR #3153, opened 2019-07-10
 Add Job ssh for kube-runtime. Create ssh key secret in cluster-config

wangdian:
Add Job ssh for kube-runtime. 
Create ssh key secret in cluster-config


src/kube-runtime/src/init/sshd.sh
```diff
@@ -20,22 +20,69 @@
...
-        echo "no private key or port provided" >&2
-    else
-        mkdir -p $CONFIG_DIR
+function prepare_ssh()
+{
+  mkdir /root/.ssh
+  touch /root/.ssh/authorized_keys
+  chmod 600 /root/.ssh/authorized_keys
+
+  mkdir -p /var/run/sshd
```
abuccts: do not need?

src/kube-runtime/src/init/sshd.sh
```diff
@@ -20,22 +20,69 @@
...
+    if [ -z "$PAI_CONTAINER_SSH_PORT" ] ; then
+        echo "no ssh port provided" >&2
+    else
+        prepare_ssh
+        prepare_job_ssh
+        prepare_user_ssh
+        start_ssh
     fi
 else
     echo "no sshd binary found" >&2
```
abuccts: install openssh server for user if there's no sshd in image?

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -255,6 +255,11 @@ const generateTaskRole = (taskRole, labels, config) => {
                   name: 'host-log',
                   mountPath: '/usr/local/pai/logs',
                 },
+                {
+                  name: 'job-ssh-secret-volume',
+                  readOnly: true,
+                  mountPath: '/etc/ssh-secret',
```
abuccts: maybe mount to `/usr/local/pai/secrets`?

src/kube-runtime/src/init/sshd.sh
wangdian: Better have this in case the files are not created.

src/kube-runtime/src/init/sshd.sh
wangdian: Ok

src/rest-server/src/models/v2/job/k8s.js
wangdian: I think mount to /etc/ssh-secret is fine because I don't think we should use ssh keys elsewhere.

wangdian: Changed mountPath to /usr/local/pai/ssh-secret. Although I think user should not use this key elsewhere.

PR #3154, opened 2019-07-10
Add PAI env variables in init scripts

abuccts:
Add PAI env variables in init scripts.

This is a temporal solution, will refine later.


src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -239,6 +239,10 @@ const generateTaskRole = (taskRole, labels, config) => {
                   name: 'PAI_CURRENT_CONTAINER_PORT',
                   value: `${Math.floor((Math.random() * 10000) + 10000)}`,
                 },
+                {
+                  name: 'PAI_CONTAINER_SSH_PORT',
+                  value: `${Math.floor((Math.random() * 10000) + 10000)}`,
+                },
```
wangdian: This is temporary solution, better add comments.

src/kube-runtime/src/parse.py
```diff
@@ -36,14 +36,35 @@ def get_container_port(envs, name):
 def export(k, v):
     print "export %s='%s'" % (k, v)
 
+def gen_static_env(framework):
```
wangdian: Static Envs should be generated in Rest-server so metrics can use them later.
This is temp solution.

PR #3158, opened 2019-07-11
Pre check vclist from yarn and remove the vc type group which is not in.

ydye:


src/rest-server/src/models/v2/group.js
```diff
@@ -412,15 +413,34 @@ const updateUserGroupAndVirtualCluster = async () => {
...
+      }
+    }
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.log(error);
+  }
+};
+
 if (config.env !== 'test') {
   (async function() {
```
debuggy: why wrap it with async function

ydye: All function is async, we should await it.

debuggy: however the groupTypeVCCheck is already async function, maybe you could use await directly

ydye: https://stackoverflow.com/questions/39679505/using-await-outside-of-an-async-function

PR #3167, opened 2019-07-11
Support stop framework api in k8s rest server

abuccts:
Support stop framework api in k8s rest server.

### DIFF
--- a/src/rest-server/src/models/v2/job/k8s.js
+++ b/src/rest-server/src/models/v2/job/k8s.js
@@ -423,6 +423,34 @@ const put = async (frameworkName, config, rawConfig) => {
   }
 };
 
+const execute = async (frameworkName, executionType) => {
+  // send request to framework controller
+  let response;
+  try {
+    response = await axios({
+      method: 'patch',
+      url: launcherConfig.frameworkPath(encodeName(frameworkName)),
+      headers: {
+        'Content-Type': 'application/merge-patch+json',
+      },
+      data: {
+        spec: {
+          executionType,
+        },
+      },
+    });
+  } catch (error) {
+    if (error.response != null) {
+      response = error.response;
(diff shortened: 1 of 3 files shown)


src/rest-server/src/routes/v2/job.js
```diff
@@ -39,6 +39,10 @@ router.route('/:frameworkName')
   /** GET /api/v2/jobs/:frameworkName - Get job */
   .get(controller.get);
 
+router.route('/:jobName/executionType')
+  /** PUT /api/v2/jobs/:frameworkName/executionType - Start or stop job */
+  .put(token.check, controller.execute);
```
yqwang-ms: patch instead of put

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -423,6 +423,34 @@ const put = async (frameworkName, config, rawConfig) => {
...
+  try {
+    response = await axios({
+      method: 'patch',
+      url: launcherConfig.frameworkPath(encodeName(frameworkName)),
+      headers: {
+        'Content-Type': 'application/merge-patch+json',
+      },
+      data: {
+        spec: {
+          executionType: `${executionType.charAt(0)}${executionType.slice(1).toLowerCase()}`,
```
yqwang-ms: Case:

image

src/rest-server/src/routes/v2/job.js
abuccts: this is rest server's api, used `PATCH` in k8s apiserver's request

src/rest-server/src/models/v2/job/k8s.js
abuccts: right, this is CamelCase

PR #3187, opened 2019-07-15
Update faq on how to move docker install path

wangdian:


docs/faq.md
```diff
@@ -129,3 +129,22 @@ Run "docker system df" to see how much space can be reclaimed. If the space is l
 - (3) Check /tmp directory. 
 
 Run command "du -h / | awk '$1~/[0-9]*G/{print $0}'" to list all the directory which consumes more than 1G spaces. Clean the data not used any more. 
+
+### Q: How to change Docker install path for PAI?
+
+A: If we install Docker on a small disk, when under disk pressure, the cleaner may activate and kill jobs to prevent system crash. We'd better to move Docker to a larger disk. We can change Docker install path through following steps:
```
squirrelsc: avoid to use "we". It can be rephrased like,

If default Docker cache path is set to a disk with small disk space, the path can be changed with below steps.

docs/faq.md
```diff
@@ -129,3 +129,22 @@ Run "docker system df" to see how much space can be reclaimed. If the space is l
...
+  ```sudo ./paictl.py cluster k8s-clean –p /path/to/config```
+4.  Modify your config file layout.yaml. Change "docker-data" to your new Docker path. See sample [layout.yaml](../blob/master/examples/cluster-configuration/layout.yaml#L55)  
+5.	Restart k8s cluster with updated config.  
+  ```sudo ./paictl.py cluster k8s-bootup –p /path/to/new/config```
+6.	Push the latest config to cluster.  
+  ```sudo ./paictl.py config push –p /path/to/config```
+7.	Restart all service.  
+  ```sudo ./paictl.py service start```
+
+Note: the legacy docker data file in the old path won’t be automatically cleaned. So if we want to free the disk space which is occupied by the old docker data, we should manually delete it.
```
squirrelsc: Please also add impact to cluster, like the cluster is unavailable during this change, and it may take long time due to pull all images again.

docs/faq.md
```diff
@@ -129,3 +129,22 @@ Run "docker system df" to see how much space can be reclaimed. If the space is l
 - (3) Check /tmp directory. 
 
 Run command "du -h / | awk '$1~/[0-9]*G/{print $0}'" to list all the directory which consumes more than 1G spaces. Clean the data not used any more. 
+
+### Q: How to change Docker install path for PAI?
+
+A: If we install Docker on a small disk, when under disk pressure, the cleaner may activate and kill jobs to prevent system crash. We'd better to move Docker to a larger disk. We can change Docker install path through following steps:
+
+1.	Change the Docker path on PAI node.
```
squirrelsc: Where is this setting?

docs/faq.md
```diff
@@ -129,3 +129,22 @@ Run "docker system df" to see how much space can be reclaimed. If the space is l
...
+A: If we install Docker on a small disk, when under disk pressure, the cleaner may activate and kill jobs to prevent system crash. We'd better to move Docker to a larger disk. We can change Docker install path through following steps:
+
+1.	Change the Docker path on PAI node.  
+2.	Stop ALL service of the cluster.  
+```sudo ./paictl.py service stop```
+3.	Destroy k8s cluster.  
+  ```sudo ./paictl.py cluster k8s-clean –p /path/to/config```
+4.  Modify your config file layout.yaml. Change "docker-data" to your new Docker path. See sample [layout.yaml](../blob/master/examples/cluster-configuration/layout.yaml#L55)  
+5.	Restart k8s cluster with updated config.  
+  ```sudo ./paictl.py cluster k8s-bootup –p /path/to/new/config```
```
squirrelsc: Why is it needed? Is this configured in OpenPAI configuration or docker configuration?

docs/faq.md
```diff
@@ -129,3 +129,22 @@ Run "docker system df" to see how much space can be reclaimed. If the space is l
 - (3) Check /tmp directory. 
 
 Run command "du -h / | awk '$1~/[0-9]*G/{print $0}'" to list all the directory which consumes more than 1G spaces. Clean the data not used any more. 
+
+### Q: How to change Docker install path for PAI?
```
squirrelsc: Do not use PAI, use OpenPAI.

docs/faq.md
wangdian: It's OpenPAI ocnfiguration

docs/faq.md
wangdian: Add reference to official Docker docs

docs/faq.md
wangdian: Done

docs/faq.md
wangdian: Done

docs/faq.md
wangdian: Done

docs/faq.md
```diff
@@ -130,21 +130,23 @@ Run "docker system df" to see how much space can be reclaimed. If the space is l
...
-A: If we install Docker on a small disk, when under disk pressure, the cleaner may activate and kill jobs to prevent system crash. We'd better to move Docker to a larger disk. We can change Docker install path through following steps:
+A: If default Docker cache path is set to a disk with small disk space, the path can be changed with below steps:
 
-1.	Change the Docker path on PAI node.  
-2.	Stop ALL service of the cluster.  
-```sudo ./paictl.py service stop```
-3.	Destroy k8s cluster.  
+1.	Stop ALL service of the cluster.  
+  ```sudo ./paictl.py service stop```
+2.	Destroy k8s cluster.
```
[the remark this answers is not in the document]
wangdian: I prefer separate step 3 and 4 into two steps. Because one is for Docker and the other one is for OpenPAI.
I can mention the docker path in step 4 is the configured path in step 3.

Change destroy to uninstall

docs/faq.md
wangdian: I prefer separate them. I will mention  step 3 configured path in step 4.

docs/faq.md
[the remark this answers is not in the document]
squirrelsc: I mean create an issue to simplify it, so that user can update configuration easier. And we have a chance to clean up old folder, if it's changed.

docs/faq.md
squirrelsc: I mean create an issue to simplify it, so that user can update configuration easier. And we have a chance to clean up old folder, if it's changed.

wangdian: I think this is an other topic and we need to discuss it later.
How to setup docker cache path is admin behavior. I think we'd better not touch it, just tell them when they change docker cache path, the OpenPAI config should be changed, too.
Any other suggestions about the doc?

PR #3216, opened 2019-07-18
[User & Group] restserver api doc update. 

ydye:


debuggy:
the api doc will be re-organized after this release, however I think this update could be as a quick patch for this release api

PR #3232, opened 2019-07-19
OpenPAI version 0.14.0 release

debuggy:


debuggy:
I will add a tag `v0.14.0` before this resolved commit. If there would be any hot fix version like `0.14.1`, it will be checked out from this tag and create a new branch called `pai-0.14.1`.

I think the release code should based on tag instead of branch.

debuggy:
close and merge from a new branch

PR #3248, opened 2019-07-24
Diawang/pure k8s log restserver

wangdian:


src/kube-runtime/src/runtime
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
abuccts: what if there's no `bash` in image?

src/kube-runtime/src/runtime.d/sshd.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
abuccts: same

src/kube-runtime/src/runtime.d/user.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
abuccts: same

src/kube-runtime/build/kube-runtime.dockerfile
```diff
@@ -20,11 +20,11 @@ FROM python:2.7-alpine3.8
 
 ARG BARRIER_DIR=/opt/frameworkcontroller/frameworkbarrier
 
-WORKDIR /usr/local/pai
+RUN mkdir -p /kube-runtime/src
```
abuccts: do not need, duplicate with `WORKDIR`

src/kube-runtime/build/kube-runtime.dockerfile
```diff
@@ -20,11 +20,11 @@ FROM python:2.7-alpine3.8
...
+WORKDIR /kube-runtime/src
 
 COPY src/ ./
 COPY --from=frameworkcontroller/frameworkbarrier:v0.3.0 $BARRIER_DIR/frameworkbarrier ./init.d
-RUN mkdir -p ./logs && \
-    chmod -R +x ./
+RUN chmod -R +x ./
 
-CMD ["/bin/sh", "-c", "/usr/local/pai/init"]
+CMD ["/bin/sh", "-c", "/kube-runtime/src/init"]
```
abuccts: You can move the sources in dockerfile

```suggestion
CMD ["/bin/sh", "-c", "mv ./* /usr/local/pai && /usr/local/pai/init"]
```

src/kube-runtime/src/runtime
wangdian: Ubuntu is our officially supported image, bash should be installed. Using /bin/sh may call dash shell and cause unexpected result.

PR #3251, opened 2019-07-25
match commented out sshd port

wyatuestc:
In ubuntu18.04, the configuration of the default port of sshd service is commented out like "#Port 22".   Thus, sed -i 's/Port.*/Port '$PAI_CONTAINER_SSH_PORT'/' missed  ’#‘ and it cannot start sshd service on $PAI_CONTAINER_SSH_PORT.

### DIFF
--- a/src/rest-server/src/templates/dockerContainerScript.mustache
+++ b/src/rest-server/src/templates/dockerContainerScript.mustache
@@ -137,7 +137,7 @@ function start_ssh_service()
   printf "%s %s\n" \
     "[INFO]" "start ssh service"
   cat /root/.ssh/{{ jobData.jobName }}.pub >> /root/.ssh/authorized_keys
-  sed -i 's/.Port.*/Port '$PAI_CONTAINER_SSH_PORT'/' /etc/ssh/sshd_config
+  sed -i 's/[# ]*Port.*/Port '$PAI_CONTAINER_SSH_PORT'/' /etc/ssh/sshd_config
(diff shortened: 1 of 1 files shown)


src/rest-server/src/templates/dockerContainerScript.mustache
```diff
@@ -137,7 +137,7 @@ function start_ssh_service()
   printf "%s %s\n" \
     "[INFO]" "start ssh service"
   cat /root/.ssh/{{ jobData.jobName }}.pub >> /root/.ssh/authorized_keys
-  sed -i 's/Port.*/Port '$PAI_CONTAINER_SSH_PORT'/' /etc/ssh/sshd_config
+  sed -i 's/.Port.*/Port '$PAI_CONTAINER_SSH_PORT'/' /etc/ssh/sshd_config
```
abuccts: need to work with the old `Port 22` line
```suggestion
  sed -i 's/[# ]*Port.*/Port '$PAI_CONTAINER_SSH_PORT'/' /etc/ssh/sshd_config
```

wyatuestc: it is better than my commit, thanks!

abuccts: could you help to update your PR branch?

wyatuestc: OK, thanks!

abuccts:
@wyatuestc Merged, thanks for you contribution.

PR #3252, opened 2019-07-25
[Rest Server] Fix auth file issue in protocol converter

abuccts:
Fix authentication file issue in protocol converter.


sunqinzheng:
BREAKING CHANGE
- The rest server should treat job config file as plain text, so accept header feature should not be implemented

PR #3259, opened 2019-07-25
[Kube Runtime] Add port conflict check

abuccts:
Add port conflict check in runtime.

### DIFF
--- a/src/kube-runtime/src/runtime.d/port.py
+++ b/src/kube-runtime/src/runtime.d/port.py
@@ -20,6 +20,7 @@
 from __future__ import print_function
 
 import os
+import re
 import sys
 import socket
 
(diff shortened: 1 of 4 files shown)


src/kube-runtime/src/runtime.d/port.py
```diff
@@ -0,0 +1,57 @@
...
+def check_port(portno):
+    """Check whether the port is in use.
+
+    Exit with code 200 if the port is already in use.
+
+    Args:
+        portno: Port number to check.
+    """
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    ret = sock.connect_ex(('localhost', portno))
```
mzmssg: why not use sock.connetc and catch exception

abuccts: because there's a `connect_ex` function which returns C `connect` system call's error code

PR #3260, opened 2019-07-26
[pylon] https location conf fix

ydye:


Gerhut:
Use `locations.conf` instead please which `.conf` is the default extension of nginx configuration.

ydye:
> Use `locations.conf` instead please which `.conf` is the default extension of nginx configuration.

done

PR #3268, opened 2019-07-28
[Rest Server] Add job route rewrite for k8s v1 api

abuccts:
Add job route rewrite for k8s v1 api.

Fixes job clone issue in k8s webportal.


mzmssg:
you doesn't implement full feature like `job.list` in v2 yarn model, so it will break yarn version?

PR #3284, opened 2019-07-30
refactor eslint of webportal

debuggy:


src/webportal/prettier.config.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
+  // Why include an unnecessary character at the end of every line? Break the habit (automatically)!
+  semi: false,
```
Binyang2014: Do we want to remove the semicolon? Seems most JS code styles use semicolon to avoid some errors.
refer: [airbnb code style](https://github.com/airbnb/javascript#semicolons)

debuggy: Actually we considered airbnb as a candidate rule, however there is a voice that the airbnb config is too much overcontrol in some places and contains more plugins than actually needed. So we choose standard instead.
Standard uses a semicolon-free style, and it has [the solution to avoid certain errors](https://standardjs.com/rules.html#semicolons). And since we use prettier as a formatter, it could [automatically resolve the issue for us](https://prettier.io/docs/en/rationale.html#semicolons).

abuccts:
I suggest to keep the semicolon and avoid `function`. I think we'd better vote for the controversial rules.

/cc @debuggy @sunqinzheng @Binyang2014 @Gerhut

src/webportal/prettier.config.js
debuggy: add semi as discussion

debuggy:
> I suggest to keep the semicolon and avoid function. I think we'd better vote for the controversial rules.

We have a discussion yesterday, and reach an agreement to keep semi. Avoid function you mean use lamda? If so, we have set this rule.

abuccts:
> > I suggest to keep the semicolon and avoid function. I think we'd better vote for the controversial rules.
> 
> @abuccts We have a discussion yesterday, and reach an agreement to keep semi. Avoid function you mean use lamda? If so, we have set this rule.

Use arrow function and avoid `function` keyword, instead of mixing both. Here's the [reference](https://eslint.org/docs/rules/func-style#allowarrowfunctions).

debuggy:
update this pr in #3604

PR #3297, opened 2019-08-01
add alert for gpu count changes

Binyang2014:
Add alert when node GPU count is changed. This may caused by driver issue.

Add a record rule to generate a new metric: `node:nvidiasmi_utilization_gpu:count`, which is calculated by: `count(nvidiasmi_utilization_gpu) by (instance)`

Add a new alert when `changes(node:nvidiasmi_utilization_gpu:count) > 0`


xudifsd:
I did not know this record, good to know.

PR #3309, opened 2019-08-02
Binyli/bug fix

Binyang2014:
Add `job_instance_uid` to distinguish the job with different retry number

For yarn version the id is applicationId, for k8s version, the id is FC_FRAMEWORK_ATTEMPT_INSTANCE_UID.

Currently, we can retrieve job_instance_uid from restAPI. Which field name is `app_id`

image

### DIFF
--- a/src/job-exporter/src/collector.py
+++ b/src/job-exporter/src/collector.py
@@ -511,6 +511,9 @@ class ContainerCollector(Collector):
         result_labels["role_name"] = inspect_info.role_name or "unknown"
         result_labels["task_index"] = inspect_info.task_index or "unknown"
 
+        # Used to distinguage jobs/tasks with same name but diffrent retry number
+        result_labels["job_instance_uid"] = inspect_info.job_instance_uid or "unknown"
+
         if inspect_info.gpu_ids:
             ids = inspect_info.gpu_ids.replace("\"", "").split(",")
             for id in ids:
--- a/src/job-exporter/src/collector.py
+++ b/src/job-exporter/src/collector.py
@@ -102,7 +102,7 @@ class ResourceGauges(object):
                 "job_name",
                 "role_name",
                 "task_index",
-                "job_instance_uid",
+                "job_instance_id", # Used to distinguish job instance with same name but defferent retry number.
                 ]
         self.service_labels = ["name"]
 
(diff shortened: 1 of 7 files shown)


src/job-exporter/src/collector.py
```diff
@@ -511,6 +512,9 @@ def parse_from_labels(inspect_info, gpu_infos):
         result_labels["role_name"] = inspect_info.role_name or "unknown"
         result_labels["task_index"] = inspect_info.task_index or "unknown"
 
+        # Used to distinguage jobs/tasks with same name but different retry number
```
xudifsd: distinguage typo

src/job-exporter/src/collector.py
```diff
@@ -101,7 +101,8 @@ def __init__(self):
                 "username",
                 "job_name",
                 "role_name",
-                "task_index"
+                "task_index",
+                "app_id",
```
xudifsd: please use a more specific name, maybe `retry_count`.

src/job-exporter/src/collector.py
Binyang2014: Thanks, will fix it

src/job-exporter/src/collector.py
Binyang2014: It's not a `retry_count`. It's a job instance id such as `application_1553664769226_0080`. This id will be different in each job retry.
I use `app_id` here because it's same with the `app_id` filed in PAI restAPI. If we want to support multi-platform (such as DLWS), I can rename it to `job_instance_uid`.

xudifsd: I see.

xudifsd: Maybe job_instance_id, or instance_id. And please add comment here.

mzmssg: Why use `uid`? it usually means user id

Binyang2014: Means unique id. Will change to `job_instance_id` here.

mzmssg: Actually I think `attempt_id` is better, but it might be messy with yarn's attempt.

src/job-exporter/src/collector.py
```diff
@@ -102,7 +102,7 @@ def __init__(self):
                 "job_name",
                 "role_name",
                 "task_index",
-                "job_instance_uid",
+                "job_instance_id", # Used to distinguish job instance with same name but defferent retry number.
```
xudifsd: different, typo

Binyang2014: fixed

PR #3330, opened 2019-08-07
[Rest-server] oidc config configurable

ydye:


ydye:
> better to have a default value

Default value is : /auth-configuration/oidc.yaml

Issue #3325, opened 2019-08-06
enable ssh deployment that supports ed25519


fanyangCS:
**What would you like to be added**:
enable ssh deployment that supports ed25519

**Why is this needed**:
some host only allows ed25519, instead of rsa

**Without this feature, how does the current module work**:

**Components that may involve changes**:

ydye:
Install paramiko through pip？

fanyangCS:
@ydye, yes if that works.

abuccts:
It's better to upgrade to Python 3, Python 2 and many related packages are deprecated.

PR #2825, opened 2019-05-24
Fix ssh issues

aboettcher:
Hi,

this pull request shows the changes we had to made to deploy a single node setup (v0.12.0). I am not sure there is really a bug or we did something wrong during deployment. It might be helpful for others to see where we struggled and what we changed to get it running.

PR #3332, opened 2019-08-07
 [Deployment] Support ECC private keys for SSH in deployment

abuccts:
* Upgrade paramiko to 2.x
* Support ECC private keys for SSH in deployment

Solves #3325


ydye:
https://github.com/microsoft/pai/pull/2825 This is a relative PR.

abuccts:
@ydye That PR is to fix "ssh key file not found" exception and not related.

@fanyangCS Using the latest version is OK, otherwise we need to fix all packages' versions.

Issue #3275, opened 2019-07-29
[AAD] Decouple PAI internal resource from external user/groups 


mzmssg:
In current AAD, we have a strong requirement that every vc need a dedicated group, which means resource acl depends on external user/groups. This issue is to decouple them.

## Design

Current user/group model:
![image](https://user-images.githubusercontent.com/11887940/62200527-35d4be80-b3b8-11e9-8938-4ff02796a843.png)

New user/group model:
![image](https://user-images.githubusercontent.com/11887940/62200554-42591700-b3b8-11e9-953c-69564ae82e03.png)
 
From implementation, basically, we will move virutal cluster out of user and into group schema: users belong to some groups, and groups define resource.

User schema(generated/updated from aad):

Key | Value | Description
-- | -- | --
username | String |  
email | String |  
group List | String array |  
password | String |  
extension | Map | ~virtualCluster: vc1,   vc2~ <br>  

Group schema(maintained by admin):

Key | Value | Description
-- | -- | --
groupName | String |  
externalName | String |  
extension | Map |  ~groupType:   vc/admin/stroage~ <br> acls: { <br>   virtualClusters: vc1,   vc2,  <br> storageConfigs:   storageConfig1, storageConfig2,  <br> admin: true/false, <br> etc<br>} 

Login: only groups with vc access could login.

## Items:

- [x] redesign user/group schema
- [x] job submission: submit,delete
- [x] user management: add/delete user, password change 
- [x] group management: add/delete group
- [x] vc management: add/delete vc
- [x] new api: update group acls
- [x] migrate existing schema to new schema
- [x] adapt webportal
- [x] fix unit test (doing)
- [ ] ~update storage logic with new schema~
- [ ] ~TBD~

PR #3336, opened 2019-08-08
Decouple resource acls from external group

mzmssg:
#3275


src/rest-server/src/config/v2/group.js
```diff
@@ -44,7 +44,7 @@ const groupCreateInputSchema = Joi.object().keys({
     .empty(''),
   extension: Joi.object()
     .pattern(/\w+/, Joi.required())
-    .default({}),
```
ydye: Please keep it

src/rest-server/src/config/v2/user.js
```diff
@@ -74,7 +74,7 @@ const userCreateInputSchema = Joi.object().keys({
     .required(),
   extension: Joi.object()
     .pattern(/\w+/, Joi.required())
-    .default({}),
```
ydye: Keep it

mzmssg: For object, it's better to keep a empty default, see https://github.com/hapijs/joi/issues/1135

ydye: Can you test this API and request it without the extension field. Could share me with the result?

src/rest-server/src/utils/manager/user/user.js
```diff
@@ -33,13 +33,13 @@ const userSchema = Joi.object().keys({
   password: Joi.string()
     .empty('')
     .default(''),
-  extension: Joi.object().pattern(/\w+/, Joi.required()).required(),
+  extension: Joi.object().pattern(/\w+/, Joi.required()).default(),
```
ydye: This filed must be a ```{}```, if you think replace ```required``` with default is better. Please set default value as ```{}``` (empty dict). This schema constraint the final structure in secret, pls check the data strictly here. If you wanna set any default value logic, you'd better put them in the controller or model level.

src/rest-server/src/utils/manager/group/group.js
```diff
@@ -29,13 +29,23 @@ const groupSchema = Joi.object().keys({
     .default(''),
   extension: Joi.object()
     .pattern(/\w+/, Joi.required())
-    .required(),
+    .keys({
+      acls: Joi.object()
```
ydye: From the perspective of OpenPAI, I think this change is fine, because we could strictly ask each opertion add the acl filed to every group, or the operation will fail.

ydye: But, I wonder if it is possible to remove this check into model level? Or in the next time, if we add new value into extension, we will have to modify the schema and add new entry into it again. Or do you have a better solution?

src/rest-server/src/utils/common.js
```diff
@@ -0,0 +1,37 @@
...
+// to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+// The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING
+// BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+const assignValueByKeyarray = (obj, keyarray, value) => {
```
ydye: Pls add jsdoc here.

src/rest-server/src/routes/v2/group.js
```diff
@@ -36,6 +36,10 @@ router.route('/:groupname/extension')
 /** Put /api/v2/group/:groupname/extension */
   .put(token.check, param.validate(groupInputSchema.groupExtensionUpdateInputSchema), groupController.updateGroupExtension);
 
+router.route('/:groupname/extension/*')
+/** Put /api/v2/group/:groupname/extension/:attribute */
+  .put(token.check, groupController.updateGroupExtensionAttr);
```
ydye: param.validate

src/rest-server/src/models/v2/user.js
```diff
@@ -86,35 +65,19 @@ const createUserIfNonExistent = async (username, userValue) => {
...
-      // admin has the permission of all groups.
-      ret = true;
-    }
-    return ret;
-  } catch (error) {
-    throw error;
-  }
+const getUserVCs = async (username) => {
+  const userItem = await getUser(username);
+  return groupModel.getGroupsVCs(userItem.grouplist);
```
ydye: The function named getUserVCs returns grouplist?

src/rest-server/src/models/v2/user.js
```diff
@@ -19,59 +19,38 @@
...
-  }
+  return await crudUser.remove(username, crudConfig);
 };
 
+// it's an inplace encrypt!
 const getEncryptPassword = async (userValue) => {
-  try {
     await user.encryptUserPassword(userValue);
     return userValue;
-  } catch (error) {
```
ydye: This time is fine. But next time, I think the unrelative change should be in another PR.

src/rest-server/deploy/legacy_user_migrate.py
```diff
@@ -138,6 +138,60 @@ def secret_data_prepare_v2_group(self, groupname):
         post_data_dict['data'] = group_dict
         return post_data_dict
 
+    def load_v2_groups(self):
+        ns_pai_group_list = self.list_all_secrets_from_namespace(self.secret_ns_group_v2)
+        decode_group_list = []
+        vc_set = set()
+        for group in ns_pai_group_list:
+            meta_dict = dict()
+            meta_dict['name'] = bytes.fromhex(group.metadata.name).decode('utf-8')
```
ydye: Tag 1

src/rest-server/deploy/legacy_user_migrate.py
```diff
@@ -138,6 +138,60 @@ def secret_data_prepare_v2_group(self, groupname):
...
+                extension_dict['acls']['virtualClusters'].extend(all_vcs)
+            elif extension_dict['groupType'] == 'vc':
+                extension_dict['acls']['virtualClusters'].append(data_dict['data']['groupname'])
+
+        return data_dict, True
+
+    def update_v2_group(self, data_dict):
+        post_data_dict = {
+            'metadata': {
+                'name': data_dict['metadata']['name'].encode('utf-8').hex()
```
ydye: In tag1, I think it's unnecessary to do decode and then encode here. Why not directly use the value from k8s?

src/rest-server/deploy/legacy_user_migrate.py
```diff
@@ -138,6 +138,60 @@ def secret_data_prepare_v2_group(self, groupname):
...
 
+    def load_v2_groups(self):
+        ns_pai_group_list = self.list_all_secrets_from_namespace(self.secret_ns_group_v2)
+        decode_group_list = []
+        vc_set = set()
+        for group in ns_pai_group_list:
+            meta_dict = dict()
+            meta_dict['name'] = bytes.fromhex(group.metadata.name).decode('utf-8')
+            group_dict = {
+                'groupname': str(base64.b64decode(group.data['groupname'].encode('utf-8')), 'utf-8'),
```
ydye: Are there any update or change happening at groupname, description or externalName? If not, you can use the value from k8s directly. And I think decode is unnecessary.

src/rest-server/deploy/legacy_user_migrate.py
```diff
@@ -295,32 +368,39 @@ def main():
...
+        vc_set = transferCli.vc_set
+        for vc in vc_set:
+            secret_post_data = transferCli.secret_data_prepare_v2_group(vc)
+            transferCli.create_secret_group_v2(secret_post_data)
+        logger.info('Legacy user data transfer from namespace v1 to namespace v2 successfully')
     else:
-      logger.info("Legacy data has already been transferred from v1 to v2. Skip it.")
+        logger.info("Legacy data has already been transferred from v1 to v2. Skip it.")
+
+    group_list, vc_set = transferCli.load_v2_groups()
```
ydye: Will this logic happen everytime if we start rest-server? I have some concerns about the result if this script executed several times. Have you consider it?

src/rest-server/src/controllers/v2/group.js
```diff
@@ -104,6 +104,29 @@ const updateGroupExtension = async (req, res, next) => {
   }
 };
 
+const updateGroupExtensionAttr = async (req, res, next) => {
+  try {
+    const groupname = req.params.groupname;
+    const attrs = req.params[0].split('/');
```
ydye: params[0]？？？？？？？ Can you provide some schema about this request?

src/rest-server/src/controllers/v2/group.js
```diff
@@ -104,6 +104,29 @@ const updateGroupExtension = async (req, res, next) => {
...
+    const updateData = req.body.data;
```
ydye: The same problem. We have no idea about this data.

src/rest-server/src/controllers/v2/group.js
```diff
@@ -168,4 +191,5 @@ module.exports = {
   updateGroupDescription,
   updateGroupExternalName,
   deleteGroup,
+  updateGroupExtensionAttr,
```
ydye: So this function can only update one key field in extension, right?

src/rest-server/src/models/v2/group.js
```diff
@@ -21,7 +21,6 @@ const authConfig = require('@pai/config/authn');
 const secretConfig = require('@pai/config/secret');
 const adapter = require('@pai/utils/manager/group/adapter/externalUtil');
 const config = require('@pai/config/index');
-const userModel = require('@pai/models/v2/user');
```
ydye: Is it necessary to do this change? Because you do the same operation in the function.

src/rest-server/src/models/v2/group.js
```diff
@@ -31,236 +30,182 @@ const crudConfig = crudGroup.initConfig(process.env.K8S_APISERVER_URI);
...
 
-const createGroup = async (groupname, groupValue) => {
-  try {
-    const ret = await crudGroup.create(groupname, groupValue, crudConfig);
-    await addGroupIntoAdminUser(groupname, groupValue);
-    return ret;
-  } catch (error) {
-    throw error;
-  }
+const getGroupVCs = async (groupname) => {
```
ydye: getGroupsVCs(['groupname']) === getGroupVCs('groupname') ?

src/rest-server/src/models/v2/group.js
```diff
@@ -31,236 +30,182 @@ const crudConfig = crudGroup.initConfig(process.env.K8S_APISERVER_URI);
...
+  return Promise.all(grouplist.map(getGroup));
+};
+
+const batchUpdateGroups = async (groupItems) => {
+  return await Promise.all(groupItems.map(async (groupItem) => {
+    await updateGroup(groupItem.groupname, groupItem);
+  }));
+};
+
+const getVCsWithGroupinfo = async (groupItems) => {
```
ydye: getVCsWithGroupInfo

src/rest-server/src/models/v2/group.js
```diff
@@ -313,41 +258,31 @@ const initGrouplistInCfg = async () => {
...
-    logger.error('Failed to create admin group configured in configuration.');
+    logger.error('Failed to create group configured in configuration.');
     // eslint-disable-next-line no-console
     console.log(error);
   }
 };
 
-const createAdminUser = async () => {
+const createDefaultAdminUser = async () => {
+  const userModel = require('@pai/models/v2/user');
```
ydye: Requre in the top

ydye:
- [ ] Deploy this code with basic mode, test all feature
- [ ] Deploy this code with AAD mode, test all feature

src/rest-server/src/controllers/v2/user.js
```diff
@@ -20,13 +20,25 @@ const userModel = require('@pai/models/v2/user');
...
 
 const getUser = async (req, res, next) => {
   try {
     const username = req.params.username;
     const userInfo = await userModel.getUser(username);
-    userInfo['admin'] = userInfo.grouplist.includes(authConfig.groupConfig.adminGroup.groupname);
-    userInfo['virtualCluster'] = userInfo['extension']['virtualCluster'] ? userInfo['extension']['virtualCluster'] : [];
+    const groupItems = await groupModel.getListGroup(userInfo.grouplist);
+    userInfo['admin'] = groupModel.getAdminWithGroupinfo(groupItems);
+    userInfo['virtualCluster'] = await groupModel.getVCsWithGroupinfo(groupItems);
```
ydye: getVCListWithGroupInfo will be much better

src/rest-server/src/controllers/v2/user.js
```diff
@@ -40,26 +52,33 @@ const getUser = async (req, res, next) => {
...
   try {
     const userList = await userModel.getAllUser();
-    let retUserList = [];
-    for (let userItem of userList) {
-      userItem['admin'] = userItem.grouplist.includes(authConfig.groupConfig.adminGroup.groupname);
-      userItem['virtualCluster'] = userItem['extension']['virtualCluster'] ? userItem['extension']['virtualCluster'] : [];
-      delete userItem['password'];
-      retUserList.push(userItem);
+    const groupInfo = await groupModel.getAllGroup();
+    const allVClist = Object.keys(await vcModel.prototype.getVcListAsyc());
```
ydye: vcList or AllVC

src/rest-server/src/controllers/v2/user.js
```diff
@@ -40,26 +52,33 @@ const getUser = async (req, res, next) => {
...
+      delete userItem.password;
+      return userItem;
+    }));
     return res.status(200).json(retUserList);
   } catch (error) {
     return next(createError.unknown(error));
   }
 };
 
+// OIDC
```
ydye: JSDOC is better

src/rest-server/deploy/legacy_user_migrate.py
mzmssg: They are different funcs, we should make them more genernal. In other words, this func shouldn't assume input is a partly encoded object.

src/rest-server/deploy/legacy_user_migrate.py
mzmssg: No, it will be checked and only be converted once, see line 164

src/rest-server/src/controllers/v2/group.js
mzmssg: I want to provide an api to update single key in a nested extension, like `/extension/acls/virtualClusters` or `/extension/acls/virtualClusters` or further `/extension/attr_a/attr_b/attr_c...`. In express, I could only find the way of match route path by '*' and then reference it by `req.params[0]`. Do you have better suggestion?

src/rest-server/src/controllers/v2/group.js
mzmssg: Because we don't know the updated attrs or its type, here I use data as a general name.

src/rest-server/src/controllers/v2/group.js
mzmssg: Yes, more specificly, only one key in a nested extension

src/rest-server/src/controllers/v2/user.js
```diff
@@ -270,9 +256,6 @@ const addGroupIntoUserGrouplist = async (req, res, next) => {
     const username = req.params.username;
     const groupname = req.body.groupname;
     let userInfo = await userModel.getUser(username);
-    if (userInfo.grouplist.includes(authConfig.groupConfig.adminGroup.groupname)) {
```
ydye: Allow the modify for admin user?

src/rest-server/src/models/v2/group.js
mzmssg: Yes, getGroupVCs is frequently used, so I provide a wrapper for it.

src/rest-server/src/models/v2/group.js
mzmssg: In currently implementation, there exists a circular dependencies: group model depends on user model (when deleting group), user model depends on group mode (when checking admin and vc). CommonJS can't handle it correctly. As a workaround, we need require model on demand, instead at top.

mzmssg: To solve it, I think we could remove the group model dependency, move related logic like deleting user grouplist into controller.

src/rest-server/src/config/v2/user.js
mzmssg: Create user:
image

Get user:
image

mzmssg: For object, we should use `default()` instead of `default({})`, they could achive the same result, using `default()` is less possible to make mistakes.

src/rest-server/src/config/v2/group.js
mzmssg: Same reason as below

src/rest-server/src/utils/manager/user/user.js
mzmssg: Same reason as above

src/rest-server/src/controllers/v2/group.js
mzmssg: See https://stackoverflow.com/questions/23876045/get-rest-of-the-path-from-requested-route-in-nodejs

src/rest-server/deploy/legacy_user_migrate.py
mzmssg: See above

src/rest-server/src/models/v2/group.js
```diff
@@ -31,236 +30,182 @@ const crudConfig = crudGroup.initConfig(process.env.K8S_APISERVER_URI);
...
+// TODO: replace updateGroup2ExnternalMapper
+// const updateExternalName2Groupname = async () => {
+//   const groupList = await getAllGroup();
+//   externalName2Groupname.clear();
+//   for (const groupItem of groupList) {
+//     externalName2Groupname[groupItem.externalName] = groupItem.groupname;
+//   }
+// };
+
+// hack for basic mode, assume every vc have a same name group
```
ydye: JSDOC

src/rest-server/src/models/v2/group.js
```diff
@@ -358,74 +293,64 @@ const createAdminUser = async () => {
...
+//       logger.info('No user\' grouplist need to be updated.');
+//     }
+//   } catch (error) {
+//     // eslint-disable-next-line no-console
+//     console.log(error);
+//   }
+// };
 
-const groupTypeVCCheck = async () => {
+// delete non-exist vc from all groups
```
ydye: jsdoc

src/rest-server/src/utils/manager/group/group.js
mzmssg: We already have two validation and defaulting: input from route, output to k8s. I'm not sure it's a good idea to introduce more default logic at other level. I think setup all defaulting at the bottom could make upper level more clean
For compatible schema change, I think it's fine, especially we are changing extension field, which is designed for changable.

src/rest-server/src/models/v2/group.js
mzmssg: Workaround for circular dependency, see http://www.ruanyifeng.com/blog/2015/11/circular-dependency.html.

User model require group model, then if we require user model in group model in turn, here will get a empty userModel.

src/rest-server/src/controllers/v2/user.js
mzmssg: There are two questions:
1. Whether allow to update user permission by update grouplist.
Admin is a group attr, not a user attr, so allow to update user admin by updating his groups make sense. If we stick origin experience, I'm fine, but the logic is disallowing to add/remove *admin group* into/from user grouplist.
2. Whether allow to update admin grouplist if no admin role change.
I think it should be. E.g If we support different group storage preference(not permission), we should also allow to customize such preference for admin, the solution is add admin into a specific group configured his preference.

ydye: sync with @scarlett2018

src/rest-server/src/models/v2/group.js
ydye: Create issue about this

src/rest-server/src/utils/manager/group/group.js
ydye: pending design？

src/rest-server/src/controllers/v2/user.js
mzmssg: Renamed getVCsWithGroupInfo

src/rest-server/src/controllers/v2/user.js
mzmssg: Keep allVClist

src/rest-server/src/controllers/v2/user.js
mzmssg: Add jsdoc for specific funcs, the comments will be kept. Will mark all jsdoc comment resolved

src/rest-server/src/models/v2/group.js
mzmssg: After checkin

PR #3370, opened 2019-08-13
Migrate CI from Travis to GitHub Actions

abuccts:
Migrate CI from Travis to GitHub Actions.


abuccts:
> Curiosity about how build banner will display after removing Travis.

Do you mean the badges on README? We should change to GitHub Actions after deleted Travis.

abuccts:
> Will it block the code merge if some checks are failed?

No, need to config settings before deleting `.travis.yml`.

PR #3373, opened 2019-08-13
add python script

hzy46:


contrib/python-sdk/openpaisdk/job.py
```diff
@@ -503,6 +512,7 @@ def connect_jupyter_interactive(self, status: dict=None):
     "one_liner": 'py-sdk-one-liner',
     "interactive_nb": 'py-sdk-notebook-interactive',
     "batch_nb": 'py-sdk-notebook-batch',
+    "python_script": 'py-sdk-python-script',
```
mydmdm: let's use `script_nb` to indicate that this tag is for `notebook (nb)`

hzy46: OK

Issue #3358, opened 2019-08-12
Job compatibility between yarn and k8s based PAI


abuccts:
Job compatibility comparisons between YARN and k8s based PAI.

TODO:
- [x] shared memory
- [x] InfiniBand

PR #3380, opened 2019-08-15
[Rest Server] Set shared memory size for k8s pod jobs

abuccts:
Set shared memory size for k8s pod jobs #3358.

### DIFF
--- a/src/rest-server/src/models/v2/job/k8s.js
+++ b/src/rest-server/src/models/v2/job/k8s.js
@@ -208,6 +208,11 @@ const generateTaskRole = (taskRole, labels, config) => {
       count: ports[port],
     };
   }
+  // get shared memory size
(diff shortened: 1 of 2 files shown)


src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -208,6 +208,11 @@ const generateTaskRole = (taskRole, labels, config) => {
       count: ports[port],
     };
   }
+  // get shared memory size
+  let shmMB = 512;
```
mzmssg: Could we on demand mount, don't set default value?

abuccts: even if we don't set the `shm-size`, there's also a default value in Docker (`shm-size=64m`)
just enlarge the default value here in case user don't know to set it

Issue #3378, opened 2019-08-15
Test aad and resource acl on k8s


mzmssg:
We have converge yarn and k8s code, but some features are not fully converged or tested on k8s. 

- [x] Tested aad login
- [ ] Tested aad acl, like admin, vc.
   known item:
   - [x] Add vc checker in k8s job submission
   - [x] Use v2 vc management
   - [ ] Update groups in k8s vc management

mzmssg:
Closed, tracked by #3382

PR #3410, opened 2019-08-20
Check available vc in k8s job submission

mzmssg:
#3378


src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -24,6 +24,7 @@ const status = require('statuses');
 const runtimeEnv = require('./runtime-env');
 const launcherConfig = require('@pai/config/launcher');
 const createError = require('@pai/utils/error');
+const userModelV2 = require('@pai/models/v2/user');
```
abuccts: you can remove v2 in variable name

mzmssg: Done

PR #3411, opened 2019-08-20
Binyli/home page

Binyang2014:
For admin, show VC utilization 
Annotation 2019-08-20 163101

Add alerts notification for admin
Annotation 2019-08-20 163200


src/webportal/src/app/components/util/job.js
```diff
@@ -94,3 +96,43 @@ export function getJobModifiedTimeString(job) {
...
+      return acc;
+    }
+    const lowGpuUsageJob = {...cur};
+    lowGpuUsageJob['gpuUsage'] = gpuUsageInfo.gpuUsage;
+    acc.push(lowGpuUsageJob);
+    return acc;
+  }, []);
+
+  // Merge long run jobs and low GPU usage jobs
+  const abnormalJobs = [...longRunJobs];
```
sunqinzheng: A shallow copy is created here. The original data will be edited.
I'm not sure if it is by design.

src/webportal/src/app/home/home/abnormal-job-list.jsx
```diff
@@ -0,0 +1,237 @@
...
+          </div>
+        );
+      },
+    },
+  ];
+
+  const stopJob = useCallback((job) => {
+    userAuth.checkToken((token) => {
+      const {name, username} = job;
+      fetch(`${webportalConfig.restServerUri}/api/v1/jobs/${username}~${name}/executionType`, {
```
sunqinzheng: I think we should extract the api call function to conn.js

src/webportal/src/app/layout/alerts.jsx
```diff
@@ -0,0 +1,120 @@
+import c from 'classnames';
+import {Panel, List, mergeStyleSets, getFocusStyle, getTheme, initializeIcons, PanelType, Stack, StackItem} from 'office-ui-fabric-react';
+import React, {useCallback, useState, useEffect} from 'react';
+
+import {initTheme} from '../components/theme';
+import config from '../config/webportal.config';
+
+initTheme();
+initializeIcons();
```
sunqinzheng: The initialization functions and missed polyfills (`corejs`, `fetch`...) should be in the root file `layout.js` instead of a component.

sunqinzheng: `initializeIcons` should only be called once but every page rely on `layout.component.js`.
Maybe we need to refactor the init function calls

src/webportal/src/app/home/home/virtual-cluster-statistics.jsx
```diff
@@ -0,0 +1,206 @@
...
+
+const VirtualClusterStatistics = ({style, virtualClusters}) => {
+  const vcNames = Object.keys(virtualClusters);
+  const {spacing} = getTheme();
+  const vcList = vcNames.map((vcName) => {
+    return {name: vcName, ...virtualClusters[vcName]};
+  });
+
+  return (
+    <Card className={t.ph5} style={{paddingRight: spacing.m, ...style}}>
```
sunqinzheng: I think we need a horizontal scrollbar if the screen is too small (mobile devices)

P.S. For mobile devices I think we could use a sticky header with an expand/collapse button and show all vc items instead of double scrolbars. But this approach needs extra efforts (implmentation, blocked by admin-lte's responsive breakpoint...), I think we could leave it in backlog and discuss it later.

src/webportal/src/app/components/util/job.js
Binyang2014: It's a mistake, thanks

src/webportal/src/app/home/home/abnormal-job-list.jsx
Binyang2014: I will extract it to conn.js. Actually I think we need to extract these common APIs to a common lib. Maybe JS SDK is doing such things.

src/webportal/src/app/layout/alerts.jsx
Binyang2014: Remove these init function. Currently, it works. We can refactor the init function call when we have time.

src/webportal/src/app/home/home/virtual-cluster-statistics.jsx
Binyang2014: Agree, we can create an issue for this

sunqinzheng: #3431

PR #3412, opened 2019-08-20
 Change the code style

AosChen:


contrib/profiler/profiler.py
```diff
@@ -0,0 +1,364 @@
...
+        container_blk_filelist.append('/sys/fs/cgroup/blkio/blkio.throttle.io_service_bytes')
+        container_net_file = '/proc/net/dev'
+    while True:
+        # 1st info about I/O and network
+        read_bytes1 = get_disk_read_bytes(container_blk_filelist)
+        write_bytes1 = get_disk_write_bytes(container_blk_filelist)
+        [network_receive1, network_transmit1] = get_network_bytes(container_net_file)
+
+        # CPU usage will cost time('period') running
+        cpu_usage = get_cpu_percent(container_cpu_filelist, period)[-1]
```
Binyang2014: Can we move the `time.sleep(period)`  from `get_cpu_percent` to the while loop. Then other collection will not depends on `get_cpu_percent` function.
I know we need to collect cpu usage twice to get the rate. We can set a `initialized` flag to solve it.

PR #3415, opened 2019-08-20
Converge virtual cluster v1/v2 api and model

mzmssg:
This pr
1. merge aad changes on v1 vc controller into v2 vc controller
2. move v1 vc model from models/vc.js to models/v1/virtual-clusters
3. change all internal vc call to v2 model
4. change external vc call(webportal) to v2 api


src/rest-server/src/models/v2/group.js
```diff
@@ -349,7 +349,7 @@ const createDefaultAdminUser = async () => {
 const deleteNonexistVCs = async () => {
   try {
     const groupInfoList = await getAllGroup();
-    const vcList = await vcModel.prototype.getVcListAsyc();
+    const vcList = await vcModel.list();
```
ydye: So the list function is an async or promise?

mzmssg: Yes, it's promise

PR #3423, opened 2019-08-21
[Webportal] Fix webpack conflict

abuccts:
Resolve webpack chunks runtime conflicts between webportal and plugins,
use another `jsonpFunction` name for webportal.

Issue left in #3244.

Reference: https://webpack.js.org/configuration/output/#outputjsonpfunction

### DIFF
--- a/src/webportal/config/webpack.common.js
+++ b/src/webportal/config/webpack.common.js
@@ -75,6 +75,7 @@ const config = (env, argv) => ({
   output: {
     path: helpers.root('dist'),
     filename: 'scripts/[name].bundle.js',
+    jsonpFunction: 'webportalWebpackJsonp',
   },
   resolve: {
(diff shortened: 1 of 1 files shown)


src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -137,7 +137,7 @@ const convertTaskDetail = async (taskStatus, ports, userName, jobName, taskRoleN
     taskState: convertState(
       taskStatus.state,
       completionStatus ? completionStatus.code : null,
-      taskStatus.retryPolicy.retryDelaySec,
+      taskStatus.retryPolicyStatus.retryDelaySec,
```
Gerhut: Is this commited by mistake?

abuccts: it's another bug

PR #3429, opened 2019-08-22
Fix bugs in pure k8s

abuccts:
* ignore modified files during deployment in git
* fix typo `retryPolicy` -> `retryPolicyStatus`
* fix incompatible default task role name
* rename "Go to Yarn Tracking Page"
* convert task role names in log path
* change `alert` to `prompt` in view token

### DIFF
--- a/src/dev-box/.gitignore
+++ b/src/dev-box/.gitignore
@@ -1,0 +1,1 @@
+build/kubectl-install.sh
--- a/src/pylon/.gitignore
+++ b/src/pylon/.gitignore
@@ -1,1 +1,2 @@
+deploy/https-config/
 .env*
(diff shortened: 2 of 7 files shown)


mzmssg:
Change the prefix to a legal name? https://github.com/microsoft/pai/blob/19817a0170b72d44ed9ce0fe2fe6f430c0d5b3f3/src/webportal/src/app/job-submission/components/task-roles.jsx#L11

mzmssg:
By the way, why not change restserver validation?

abuccts:
> By the way, why not change restserver validation?

Why need to change validation? The previous name still works.

mzmssg:
> fix incompatible default task role name

Not sure your `still works`. If it works, why we need to fix the incompatibility?

PR #3437, opened 2019-08-23
Add cluster-type and related config

wangdian:
Use [service name]_[cluster type].yaml file for default config for cluster type.
If not found cluster type config file, use [service name].yaml instead


ydye:
why not add this type into cluster.coomon?

wangdian:
The cluster-type is in cluster.common

cluster:
  common:
      cluster-id: pai_ut
      cluster-type: yarn

src/cluster/config/cluster.py
```diff
@@ -50,6 +50,8 @@ def merge_service_configuration(self, overwrite_srv_cfg, default_srv_cfg):
     def validation_common(self, common_configuration):
         if "cluster-id" not in common_configuration:
             return False, "cluster-id is miss in service-configuration.yaml -> cluster -> common -> cluster-id"
+        if "cluster-type" not in common_configuration or common_configuration["cluster-type"] not in ["yarn", "k8s"]:
```
abuccts: does `cluster-type` has a default value for the previous config without this field?

src/rest-server/config/rest-server_k8s.yaml
```diff
@@ -0,0 +1,25 @@
+# Copyright (c) Microsoft Corporation
```
abuccts: use `.k8s.yaml` in filename

src/webportal/config/webportal_k8s.yaml
```diff
@@ -0,0 +1,19 @@
+# Copyright (c) Microsoft Corporation
```
abuccts: is this file included in this PR?

src/rest-server/config/rest-server_k8s.yaml
wangdian: Ok

src/webportal/config/webportal_k8s.yaml
wangdian: Yes, will change to .k8s.yaml.

src/cluster/config/cluster.py
wangdian: The previous config does not have cluster-type.  Please refer to forward_compatibility.py, when use old config file, use "yarn" as default cluster-type

PR #3445, opened 2019-08-26
[hotfix] fix dashboard

hzy46:
fix the port issues for k8s dashboard:
image

### DIFF
--- a/deployment/k8sPaiLibrary/template/dashboard-service.yaml.template
+++ b/deployment/k8sPaiLibrary/template/dashboard-service.yaml.template
@@ -27,6 +27,3 @@ metadata:
 spec:
   selector:
     k8s-app: kubernetes-dashboard
-  ports:
-  - port: 80
-    targetPort: 9090


deployment/k8sPaiLibrary/template/dashboard-service.yaml.template
```diff
@@ -27,6 +27,3 @@ metadata:
 spec:
   selector:
     k8s-app: kubernetes-dashboard
-  ports:
-  - port: 80
```
abuccts: why remove? and you should use `paictl` to deploy instead of `kubectl`

hzy46: If it is not removed, an error will be raised when deployed. But the CI shows this should not be removed. I'm trying to figure out why the error happens when I deploy it.

abuccts: Please __clean all services and k8s__ before your deployment.
I guess the k8s or dashboard's versions may be changed by others.

/cc @ydye to help.

hzy46: I used "./paictl.py service stop" and "./paictl.py cluster k8s-clean" to clean the cluster before I deploy. But the problem remains.  It occurs only on int bed. Another bed doesn't have this issue.

hzy46:
Close this issue since it is not a bug but an environmental problem.

PR #3446, opened 2019-08-26
[hotfix] fix dashboard

hzy46:
same as https://github.com/microsoft/pai/pull/3445


hzy46:
Close this since it is not a bug but an environmental problem.

PR #3461, opened 2019-08-28
not show alert when alert-manager not deployed

Binyang2014:


debuggy:
do you open a PR to master with same change?

Binyang2014:
> do you open a PR to master with same change?

Will create an other PR for this

PR #3498, opened 2019-08-30
[hotfix] check group existence when updating user grouplist

mzmssg:
1. check groups when creating user, updating user grouplist, adding group to user grouplist.
2. remove nonexistent groups from user grouplist every time restserver start.

### DIFF
--- a/src/rest-server/src/controllers/v2/user.js
+++ b/src/rest-server/src/controllers/v2/user.js
@@ -126,40 +126,47 @@ const updateUserGroupListFromExternal = async (req, res, next) => {
 };
 
 const createUser = async (req, res, next) => {
+  if (!req.user.admin) {
+    next(createError('Forbidden', 'ForbiddenUserError', `Non-admin is not allow to do this operation.`));
+  }
+  let grouplist;
   try {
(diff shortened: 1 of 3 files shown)


src/rest-server/src/controllers/v2/user.js
```diff
@@ -126,40 +126,48 @@ const updateUserGroupListFromExternal = async (req, res, next) => {
...
-      }
+    return next(createError.unknown(error));
+  }
+  if (grouplist.length !== req.body.virtualCluster.length) {
+    next(createError('Bad Request', 'NoVirtualClusterError', `Try to update: ${req.body.virtualCluster}, but found ${grouplist}`));
+  }
+  if (!grouplist.includes(authConfig.groupConfig.defaultGroup.groupname)) {
+    grouplist.push(authConfig.groupConfig.defaultGroup.groupname);
+  }
+  if (req.body.admin) {
```
abuccts: why is there an `admin` filed in body?

ydye: Experience consideration. @scarlett2018

PR #3525, opened 2019-09-02
Fix helper docs on the webportal 

hzy46:
same as https://github.com/microsoft/pai/pull/3524


hzy46:
close this to use a cherry-pick instead

PR #3556, opened 2019-09-04
fix bug: job always show success even it failed

Binyang2014:


yqwang-ms:
Could we also add below stuffs in all runtime bash? So that error will never be ignored again:
```
set -o errexit
set -o nounset
set -o pipefail
```

src/kube-runtime/src/runtime
```diff
@@ -55,12 +55,15 @@ ${PAI_RUNTIME_DIR}/precommands.sh > >(tee -a ${PAI_LOG_DIR}/runtime.pai.stdout)
...
 wait ${USER_PID}
+USER_EXIT_CODE=$?
+echo "[INFO] USER COMMAND END" >> ${PAI_LOG_DIR}/user.pai.stderr
 
 # execute postCommands generated by plugin
-${PAI_RUNTIME_DIR}/postcommands.sh > >(tee -a ${PAI_LOG_DIR}/runtime.pai.stdout) 2> >(tee -a ${PAI_LOG_DIR}/runtime.pai.stderr >&2)
\ No newline at end of file
+${PAI_RUNTIME_DIR}/postcommands.sh > >(tee -a ${PAI_LOG_DIR}/runtime.pai.stdout) 2> >(tee -a ${PAI_LOG_DIR}/runtime.pai.stderr >&2)
+echo "container will exit with code ${USER_EXIT_CODE}" >> ${PAI_LOG_DIR}/runtime.pai.stderr
+exit ${USER_EXIT_CODE}
```
abuccts: what if `precommands.sh` or `postcommands.sh` failed? better to add `|| exit $?` at each line's ending

yqwang-ms: That likes "set -o errexit", so I suggest add them all in header.

Binyang2014: Dian create another PR for this: https://github.com/microsoft/pai/pull/3559, will close this PR

PR #3559, opened 2019-09-04
Add job exit code

wangdian:
Add job exit code


yqwang-ms:
Could we also add below stuffs in all runtime bash? So that error will never be ignored again:
```
set -o errexit
set -o nounset
set -o pipefail
```

wangdian:
Thanks yuqi.
Seems it is a better way to handle job exits.

src/kube-runtime/src/init.d/initializer.py
```diff
@@ -100,10 +100,10 @@ def init_plugins(jobconfig, commands, plugins_path, runtime_path, taskrole):
             run_script("{}/{}".format(plugin_base_path, plugin_desc["init-script"]), parameters, plugin_scripts)        
 
         if os.path.isfile(plugin_scripts[0]):
-            commands[0].append("/bin/bash {}".format(plugin_scripts[0]))
+            commands[0].append("/bin/bash {} || exit $?".format(plugin_scripts[0]))
```
abuccts: not need?

src/kube-runtime/src/runtime
```diff
@@ -55,12 +58,13 @@ ${PAI_RUNTIME_DIR}/precommands.sh > >(tee -a ${PAI_LOG_DIR}/runtime.pai.stdout)
 # Put verbose output to user.pai.all, stdout to user.pai.stdout, stderr to user.pai.stderr
 # execute user commands
 # priority=100
-echo "[INFO] USER COMMAND START"
+echo "[INFO] USER COMMAND START" >> ${PAI_LOG_DIR}/user.pai.stderr
 ${PAI_RUNTIME_DIR}/user.sh > >(tee -a ${PAI_LOG_DIR}/user.pai.stdout ${PAI_USER_LOG_FILE}) 2> >(tee -a ${PAI_LOG_DIR}/user.pai.stderr ${PAI_USER_LOG_FILE} >&2) &
```
abuccts: why do we need to run user command in background?

src/kube-runtime/src/init.d/initializer.py
Binyang2014: Since we already have `set -o errexit`, do we still need `|| exit $?` here?

src/kube-runtime/src/runtime.d/user.sh
```diff
@@ -17,5 +17,8 @@
 # DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
+set -o errexit
+#set -o nounset
```
yqwang-ms: Why remove this?

src/kube-runtime/src/init.d/initializer.py
wangdian: This is needed because the file doesn't have set -o errexit

wangdian: This is needed because the generated file doesn't have set -o errexit

wangdian:
@fanyangCS  The install of openssh is handled by ssh plugin.

src/kube-runtime/src/runtime.d/user.sh
wangdian: Currently remove this because the injected code may break this and cause error. Temporarily remove this.
Will enable this after change webportal code.

src/kube-runtime/src/runtime
wangdian: Run user command in background is from original kube-runtime.
The purpose is user command may take long time and we may want to do something(e.g monitor user process) before wait.
Leave it here for future implementation.

src/kube-runtime/src/runtime.d/user.sh
yqwang-ms: Why runtime depends on webportal ?

wangdian: It's because webportal may inject some codes in user commands. Removed the codes that may cause error.

fanyangCS: Can you be specific?

PR #3564, opened 2019-09-05
Add user SSH on job submission page. Generate SSH key pair on rest se…

wangdian:
Add user SSH on job submission page.
Generate SSH key pair on rest server

### DIFF
--- a/src/webportal/src/app/env.js.template
+++ b/src/webportal/src/app/env.js.template
@@ -1,4 +1,5 @@
 window.ENV = {
+  restServerLauncherType: '${REST_SERVER_LAUNCHER_TYPE}',
   restServerUri: '${REST_SERVER_URI}',
   prometheusUri: '${PROMETHEUS_URI}/prometheus',
   yarnWebPortalUri: '${YARN_WEB_PORTAL_URI}',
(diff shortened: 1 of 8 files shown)


src/webportal/src/app/job-submission/utils/constants.js
```diff
@@ -144,3 +144,16 @@ export const DOCKER_OPTIONS = [
   },
 ];
 export const DEFAULT_DOCKER_URI = 'openpai/tensorflow-py36-cu90';
+// For PAI runtime only
+export const PAI_PLUGIN = 'com.microsoft.pai.runtimeplugin';
```
sunqinzheng: I'm not sure if task's ssh connection will be a feature that could only be accessed by web portal.
If not, I think we should add some documents about job protocol's extra field format and move some logics to rest server.

scarlett2018:
**TODO**: add document for usage instruction

src/webportal/src/app/env.js.template
```diff
@@ -1,4 +1,5 @@
 window.ENV = {
+  restServerLauncherType: '${REST_SERVER_LAUNCHER_TYPE}',
```
sunqinzheng: I have added a environment variable `launcherType` and it is already checked in. Please used that instead.

wangdian:
Separate this to multiple PRs, close this.
Please refer to #3596

PR #3569, opened 2019-09-05
[Examples] Update marketplace job to be compatible with k8s

abuccts:
Update marketplace job to be compatible with k8s.


marketplace-v2/keras-tensorflow-mnist.yaml
```diff
@@ -33,4 +33,5 @@ taskRoles:
       memoryMB: 8192
       gpu: 1
     commands:
+      - rm /usr/local/cuda/lib64/stubs/libcuda.so.1
```
Binyang2014: Add comments for this line?

PR #3575, opened 2019-09-05
Change rest-server to support errorSpec

Binyang2014:


src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -615,6 +649,102 @@ const getSshInfo = async (frameworkName) => {
   throw createError('Not Found', 'NoJobSshInfoError', `SSH info of job ${frameworkName} is not found.`);
 };
 
+const extractDiagnosticsInfo = (diag) => {
+  if (_.isEmpty(diag)) {
+    return null;
+  }
+  // The pattern for diagnostics is "podPattern unmatched: json" or "podPattern nmatched: json"
```
yqwang-ms: >nmatched [](start = 80, length = 8)

matched

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -615,6 +649,102 @@ const getSshInfo = async (frameworkName) => {
...
+  }
+  return null;
+};
+
+const extractRuntimeOutput = (diag) => {
+  if (_.isEmpty(diag)) {
+    return null;
+  }
+
+  // Do not container user container info
```
yqwang-ms: >container [](start = 12, length = 9)

typo

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -615,6 +649,102 @@ const getSshInfo = async (frameworkName) => {
...
+  if (_.isNil(diag.containers) || diag.containers.length < 2) {
+    return null;
+  }
+  const message = diag.containers[1].message;
```
yqwang-ms: Do not use index to locate user container, use the container name.

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -615,6 +649,102 @@ const getSshInfo = async (frameworkName) => {
...
+  return null;
+};
+
+const extractRuntimeOutput = (diag) => {
+  if (_.isEmpty(diag)) {
+    return null;
+  }
+
+  // Do not container user container info
+  if (_.isNil(diag.containers) || diag.containers.length < 2) {
```
yqwang-ms: Runtime code is in both init and app container, so why only app container has PAI_RUNTIME_ERROR_START/END?

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -615,6 +649,102 @@ const getSshInfo = async (frameworkName) => {
...
+  const obj = {};
+  diag.containers.forEach((container) => {
+    let message = container.message;
+    if (_.isEmpty(message)) {
+      return;
+    }
+    const match = message.match(anchor);
+    if (match !== null) {
+      message = message.substring(0, match.index);
+    }
```
yqwang-ms: In which cases, the message is not empty?

yqwang-ms: In yarn, we need this because, we use it to catch YarnContainerScripts stderr, in k8s, do we still need?


     a) Get error message from container stderr (ExitCodeException Message)
     ExitCodeException exitCode=127: YarnContainerScripts/0.sh: line 58: [: -gt: unary operator expected
             at org.apache.hadoop.util.Shell.runCommand(Shell.java:998)
             at org.apache.hadoop.util.Shell.run(Shell.java:884)
             at org.apache.hadoop.util.Shell$ShellCommandExecutor.execute


---
In reply to: [321554938](https://github.com/microsoft/pai/pull/3575#discussion_r321554938) [](ancestors = 321554938)

Binyang2014: If all the message here is extracted from pod `termination message`. 
Then the message will be empty.
I'm not sure if controller customize the message here. If not, will remove this function.
And `appExitMessages.container` will be null in k8s version

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -151,6 +172,8 @@ const convertTaskDetail = async (taskStatus, ports, userName, jobName, taskRoleN
 
 const convertFrameworkDetail = async (framework) => {
   const completionStatus = framework.status.attemptStatus.completionStatus;
+  const diagnostics = completionStatus ? completionStatus.diagnostics : null;
+  const diagInfo = extractDiagnosticsInfo(diagnostics);
   const detail = {
```
yqwang-ms: diagnostics may only contain controller reported msg, such as "Pod spec invalid: xxxx". So the msg is lost after this line.
Suggest to rename extractDiagnosticsInfo as extractPodCompletionStatus, and it should not be used as the whole diag.

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -177,16 +200,16 @@ const convertFrameworkDetail = async (framework) => {
...
-      appExitMessages: {
-        container: null,
-        runtime: null,
-        launcher: null,
-      },
-      appExitTriggerMessage: completionStatus ? completionStatus.diagnostics : null,
-      appExitTriggerTaskRoleName: null, // TODO
-      appExitTriggerTaskIndex: null, // TODO
+      appExitSpec: completionStatus ? generateExitSpec(completionStatus.code) : generateExitSpec(null),
+      appExitDiagnostics: diagInfo,
```
yqwang-ms: >diagInfo [](start = 26, length = 8)

should be the whole diagnostics

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -177,16 +200,16 @@ const convertFrameworkDetail = async (framework) => {
...
+      appExitMessages: diagInfo ? {
```
yqwang-ms: >diagInfo [](start = 23, length = 8)

diagnostics

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -177,16 +200,16 @@ const convertFrameworkDetail = async (framework) => {
...
+        container: extractContainerOutput(diagInfo),
+        runtime: extractRuntimeOutput(diagInfo),
+        launcher: extractLauncherOutput(completionStatus.code, diagInfo),
```
yqwang-ms: >diagInfo [](start = 63, length = 8)

diagnostics

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: Actually, for init container, I just want to return specific code for specific error and do not use pattern match. Since it all controlled by ourselves. But it's OK to leverage the same exit message format with user container. Will add this part.

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: Will fix it

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -281,6 +304,10 @@ const generateTaskRole = (taskRole, labels, config) => {
                   subPath: `${labels.userName}/${labels.jobName}/${convertName(taskRole)}`,
                   mountPath: '/usr/local/pai/logs',
                 },
+                {
+                  name: 'user-exit-spec',
```
yqwang-ms: >user-exit-spec [](start = 25, length = 14)

job-exit-spec?

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -349,6 +377,12 @@ const generateTaskRole = (taskRole, labels, config) => {
                 secretName: 'job-ssh-secret',
               },
             },
+            {
+              name: 'user-exit-spec',
+              configMap: {
+                name: 'user-exit-spec-configuration',
+              },
```
yqwang-ms: job-exit-spec?

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: If we do not need this. Will leave it as null for API compatibility

yqwang-ms: it is raw or substring of termination message, so let's remove it

---
In reply to: [321557904](https://github.com/microsoft/pai/pull/3575#discussion_r321557904) [](ancestors = 321557904)

src/rest-server/src/models/v2/job/k8s.js
yqwang-ms: ok, but it equivalent to empty exit message.

---
In reply to: [321558507](https://github.com/microsoft/pai/pull/3575#discussion_r321558507) [](ancestors = 321558507)

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -615,6 +649,102 @@ const getSshInfo = async (frameworkName) => {
...
+    return null;
+  }
+  return obj;
+};
+
+const extractLauncherOutput = (diag, code) => {
+  if (_.isEmpty(diag) || code > 0) {
+    return null;
+  }
+  return diag;
```
yqwang-ms: >diag [](start = 9, length = 4)

return this, when PAI_RUNTIME_ERROR_START cannot be found, instead of  code comparing with 0.
Such as when SIGKILL received

src/rest-server/src/models/v2/job/k8s.js
yqwang-ms: >appExitDiagnostics [](start = 6, length = 18)

This should contain the whole raw diagnostics

src/rest-server/src/models/v2/job/k8s.js
yqwang-ms: >appExitMessages [](start = 6, length = 15)

This should contain the exit info after split the diagnostics to be launcher and runtime. 

And if runtime PAI_RUNTIME_ERROR_START cannot be found, the whole raw diagnostics should be just appExitMessages.launcher, and appExitMessages.runtime =null.
Otherwise, appExitMessages.launcher=null, appExitMessages.runtime =extracted.

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -615,6 +648,100 @@ const getSshInfo = async (frameworkName) => {
   throw createError('Not Found', 'NoJobSshInfoError', `SSH info of job ${frameworkName} is not found.`);
 };
 
+const generateExitDiagnostics = (diag) => {
+  const podCompletionStatus = extractPodCompletionStatus(diag);
```
yqwang-ms: Why not directly use diag?

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -177,16 +199,16 @@ const convertFrameworkDetail = async (framework) => {
...
-        runtime: null,
-        launcher: null,
-      },
-      appExitTriggerMessage: completionStatus ? completionStatus.diagnostics : null,
-      appExitTriggerTaskRoleName: null, // TODO
-      appExitTriggerTaskIndex: null, // TODO
+        runtime: extractRuntimeOutput(diagnostics),
+        launcher: extractLauncherOutput(diagnostics),
+      } : null,
+      appExitTriggerMessage: completionStatus && completionStatus.trigger ? yaml.safeDump(completionStatus.trigger) : null,
```
yqwang-ms: >appExitTriggerMessage [](start = 6, length = 21)

Why not just use "completionStatus.trigger.message"

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -615,6 +648,100 @@ const getSshInfo = async (frameworkName) => {
...
+    const anchor1 = /\[PAI_RUNTIME_ERROR_START\]/;
+    const anchor2 = /\[PAI_RUNTIME_ERROR_END\]/;
+    const match1 = message.match(anchor1);
+    const match2 = message.match(anchor2);
+    if (match1 !== null && match2 !== null) {
+      const start = match1.index + match1[0].length;
+      const end = match2.index;
+      const output = message.substring(start, end).trim();
+      res = yaml.safeLoad(output);
+      break;
```
yqwang-ms: Will you include the container name in your runtime message? init or app

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: Since message in diag is messy。 It container string such as: "userLogs: \"logs\"\\n". multi escape character is used. 
If we use it directly, we need to change web portal to make this message show neatly.
Since this filed just a string and we don't promise any format here. Try to format this string in rest-server for backward compatibility

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -615,6 +648,100 @@ const getSshInfo = async (frameworkName) => {
...
+
+const extractLauncherOutput = (diag) => {
+  const anchor = /\[PAI_RUNTIME_ERROR_START\]/;
+  const match = diag.match(anchor);
+  if (match != null) {
+    return null;
+  }
+  return diag;
+};
+
```
yqwang-ms: merge extractLauncherOutput with extractLauncherOutput, since they depend on each other and you just need to match once?

to be generateAppExitMessages?

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: Thanks, I don't notice this field

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: You mean merge extractLauncherOutput  and extractRuntimeOutput here?

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: That's OK for me

src/rest-server/src/models/v2/job/k8s.js
yqwang-ms: If so, can we also merge generateExitDiagnostics with extractXXX? They depends on each other and reduce the match times.

---
In reply to: [321640867](https://github.com/microsoft/pai/pull/3575#discussion_r321640867) [](ancestors = 321640867)

yqwang-ms:
Also change getUserFailureHintItems in src/webportal/src/app/job/job-view/fabric/job-detail/components/summary.jsx

To be first get runtimeOutput if not find, then fallback to launcherOutput. (Instead of judging by code > 0, like

    if (code > 0) {
      if (runtimeOutput && runtimeOutput.reason) {
        reason.push(
          <div key='runtime-reason'>{runtimeOutput.reason}</div>,
        );
      }
    } else {
      const launcherOutput = get(jobInfo, 'jobStatus.appExitMessages.launcher');
      if (launcherOutput) {
        reason.push(
          <div key='launcher-reason'>{launcherOutput}</div>,
        );
      }
    }
)

Binyang2014:
> Also change getUserFailureHintItems in src/webportal/src/app/job/job-view/fabric/job-detail/components/summary.jsx
> 
> To be first get runtimeOutput, if not find, then fallback to launcherOutput. (Instead of judging by code > 0, like
> 
> ```
> if (code > 0) {
>   if (runtimeOutput && runtimeOutput.reason) {
>     reason.push(
>       <div key='runtime-reason'>{runtimeOutput.reason}</div>,
>     );
>   }
> } else {
>   const launcherOutput = get(jobInfo, 'jobStatus.appExitMessages.launcher');
>   if (launcherOutput) {
>     reason.push(
>       <div key='launcher-reason'>{launcherOutput}</div>,
>     );
>   }
> }
> ```
> 
> )

Hope it will not break the previous behavior. And I will create an other issue for this change

yqwang-ms:
Better to use this PR. Can you distinguish k8s or yarn, then choose different logic? If not, let's just change to current logic 

---
In reply to: [528820112](https://github.com/microsoft/pai/pull/3575#issuecomment-528820112) [](ancestors = 528820112)

Binyang2014:
> Better to use this PR. Can you distinguish k8s or yarn, then choose different logic? If not, let's just change to current logic
> 
> In reply to: [528820112](https://github.com/microsoft/pai/pull/3575#issuecomment-528820112) [](ancestors = 528820112)


We can distinguish yarn or k8s in webportal currently. This piece of code just checked in. https://github.com/microsoft/pai/blob/30987a3efa2c6c6c0cd051b76f5798f54d204ee9/src/webportal/src/app/env.js.template#L14

yqwang-ms:
>         maxRetryCount: 0,

change this to false, this is another issue but let's fix it here

---
Refers to: src/rest-server/src/models/v2/job/k8s.js:265 in 9c3cb16. [](commit_id = 9c3cb161d5f120fcf8a14034b0148fbb78c2ce0e, deletion_comment = False)

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -615,6 +649,99 @@ const getSshInfo = async (frameworkName) => {
...
+  let podCompletionStatus = null;
+  try {
+    podCompletionStatus = JSON.parse(matches[1]);
+  } catch (error) {
+    logger.warn('Get diagnostics info failed', error);
+    return exitDiagnostics;
+  }
+
+  const summmaryInfo = diag.substring(0, matches.index + 'matched:'.length);
```
yqwang-ms: >matches.index [](start = 41, length = 13)

What is this matches.index? multiple values?

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -615,6 +649,99 @@ const getSshInfo = async (frameworkName) => {
...
+
+  // Get runtime output, set launcher output to null. Otherwise, treat all message as launcher output
+  exitDiagnostics.runtime = extractRuntimeOutput(podCompletionStatus);
+  if (exitDiagnostics.runtime !== null) {
+    exitDiagnostics.launcher = null;
+    return exitDiagnostics;
+  }
+
+  exitDiagnostics.launcher = exitDiagnostics.diagnosticsSummary;
+  return exitDiagnostics;
```
yqwang-ms: Move this to line 680?

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -615,6 +649,99 @@ const getSshInfo = async (frameworkName) => {
...
+    const anchor2 = /\[PAI_RUNTIME_ERROR_END\]/;
+    const match1 = message.match(anchor1);
+    const match2 = message.match(anchor2);
+    if (match1 !== null && match2 !== null) {
+      const start = match1.index + match1[0].length;
+      const end = match2.index;
+      const output = message.substring(start, end).trim();
+      res = {
+        ...yaml.safeLoad(output),
+        name: container.name,
```
yqwang-ms: try catch?

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: index mean the postion where patten is found. refer: [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match)

---
In reply to: [323575627](https://github.com/microsoft/pai/pull/3575#discussion_r323575627) [](ancestors = 323575627)

PR #3584, opened 2019-09-06
[Log manager] Access Log Manager through Pylon

TongboZhang:
- fix pylon
- fix rest-server

### DIFF
--- a/src/pylon/deploy/pylon-config/location.conf.template
+++ b/src/pylon/deploy/pylon-config/location.conf.template
@@ -80,6 +80,11 @@ location ~ ^/alert-manager(.*)$ {
   proxy_pass {{ALERT_MANGER_URI}}/alert-manager$1$is_args$args;
 }
 
+# log-manager server
+location ~ ^/log-manager/([^/]+):(\d+)/(.*)$ {
+  proxy_pass http://$1:$2/$3$is_args$args;
+}
+
 #
 # YARN web portal
 #
--- a/src/rest-server/src/models/v2/job/k8s.js
+++ b/src/rest-server/src/models/v2/job/k8s.js
@@ -144,7 +144,7 @@ const convertTaskDetail = async (taskStatus, ports, userName, jobName, taskRoleN
     containerIp: taskStatus.attemptStatus.podHostIP,
     containerPorts,
     containerGpus,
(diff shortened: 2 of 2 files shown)


src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -144,7 +144,7 @@ const convertTaskDetail = async (taskStatus, ports, userName, jobName, taskRoleN
     containerIp: taskStatus.attemptStatus.podHostIP,
     containerPorts,
     containerGpus,
-    containerLog: `http://${taskStatus.attemptStatus.podHostIP}:${process.env.LOG_MANAGER_PORT}/logs/${userName}/${jobName}/${taskRoleName}/${taskStatus.attemptStatus.podUID}/`,
+    containerLog: `https://${process.env.WEBPORTAL_URL}/log-manager/${taskStatus.attemptStatus.podHostIP}:${process.env.LOG_MANAGER_PORT}/logs/${userName}/${jobName}/${taskRoleName}/${taskStatus.attemptStatus.podUID}/`,
```
abuccts: there shouldn't be a change, you need to add a `sub_filter` section [here](https://github.com/microsoft/pai/blob/30987a3efa2c6c6c0cd051b76f5798f54d204ee9/src/pylon/deploy/pylon-config/location.conf.template#L274-L276)

src/pylon/deploy/pylon-config/location.conf.template
```diff
@@ -80,6 +80,11 @@ location ~ ^/alert-manager(.*)$ {
   proxy_pass {{ALERT_MANGER_URI}}/alert-manager$1$is_args$args;
 }
 
+# log-manager server
+location ~ ^/log-manager/([^/]+):(\d+)/(.*)$ {
+  proxy_pass http://$1:$2/$3$is_args$args;
```
abuccts: because it uses ip of each node, you'd better use a central ip and make the node ip in the path
otherwise, you may not write a `sub_filter` to match each node's ip

abuccts: e.g. `node-ip:log-port/log-path` -> `pylon-address/log-manager/node-ip/log-path`

src/rest-server/src/models/v2/job/k8s.js
abuccts: 1. what if there's no https? http should work
2. what if there's no pylon? webportal should work at its port
3. webportal ip may be different from pylon ip

PR #3591, opened 2019-09-09
Fix version change in webportal

hzy46:
Currently we use `cp -arf "../../../version" "../version"` to copy the PAI version to `src/webportal/` , here is a small problem:

- If it is the first time to run `cp -arf "../../../version" "../version"` , a `version` folder will be created under `src/webportal`
- If you run the  `cp -arf "../../../version" "../version"` again,  it will copy the `version` folder under `src/webportal/version/`. Thus the folder structure becomes `src/webportal/version/version`, which is not desired.

To reproduce:
- build and deploy the cluster
- change the PAI.VERSION to make a new version
- build and deploy the cluster, but the version will not change in webportal

### DIFF
--- a/src/webportal/build/build-pre.sh
+++ b/src/webportal/build/build-pre.sh
@@ -21,7 +21,8 @@ pushd $(dirname "$0") > /dev/null
 
 mkdir -p "../dependency"
 cp -arf "../../../docs" "../../../examples" "../dependency"
-cp -arf "../../../version" "../version"
+mkdir -p "../version"
+cp -arf ../../../version/* "../version/"
 if [ "$ATTACH_COMMIT_ID" = true ]; then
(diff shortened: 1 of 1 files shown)


debuggy:
Why would this ```cp``` command run two times?

hzy46:
> Why would this `cp` command run two times?

For example, I built and deployed the cluster on 2pm. After some code changes, I built it again on 3pm using the same directory. Then this problem would occurred.

debuggy:
> 
> 
> > Why would this `cp` command run two times?
> 
> For example, I built and deployed the cluster on 2pm. After some code changes, I built it again on 3pm using the same directory. Then this problem would occur.

ok, I got it.

hzy46:
The way in the script to make a "version" folder seems to conflict with the Jenkins pipeline. Close this PR to find a better fix or modify the Jenkins file.

PR #3599, opened 2019-09-09
[Rest Server] Change to default k8s scheduler

abuccts:
Change to default k8s scheduler.

### DIFF
--- a/src/rest-server/src/models/v2/virtual-cluster/k8s.js
+++ b/src/rest-server/src/models/v2/virtual-cluster/k8s.js
@@ -95,7 +95,7 @@ const getNodeResource = async () => {
     const nodes = (await axios({
       method: 'get',
       url: `${kubernetesConfig.apiserver.uri}/api/v1/nodes`,
-    })).data.items.filter(node => node.metadata.labels['pai-worker'] === 'true');
+    })).data.items.filter((node) => node.metadata.labels['pai-worker'] === 'true');
     for (let node of nodes) {
       const nodeName = node.metadata.name;
       const gpuNumber = parseInt(node.status.capacity['nvidia.com/gpu']);
@@ -170,7 +170,7 @@ const getVcList = async () => {
     const nodes = (await axios({
       method: 'get',
       url: `${kubernetesConfig.apiserver.uri}/api/v1/nodes`,
-    })).data.items.filter(node => node.metadata.labels['pai-worker'] === 'true');
+    })).data.items.filter((node) => node.metadata.labels['pai-worker'] === 'true');
     vcInfos['default'].resourcesTotal = {
       cpu: nodes.reduce((sum, node) => sum + parseInt(node.status.capacity.cpu), 0),
       memory: nodes.reduce((sum, node) => sum + k8s.convertMemoryMb(node.status.capacity.memory), 0),
--- a/src/rest-server/src/models/v2/virtual-cluster/k8s.js
+++ b/src/rest-server/src/models/v2/virtual-cluster/k8s.js
@@ -18,6 +18,8 @@
 // module dependencies
 const createError = require('@pai/utils/error');
 const vcConfig = require('@pai/config/vc');
+const launcherConfig = require('@pai/config/launcher');
+const kubernetesConfig = require('@pai/config/kubernetes');
 const k8s = require('@pai/utils/k8sUtils');
 const axios = require('axios');
 const yaml = require('js-yaml');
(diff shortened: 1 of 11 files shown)


src/rest-server/src/models/v2/virtual-cluster/k8s.js
```diff
@@ -81,12 +83,28 @@ const getNodeResource = async () => {
...
+        gpuTotal: vcData.clusterNodeGpu[node].gpu,
+        gpuUsed: 0,
+        gpuAvaiable: vcData.clusterNodeGpu[node].gpu,
+      };
+    }
+  } else {
+    const nodes = (await axios({
+      method: 'get',
+      url: `${kubernetesConfig.apiserver.uri}/api/v1/nodes`,
+    })).data.items.filter((node) => node.metadata.labels['pai-worker'] === 'true');
```
yqwang-ms: Will this be called only once at init or every time refresh the VC page?

src/rest-server/src/models/v2/virtual-cluster/k8s.js
```diff
@@ -138,18 +160,31 @@ const getVcList = async () => {
...
+    for (let vc of Object.keys(vcData.virtualCellCapacity)) {
+      vcInfos[vc].resourcesTotal.memory = vcData.virtualCellCapacity[vc].resourcesTotal.memory;
+      vcInfos[vc].resourcesTotal.cpu = vcData.virtualCellCapacity[vc].resourcesTotal.cpu;
+      vcInfos[vc].resourcesTotal.gpu = vcData.virtualCellCapacity[vc].resourcesTotal.gpu;
+    }
+  } else {
+    const nodes = (await axios({
+      method: 'get',
+      url: `${kubernetesConfig.apiserver.uri}/api/v1/nodes`,
+    })).data.items.filter((node) => node.metadata.labels['pai-worker'] === 'true');
```
yqwang-ms: Will this be called only once at init or every time refresh the VC page?

src/rest-server/src/routes/v2/virtual-cluster.js
```diff
@@ -41,11 +37,11 @@ router.route('/:virtualClusterName')
   .delete(token.check, controller.remove);
 
 router.route('/:virtualClusterName/status')
-  /** PUT /api/v2/virtual-clusters/:virtualClusterName - Change virtual cluster status (running or stopped) */
+  /** PUT /api/v2/virtual-clusters/:virtualClusterName/status - Change virtual cluster status (running or stopped) */
   .put(token.check, param.validate(vcConfig.vcStatusPutInputSchema), controller.updateStatus);
 
-router.route('/:virtualClusterName/resourceUnits')
-  /** GET /api/v2/virtual-clusters/:virtualClusterName/resourceUnits - Get virtual cluster available resourceUnits */
+router.route('/:virtualClusterName/units')
```
yqwang-ms: >units [](start = 35, length = 5)

why rename this? but keep controller.getResourceUnits

src/rest-server/src/models/v2/virtual-cluster/k8s.js
abuccts: every time when api is called, there may be node addition/deletion so getting the results at the beginning may not consistent

src/rest-server/src/models/v2/virtual-cluster/k8s.js
abuccts: same

src/rest-server/src/routes/v2/virtual-cluster.js
abuccts: just renamed the api, current api uses dash `-` but `resourceUnits` is a camel style
the code uses camel case so the variable name is fine

src/rest-server/src/models/v2/virtual-cluster/k8s.js
abuccts: getting used resources or running pods also call apiserver every time when the same api is called in previous implementation, so I assume it's ok for apiserver?

src/rest-server/src/routes/v2/virtual-cluster.js
```diff
@@ -28,10 +28,6 @@ router.route('/')
   /** GET /api/v2/virtual-clusters - Return cluster virtual cluster info */
   .get(controller.list);
 
-router.route('/nodeResource')
```
abuccts: also changed api /api/v2/virtual-clusters/nodeResource to /api/v2/virtual-clusters?nodes,
in case there's a vc called nodeResource

src/rest-server/src/models/v2/virtual-cluster/k8s.js
yqwang-ms: ok, current is fine. we would better to watch and cache it in future.

src/rest-server/src/routes/v2/virtual-cluster.js
yqwang-ms: So why not resource-units

abuccts: updated

yqwang-ms:
>         maxRetryCount: 0,

change this to false, this is another issue but let's fix it here

---
Refers to: src/rest-server/src/models/v2/job/k8s.js:244 in da3872e. [](commit_id = da3872e6f6d354124ae1ecaf1004e131fd3c6d68, deletion_comment = False)

yqwang-ms:
>     containerId: taskStatus.attemptStatus.podName,

use podUID here, besides, add another field which equals to the backend framework name (for easy debug)?

---
Refers to: src/rest-server/src/models/v2/job/k8s.js:145 in da3872e. [](commit_id = da3872e6f6d354124ae1ecaf1004e131fd3c6d68, deletion_comment = False)

yqwang-ms:
>               image: config.prerequisites.dockerimage[config.taskRoles[taskRole].dockerImage].uri,

Change this to "app" container? align with k8s: [https://kubernetes.io/docs/concepts/workloads/pods/init-containers/](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/)
Also @Binyang2014  to confirm no breaking 

---
Refers to: src/rest-server/src/models/v2/job/k8s.js:294 in bfa3197. [](commit_id = bfa319758d88100b84ab4a4bcc71cfcb770f985f, deletion_comment = False)

PR #3623, opened 2019-09-16
add job history rest-server & webportal, query by frameworkcontroller…

doubleyoung18:
add job history rest-server & webportal, query by frameworkcontroller…


src/rest-server/src/server.js
```diff
@@ -35,4 +35,11 @@ app.listen(config.serverPort, () => {
   logger.info('RESTful API server starts on port %d', config.serverPort);
 });
 
-module.exports = app;
+esApp.listen(config.esPort, () => {
```
sunqinzheng: I think the elastic search apis should be a sub router of the current rest server so that the third party extensions and rest api users could access these apis easily.

sunqinzheng: Another solution is just like the prometheus server. We just expose the elastic search's endpoint and let user customizes their queries.

src/webportal/src/app/job/job-view/fabric/job-detail/components/summary.jsx
```diff
@@ -94,8 +94,10 @@ export default class Summary extends React.Component {
       modalTitle: '',
       autoReloadInterval: 10 * 1000,
       hideDialog: true,
+      currentAttemptID: this.props.jobInfo.jobStatus.retries,
```
sunqinzheng: I will implement an independent job attempts page. Please seperate the frontend works and check in the api changes only.

debuggy:
the API endpoints added here should be included in API doc

src/rest-server/src/models/v2/history.js
```diff
@@ -0,0 +1,420 @@
...
+                  'ObjectSnapshot.metadata.uid.keyword': podUID,
+                },
+              },
+            ],
+          },
+        },
+      },
+    },
+    'sort': {
+      'CollectTime': 'desc',
```
yqwang-ms: >CollectTime [](start = 7, length = 11)

Rename all places:
CollectTime -> collectTime
ObjectSnapshot -> objectSnapshot

src/rest-server/src/models/v2/history.js
```diff
@@ -0,0 +1,420 @@
...
+        status: 404,
+        data: {
+          message: `The specified ${index} is not found`,
+        },
+      };
+    } else {
+      let resultObj = {'items': []};
+      for (let i = 0; i < aggResults.length; i++) {
+        resultObj['items'].push(aggResults[i].CollectTime_sort.buckets[0].top.hits.hits[0]._source.ObjectSnapshot);
+      }
```
yqwang-ms: May can be replaced by:

        "CollectTime_sort": {
          "terms": {
            "field": "collectTime",
            **"size": 1,**
            "order": {
              "_key": "desc"
            }
          },

yqwang-ms: Or:

        "CollectTime_sort": {
          "top_hits": {
            "sort": [
              {
                "collectTime": {
                  "order": "desc"
                }
              }
            ],
            "size": 1
          }
        }

---
In reply to: [339911150](https://github.com/microsoft/pai/pull/3623#discussion_r339911150) [](ancestors = 339911150)

yqwang-ms:
Close it as it is refined to be https://github.com/microsoft/pai/pull/3831

PR #3649, opened 2019-09-18
Add the analyze part, and change the code structure

AosChen:
@abuccts Plz review it, thanks.


contrib/profiler/utils.py
```diff
@@ -0,0 +1,119 @@
+class Advise:
```
Binyang2014: Change name to Adviser?

contrib/profiler/utils.py
```diff
@@ -0,0 +1,119 @@
...
+    def __init__(self, advise_list=None):
+        self._advise = ['The model does not need to be optimized.',
```
Binyang2014: change to `_advice`? advise is a verb

contrib/profiler/profiler.py
```diff
@@ -188,148 +143,188 @@ def get_network_bytes(filename):
...
-           sum(cpu_when_gpu_low) / length_gpu_low,
-           sum(mem_when_gpu_low) / length_gpu_low,
-           sum(disk_read_when_gpu_low) / length_gpu_low))
-    return [max_gpu, sum_gpu / count, max_cpu, sum_cpu / count, max_mem, sum_mem / count, max_read, sum_read,
-            sum(cpu_when_gpu_low) / length_gpu_low,
-            sum(mem_when_gpu_low) / length_gpu_low,
-            sum(disk_read_when_gpu_low) / length_gpu_low]
+def analyze_samples(sample_list, advise):
+    sample_list = np.array(sample_list, dtype=np.float)
+    used_gpu_num = (sample_list.shape[1] - 7) / 4
```
Binyang2014: What's the meaning of 7 and 4...? Can you add some comments for the data format of `sample_list`?

contrib/profiler/profiler.py
```diff
@@ -188,148 +143,188 @@ def get_network_bytes(filename):
...
+        gpu_mem_usage.append(sample_list[:, 9 + i * 4] / sample_list[:, 10 + i * 4])
+
+    if used_gpu_num >= 2:
+        # multiple GPUs, analyze whether each GPU has the same memory usage
+        gpu_mem_avg_0 = np.average(gpu_mem_usage[0])
+        gpu_mem_avg_1 = np.average(gpu_mem_usage[1])
+        if abs(gpu_mem_avg_0 - gpu_mem_avg_1) > 0.15:
+            advise.add_times(index=1)
+
+    if np.average(gpu_usage[0]) >= 90:
```
Binyang2014: should be 0.90?

contrib/profiler/profiler.py
```diff
@@ -188,148 +143,188 @@ def get_network_bytes(filename):
...
+        for i in range(cpu_usage.shape[0]):
+            cpu_slide.append(slide_windows.get_data(cpu_usage[i]))
+        cpu_slide_copy = cpu_slide.copy()
+        cpu_slide_copy.sort()
+        cpu_std_max = cpu_slide_copy[int(len(cpu_slide_copy) * 0.8)]
+        cpu_std_min = cpu_slide_copy[int(len(cpu_slide_copy) * 0.2)]
+
+        gpu_usage_up_down = [0]
+        for i in range(1, len(gpu_usage[0])):
+            if gpu_usage[0][i] > gpu_usage[0][i - 1]:
```
Binyang2014: Do we need to add a threshold to descide up or down? If the change is minor, shall we ignore it?

contrib/profiler/profiler.py
AosChen: 90% is from the testing of our model. The value can be changed if we can get more testing data.

contrib/profiler/profiler.py
AosChen: According to the testing data, the change in the values of GPU usage is either zero or not minor. I think we don't need to set a threshold.

contrib/profiler/profiler.py
Binyang2014: Maybe change 7 to `GPU_statistics_offset`, change 4 to ` statistics_num_per_GPU`  is more readable.
Also, create a reasonable name for 9, 10 in following code

---
In reply to: [329445955](https://github.com/microsoft/pai/pull/3649#discussion_r329445955) [](ancestors = 329445955)

contrib/profiler/profiler.py
Binyang2014: I mean for `gpu_mem_usage`, you use threshold as 0.80. But here you use threshold as 90. The unit is different?

---
In reply to: [332371115](https://github.com/microsoft/pai/pull/3649#discussion_r332371115) [](ancestors = 332371115)

contrib/profiler/profiler.py
Binyang2014: OK, fine

---
In reply to: [332372450](https://github.com/microsoft/pai/pull/3649#discussion_r332372450) [](ancestors = 332372450)

contrib/profiler/profiler.py
AosChen: Yes, the 'gpu_mem_usage' is calculated by myself, and the value if from 0 to 1.
But the 'gpu_usage'  is gotten from the NVIDIA's API, it returns the numerator of the percent.

Binyang2014: Got it, thanks

---
In reply to: [332808207](https://github.com/microsoft/pai/pull/3649#discussion_r332808207) [](ancestors = 332808207)

Issue #3635, opened 2019-09-17
Dashboard and Grafana cannot be visited behind Azure Gateway


abuccts:
Dashboard and Grafana cannot be visited behind Azure Gateway, requests will be redirected to internal IP.

PR #3652, opened 2019-09-18
[Deployment] Add gateway uri in configuration

abuccts:
In some network environments, pylon may not have a public address and the cluster can only be visited through a gateway #3635.

### DIFF
--- a/examples/cluster-configuration/services-configuration.yaml
+++ b/examples/cluster-configuration/services-configuration.yaml
@@ -22,6 +22,9 @@
 #    cluster-id: pai
 #    cluster-type: yarn
 #
+#    # Gateway address, used to visit the cluster from public network
+#    gateway-uri: "https://pai.domain.com"
+#
 #    # HDFS, zookeeper data path on your cluster machine.
 #    data-path: "/datastorage"
 #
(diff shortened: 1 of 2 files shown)


examples/cluster-configuration/services-configuration.yaml
```diff
@@ -22,6 +22,9 @@
 #    cluster-id: pai
 #    cluster-type: yarn
 #
+#    # Gateway address, used to visit the cluster from public network
+#    gateway-uri: "https://pai.domain.com"
```
Binyang2014: Can it be IP address? Not just domain name

examples/cluster-configuration/services-configuration.yaml
```diff
@@ -22,6 +22,9 @@
 #    cluster-id: pai
 #    cluster-type: yarn
 #
+#    # Gateway address, used to visit the cluster from public network
```
Binyang2014: Add comment for the default value if this field is missing?

examples/cluster-configuration/services-configuration.yaml
Binyang2014: Do we need to point the scheme names here. Or use can just config the gateway address. Then both http and https are supported . If gateway supports multi protocol.

ydye: what‘s difference with pylon.domain

abuccts: pylon may not have a public ip or domain
```
user <-> gateway <-> pylon
```

abuccts:
Abandoned, #3659 & #3663

PR #3659, opened 2019-09-19
fix grafana can not be accessed behind the gateway

Binyang2014:
Before this change, grafana access Prometheus from the front-end. And it doesn't work in the env with the gateway. Change it to though back-end proxy.
The request flow is changed from "Grafana->Gateway->Pylon->Prometheus" to  "Grafana->->Prometheus"


xudifsd:
LGTM

PR #3664, opened 2019-09-19
Foreground stop all frameworks

abuccts:
Froeground stop all frameworks before delete controller crd.

### DIFF
--- a/src/frameworkcontroller/deploy/delete.sh
+++ b/src/frameworkcontroller/deploy/delete.sh
@@ -19,8 +19,18 @@
 
 pushd $(dirname "$0") > /dev/null
 
+APISERVER=$(kubectl config view | grep server | cut -f 2- -d ":" | tr -d " ")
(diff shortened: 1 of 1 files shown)


src/frameworkcontroller/deploy/delete.sh
```diff
@@ -19,8 +19,18 @@
...
+    curl -X DELETE $FRAMEWORK_ENDPOINT \
+        -H "Content-Type: application/json" \
+        -d '{"kind":"DeleteOptions","apiVersion":"v1","propagationPolicy":"Foreground"}' > /dev/null 2>&1
+    sleep 5
+done
+
 /bin/bash stop.sh || exit $?
 
-kubectl delete customresourcedefinitions frameworks.frameworkcontroller.microsoft.com
+kubectl delete customresourcedefinitions frameworks.frameworkcontroller.microsoft.com || exit $?
```
yqwang-ms: Move above /bin/bash stop.sh to best effort avoid race condition

PR #3679, opened 2019-09-27
Binyli/gang schedule

Binyang2014:


src/kube-runtime/src/init.d/parser.py
```diff
@@ -83,8 +83,11 @@ def generate_runtime_env(framework):
             def get_port_base(x):
                 return int(ports[x]["start"]) + int(ports[x]["count"]) * int(index)
 
-            # export ip/port for task role
-            export("PAI_HOST_IP_{}_{}".format(name, index), current_ip)
+            # export ip/port for task role, current ip maybe None for non-gang-allocation
+            if current_ip:
```
abuccts: if `current_ip` is null, assign it a fake value like `0.0.0.0`? so that you can keep other code

PR #3684, opened 2019-09-30
HiveD algorithm refactor

zhypku:
- In-VC scheduling no longer relies on cells (and buddy alloc).
- Wrote a topology-aware scheduling algorithm, which is used for in-VC scheduling and opportunistic pod scheduling.
- Cell priority mechanism simplified: only keeps a free cell list for the physical cluster for regular pod scheduling, and a full cell list for opportunistic pod scheduling. They are decoupled now.

### DIFF
--- a/subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
+++ b/subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
@@ -44,13 +44,13 @@ type Cell interface {
 	AtOrHigherThanNode() bool
 
 	IncreaseFreeGpuNum(int32)
-	GetFreeGpuNumForPriority(CellPriority) int32
+	GetFreeGpuNumAtPriority(CellPriority) int32
 	GetUsedGpuNumAtPriority(CellPriority) int32
 	IncreaseUsedGpuNumAtPriority(CellPriority, int32)
 	GetUsedGpuNumSamePriority() int32
 	GetUsedGpuNumOtherPriority() int32
-	SetUsedGpuNumForPriority(CellPriority)
-	SetUsedGpuNumAllPriority(CellPriority)
+	UpdateUsedGpuNumForPriority(CellPriority)
+	UpdateUsedGpuNumAllPriority(CellPriority)
 }
 
 type GenericCell struct {
@@ -61,14 +61,14 @@ type GenericCell struct {
 	children           CellList // pointer to its children cells
 	atOrHigherThanNode bool     // true if the cell is at or higher than node level
 
-	totalGpuNum           int32                  // total GPU number of a cell
-	freeGpuNum            int32                  // free GPU number of a cell
-	freeGpuNumForPriority int32                  // free GPU number for a certain priority (lower priority considered as free)
-	usedGpuNumAtPriority  map[CellPriority]int32 // GPU number used by each priority
-	// GPU number used by the same priority (used for sorting cells in topologyAwareScheduler)
+	totalGpuNum          int32                  // total GPU number of a cell
+	freeGpuNum           int32                  // free GPU number of a cell
+	freeGpuNumAtPriority int32                  // free GPU number at a certain priority (lower priority considered as free)
+	usedGpuNumAtPriority map[CellPriority]int32 // GPU number used by each priority
+	// GPU number used by the same priority as that of the pod to be scheduled in topologyAwareScheduler (for sorting cells)
 	usedGpuNumSamePriority int32
-	// GPU number used by other priorities (used for sorting cells in topologyAwareScheduler)
-	usedGpuNumOtherPriority int32
+	// GPU number used by the lower priorities than that of the pod to be scheduled in topologyAwareScheduler (for sorting cells)
+	usedGpuNumLowerPriority int32
 }
 
 func (c *GenericCell) GetChain() CellChain {
@@ -111,8 +111,8 @@ func (c *GenericCell) IncreaseFreeGpuNum(delta int32) {
 	c.freeGpuNum += delta
 }
 
-func (c *GenericCell) GetFreeGpuNumForPriority(p CellPriority) int32 {
-	return c.freeGpuNumForPriority
+func (c *GenericCell) GetFreeGpuNumAtPriority(p CellPriority) int32 {
+	return c.freeGpuNumAtPriority
 }
 
 func (c *GenericCell) GetUsedGpuNumAtPriority(p CellPriority) int32 {
@@ -131,29 +131,29 @@ func (c *GenericCell) GetUsedGpuNumSamePriority() int32 {
 }
 
 func (c *GenericCell) GetUsedGpuNumOtherPriority() int32 {
-	return c.usedGpuNumOtherPriority
+	return c.usedGpuNumLowerPriority
 }
 
-func (c *GenericCell) SetUsedGpuNumForPriority(p CellPriority) {
+func (c *GenericCell) UpdateUsedGpuNumForPriority(p CellPriority) {
 	c.usedGpuNumSamePriority = c.usedGpuNumAtPriority[p]
-	c.usedGpuNumOtherPriority = 0
-	c.freeGpuNumForPriority = c.totalGpuNum
+	c.usedGpuNumLowerPriority = 0
+	c.freeGpuNumAtPriority = c.totalGpuNum
 	for priority, n := range c.usedGpuNumAtPriority {
 		if priority != p {
-			c.usedGpuNumOtherPriority += n
+			c.usedGpuNumLowerPriority += n
 		}
 		if priority >= p {
-			c.freeGpuNumForPriority -= n
+			c.freeGpuNumAtPriority -= n
 		}
 	}
 }
 
-func (c *GenericCell) SetUsedGpuNumAllPriority(p CellPriority) {
+func (c *GenericCell) UpdateUsedGpuNumAllPriority(p CellPriority) {
 	c.usedGpuNumSamePriority = c.totalGpuNum - c.freeGpuNum
-	c.freeGpuNumForPriority = c.totalGpuNum
+	c.freeGpuNumAtPriority = c.totalGpuNum
 	for priority, n := range c.usedGpuNumAtPriority {
 		if priority >= p {
-			c.freeGpuNumForPriority -= n
+			c.freeGpuNumAtPriority -= n
 		}
 	}
 }
@@ -161,13 +161,13 @@ func (c *GenericCell) SetUsedGpuNumAllPriority(p CellPriority) {
 // PhysicalCell defines a cell in the physical cluster.
 type PhysicalCell struct {
 	GenericCell
-	nodes          []string     // node names inside the cell
-	gpuIndices     []int32      // [-1] for cells at levels higher than node
-	pods           []*core.Pod  // pods running in this cell or its children
-	virtualCell    *VirtualCell // points to the bound virtual cell
-	tmpVirtualCell *VirtualCell // points to the temporarily bound virtual cell (before the binding is confirmed)
-	split          bool         // true when the cell has been split
-	reserved       bool         // true when this is a reserved cell
+	nodes               []string     // node names inside the cell
+	gpuIndices          []int32      // [-1] for cells at levels higher than node
+	pods                []*core.Pod  // pods running in this cell or its children
+	virtualCell         *VirtualCell // points to the bound virtual cell
+	preBoundVirtualCell *VirtualCell // points to the temporarily bound virtual cell (before the binding is confirmed)
+	split               bool         // true when the cell has been split
+	reserved            bool         // true when this is a reserved cell
 }
 
 func NewPhysicalCell(c CellChain, l CellLevel, g bool, n int32) *PhysicalCell {
@@ -237,12 +237,12 @@ func (c *PhysicalCell) SetVirtualCell(vc *VirtualCell) {
 	c.virtualCell = vc
 }
 
-func (c *PhysicalCell) GetTmpVirtualCell() *VirtualCell {
-	return c.tmpVirtualCell
+func (c *PhysicalCell) GetPreBoundVirtualCell() *VirtualCell {
+	return c.preBoundVirtualCell
 }
 
-func (c *PhysicalCell) SetTmpVirtualCell(vc *VirtualCell) {
-	c.tmpVirtualCell = vc
+func (c *PhysicalCell) SetPreBoundVirtualCell(vc *VirtualCell) {
+	c.preBoundVirtualCell = vc
 }
 
 func (c *PhysicalCell) IsSplit() bool {
@@ -264,12 +264,12 @@ func (c *PhysicalCell) SetReserved(reserved bool) {
 // VirtualCell defines a cell in a VC.
 type VirtualCell struct {
 	GenericCell
-	vc              api.VirtualClusterName // name of its VC
-	rid             api.ReservationId      // reservation ID
-	indexInChain    int32                  // index of the cell in the ChainCellList it belongs to (assigned in initialization)
-	preAssignedCell *VirtualCell           // top level cell of this cell chain
-	physicalCell    *PhysicalCell          // points to the bound physical cell
-	tmpPhysicalCell *PhysicalCell          // points to the temporarily bound physical cell (before the binding is confirmed)
+	vc                   api.VirtualClusterName // name of its VC
+	rid                  api.ReservationId      // reservation ID
+	indexInChain         int32                  // index of the cell in the ChainCellList it belongs to (assigned in initialization)
+	preAssignedCell      *VirtualCell           // top level cell of this cell chain
+	physicalCell         *PhysicalCell          // points to the bound physical cell
+	preBoundPhysicalCell *PhysicalCell          // points to the temporarily bound physical cell (before the binding is confirmed)
 }
 
 func NewVirtualCell(vc api.VirtualClusterName,
@@ -331,10 +331,10 @@ func (c *VirtualCell) SetPhysicalCell(pc *PhysicalCell) {
 	c.physicalCell = pc
 }
 
-func (c *VirtualCell) GetTmpPhysicalCell() *PhysicalCell {
-	return c.tmpPhysicalCell
+func (c *VirtualCell) GetPreBoundVirtualCell() *PhysicalCell {
+	return c.preBoundPhysicalCell
 }
 
-func (c *VirtualCell) SetTmpPhysicalCell(pc *PhysicalCell) {
-	c.tmpPhysicalCell = pc
+func (c *VirtualCell) SetPreBoundVirtualCell(pc *PhysicalCell) {
+	c.preBoundPhysicalCell = pc
 }
--- a/subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
+++ b/subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
@@ -54,18 +54,20 @@ type Cell interface {
 }
 
 type GenericCell struct {
-	chain                  CellChain
-	level                  CellLevel
-	priority               CellPriority // priority of a cell is max of those of its children and itself
-	parent                 Cell         // pointer to its parent cell
-	children               CellList     // pointer to its children cells
-	atOrHigherThanNode     bool
+	chain              CellChain
+	level              CellLevel
+	priority           CellPriority
+	parent             Cell     // pointer to its parent cell
+	children           CellList // pointer to its children cells
+	atOrHigherThanNode bool     // true if the cell is at or higher than node level
 
-	totalGpuNum int32
-	freeGpuNum int32
-	freeGpuNumForPriority int32
-	usedGpuNumAtPriority   map[CellPriority]int32
+	totalGpuNum           int32                  // total GPU number of a cell
+	freeGpuNum            int32                  // free GPU number of a cell
+	freeGpuNumForPriority int32                  // free GPU number for a certain priority (lower priority considered as free)
+	usedGpuNumAtPriority  map[CellPriority]int32 // GPU number used by each priority
+	// GPU number used by the same priority (used for sorting cells in topologyAwareScheduler)
 	usedGpuNumSamePriority int32
+	// GPU number used by other priorities (used for sorting cells in topologyAwareScheduler)
 	usedGpuNumOtherPriority int32
 }
 
@@ -159,24 +161,24 @@ func (c *GenericCell) SetUsedGpuNumAllPriority(p CellPriority) {
 // PhysicalCell defines a cell in the physical cluster.
 type PhysicalCell struct {
 	GenericCell
-	nodes       []string     // node names inside the cell
-	gpuIndices  []int32      // [-1] for cells at levels higher than node
-	pods        []*core.Pod  // pods running in this cell or its children
-	virtualCell *VirtualCell // points to the bound virtual cell
+	nodes          []string     // node names inside the cell
+	gpuIndices     []int32      // [-1] for cells at levels higher than node
+	pods           []*core.Pod  // pods running in this cell or its children
+	virtualCell    *VirtualCell // points to the bound virtual cell
 	tmpVirtualCell *VirtualCell // points to the temporarily bound virtual cell (before the binding is confirmed)
-	split bool
-	reserved    bool         // true when this is a reserved cell
+	split          bool         // true when the cell has been split
+	reserved       bool         // true when this is a reserved cell
 }
 
 func NewPhysicalCell(c CellChain, l CellLevel, g bool, n int32) *PhysicalCell {
 	return &PhysicalCell{
 		GenericCell: GenericCell{
-			chain:              c,
-			level:              l,
-			priority:           freePriority,
-			atOrHigherThanNode: g,
-			totalGpuNum:n,
-			freeGpuNum:n,
+			chain:                c,
+			level:                l,
+			priority:             freePriority,
+			atOrHigherThanNode:   g,
+			totalGpuNum:          n,
+			freeGpuNum:           n,
 			usedGpuNumAtPriority: map[CellPriority]int32{},
 		},
 		pods: []*core.Pod{},
@@ -267,7 +269,7 @@ type VirtualCell struct {
 	indexInChain    int32                  // index of the cell in the ChainCellList it belongs to (assigned in initialization)
 	preAssignedCell *VirtualCell           // top level cell of this cell chain
 	physicalCell    *PhysicalCell          // points to the bound physical cell
-	tmpPhysicalCell *PhysicalCell // points to the temporarily bound physical cell (before the binding is confirmed)
+	tmpPhysicalCell *PhysicalCell          // points to the temporarily bound physical cell (before the binding is confirmed)
 }
 
 func NewVirtualCell(vc api.VirtualClusterName,
@@ -280,11 +282,11 @@ func NewVirtualCell(vc api.VirtualClusterName,
 
 	return &VirtualCell{
 		GenericCell: GenericCell{
-			chain:              c,
-			level:              l,
-			priority:           freePriority,
-			atOrHigherThanNode: g,
-			totalGpuNum:n,
+			chain:                c,
+			level:                l,
+			priority:             freePriority,
+			atOrHigherThanNode:   g,
+			totalGpuNum:          n,
 			usedGpuNumAtPriority: map[CellPriority]int32{},
 		},
 		vc:              vc,
--- a/subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
+++ b/subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
@@ -48,7 +48,7 @@ type Cell interface {
 	GetUsedGpuNumAtPriority(CellPriority) int32
 	IncreaseUsedGpuNumAtPriority(CellPriority, int32)
 	GetUsedGpuNumSamePriority() int32
-	GetUsedGpuNumOtherPriority() int32
+	GetUsedGpuNumHigherPriority() int32
 	UpdateUsedGpuNumForPriority(CellPriority)
 	UpdateUsedGpuNumAllPriority(CellPriority)
 }
@@ -67,8 +67,8 @@ type GenericCell struct {
 	usedGpuNumAtPriority map[CellPriority]int32 // GPU number used by each priority
 	// GPU number used by the same priority as that of the pod to be scheduled in topologyAwareScheduler (for sorting cells)
 	usedGpuNumSamePriority int32
-	// GPU number used by the lower priorities than that of the pod to be scheduled in topologyAwareScheduler (for sorting cells)
-	usedGpuNumLowerPriority int32
+	// GPU number used by higher priorities than that of the pod to be scheduled in topologyAwareScheduler (for sorting cells)
+	usedGpuNumHigherPriority int32
 }
 
 func (c *GenericCell) GetChain() CellChain {
@@ -130,17 +130,17 @@ func (c *GenericCell) GetUsedGpuNumSamePriority() int32 {
 	return c.usedGpuNumSamePriority
 }
 
-func (c *GenericCell) GetUsedGpuNumOtherPriority() int32 {
-	return c.usedGpuNumLowerPriority
+func (c *GenericCell) GetUsedGpuNumHigherPriority() int32 {
+	return c.usedGpuNumHigherPriority
 }
 
 func (c *GenericCell) UpdateUsedGpuNumForPriority(p CellPriority) {
 	c.usedGpuNumSamePriority = c.usedGpuNumAtPriority[p]
-	c.usedGpuNumLowerPriority = 0
+	c.usedGpuNumHigherPriority = 0
 	c.freeGpuNumAtPriority = c.totalGpuNum
 	for priority, n := range c.usedGpuNumAtPriority {
-		if priority != p {
-			c.usedGpuNumLowerPriority += n
+		if priority > p {
+			c.usedGpuNumHigherPriority += n
 		}
 		if priority >= p {
 			c.freeGpuNumAtPriority -= n
--- a/subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
+++ b/subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
@@ -63,7 +63,7 @@ type GenericCell struct {
 
 	totalGpuNum          int32                  // total GPU number of a cell
 	freeGpuNum           int32                  // free GPU number of a cell
-	freeGpuNumAtPriority int32                  // free GPU number at a certain priority (lower priority considered as free)
+	freeGpuNumAtPriority int32                  // free GPU number at the priority of the pod to be scheduled (lower priority considered as free)
 	usedGpuNumAtPriority map[CellPriority]int32 // GPU number used by each priority
 	// GPU number used by the same priority as that of the pod to be scheduled in topologyAwareScheduler (for sorting cells)
 	usedGpuNumSamePriority int32
--- a/subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
+++ b/subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
@@ -51,6 +51,14 @@ type Cell interface {
 	UpdateUsedGpuNumAllPriority(CellPriority)
 }
 
+func CellEqual(c1 Cell, c2 Cell) bool {
+	if c1 == nil || c2 == nil {
+		return c1 == nil && c2 == nil
+	} else {
+		return c1.GetName() == c2.GetName()
+	}
+}
+
 type GenericCell struct {
 	chain              CellChain
 	level              CellLevel
--- a/subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
+++ b/subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
@@ -25,7 +25,8 @@ package algorithm
 import (
 	"fmt"
 	"github.com/microsoft/hivedscheduler/pkg/api"
-	"k8s.io/apimachinery/pkg/types"
+	"github.com/microsoft/hivedscheduler/pkg/internal"
+	core "k8s.io/api/core/v1"
 )
 
 // A Cell represents a set of GPUs affinitized by their interconnection topology.
@@ -40,14 +41,32 @@ type Cell interface {
 	SetParent(Cell)
 	GetChildren() CellList
 	SetChildren(CellList)
+	AtOrHigherThanNode() bool
+
+	IncreaseFreeGpuNum(int32)
+	GetFreeGpuNumForPriority(CellPriority) int32
+	GetUsedGpuNumAtPriority(CellPriority) int32
+	IncreaseUsedGpuNumAtPriority(CellPriority, int32)
+	GetUsedGpuNumSamePriority() int32
+	GetUsedGpuNumOtherPriority() int32
+	SetUsedGpuNumForPriority(CellPriority)
+	SetUsedGpuNumAllPriority(CellPriority)
 }
 
 type GenericCell struct {
-	chain    CellChain
-	level    CellLevel
-	priority CellPriority // priority of a cell is max of those of its children and itself
-	parent   Cell         // pointer to its parent cell
-	children CellList     // pointer to its children cells
+	chain                  CellChain
+	level                  CellLevel
+	priority               CellPriority // priority of a cell is max of those of its children and itself
+	parent                 Cell         // pointer to its parent cell
+	children               CellList     // pointer to its children cells
+	atOrHigherThanNode     bool
+
+	totalGpuNum int32
+	freeGpuNum int32
+	freeGpuNumForPriority int32
+	usedGpuNumAtPriority   map[CellPriority]int32
+	usedGpuNumSamePriority int32
+	usedGpuNumOtherPriority int32
 }
 
 func (c *GenericCell) GetChain() CellChain {
@@ -82,28 +101,85 @@ func (c *GenericCell) SetChildren(children CellList) {
 	c.children = children
 }
 
+func (c *GenericCell) AtOrHigherThanNode() bool {
+	return c.atOrHigherThanNode
+}
+
+func (c *GenericCell) IncreaseFreeGpuNum(delta int32) {
+	c.freeGpuNum += delta
+}
+
+func (c *GenericCell) GetFreeGpuNumForPriority(p CellPriority) int32 {
+	return c.freeGpuNumForPriority
+}
+
+func (c *GenericCell) GetUsedGpuNumAtPriority(p CellPriority) int32 {
+	return c.usedGpuNumAtPriority[p]
+}
+
+func (c *GenericCell) IncreaseUsedGpuNumAtPriority(p CellPriority, delta int32) {
+	c.usedGpuNumAtPriority[p] += delta
+	if c.usedGpuNumAtPriority[p] == 0 {
+		delete(c.usedGpuNumAtPriority, p)
+	}
+}
+
+func (c *GenericCell) GetUsedGpuNumSamePriority() int32 {
+	return c.usedGpuNumSamePriority
+}
+
+func (c *GenericCell) GetUsedGpuNumOtherPriority() int32 {
+	return c.usedGpuNumOtherPriority
+}
+
+func (c *GenericCell) SetUsedGpuNumForPriority(p CellPriority) {
+	c.usedGpuNumSamePriority = c.usedGpuNumAtPriority[p]
+	c.usedGpuNumOtherPriority = 0
+	c.freeGpuNumForPriority = c.totalGpuNum
+	for priority, n := range c.usedGpuNumAtPriority {
+		if priority != p {
+			c.usedGpuNumOtherPriority += n
+		}
+		if priority >= p {
+			c.freeGpuNumForPriority -= n
+		}
+	}
+}
+
+func (c *GenericCell) SetUsedGpuNumAllPriority(p CellPriority) {
+	c.usedGpuNumSamePriority = c.totalGpuNum - c.freeGpuNum
+	c.freeGpuNumForPriority = c.totalGpuNum
+	for priority, n := range c.usedGpuNumAtPriority {
+		if priority >= p {
+			c.freeGpuNumForPriority -= n
+		}
+	}
+}
+
 // PhysicalCell defines a cell in the physical cluster.
 type PhysicalCell struct {
 	GenericCell
 	nodes       []string     // node names inside the cell
 	gpuIndices  []int32      // [-1] for cells at levels higher than node
-	pods        []types.UID  // pods running in this cell or its children
+	pods        []*core.Pod  // pods running in this cell or its children
 	virtualCell *VirtualCell // points to the bound virtual cell
+	tmpVirtualCell *VirtualCell // points to the temporarily bound virtual cell (before the binding is confirmed)
+	split bool
 	reserved    bool         // true when this is a reserved cell
-
-	// a ChainCellList with this cell as the top.
-	// used for pod placement inside a cell and cell allocation in a reserved one.
-	internalCellList ChainCellList
 }
 
-func NewPhysicalCell(c CellChain, l CellLevel) *PhysicalCell {
+func NewPhysicalCell(c CellChain, l CellLevel, g bool, n int32) *PhysicalCell {
 	return &PhysicalCell{
 		GenericCell: GenericCell{
-			chain:    c,
-			level:    l,
-			priority: freePriority,
+			chain:              c,
+			level:              l,
+			priority:           freePriority,
+			atOrHigherThanNode: g,
+			totalGpuNum:n,
+			freeGpuNum:n,
+			usedGpuNumAtPriority: map[CellPriority]int32{},
 		},
-		pods: []types.UID{},
+		pods: []*core.Pod{},
 	}
 }
 
@@ -124,14 +200,11 @@ func (c *PhysicalCell) SetPhysicalResources(nodes []string, gpuIndices []int32) 
 	c.gpuIndices = gpuIndices
 }
 
-func (c *PhysicalCell) AddPod(pod types.UID) {
+func (c *PhysicalCell) AddPod(pod *core.Pod) {
 	c.pods = append(c.pods, pod)
-	if c.parent != nil {
-		c.parent.(*PhysicalCell).AddPod(pod)
-	}
 }
 
-func (c *PhysicalCell) DeletePod(pod types.UID) {
+func (c *PhysicalCell) DeletePod(pod *core.Pod) {
 	idx := -1
 	for i, p := range c.pods {
 		if pod == p {
@@ -140,12 +213,9 @@ func (c *PhysicalCell) DeletePod(pod types.UID) {
 		}
 	}
 	if idx == -1 {
-		panic(fmt.Sprintf("Error when deleting pod %v: not exist in cell %v!", pod, c.GetName()))
+		panic(fmt.Sprintf("Error when deleting pod %v: not exist in cell %v!", internal.Key(pod), c.GetName()))
 	} else {
 		c.pods = append(c.pods[:idx], c.pods[idx+1:]...)
-	}
-	if c.parent != nil {
-		c.parent.(*PhysicalCell).DeletePod(pod)
 	}
 }
 
@@ -153,22 +223,8 @@ func (c *PhysicalCell) HasPod() bool {
 	return len(c.pods) != 0
 }
 
-func (c *PhysicalCell) InitPhysicalCellList() {
-	c.internalCellList = NewChainCellList(c.level)
-	c.internalCellList[c.level] = append(c.internalCellList[c.level], c)
-	for level := c.level - 1; level >= CellLevel(1); level-- {
-		for _, cc := range c.internalCellList[level+1] {
-			c.internalCellList[level] = append(c.internalCellList[level], cc.GetChildren()...)
-		}
-	}
-}
-
-func (c *PhysicalCell) GetPhysicalCellList() ChainCellList {
-	return c.internalCellList
-}
-
-func (c *PhysicalCell) ClearPhysicalCellList() {
-	c.internalCellList = nil
+func (c *PhysicalCell) GetPods() []*core.Pod {
+	return c.pods
 }
 
 func (c *PhysicalCell) GetVirtualCell() *VirtualCell {
@@ -177,6 +233,22 @@ func (c *PhysicalCell) GetVirtualCell() *VirtualCell {
 
 func (c *PhysicalCell) SetVirtualCell(vc *VirtualCell) {
 	c.virtualCell = vc
+}
+
+func (c *PhysicalCell) GetTmpVirtualCell() *VirtualCell {
+	return c.tmpVirtualCell
+}
+
+func (c *PhysicalCell) SetTmpVirtualCell(vc *VirtualCell) {
+	c.tmpVirtualCell = vc
+}
+
+func (c *PhysicalCell) IsSplit() bool {
+	return c.split
+}
+
+func (c *PhysicalCell) SetSplit(split bool) {
+	c.split = split
 }
 
 func (c *PhysicalCell) IsReserved() bool {
@@ -195,14 +267,25 @@ type VirtualCell struct {
 	indexInChain    int32                  // index of the cell in the ChainCellList it belongs to (assigned in initialization)
 	preAssignedCell *VirtualCell           // top level cell of this cell chain
 	physicalCell    *PhysicalCell          // points to the bound physical cell
+	tmpPhysicalCell *PhysicalCell // points to the temporarily bound physical cell (before the binding is confirmed)
 }
 
-func NewVirtualCell(vc api.VirtualClusterName, c CellChain, l CellLevel, i int32, pac *VirtualCell) *VirtualCell {
+func NewVirtualCell(vc api.VirtualClusterName,
+	c CellChain,
+	l CellLevel,
+	g bool,
+	n int32,
+	i int32,
+	pac *VirtualCell) *VirtualCell {
+
 	return &VirtualCell{
 		GenericCell: GenericCell{
-			chain:    c,
-			level:    l,
-			priority: freePriority,
+			chain:              c,
+			level:              l,
+			priority:           freePriority,
+			atOrHigherThanNode: g,
+			totalGpuNum:n,
+			usedGpuNumAtPriority: map[CellPriority]int32{},
 		},
 		vc:              vc,
 		indexInChain:    i,
@@ -244,4 +327,12 @@ func (c *VirtualCell) GetPhysicalCell() *PhysicalCell {
 
 func (c *VirtualCell) SetPhysicalCell(pc *PhysicalCell) {
 	c.physicalCell = pc
+}
+
+func (c *VirtualCell) GetTmpPhysicalCell() *PhysicalCell {
+	return c.tmpPhysicalCell
+}
+
+func (c *VirtualCell) SetTmpPhysicalCell(pc *PhysicalCell) {
+	c.tmpPhysicalCell = pc
 }
--- a/subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
+++ b/subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
@@ -25,8 +25,6 @@ package algorithm
 import (
 	"fmt"
 	"github.com/microsoft/hivedscheduler/pkg/api"
-	"github.com/microsoft/hivedscheduler/pkg/internal"
-	core "k8s.io/api/core/v1"
 )
 
 // A Cell represents a set of GPUs affinitized by their interconnection topology.
@@ -161,13 +159,13 @@ func (c *GenericCell) UpdateUsedGpuNumAllPriority(p CellPriority) {
 // PhysicalCell defines a cell in the physical cluster.
 type PhysicalCell struct {
 	GenericCell
-	nodes               []string     // node names inside the cell
-	gpuIndices          []int32      // [-1] for cells at levels higher than node
-	pods                []*core.Pod  // pods running in this cell or its children
-	virtualCell         *VirtualCell // points to the bound virtual cell
-	preBoundVirtualCell *VirtualCell // points to the temporarily bound virtual cell (before the binding is confirmed)
-	split               bool         // true when the cell has been split
-	reserved            bool         // true when this is a reserved cell
+	nodes               []string             // node names inside the cell
+	gpuIndices          []int32              // [-1] for cells at levels higher than node
+	affinityGroups      []*AlgoAffinityGroup // affinity groups using this cell
+	virtualCell         *VirtualCell         // points to the bound virtual cell
+	preBoundVirtualCell *VirtualCell         // points to the temporarily bound virtual cell (before the binding is confirmed)
+	split               bool                 // true when the cell has been split
+	reserved            bool                 // true when this is a reserved cell
 }
 
 func NewPhysicalCell(c CellChain, l CellLevel, g bool, n int32) *PhysicalCell {
@@ -181,7 +179,7 @@ func NewPhysicalCell(c CellChain, l CellLevel, g bool, n int32) *PhysicalCell {
 			freeGpuNum:           n,
 			usedGpuNumAtPriority: map[CellPriority]int32{},
 		},
-		pods: []*core.Pod{},
+		affinityGroups: []*AlgoAffinityGroup{},
 	}
 }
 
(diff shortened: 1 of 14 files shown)


subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/intra_vc_scheduler.go
```diff
@@ -54,27 +74,27 @@ func (s *defaultIntraVCScheduler) getReservedCellList() map[api.ReservationId]Ch
...
+	if sr.reservationId != "" {
+		scheduler = s.schedulersReserved[sr.reservationId]
 	} else {
-		fullCellList = s.virtualNonReservedCellList[cr.Chain]
+		scheduler = s.schedulersNonReserved[sr.chain]
 	}
-
-	vcCellView := getCellViewWithPriority(fullCellList, cr.Priority)
-	c := buddyAlloc(vcCellView, cr.Level)
-	if c != nil {
```
yqwang-ms: Previous cell based intra VC scheduling (assume we can support a fraction of cell allocation inside VC) may better than new algo in some cases, so can we also keep original algo, and let user to choose to use which algo by themselves?

If so, pls make comments to compare new and old algo by pros/cons, suitable scenarios, etc.

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
```diff
@@ -0,0 +1,284 @@
...
+
+import (
+	"fmt"
+	"sort"
+)
+
+// topologyAwareScheduler can schedule a set of pods on a cluster view.
+type topologyAwareScheduler struct {
+	clusterView              CellList  // a list of node- (or lower-)level cells
+	podGpuNumbers            []int32   // GPU number of each pod to schedule
```
yqwang-ms: For clusterView and currentSolutionIndices:

One field would better to only mean one thing, so split to 2 fields or remove the field and pass by return/args.

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
```diff
@@ -0,0 +1,284 @@
...
+
+package algorithm
+
+import (
+	"fmt"
+	"sort"
+)
+
+// topologyAwareScheduler can schedule a set of pods on a cluster view.
+type topologyAwareScheduler struct {
```
yqwang-ms: Brief comment what is the behavior of this algo

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/common/types.go
```diff
@@ -41,6 +41,10 @@ func NewSet(items ...T) Set {
 	return s
 }
 
+func (s Set) Items() map[T]Empty {
+	return s.items
+}
```
yqwang-ms: remove?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/api/types.go
```diff
@@ -96,15 +96,20 @@ type PodBindInfo struct {
...
-	VirtualCellIndex      int32    `yaml:"virtualCellIndex"`
+	GpuIsolation      []int32                   `yaml:"gpuIsolation"`
+	CellChain         string                    `yaml:"cellChain"`
+	GroupScheduleInfo []GroupMemberScheduleInfo `yaml:"groupScheduleInfo"`
+}
+
+type GroupMemberScheduleInfo struct {
+	GpuNumber     int32              `yaml:"gpuNumber"`
+	PodPlacements []PodPlacementInfo `yaml:"podPlacements"`
+}
```
yqwang-ms: To align with above AffinityGroup/AffinityGroupMember, how about this?

 AffinityGroupBindInfo []AffinityGroupMemberBindInfo
}

type AffinityGroupMemberBindInfo struct {
 GpuNumber     int32
 PodPlacements []PodPlacementInfo
}

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/api/types.go
```diff
@@ -96,15 +96,20 @@ type PodBindInfo struct {
...
-	VirtualCellLevel      int32    `yaml:"virtualCellLevel"`
-	VirtualCellIndex      int32    `yaml:"virtualCellIndex"`
+	GpuIsolation      []int32                   `yaml:"gpuIsolation"`
+	CellChain         string                    `yaml:"cellChain"`
+	GroupScheduleInfo []GroupMemberScheduleInfo `yaml:"groupScheduleInfo"`
+}
+
+type GroupMemberScheduleInfo struct {
+	GpuNumber     int32              `yaml:"gpuNumber"`
+	PodPlacements []PodPlacementInfo `yaml:"podPlacements"`
```
yqwang-ms: It can be calculated from PodPlacements, so why need it to persist and recover?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/api/types.go
```diff
@@ -96,15 +96,20 @@ type PodBindInfo struct {
...
+}
+
+type PodPlacementInfo struct {
+	PhysicalNode          string  `yaml:"physicalNodes"`
```
yqwang-ms: >physicalNodes [](start = 38, length = 13)

physicalNode?

yqwang-ms:
> }

Remove?

---
Refers to: subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/api/types.go:118 in 93bfc6e. [](commit_id = 93bfc6ee74f052e1be9a0e7586d1dc4519910df2, deletion_comment = False)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/api/types.go
```diff
@@ -96,15 +96,20 @@ type PodBindInfo struct {
...
+	PhysicalGpuIndices    []int32 `yaml:"physicalGpuIndices"`
+	VirtualGpuCellIndices []int32 `yaml:"virtualGpuCellIndices"`
```
yqwang-ms: >VirtualGpuCellIndices [](start = 1, length = 21)

Why not VirtualCellIndices?
And then:
type PodPlacementInfo struct {
 Node          string
 GpuIndices    []int32
 CellIndices []int32
}

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/types.go
```diff
@@ -34,12 +35,12 @@ type (
...
-	ReservationId api.ReservationId
-	Chain         CellChain
-	Level         CellLevel
-	Priority      CellPriority
+type schedulingRequest struct {
+	vc            api.VirtualClusterName
+	reservationId api.ReservationId
+	chain         CellChain
+	podGpuNumbers map[int32]int32
+	priority      CellPriority
```
yqwang-ms: Comment map key and val

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/types.go
```diff
@@ -57,11 +58,49 @@ func (cl CellList) String() string {
 	return strings.Join(names, ", ")
 }
 
+func (cl CellList) removeCell(c Cell) CellList {
+	index := -1
+	for i, cc := range cl {
+		if cc == c {
```
yqwang-ms: >if cc == c { [](start = 2, length = 12)

How does it compare object? By value or reference, deep compare or not?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/types.go
```diff
@@ -57,11 +58,49 @@ func (cl CellList) String() string {
 	return strings.Join(names, ", ")
 }
 
+func (cl CellList) removeCell(c Cell) CellList {
```
yqwang-ms: >removeCell [](start = 19, length = 10)

removeCell -> remove?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
```diff
@@ -40,14 +41,34 @@ type Cell interface {
...
+	children           CellList // pointer to its children cells
+	atOrHigherThanNode bool     // true if the cell is at or higher than node level
+
+	totalGpuNum           int32                  // total GPU number of a cell
+	freeGpuNum            int32                  // free GPU number of a cell
+	freeGpuNumForPriority int32                  // free GPU number for a certain priority (lower priority considered as free)
+	usedGpuNumAtPriority  map[CellPriority]int32 // GPU number used by each priority
+	// GPU number used by the same priority (used for sorting cells in topologyAwareScheduler)
+	usedGpuNumSamePriority int32
+	// GPU number used by other priorities (used for sorting cells in topologyAwareScheduler)
```
yqwang-ms: >other priorities [](start = 23, length = 16)

Does it only include "lower priorities"

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/types.go
```diff
@@ -76,9 +115,16 @@ func (ccl ChainCellList) String() string {
 	return str
 }
 
+func (ccl ChainCellList) removeCell(c Cell, l CellLevel) {
```
yqwang-ms: >removeCell [](start = 25, length = 10)

removeCell -> remove?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/types.go
```diff
@@ -76,9 +115,16 @@ func (ccl ChainCellList) String() string {
...
+	klog.Infof("Cell removed from cell list: %v", c.GetName())
+}
+
 type AlgoAffinityGroup struct {
-	cell               *PhysicalCell
-	unallocatedPodNums map[int32]int32 // GpuNum -> PodNum
+	cell                 *PhysicalCell
+	unallocatedPodNums   map[int32]int32      // GpuNum -> PodNum
+	physicalGpuPlacement map[int32][]CellList // GpuNum -> a list of pods -> a list of GPUs of each pod
+	vcGpuPlacement       map[int32][]CellList // same as above
```
yqwang-ms: >same as above [](start = 46, length = 13)

So what is the difference with above? Comment it.

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/intra_vc_scheduler.go
```diff
@@ -28,22 +28,42 @@ import (
...
+// It stores two maps of ChainCellList, one for reserved cells, the other for non-reserved ones.
+// It should be able to return a set of GPU placements in the VC for a scheduling request.
 type intraVCScheduler interface {
 	getNonReservedCellList() map[CellChain]ChainCellList
 	getReservedCellList() map[api.ReservationId]ChainCellList
 
-	// Cell allocation inside a VC. We use buddy alloc for default
-	// (see defaultIntraVCScheduler).
-	allocateCell(CellRequest) *VirtualCell
+	// Scheduling a series of pods inside a VC. We use topologyAwareScheduler by default.
```
yqwang-ms: >a series of pods [](start = 15, length = 16)

The "series of pods" must be the same as "pods in the same affinity group", if so why not directly use latter?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/types.go
yqwang-ms: The podGpuNumbers must be the same as "pods in the same affinity group", if so why not directly use latter?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/intra_vc_scheduler.go
```diff
@@ -28,22 +28,42 @@ import (
...
+	schedule(schedulingRequest) map[int32][]CellList
 }
 
 type defaultIntraVCScheduler struct {
 	virtualNonReservedCellList map[CellChain]ChainCellList
 	virtualReservedCellList    map[api.ReservationId]ChainCellList
+	schedulersNonReserved      map[CellChain]*topologyAwareScheduler
+	schedulersReserved         map[api.ReservationId]*topologyAwareScheduler
```
yqwang-ms: nonReservedSchedulers
reservedSchedulers

yqwang-ms: What is the benefits to make a dedicated scheduler for each CellChain and ReservationId?
Why not share the same one?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/intra_vc_scheduler.go
```diff
@@ -54,27 +74,27 @@ func (s *defaultIntraVCScheduler) getReservedCellList() map[api.ReservationId]Ch
...
-	vcCellView := getCellViewWithPriority(fullCellList, cr.Priority)
-	c := buddyAlloc(vcCellView, cr.Level)
-	if c != nil {
-		return c.(*VirtualCell) // downgrade pod running on c
-	} else {
+	var result map[int32][]CellList
+	if scheduler != nil {
+		result = scheduler.schedule(sr.podGpuNumbers, sr.priority)
+	}
+	if scheduler == nil || result == nil {
```
yqwang-ms: >scheduler == nil ||  [](start = 4, length = 20)

"scheduler == nil || " can be removed

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/intra_vc_scheduler.go
```diff
@@ -54,27 +74,27 @@ func (s *defaultIntraVCScheduler) getReservedCellList() map[api.ReservationId]Ch
...
+			str = fmt.Sprintf("chain %v", sr.chain)
 		}
-		klog.Infof("Insufficient quota in VC %v for cell request: %v, level %v, priority %v",
-			cr.VC, str, cr.Level, cr.Priority)
-		return nil
+		klog.Infof("Insufficient quota in VC %v for scheduling request: %v, GPU numbers %v, priority %v",
+			sr.vc, str, sr.podGpuNumbers, sr.priority)
 	}
+	klog.Infof("%v", result)
+	return result
```
yqwang-ms: format the log

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
```diff
@@ -40,14 +41,34 @@ type Cell interface {
...
+	chain              CellChain
+	level              CellLevel
+	priority           CellPriority
+	parent             Cell     // pointer to its parent cell
+	children           CellList // pointer to its children cells
+	atOrHigherThanNode bool     // true if the cell is at or higher than node level
+
+	totalGpuNum           int32                  // total GPU number of a cell
+	freeGpuNum            int32                  // free GPU number of a cell
+	freeGpuNumForPriority int32                  // free GPU number for a certain priority (lower priority considered as free)
```
yqwang-ms: >freeGpuNumForPriority [](start = 1, length = 21)

Align with usedGpuNumAtPriority, so rename to freeGpuNumAtPriority?

And why it is not a map, and what is difference with freeGpuNum?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
```diff
@@ -40,14 +41,34 @@ type Cell interface {
...
+	usedGpuNumAtPriority  map[CellPriority]int32 // GPU number used by each priority
+	// GPU number used by the same priority (used for sorting cells in topologyAwareScheduler)
```
yqwang-ms: >same [](start = 27, length = 4)

same with what? GenericCell.priority?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
yqwang-ms: >other [](start = 23, length = 5)

Can it be greater than GenericCell.priority?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
```diff
@@ -82,28 +103,85 @@ func (c *GenericCell) SetChildren(children CellList) {
...
+func (c *GenericCell) GetFreeGpuNumForPriority(p CellPriority) int32 {
+	return c.freeGpuNumForPriority
+}
+
+func (c *GenericCell) GetUsedGpuNumAtPriority(p CellPriority) int32 {
+	return c.usedGpuNumAtPriority[p]
+}
+
+func (c *GenericCell) IncreaseUsedGpuNumAtPriority(p CellPriority, delta int32) {
+	c.usedGpuNumAtPriority[p] += delta
```
yqwang-ms: >usedGpuNumAtPriority[p] [](start = 3, length = 23)

What if usedGpuNumAtPriority[p] does not exist

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
```diff
@@ -82,28 +103,85 @@ func (c *GenericCell) SetChildren(children CellList) {
...
+
+func (c *GenericCell) GetUsedGpuNumSamePriority() int32 {
+	return c.usedGpuNumSamePriority
+}
+
+func (c *GenericCell) GetUsedGpuNumOtherPriority() int32 {
+	return c.usedGpuNumOtherPriority
+}
+
+func (c *GenericCell) SetUsedGpuNumForPriority(p CellPriority) {
```
yqwang-ms: >SetUsedGpuNumForPriority [](start = 22, length = 24)

UpdateUsedGpuNumForPriority

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
```diff
@@ -82,28 +103,85 @@ func (c *GenericCell) SetChildren(children CellList) {
...
+		if priority != p {
+			c.usedGpuNumOtherPriority += n
+		}
+		if priority >= p {
+			c.freeGpuNumForPriority -= n
+		}
+	}
+}
+
+func (c *GenericCell) SetUsedGpuNumAllPriority(p CellPriority) {
```
yqwang-ms: >SetUsedGpuNumAllPriority [](start = 22, length = 24)

UpdateUsedGpuNumAllPriority

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
```diff
@@ -82,28 +103,85 @@ func (c *GenericCell) SetChildren(children CellList) {
...
-	reserved    bool         // true when this is a reserved cell
-
-	// a ChainCellList with this cell as the top.
-	// used for pod placement inside a cell and cell allocation in a reserved one.
-	internalCellList ChainCellList
+	nodes          []string     // node names inside the cell
+	gpuIndices     []int32      // [-1] for cells at levels higher than node
+	pods           []*core.Pod  // pods running in this cell or its children
+	virtualCell    *VirtualCell // points to the bound virtual cell
+	tmpVirtualCell *VirtualCell // points to the temporarily bound virtual cell (before the binding is confirmed)
```
yqwang-ms: >tmpVirtualCell [](start = 1, length = 14)

preboundVirtualCell

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
```diff
@@ -195,14 +269,25 @@ type VirtualCell struct {
 	indexInChain    int32                  // index of the cell in the ChainCellList it belongs to (assigned in initialization)
 	preAssignedCell *VirtualCell           // top level cell of this cell chain
 	physicalCell    *PhysicalCell          // points to the bound physical cell
+	tmpPhysicalCell *PhysicalCell          // points to the temporarily bound physical cell (before the binding is confirmed)
```
yqwang-ms: >tmpPhysicalCell [](start = 1, length = 15)

preboundPhysicalCell

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -29,16 +29,19 @@ import (
...
+	"math"
 	"sync"
 )
 
 type HivedAlgorithm struct {
-	// intra-VC scheduler for each VC
+	// scheduler in each VC
 	vcSchedulers map[api.VirtualClusterName]intraVCScheduler
+	// scheduler for opportunistic pods
+	opportunisticSchedulers map[CellChain]*topologyAwareScheduler
```
yqwang-ms: Comment why opportunistic use a different scheduler?
Why it hard codes to be topologyAwareScheduler?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -529,242 +599,209 @@ func (h *HivedAlgorithm) findVirtualCell(
...
+					[]int32, len(podPhysicalPlacements[i]))
+				gms.PodPlacements[i].VirtualGpuCellIndices = make(
+					[]int32, len(podPhysicalPlacements[i]))
+				for j := 0; j < len(podPhysicalPlacements[i]); j++ {
+					pGpu := podPhysicalPlacements[i][j].(*PhysicalCell)
+					nodes, gpuIndices := pGpu.GetPhysicalPlacement()
+					gms.PodPlacements[i].PhysicalNode = nodes[0]
+					gms.PodPlacements[i].PhysicalGpuIndices[j] = gpuIndices[0]
+					if newGroup && pGpu.HasPod() {
+						preemptionVictims.Add(pGpu.GetPods()[0])
```
yqwang-ms: >(pGpu.GetPods()[0]) [](start = 27, length = 19)

Why not add all pods

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -529,242 +599,209 @@ func (h *HivedAlgorithm) findVirtualCell(
...
+		)
+		preemptionVictims := common.NewSet()
+		gsi := make([]api.GroupMemberScheduleInfo, len(groupPhysicalPlacement))
+		groupMemberIndex := 0
+		for podGpuNum, podPhysicalPlacements := range groupPhysicalPlacement {
+			gms := api.GroupMemberScheduleInfo{
+				GpuNumber:     podGpuNum,
+				PodPlacements: make([]api.PodPlacementInfo, len(podPhysicalPlacements)),
+			}
+			for i := int32(0); i < int32(len(podPhysicalPlacements)); i++ {
```
yqwang-ms: >i [](start = 7, length = 1)

podIdx

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -529,242 +599,209 @@ func (h *HivedAlgorithm) findVirtualCell(
...
+				gms.PodPlacements[i].PhysicalGpuIndices = make(
+					[]int32, len(podPhysicalPlacements[i]))
+				gms.PodPlacements[i].VirtualGpuCellIndices = make(
+					[]int32, len(podPhysicalPlacements[i]))
+				for j := 0; j < len(podPhysicalPlacements[i]); j++ {
```
yqwang-ms: >j [](start = 8, length = 1)

cellIdx

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -529,242 +599,209 @@ func (h *HivedAlgorithm) findVirtualCell(
...
+					pGpu := podPhysicalPlacements[i][j].(*PhysicalCell)
+					nodes, gpuIndices := pGpu.GetPhysicalPlacement()
+					gms.PodPlacements[i].PhysicalNode = nodes[0]
```
yqwang-ms: >.PhysicalNode = nodes[0] [](start = 25, length = 24)

when iterate next cell, previous assignment is overridden. Is it by design?
Can we ensure gms.PodPlacements[i].PhysicalNode is immutable after first assignment?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -529,242 +599,209 @@ func (h *HivedAlgorithm) findVirtualCell(
...
+					gms.PodPlacements[i].PhysicalGpuIndices[j] = gpuIndices[0]
```
yqwang-ms: >gpuIndices[0] [](start = 50, length = 13)

Why only choose the first gpu?

yqwang-ms: Or they must be one gpu cells? So do we have to reuse the cell structure? It may introduce confuse, why not use gpu directly.

---
In reply to: [332835528](https://github.com/microsoft/pai/pull/3684#discussion_r332835528) [](ancestors = 332835528)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -88,68 +94,37 @@ func (h *HivedAlgorithm) Schedule(pod *core.Pod, suggestedNodes []string) intern
...
+		// gpu number -> a set of pods -> a set of GPUs of each pod
+		groupPhysicalPlacement map[int32][]CellList
+		groupVcPlacement       map[int32][]CellList
+		podIndex               int32
+		newGroup               bool
+	)
+	if group := h.allocatedAffinityGroups[s.AffinityGroup.Name]; group == nil {
+		newGroup = true
+		klog.Infof("Scheduling new affinity group %v", s.AffinityGroup.Name)
+		groupPhysicalPlacement, groupVcPlacement = h.scheduleNewAffinityGroup(pod, s)
```
yqwang-ms: >scheduleNewAffinityGroup [](start = 47, length = 24)

Why not directly generate and return the GroupMemberScheduleInfo

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm_test.go
```diff
@@ -195,98 +188,91 @@ var pss = map[types.UID]api.PodSchedulingSpec{
...
-	}, "pod3": { // share the same cell with pod4
-		VirtualCluster: "VC1",
-		Priority:       1,
-		ReservationId:  "",
-		GpuType:        "DGX2-V100",
-		GpuNumber:      1,
-		AffinityGroup:  group3,
-	}, "pod4": { // share the same cell with pod3
+	}, "pod3": { // non-buddy of pod 1 & 2 (avoidance of preemption)
 		VirtualCluster: "VC1",
```
yqwang-ms: Test to allocate a fraction of cell inside VC?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -29,16 +29,19 @@ import (
 	"github.com/microsoft/hivedscheduler/pkg/internal"
 	core "k8s.io/api/core/v1"
 	"k8s.io/klog"
+	"math"
 	"sync"
 )
 
 type HivedAlgorithm struct {
-	// intra-VC scheduler for each VC
+	// scheduler in each VC
```
yqwang-ms: Comment the difference, relationship and responsibility with topologyAwareScheduler

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
```diff
@@ -0,0 +1,284 @@
...
+	// pack pods cross different priorities, or inside each priority. the former is for opportunistic
+	// pod scheduling (stay away from regular pods), because regular pods can avoid preempting opportunistic pods
+	// only among buddy cells. the latter is for intra-VC scheduling, because high-priority can avoid preemption
+	// in the whole cluster view, and hence we can pack pods with different priorities.
+	crossPriorityPack bool
+}
+
+// newTopologyAwareScheduler initializes the scheduler by extracting node-level cells
+// (lower-level if no node-level) from a chain cell list.
+func newTopologyAwareScheduler(ccl ChainCellList, crossPriorityPack bool) *topologyAwareScheduler {
```
yqwang-ms: >newTopologyAwareScheduler [](start = 5, length = 25)

Make public

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
```diff
@@ -0,0 +1,284 @@
...
+		}
+	}
+	clusterView := make(CellList, len(ccl[l]))
+	copy(clusterView, ccl[l])
+	return &topologyAwareScheduler{
+		clusterView:       clusterView,
+		crossPriorityPack: crossPriorityPack}
+}
+
+func (t *topologyAwareScheduler) schedule(
```
yqwang-ms: >schedule [](start = 33, length = 8)

Make public

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
```diff
@@ -0,0 +1,284 @@
...
+func sortGpuNumbers(n []int32) {
+	tmp := make([]int, len(n))
+	for i := 0; i < len(n); i++ {
+		tmp[i] = int(n[i])
+	}
+	sort.Ints(tmp)
+	for i := 0; i < len(tmp); i++ {
+		n[i] = int32(tmp[i])
+	}
+}
```
yqwang-ms: Why not directly use sort.Ints(tmp)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
```diff
@@ -0,0 +1,284 @@
...
+// topologyAwareScheduler can schedule a set of pods on a cluster view.
+type topologyAwareScheduler struct {
+	clusterView              CellList  // a list of node- (or lower-)level cells
+	podGpuNumbers            []int32   // GPU number of each pod to schedule
+	currentSolutionIndices   []int32   // indices of currently picked nodes (or GPUs)
+	currentLCA               CellList  // current LCA (lowest common ancestor) of currently picked GPUs
+	lowestLCASolution        CellList  // GPUs with the lowest LCA
+	lowestLCASolutionIndices []int32   // Indices of the GPUs with the lowest LCA
+	lowestLCALevel           CellLevel // level of the lowest LCA
+	// pack pods cross different priorities, or inside each priority. the former is for opportunistic
```
yqwang-ms: >// pack pods cross different priorities, or inside each priority. the former is for opportunistic [](start = 1, length = 97)

If so, why opportunistic's crossPriorityPack = false

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
```diff
@@ -0,0 +1,284 @@
...
+
+// topologyAwareScheduler can schedule a set of pods on a cluster view.
+type topologyAwareScheduler struct {
+	clusterView              CellList  // a list of node- (or lower-)level cells
+	podGpuNumbers            []int32   // GPU number of each pod to schedule
+	currentSolutionIndices   []int32   // indices of currently picked nodes (or GPUs)
+	currentLCA               CellList  // current LCA (lowest common ancestor) of currently picked GPUs
+	lowestLCASolution        CellList  // GPUs with the lowest LCA
+	lowestLCASolutionIndices []int32   // Indices of the GPUs with the lowest LCA
+	lowestLCALevel           CellLevel // level of the lowest LCA
```
yqwang-ms: Do not store temp var as field, move them into func temp var.

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
```diff
@@ -0,0 +1,284 @@
...
+func (t *topologyAwareScheduler) schedule(
+	podGpuNumbers map[int32]int32,
+	p CellPriority) map[int32][]CellList {
+
+	t.podGpuNumbers = []int32{}
+	for gpuNum, podNum := range podGpuNumbers {
+		for i := int32(0); i < podNum; i++ {
+			t.podGpuNumbers = append(t.podGpuNumbers, gpuNum)
+		}
+	}
```
yqwang-ms: If there are 1000 pods, it will make us not efficient. Why need to flatten it?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/types.go
```diff
@@ -57,11 +58,49 @@ func (cl CellList) String() string {
...
+		return true
+	} else if cl[i].GetUsedGpuNumSamePriority() < cl[j].GetUsedGpuNumSamePriority() {
+		return false
+	} else if cl[i].GetUsedGpuNumOtherPriority() < cl[j].GetUsedGpuNumOtherPriority() {
+		return true
+	} else {
+		return false
+	}
+}
+
```
yqwang-ms: Refer it in the scheduling policy comments

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
```diff
@@ -0,0 +1,284 @@
...
+
+// findNodesForPods finds a set of nodes that can accommodate the GPU requirements of the pods.
+func (t *topologyAwareScheduler) findNodesForPods(p CellPriority) bool {
+	sort.Sort(t.clusterView)
+	t.currentSolutionIndices = make([]int32, len(t.podGpuNumbers))
+	podIndex := 0
+	pickedGpuNum := int32(0)
+	var node Cell
+	for nodeIndex := 0; nodeIndex < len(t.clusterView); {
+		node = t.clusterView[nodeIndex]
```
yqwang-ms: Can you ensure the algo always can find the solution, regardless the iteration order?

For example:
1. clusterView = 2GPU Node, 1GPU Node
2. podGpuNumbers = 1GPU Pod, 2GPU Pod

First 1GPU Pod may allocate to 2GPU Node, but the latter pod cannot be fitted anymore.

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
```diff
@@ -0,0 +1,284 @@
...
+	"sort"
+)
+
+// topologyAwareScheduler can schedule a set of pods on a cluster view.
+type topologyAwareScheduler struct {
+	clusterView              CellList  // a list of node- (or lower-)level cells
+	podGpuNumbers            []int32   // GPU number of each pod to schedule
+	currentSolutionIndices   []int32   // indices of currently picked nodes (or GPUs)
+	currentLCA               CellList  // current LCA (lowest common ancestor) of currently picked GPUs
+	lowestLCASolution        CellList  // GPUs with the lowest LCA
```
yqwang-ms: >lowest LCA [](start = 53, length = 10)

What is lowest lowest common ancestor

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
```diff
@@ -0,0 +1,284 @@
...
+	selectedNodes := make(CellList, len(t.podGpuNumbers))
+	for i := 0; i < len(t.currentSolutionIndices); i++ {
+		selectedNodes[i] = t.clusterView[t.currentSolutionIndices[i]]
+	}
+	nodeGpus := map[Cell]CellList{}
+	podPlacements := map[int32][]CellList{}
+	for podIndex := 0; podIndex < len(t.podGpuNumbers); podIndex++ {
+		gpuNumber := t.podGpuNumbers[podIndex]
+		node := selectedNodes[podIndex]
+		nodeGpus[node] = t.findGpusInNode(node, gpuNumber, priority, nodeGpus[node])
```
yqwang-ms: >findGpusInNode [](start = 21, length = 14)

Why decouple findNodesForPods and findGpusInNode? They may leverage each other in future.

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/api/types.go
zhypku: Fixed. But why we store CellChain in each AffinityGroupMemberBindInfo? All members should share a common chain.

---
In reply to: [332493391](https://github.com/microsoft/pai/pull/3684#discussion_r332493391) [](ancestors = 332493391)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/types.go
zhypku: You're right. But this schedulingRequest is an input to a scheduling algorithm. It seems that a scheduling algorithm does not have to be aware of the concept of affinity group. If you think it's better to rename, I'm ok :)

---
In reply to: [332507384](https://github.com/microsoft/pai/pull/3684#discussion_r332507384) [](ancestors = 332507384)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/types.go
zhypku: By reference. The type Cell is all pointers. I've tested. 

---
In reply to: [332499806](https://github.com/microsoft/pai/pull/3684#discussion_r332499806) [](ancestors = 332499806)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/types.go
yqwang-ms: Let's rename it. podGpuNumbers has less meaning than affinityGroup

---
In reply to: [334217302](https://github.com/microsoft/pai/pull/3684#discussion_r334217302) [](ancestors = 334217302,332507384)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/api/types.go
yqwang-ms: OK, my mistake, let's keep CellChain         

---
In reply to: [334216816](https://github.com/microsoft/pai/pull/3684#discussion_r334216816) [](ancestors = 334216816,332493391)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/types.go
yqwang-ms: This means that you need to be aware that cc and c does not make a copy to each other. Is it too strict?

---
In reply to: [334217427](https://github.com/microsoft/pai/pull/3684#discussion_r334217427) [](ancestors = 334217427,332499806)

zhypku: I mean, as long as cc and c point to the same object, they are considered equal. It doesn't matter if cc and c are copies or whatever.

---
In reply to: [334217590](https://github.com/microsoft/pai/pull/3684#discussion_r334217590) [](ancestors = 334217590,334217427,332499806)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
zhypku: sort.Ints([]int) only accepts []int, not []int32... Go does not have build-in sort for int32. :(

---
In reply to: [332855059](https://github.com/microsoft/pai/pull/3684#discussion_r332855059) [](ancestors = 332855059)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/intra_vc_scheduler.go
zhypku: The scheduler operates on a cluster view. If we use the same one for all chains, it needs to calculate the cluster view every time the scheduler is invoked

---
In reply to: [332510858](https://github.com/microsoft/pai/pull/3684#discussion_r332510858) [](ancestors = 332510858)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/types.go
yqwang-ms: I mean the object is copied to be another different object. Pls comment the method that it is removed by reference.

---
In reply to: [334218030](https://github.com/microsoft/pai/pull/3684#discussion_r334218030) [](ancestors = 334218030,334217590,334217427,332499806)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
zhypku: I've made the class stateless (it only stores a cluster view)

---
In reply to: [332858662](https://github.com/microsoft/pai/pull/3684#discussion_r332858662) [](ancestors = 332858662)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
zhypku: It will be equivalent to usedGpuNumAtPriority[p] = delta (empty value considered as 0)

---
In reply to: [332820927](https://github.com/microsoft/pai/pull/3684#discussion_r332820927) [](ancestors = 332820927)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/api/types.go
zhypku: Renamed. But the GPU indices is for physical, cell indices is for virtual. I'm afraid your simplification would lead to confusion

---
In reply to: [332496961](https://github.com/microsoft/pai/pull/3684#discussion_r332496961) [](ancestors = 332496961)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
zhypku: I wrote it wrong... They are in the wrong order. I've corrected and refined the comment.

---
In reply to: [332856608](https://github.com/microsoft/pai/pull/3684#discussion_r332856608) [](ancestors = 332856608)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
zhypku: It is a single number because it is for a specific priority (input of GetFreeGpuNumAtPriority()).
freeGpuNum is the number of all free GPUs. freeGpuNumForPriority considers GPUs used by lower priorities also free.

---
In reply to: [332808421](https://github.com/microsoft/pai/pull/3684#discussion_r332808421) [](ancestors = 332808421)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
zhypku: yes. renamed.

---
In reply to: [332501702](https://github.com/microsoft/pai/pull/3684#discussion_r332501702) [](ancestors = 332501702)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
```diff
@@ -0,0 +1,308 @@
...
+package algorithm
+
+import (
+	"fmt"
+	"sort"
+)
+
+// topologyAwareScheduler can schedule a set of pods on a cluster view.
+// It tries to place pods to nodes with fewer free GPUs (i.e., packing), while trying to avoid preemptions.
+// Inside each node, it tries to allocate GPUs with better affinity.
```
yqwang-ms: Because packing is not aware affinity, so comment:
... first tries to place pods... then inside each node

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/types.go
```diff
@@ -76,17 +117,26 @@ func (ccl ChainCellList) String() string {
...
+	klog.Infof("Cell removed from cell list: %v", c.GetName())
+}
+
 type AlgoAffinityGroup struct {
-	cell               *PhysicalCell
-	unallocatedPodNums map[int32]int32 // GpuNum -> PodNum
+	cell                 *PhysicalCell
+	unallocatedPodNums   map[int32]int32      // GpuNum -> PodNum
+	physicalGpuPlacement map[int32][]CellList // GpuNum -> a list of pods -> a list of physical GPUs of each pod
+	vcGpuPlacement       map[int32][]CellList // GpuNum -> a list of pods -> a list of virtual GPUs of each pod
```
yqwang-ms: >vcGpuPlacement [](start = 1, length = 14)

virtualGpuPlacement?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
```diff
@@ -0,0 +1,308 @@
...
+			copy(gpus[index-offset:nextIndex-offset-1], gpus[index+1:nextIndex])
+		} else {
+			copy(gpus[index-offset:], gpus[index+1:])
+		}
+	}
+	return gpus[:len(gpus)-len(indices)]
+}
+
+func sortGpuNumbers(n []int32) {
+	tmp := make([]int, len(n))
```
yqwang-ms: Move to common?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
```diff
@@ -40,14 +41,34 @@ type Cell interface {
...
+	chain              CellChain
+	level              CellLevel
+	priority           CellPriority
+	parent             Cell     // pointer to its parent cell
+	children           CellList // pointer to its children cells
+	atOrHigherThanNode bool     // true if the cell is at or higher than node level
+
+	totalGpuNum          int32                  // total GPU number of a cell
+	freeGpuNum           int32                  // free GPU number of a cell
+	freeGpuNumAtPriority int32                  // free GPU number at a certain priority (lower priority considered as free)
```
yqwang-ms: >certain priority [](start = 69, length = 16)

Is it the "priority as that of the pod to be scheduled"? Maybe we should have a better name.

zhypku: yes, it is. I refined the comment.

---
In reply to: [334401987](https://github.com/microsoft/pai/pull/3684#discussion_r334401987) [](ancestors = 334401987)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -529,242 +600,210 @@ func (h *HivedAlgorithm) findVirtualCell(
...
-	for l := CellLevel(1); l <= top; l++ {
-		for _, c := range fullCellList[l] {
-			cc := c.(*PhysicalCell)
-			if !cc.HasPod() {
-				if p := cc.GetParent(); p == nil || l == top || p.(*PhysicalCell).HasPod() {
-					cellView[l] = append(cellView[l], cc)
-				}
+		} else {
+			return internal.PodScheduleResult{
+				PodPreemptInfo: &internal.PodPreemptInfo{},
```
yqwang-ms: >PodPreemptInfo: &internal.PodPreemptInfo{}, [](start = 4, length = 43)

After you returns victims for the specific node, please replace my previous comment:
 // Including all victim Pods in the victim cell, although the default scheduler
 // will only preempt victims on one node.

to be

 // Only need to include the victim Pods of current preemptor Pod.

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -529,242 +600,231 @@ func (h *HivedAlgorithm) findVirtualCell(
...
-			} else {
-				break
+		} else {
+			var victimPods []*core.Pod
+			for v := range preemptionVictims.Items() {
+				victimPods = append(victimPods, v.(*core.Pod))
+			}
+			return internal.PodScheduleResult{
+				PodPreemptInfo: &internal.PodPreemptInfo{VictimPods: victimPods},
 			}
```
yqwang-ms: Have you tested it?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/cell.go
yqwang-ms: Do not put temp var as fields 

---
In reply to: [332820470](https://github.com/microsoft/pai/pull/3684#discussion_r332820470) [](ancestors = 332820470)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/intra_vc_scheduler.go
yqwang-ms: Will consider in future

---
In reply to: [332453251](https://github.com/microsoft/pai/pull/3684#discussion_r332453251) [](ancestors = 332453251)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/intra_vc_scheduler.go
yqwang-ms: A group may need to relax to across multiple chains.
Will consider in future.

---
In reply to: [334218248](https://github.com/microsoft/pai/pull/3684#discussion_r334218248) [](ancestors = 334218248,332510858)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
yqwang-ms: OT jobs may cannot ensure this.
Will consider in future.

---
In reply to: [332868715](https://github.com/microsoft/pai/pull/3684#discussion_r332868715) [](ancestors = 332868715)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
yqwang-ms: Aware topology when findNodesForPods for packing.
Will consider in future.

---
In reply to: [332870613](https://github.com/microsoft/pai/pull/3684#discussion_r332870613) [](ancestors = 332870613)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
yqwang-ms: Wait until all victims in the group cleared

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/scheduler/scheduler.go
```diff
@@ -520,7 +520,22 @@ func (s *HivedScheduler) filterRoutine(args ei.ExtenderArgs) *ei.ExtenderFilterR
 			PodState:          internal.PodPreempting,
 			PodScheduleResult: &result,
 		}
-		return &ei.ExtenderFilterResult{}
+
+		// Return FailedNodes to tell K8S Default Scheduler that preemption may help.
+		failedNodeReasons := map[string]string{}
+		for _, victim := range result.PodPreemptInfo.VictimPods {
```
yqwang-ms: pls help to double check this

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
yqwang-ms: And tell to preempt only if the whole group has sufficient allocated lower priority pods.

---
In reply to: [335258080](https://github.com/microsoft/pai/pull/3684#discussion_r335258080) [](ancestors = 335258080)

yqwang-ms: The principle is that we are scheduling the whole group (as a gang) instead of the individual pod inside it.

---
In reply to: [335270249](https://github.com/microsoft/pai/pull/3684#discussion_r335270249) [](ancestors = 335270249,335258080)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -529,242 +600,231 @@ func (h *HivedAlgorithm) findVirtualCell(
...
+			panic(fmt.Sprintf("[%v]: node %v picked by algorithm but not in K8S candidates",
+				internal.Key(pod), selectedNode))
+		}
+		// collect preemption victims
+		preemptionVictims := common.NewSet()
+		for _, gpu := range groupPhysicalPlacement[currentGpuNum][currentPodIndex] {
+			pGpu := gpu.(*PhysicalCell)
+			if pGpu.HasPod() {
+				preemptionVictims.Add(pGpu.GetPods()[0])
 			}
```
yqwang-ms: Return all victim Pods inside the victim affinity group to do gang preemption. So that different allocated affinity groups have not placement overlap/conflict.

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
```diff
@@ -0,0 +1,298 @@
...
+		clusterView:       clusterView,
+		crossPriorityPack: crossPriorityPack}
+}
+
+func (t *topologyAwareScheduler) Schedule(
+	podGpuNumbers map[int32]int32,
+	p CellPriority) map[int32][]CellList {
+
+	// GPU numbers of the pods to schedule
+	var sortedPodGpuNumbers []int32
```
yqwang-ms: >var sortedPodGpuNumbers []int32 [](start = 1, length = 31)

Any difference with previous:
sortedPodGpuNumbers := []int32{}?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -88,68 +97,34 @@ func (h *HivedAlgorithm) Schedule(pod *core.Pod, suggestedNodes []string) intern
...
-					},
-					VirtualCellLevel: vcl,
-					VirtualCellIndex: vci,
-				},
-			}
+	if group := h.allocatedAffinityGroups[s.AffinityGroup.Name]; group == nil {
+		klog.Infof("Scheduling new affinity group %v", s.AffinityGroup.Name)
+		groupPhysicalPlacement, groupVirtualPlacement = h.scheduleNewAffinityGroup(pod, s)
+		if groupPhysicalPlacement != nil {
 		}
```
yqwang-ms: Empty if?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -159,58 +134,58 @@ func (h *HivedAlgorithm) AddAllocatedPod(pod *core.Pod) {
...
-				str = fmt.Sprintf("reservation %v", s.ReservationId)
-			} else {
-				str = fmt.Sprintf("chain %v", chain)
+		for _, gms := range info.AffinityGroupBindInfo {
+			gpuNumber := int32(len(gms.PodPlacements[0].PhysicalGpuIndices))
+			newGroup.physicalGpuPlacement[gpuNumber] = make([]CellList, len(gms.PodPlacements))
+			if hasVirtualPlacement {
+				newGroup.virtualGpuPlacement[gpuNumber] = make([]CellList, len(gms.PodPlacements))
+			}
+			for i := int32(0); i < int32(len(gms.PodPlacements)); i++ {
```
yqwang-ms: >i [](start = 7, length = 1)

podIndex

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -159,58 +134,58 @@ func (h *HivedAlgorithm) AddAllocatedPod(pod *core.Pod) {
...
+				newGroup.physicalGpuPlacement[gpuNumber][i] = make(
+					CellList, len(gms.PodPlacements[i].PhysicalGpuIndices))
+				if hasVirtualPlacement {
+					newGroup.virtualGpuPlacement[gpuNumber][i] = make(
+						CellList, len(gms.PodPlacements[i].PhysicalGpuIndices))
+				}
+				node := gms.PodPlacements[i].PhysicalNode
+				for j := int32(0); j < int32(len(gms.PodPlacements[i].PhysicalGpuIndices)); j++ {
```
yqwang-ms: >j [](start = 8, length = 1)

gpuIndex

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -219,29 +194,47 @@ func (h *HivedAlgorithm) DeleteAllocatedPod(pod *core.Pod) {
...
 			"[%v]: group %v not exists when deleting pod", internal.Key(pod), s.AffinityGroup.Name))
-	} else if !group.cell.HasPod() {
-		delete(h.allocatedAffinityGroups, s.AffinityGroup.Name)
-		group.cell.ClearPhysicalCellList()
-		confirmReleasedCell(group.cell)
+	} else {
+		allocatedPods := group.allocatedPods[s.GpuNumber]
+		idx := -1
+		for i, p := range allocatedPods {
+			if pod == p {
```
yqwang-ms: >if pod == p { [](start = 3, length = 13)

Pod should be identified by its UID instead of reference

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
```diff
@@ -0,0 +1,298 @@
...
+// findLCA finds the lowest common ancestor of two cells (nil if they have no LCA).
+func findLCA(lower Cell, higher Cell) Cell {
+	for lower.GetLevel() < higher.GetLevel() {
+		if lower.GetParent() == nil {
+			return nil
+		}
+		lower = lower.GetParent()
+	}
+	if lower == higher {
+		return lower
```
yqwang-ms: Is it safe to compare the refer of cell?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -219,29 +194,47 @@ func (h *HivedAlgorithm) DeleteAllocatedPod(pod *core.Pod) {
...
+				allPodsDeleted = false
+				break
+			}
+		}
+		if allPodsDeleted {
+			for _, podPlacements := range group.physicalGpuPlacement {
+				for _, podPlacement := range podPlacements {
+					for _, gpu := range podPlacement {
+						pGpu := gpu.(*PhysicalCell)
+						pGpu.DeleteAffinityGroup(group)
```
yqwang-ms: >DeleteAffinityGroup [](start = 11, length = 19)

Why not move Add/DeleteAffinityGroup into confirmAdd/Release?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -529,242 +632,259 @@ func (h *HivedAlgorithm) findVirtualCell(
...
-	return minPriorityCell(cellView[level])
-}
-
-func getGroupCellSize(a *api.AffinityGroup) int32 {
-	n := int32(0)
-	for _, m := range a.Members {
-		n += common.NextPowerOf2(m.GpuNumber) * m.PodNumber
+	for _, c := range freeList[level] {
+		fmt.Println(c.GetName())
 	}
```
yqwang-ms: Just for testing?
We should use log instead of Println

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
zhypku: it's a mistake. I forgot to delete it :)

---
In reply to: [335884633](https://github.com/microsoft/pai/pull/3684#discussion_r335884633) [](ancestors = 335884633)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -291,194 +294,295 @@ func (h *HivedAlgorithm) initReservations() {
...
+	h.validateSchedulingRequest(sr, pod)
+	if sr.reservationId != "" {
+		klog.Infof("Use reservation %v", s.ReservationId)
+		sr.chain = h.reservedCells[sr.vc][sr.reservationId].GetChain()
+		physicalPlacement, virtualPlacement = h.processSchedulingRequest(sr)
+	} else {
+		physicalPlacement, virtualPlacement = h.scheduleAffinityGroupForGpuType(sr, s.GpuType, pod)
+	}
+	if physicalPlacement != nil {
+		klog.Infof("Succeeded scheduling group %v", s.AffinityGroup.Name)
```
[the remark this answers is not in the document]
zhypku: Seems that it should be "succeed in doing something" ...

---
In reply to: [335888642](https://github.com/microsoft/pai/pull/3684#discussion_r335888642) [](ancestors = 335888642)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
zhypku: I think it is. We only create cell objects during initialization.
Another way is to create a unique identifier for each cell

---
In reply to: [335902545](https://github.com/microsoft/pai/pull/3684#discussion_r335902545) [](ancestors = 335902545)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
zhypku: sorry. forgot to delete it

---
In reply to: [335910134](https://github.com/microsoft/pai/pull/3684#discussion_r335910134) [](ancestors = 335910134)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
zhypku: Changed to comparing cellName. 

---
In reply to: [335944279](https://github.com/microsoft/pai/pull/3684#discussion_r335944279) [](ancestors = 335944279,335902545)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -291,194 +290,302 @@ func (h *HivedAlgorithm) initReservations() {
...
+				if buddy.(*PhysicalCell).GetVirtualCell() != nil {
+					allBuddyFree = false
+					break
+				}
+			}
+			if !allBuddyFree {
+				terminate = true
+			} else {
+				for _, buddy := range parent.GetChildren() {
+					if buddy != c {
```
yqwang-ms: >if buddy != c { [](start = 5, length = 15)

Compare name?

yqwang-ms: Maybe you can give a Equal method of cell, so you will not forget.

---
In reply to: [335963626](https://github.com/microsoft/pai/pull/3684#discussion_r335963626) [](ancestors = 335963626)

yqwang-ms: Seems cellA == CellB cannot be replaced by cellA.Name == CellB.Name. For example, CellB may be null. How will you handle this? Make a CellEqual(c1, c2)?

---
In reply to: [335964282](https://github.com/microsoft/pai/pull/3684#discussion_r335964282) [](ancestors = 335964282,335963626)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/topology_aware_scheduler.go
```diff
@@ -0,0 +1,375 @@
...
+			}
+		}
+		if priority >= p {
+			n.freeGpuNumAtPriority -= num
+		}
+	}
+}
+
+type clusterView []*node
+
```
yqwang-ms: Move the type to top

PR #3687, opened 2019-09-30
[Rest Server] Exclude unready node resources in virtual cluster

abuccts:
Exclude unready node resources in virtual cluster.

### DIFF
--- a/src/rest-server/src/models/v2/virtual-cluster/k8s.js
+++ b/src/rest-server/src/models/v2/virtual-cluster/k8s.js
@@ -32,6 +32,26 @@ const vcData = {
 };
 
 
+const fetchNodes = async () => {
+  const nodes = await axios({
+    method: 'get',
+    url: `${kubernetesConfig.apiserver.uri}/api/v1/nodes`,
+  });
+  return nodes.data.items.filter((node) => {
+    if (node.metadata.labels['pai-worker'] !== 'true') {
+      return false;
+    }
+    // check node readiness
+    for (let i = node.status.conditions.length - 1; i >= 0; i --) {
+      const condition = node.status.conditions[i];
+      if (condition.type === 'Ready' && condition.status === 'True') {
+        return true;
+      }
+    }
+    return false;
+  });
+};
+
 const getResourceUnits = () => {
   return vcData.resourceUnits;
 };
@@ -92,10 +112,7 @@ const getNodeResource = async () => {
       };
     }
   } else {
-    const nodes = (await axios({
-      method: 'get',
-      url: `${kubernetesConfig.apiserver.uri}/api/v1/nodes`,
-    })).data.items.filter((node) => node.metadata.labels['pai-worker'] === 'true');
+    const nodes = await fetchNodes();
     for (let node of nodes) {
       const nodeName = node.metadata.name;
       const gpuNumber = k8s.atoi(node.status.capacity['nvidia.com/gpu']);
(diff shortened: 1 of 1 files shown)


src/rest-server/src/models/v2/virtual-cluster/k8s.js
```diff
@@ -32,6 +32,26 @@ const vcData = {
 };
 
 
+const fetchNodes = async () => {
+  const nodes = await axios({
+    method: 'get',
+    url: `${kubernetesConfig.apiserver.uri}/api/v1/nodes`,
+  });
+  return nodes.data.items.filter((node) => {
+    if (node.metadata.labels['pai-worker'] !== 'true') {
```
fanyangCS: filter with "pai-worker" might be unreliable. what if the admin removes the label? right now FC can still use a node without pai-worker.

abuccts: `pai-worker` label is used to remove master nodes and machines not maintained by pai,
since the quota is shown on pai's webportal

for readiness check, the following field is used:
```jsonp
{
  "type": "Ready",
  "status": "True",
  "lastHeartbeatTime": "2019-09-30T07:26:19Z",
  "lastTransitionTime": "2019-09-30T07:12:51Z",
  "reason": "KubeletReady",
  "message": "kubelet is posting ready status"
}
```
when the node is preempted or kubelet is killed, it will become
```jsonp
{
  "type": "Ready",
  "status": "Unknown",
  ...
}
```

abuccts: sounds like we need to add a node selector when submitting jobs from pai?

currently, jobs may be scheduled to a master node, which is not by design
it is scheduled by default scheduler or hived scheduler, instead of framework controller

fanyangCS: In this PR, maybe we can just fix the "unknown" node bug. i.e., when you calculate the available nodes, please excludes those unhealthy nodes.

abuccts: yes, the pai-worker label filter is added at the beginning, just added a check for not ready nodes and moved it to a function

src/rest-server/src/models/v2/virtual-cluster/k8s.js
```diff
@@ -92,10 +112,7 @@ const getNodeResource = async () => {
       };
     }
   } else {
-    const nodes = (await axios({
-      method: 'get',
-      url: `${kubernetesConfig.apiserver.uri}/api/v1/nodes`,
-    })).data.items.filter((node) => node.metadata.labels['pai-worker'] === 'true');
```
abuccts: we excluded master nodes previously

src/rest-server/src/models/v2/virtual-cluster/k8s.js
```diff
@@ -32,6 +32,26 @@ const vcData = {
...
+  return nodes.data.items.filter((node) => {
+    if (node.metadata.labels['pai-worker'] !== 'true') {
+      return false;
+    }
+    // check node readiness
+    for (let i = node.status.conditions.length - 1; i >= 0; i --) {
+      const condition = node.status.conditions[i];
+      if (condition.type === 'Ready' && condition.status === 'True') {
+        return true;
+      }
```
Binyang2014: Here we only count the node which condition is ready. How about the node in other condition, such as `MemoryPressure`, `OutOfDisk`. Refer to https://kubernetes.io/docs/concepts/architecture/nodes/#condition
Can we just filter out the nodes which 
```js
condition.type === 'Ready' && (condition.status === 'False' || condition.status === 'Unknown')
```

PR #3695, opened 2019-10-09
clean ${PAI_WORK_DIR} before mv content to this folder

Binyang2014:
In prod env, we saw when node down and up. The pod running on this node
may be restarted by kubelet. Even though the `restartPolicy` set to
`Never`. If that happens, the init container script will failed when
executing `mv` command. To prevent this, we clean the `${PAI_WORK_DIR}`
before mv content to this folder.

### DIFF
--- a/src/kube-runtime/src/init
+++ b/src/kube-runtime/src/init
@@ -89,6 +89,7 @@ PAI_LOG_DIR=${PAI_WORK_DIR}/logs/${FC_POD_UID}
 # Move previous logs to another folder. Notice: for init.log, some part will append to previous log file
 LOG_FILES=$(find $PAI_LOG_DIR -maxdepth 1 -type f)
 if [[ ! -z $LOG_FILES ]]; then
+  echo "[WARN] $PAI_LOG_DIR not empty, contains previous results"
   RANDOM=$$
   PRE_LOG_DIR=$PAI_LOG_DIR/prelog_$RANDOM
   while [[ -d PRE_LOG_DIR=$PAI_LOG_DIR/prelog_$RANDOM ]]
--- a/src/kube-runtime/src/init
+++ b/src/kube-runtime/src/init
@@ -20,6 +20,7 @@
 set -o errexit
 set -o nounset
 set -o pipefail
+set -x
 
 CHILD_PROCESS="UNKNOWN"
 
@@ -77,6 +78,10 @@ PAI_INIT_DIR=${PAI_WORK_DIR}/init.d
 PAI_RUNTIME_DIR=${PAI_WORK_DIR}/runtime.d
 
 PAI_LOG_DIR=${PAI_WORK_DIR}/logs/${FC_POD_UID}
+
+# Clean ${PAI_WORK_DIR} and ${PAI_LOG_DIR} since they may contain last execution content. (rarely happen, but seen in real world)
+rm -rf ${PAI_WORK_DIR}/*
+rm -rf ${PAI_LOG_DIR}/*
 
 # Move all runtime sources to PAI_WORK_DIR
 mv ./* ${PAI_WORK_DIR}
--- a/src/kube-runtime/src/init
+++ b/src/kube-runtime/src/init
@@ -79,9 +79,16 @@ PAI_RUNTIME_DIR=${PAI_WORK_DIR}/runtime.d
 
 PAI_LOG_DIR=${PAI_WORK_DIR}/logs/${FC_POD_UID}
 
-# Clean ${PAI_WORK_DIR} and ${PAI_LOG_DIR} since they may contain last execution content. (rarely happen, but seen in real world)
+# Move previous logs to another folder
+LOG_FILES=$(find $PAI_LOG_DIR -maxdepth 1 -type f)
+if [[ ! -z $LOG_FILES ]]; then
+  PRE_LOG_DIR=$PAI_LOG_DIR/prelog_$(date +%s)
+  mkdir $PRE_LOG_DIR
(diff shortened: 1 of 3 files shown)


yqwang-ms:
We should make our runtime **Idempotent**, since k8s restartPolicy=Never seems to be at least once execution like FC. So please confirm it and refine other places in runtime code.

src/kube-runtime/src/init
```diff
@@ -78,6 +79,10 @@ PAI_RUNTIME_DIR=${PAI_WORK_DIR}/runtime.d
 
 PAI_LOG_DIR=${PAI_WORK_DIR}/logs/${FC_POD_UID}
 
+# Clean ${PAI_WORK_DIR} and ${PAI_LOG_DIR} since they may contain last execution content. (rarely happen, but seen in real world)
+rm -rf ${PAI_WORK_DIR}/*
+rm -rf ${PAI_LOG_DIR}/*
```
abuccts: do you need to clean log dir? keep the previous log may be helpful

src/kube-runtime/src/init
```diff
@@ -78,6 +79,10 @@ PAI_RUNTIME_DIR=${PAI_WORK_DIR}/runtime.d
 
 PAI_LOG_DIR=${PAI_WORK_DIR}/logs/${FC_POD_UID}
```
yqwang-ms: We may need to keep all history and logs for tracking. So to achieve Idempotent, can you use a different folder each time?

Binyang2014: Will move old logs to ${PAI_LOG_DIR}/pre_{index}

---
In reply to: [332809405](https://github.com/microsoft/pai/pull/3695#discussion_r332809405) [](ancestors = 332809405)

src/kube-runtime/src/init
Binyang2014: Sure, will change the code

---
In reply to: [332808965](https://github.com/microsoft/pai/pull/3695#discussion_r332808965) [](ancestors = 332808965)

src/kube-runtime/src/init
```diff
@@ -53,14 +60,15 @@ function exit_handler()
...
-
-  echo "Unknown exit code, platform error"
-  exit 1
+  case $EXIT_CODE in
+    130|131|132|134|135|136|137|139|141|143)
+      exit $EXIT_CODE
+      ;;
+    *)
+      echo "Unknown exit code, platform error"
+      exit 248
```
yqwang-ms: Expose or log orignal exitcode

src/kube-runtime/src/init
```diff
@@ -78,6 +86,17 @@ PAI_RUNTIME_DIR=${PAI_WORK_DIR}/runtime.d
...
+# Move previous logs to another folder. Notice: for init.log, some part will append to previous log file
+LOG_FILES=$(find $PAI_LOG_DIR -maxdepth 1 -type f)
+if [[ ! -z $LOG_FILES ]]; then
+  PRE_LOG_DIR=$PAI_LOG_DIR/prelog_$(date +%s)
+  mkdir $PRE_LOG_DIR
+  mv $LOG_FILES $PRE_LOG_DIR
+fi
+
+# Clean ${PAI_WORK_DIR} since it may contain last execution content. (rarely happen, but seen in real world)
+rm -rf ${PAI_WORK_DIR}/*
```
yqwang-ms: use a different folder each time: Also for PAI_WORK_DIR

src/kube-runtime/src/init
Binyang2014: logged in line 41

Binyang2014: logged in line 41

src/kube-runtime/src/init
Binyang2014: Hmm, this folder is created by and mounted by k8s: https://github.com/microsoft/pai/blob/2a2a6e25cca3bc2b272e2faf88d000d6919b9a7b/src/rest-server/src/models/v2/job/k8s.js#L318
The folder name is fixed during pod creation. We can't assign arbitrary name for different container.

And this folder content will no be persisted and always be the PAI runtime source code. (Also this folder will be deleted when deleting pod). So I think it's not necessary to use the different folder each time

src/kube-runtime/src/init
```diff
@@ -78,6 +86,17 @@ PAI_RUNTIME_DIR=${PAI_WORK_DIR}/runtime.d
 
 PAI_LOG_DIR=${PAI_WORK_DIR}/logs/${FC_POD_UID}
 
+# Move previous logs to another folder. Notice: for init.log, some part will append to previous log file
+LOG_FILES=$(find $PAI_LOG_DIR -maxdepth 1 -type f)
+if [[ ! -z $LOG_FILES ]]; then
+  PRE_LOG_DIR=$PAI_LOG_DIR/prelog_$(date +%s)
```
yqwang-ms: >$(date +%s) [](start = 34, length = 11)

time in second may be the same. So why not just use the LOG_FILES count as suffix?

src/kube-runtime/src/init
```diff
@@ -78,6 +86,22 @@ PAI_RUNTIME_DIR=${PAI_WORK_DIR}/runtime.d
 
 PAI_LOG_DIR=${PAI_WORK_DIR}/logs/${FC_POD_UID}
 
+# Move previous logs to another folder. Notice: for init.log, some part will append to previous log file
+LOG_FILES=$(find $PAI_LOG_DIR -maxdepth 1 -type f)
+if [[ ! -z $LOG_FILES ]]; then
+  RANDOM=$$
```
yqwang-ms: Can we print warning log here?

PR #3698, opened 2019-10-09
Add CRUD for Storage, Set storage configs for group, Get user storage…

wangdian:
Add CRUD for Storage
Set storage configs for group
Get user storage config in user info


src/rest-server/src/utils/manager/storage/storage.js
```diff
@@ -0,0 +1,119 @@
...
+          dataStore: Joi.string().optional(),
+          containerName: Joi.string().required(),
+          accountName: Joi.string().required(),
+          key: Joi.string().required(),
+        }).required(),
+      })
+      .when('type', {
+        is: 'hdfs',
+        then: Joi.object({
+          namenode: Joi.string().required(),
```
mydmdm: could we also include information such as webhdfs uri (or port)?

wangdian: I will add a field "extension" for all kinds of storage servers

PR #3709, opened 2019-10-10
Support k8s services deployment in RBAC

abuccts:
Support framework controller and hived scheduler deployment in RBAC.

### DIFF
--- a/deployment/k8sPaiLibrary/monitorTool/update_resource.py
+++ b/deployment/k8sPaiLibrary/monitorTool/update_resource.py
@@ -1,0 +1,83 @@
+# Copyright (c) Microsoft Corporation
+# All rights reserved.
+#
+# MIT License
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
+# documentation files (the "Software"), to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and
+# to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+# The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING
+# BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+# DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+"""
+Delete k8s resources using AppsV1Api.
+"""
+
+from __future__ import print_function
+
+import time
+import argparse
+from kubernetes import client, config
+from kubernetes.client.rest import ApiException
+
+
+def get_api_resources(apps_v1_api):
+    api_resources = {
+        'daemonset': {
+            'list': apps_v1_api.list_namespaced_daemon_set,
+            'delete': apps_v1_api.delete_namespaced_daemon_set
+        },
+        'statefulset': {
+            'list': apps_v1_api.list_namespaced_stateful_set,
+            'delete': apps_v1_api.delete_namespaced_stateful_set,
+        },
+    }
+    return api_resources
+
+
+def delete_resource(apps_v1_api, api_resource, name, namespace='default'):
+    api_resources = get_api_resources(apps_v1_api)
+    body = client.V1DeleteOptions(
+        propagation_policy='Foreground',
+        grace_period_seconds=0)
+
+    if api_resource in api_resources:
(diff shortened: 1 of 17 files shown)


src/frameworkcontroller/deploy/frameworkcontroller.yaml.template
```diff
@@ -30,12 +30,15 @@ spec:
       labels:
         app: frameworkcontroller
     spec:
+      serviceAccountName: frameworkcontroller-account
```
yqwang-ms: We need to make the same change for frameworkbarrier:

https://github.com/microsoft/frameworkcontroller/blob/77ec4abbdc194f8964d68cfd5b9780a7a87b06e6/example/framework/extension/frameworkbarrier.yaml#L60-L83

yqwang-ms: And even if we do not support RBAC, we should remove the serviceAccountName in this line:
https://github.com/microsoft/pai/blob/f3cb0f73f36dd38c5edac873b29356593e7d433f/src/rest-server/src/models/v2/job/k8s.js#L298

It does not exist, so job will fail.
Why we have not met this issue, maybe because I have already created the account. And M60 bed does not need barrier.

src/frameworkcontroller/deploy/rbac.yaml
```diff
@@ -0,0 +1,67 @@
...
+  name: frameworkbarrier
+  namespace: default
+subjects:
+  - kind: ServiceAccount
+    name: frameworkbarrier
+    namespace: default
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: frameworkbarrier
```
yqwang-ms: Align with above naming, such as frameworkbarrier-account, frameworkbarrier-role...?

abuccts: updated

deployment/k8sPaiLibrary/monitorTool/update_resource.py
```diff
@@ -0,0 +1,83 @@
...
+    parser.add_argument('--operation', help='operation type', default='delete')
+    parser.add_argument('--resource', help='api resource type')
+    parser.add_argument('--name', help='resource name')
+    parser.add_argument('--namespace', help='resource namespace', default='default')
+    args = parser.parse_args()
+
+    config.load_kube_config()
+    apps_v1_api = client.AppsV1Api()
+    if (args.operation == 'delete'):
+        delete_resource(apps_v1_api, args.resource, args.name, args.namespace)
```
yqwang-ms: Better move to maintaintool?

deployment/k8sPaiLibrary/monitorTool/update_resource.py
```diff
@@ -0,0 +1,83 @@
...
+            except ApiException as e:
+                continue
+            time.sleep(5)
+    else:
+        raise Exception('Unsupported resource {}.'.format(api_resource))
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Update service resources.')
+    parser.add_argument('--operation', help='operation type', default='delete')
```
yqwang-ms: >, default='delete' [](start = 60, length = 18)

default should not be delete, too dangerous

src/frameworkcontroller/deploy/stop.sh.template
```diff
@@ -19,17 +19,31 @@
...
-APISERVER=$(kubectl config view | grep server | cut -f 2- -d ":" | tr -d " ")
-until ! kubectl get sts | grep -q "frameworkcontroller-sts"; do
-    echo 'Trying to stop framework controller ...'
-    curl -X DELETE $APISERVER/apis/apps/v1/namespaces/default/statefulsets/frameworkcontroller-sts \
-        -H "Content-Type: application/json" \
-        -d '{"kind":"DeleteOptions","apiVersion":"v1","propagationPolicy":"Foreground"}' > /dev/null 2>&1
-    sleep 5
-done
+PYTHONPATH="../../../deployment" python -m k8sPaiLibrary.monitorTool.update_resource \
+    --resource statefulset --name frameworkcontroller-sts
```
yqwang-ms: Also for delete.sh

src/frameworkcontroller/deploy/rbac.yaml
```diff
@@ -0,0 +1,67 @@
...
+  name: frameworkbarrier-role-binding
+  namespace: default
+subjects:
+  - kind: ServiceAccount
+    name: frameworkbarrier-account
+    namespace: default
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: frameworkbarrier-role
```
Binyang2014: How about move these lines to kube-runtime/deploy and change the service account name to kuberuntime-account.
Since the cert and token is mounted to kube-runtime container. And other commands in kube-runtime such as `curl $KUBE_APISERVER_ADDRESS/apis/frameworkcontroller.microsoft.com/v1/namespaces/default/frameworks/$FC_FRAMEWORK_NAME` need to access API server. Also need the token and cert.

src/frameworkcontroller/deploy/stop.sh.template
```diff
@@ -19,17 +19,31 @@
...
+    kubectl delete serviceaccount frameworkbarrier || exit $?
+fi
+
+if kubectl get clusterrole | grep -q "frameworkbarrier"; then
+    kubectl delete clusterrole frameworkbarrier || exit $?
+fi
+
+if kubectl get clusterrolebinding | grep -q "frameworkbarrier"; then
+    kubectl delete clusterrolebinding frameworkbarrier || exit $?
+fi
```
yqwang-ms: This may fail jobs, so let's move it to runtime delete sh?

PR #3716, opened 2019-10-11
Enable rbac for prometheus

Binyang2014:


xudifsd:
you do no need to declare `ca_file` in config of prometheus?

Binyang2014:
Already declared here: [https://github.com/microsoft/pai/blob/7953d59a4abd892e9fc878aa0e92a6a940988e83/src/prometheus/deploy/prometheus-configmap.yaml.template#L40-L42](https://github.com/microsoft/pai/blob/7953d59a4abd892e9fc878aa0e92a6a940988e83/src/prometheus/deploy/prometheus-configmap.yaml.template#L40-L42). :)

---
In reply to: [540995431](https://github.com/microsoft/pai/pull/3716#issuecomment-540995431) [](ancestors = 540995431)

src/prometheus/deploy/start.sh.template
```diff
@@ -25,6 +25,7 @@ kubectl create configmap prometheus-alert --from-file=alerting --dry-run -o yaml
 kubectl create configmap prometheus-record --from-file=recording --dry-run -o yaml | kubectl apply --overwrite=true -f - || exit $?
 kubectl apply --overwrite=true -f prometheus-configmap.yaml || exit $?
 kubectl apply --overwrite=true -f prometheus-deployment.yaml || exit $?
+kubectl apply --overwrite=true -f rbac.yaml || exit $?
```
abuccts: you need to create account and role binding before `prometheus-deployment.yaml`

Binyang2014: fixed

---
In reply to: [334220527](https://github.com/microsoft/pai/pull/3716#discussion_r334220527) [](ancestors = 334220527)

ydye:
Shoule be careful about that the doamin is bind with CA. So when you deployed your service in an RBAC enabled cluster, and you get the ca and token from k8s's mounting (default path), you should get api-servers' address through the env.
Please follow this source code. : https://github.com/microsoft/pai/blob/master/src/rest-server/src/config/kubernetes.js#L44

Binyang2014:
Thanks, fixed

---
In reply to: [541487859](https://github.com/microsoft/pai/pull/3716#issuecomment-541487859) [](ancestors = 541487859)

PR #3719, opened 2019-10-12
[Rest Server] Fix apiserver requests in RBAC

abuccts:
Fix apiserver requests in RBAC.

### DIFF
--- a/src/rest-server/src/models/v2/job/k8s.js
+++ b/src/rest-server/src/models/v2/job/k8s.js
@@ -17,12 +17,14 @@
 
 
 // module dependencies
+const {Agent} = require('https');
 const axios = require('axios');
 const yaml = require('js-yaml');
 const base32 = require('base32');
 const status = require('statuses');
 const runtimeEnv = require('./runtime-env');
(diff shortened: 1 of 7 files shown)


fanyangCS:
I assume to change will also work with non rbac setting?

abuccts:
> I assume to change will also work with non rbac setting?

yes.

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -149,6 +151,8 @@ const convertTaskDetail = async (taskStatus, ports, userName, jobName, taskRoleN
     const pod = (await axios({
       method: 'get',
       url: launcherConfig.podPath(taskStatus.attemptStatus.podName),
+      headers: launcherConfig.requestHeaders,
+      httpsAgent: apiserver.ca && new Agent({ca: apiserver.ca}),
```
ydye: Token?
https://github.com/microsoft/pai/blob/master/src/rest-server/src/utils/manager/user/crudK8sSecret.js#L63

PR #3732, opened 2019-10-14
Add InfiniBand device plugin in HCA mode for k8s

abuccts:
Add InfiniBand device plugin in HCA mode for k8s.
Reference: https://github.com/Mellanox/k8s-rdma-sriov-dev-plugin.


fanyangCS:
could you use a separate service for the deployment of IB? shouldn't be coupled with hived.

fanyangCS:
I suggest to rename device-plugin to more specific name like "InfiniBand Plugin".

abuccts:
> I suggest to rename device-plugin to more specific name like "InfiniBand Plugin". In the future we may deploy more plugin like nvidia gpu plugin, amd gpu plugin, and the fuse plugin

I planned to move nvidia device plugin to this folder.
There's a merge conflict with rbac delete, so separate to another PR.

PR #3736, opened 2019-10-16
Fix runtime bug: failed to finish init container due to rm failed

Binyang2014:


src/kube-runtime/src/init
```diff
@@ -86,8 +86,8 @@ PAI_RUNTIME_DIR=${PAI_WORK_DIR}/runtime.d
 
 PAI_LOG_DIR=${PAI_WORK_DIR}/logs/${FC_POD_UID}
 
-# Move previous logs to another folder. Notice: for init.log, some part will append to previous log file
-LOG_FILES=$(find $PAI_LOG_DIR -maxdepth 1 -type f)
+# Move previous logs to another folder. Notice: exclude for init.log, new log will append to previous log file
+LOG_FILES=$(find $PAI_LOG_DIR -maxdepth 1 -type f ! -name "init.log")
 if [[ ! -z $LOG_FILES ]]; then
   echo "[WARN] $PAI_LOG_DIR not empty, contains previous results"
   RANDOM=$$
```
abuccts: `$RAMDOM` is a built-in variable, you can use it as random number directly
```suggestion
  IDX=$RANDOM
```

Binyang2014: Fixed

src/kube-runtime/src/init
```diff
@@ -86,11 +86,10 @@ PAI_RUNTIME_DIR=${PAI_WORK_DIR}/runtime.d
...
-# Move previous logs to another folder. Notice: for init.log, some part will append to previous log file
-LOG_FILES=$(find $PAI_LOG_DIR -maxdepth 1 -type f)
+# Move previous logs to another folder. Notice: exclude for init.log, new log will append to previous log file
+LOG_FILES=$(find $PAI_LOG_DIR -maxdepth 1 -type f ! -name "init.log")
 if [[ ! -z $LOG_FILES ]]; then
   echo "[WARN] $PAI_LOG_DIR not empty, contains previous results"
-  RANDOM=$$
   PRE_LOG_DIR=$PAI_LOG_DIR/prelog_$RANDOM
   while [[ -d PRE_LOG_DIR=$PAI_LOG_DIR/prelog_$RANDOM ]]
   do
```
abuccts: these `$RANDOM` have different values, you need to fix the value at the beginning

Binyang2014: Thx, fixed it

PR #3737, opened 2019-10-16
Fix restserver can not access framework in RBAC env

Binyang2014:


Binyang2014:
dup with #3719 close it

PR #3741, opened 2019-10-16
support aks env

Binyang2014:
AKS doesn't export nvdia-smi to pod. We need to mount and set ENV manually.
Issue: https://github.com/Azure/AKS/issues/1271
This change will not break the on-premise env. Since in job-exporter, we append
the /var/drivers/nvidia/current to the `PATH` env. It should behind the `nvidia path`
set by nvida-plugin.

### DIFF
--- a/src/job-exporter/deploy/job-exporter.yaml.template
+++ b/src/job-exporter/deploy/job-exporter.yaml.template
@@ -93,9 +93,15 @@ spec:
         - name: device-mount
           hostPath:
             path: /dev
+        {%- if cluster_cfg['rest-server']['launcher-type'] == "k8s" %}
+        - name: driver-path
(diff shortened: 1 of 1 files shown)


fanyangCS:
so I guess the same trick applies to job container as well?

Binyang2014:
Sure. We can use same trick. We need to mount nvdia lib when creating framework in rest-server and export the PATH to job container in runtime. Then  `nvdia-smi` can be used in job container

---
In reply to: [542672868](https://github.com/microsoft/pai/pull/3741#issuecomment-542672868) [](ancestors = 542672868)

Binyang2014:
Close this PR, dup with https://github.com/microsoft/pai/pull/3749

PR #3742, opened 2019-10-17
[Rest Server] Add pod priority to support FIFO job scheduling

### LINKED ISSUE #3704: P1 - Job Near FIFO scheduling

**Why is this needed**:

**Potential Solutions**:
1. [**LongTerm**][Strict FIFO]: Support workflow/jobdependeny by user or PAI (such as Aether https://github.com/microsoft/pai/issues/3702) - **Pending**
2. [**Picked**][Near FIFO]: RestServer leverage [Pod priority](https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/) to nearly ensure Job FIFO scheduling in default scheduler - **Works: near FIFO, need to disable Priority Admission Controller and set per job priority decreasingly, [disable k8s preemption](https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#how-to-disable-preemption), a little hack and [small chances to starve later jobs](https://github.com/kubernetes/kubernetes/issues/71486#issuecomment-445567835)**
Such as PodPriority=-(MaxJobPriority-JobPriority, JobCreationTime)
     ```
     Given JobPriority range: [-1, 126]
     PodPriority int32bits =
     FixToNegative: uint1bits = (Fix to 1) | 
     JobPriority: uint8bits = (126 - JobPriority) | 
     JobCreationTime: uint23bits = (UnixEpoch % (2^23))
     ```
   This makes sure job Pod priority is by default lower than system services Pod priority, and k8s reserved priorities.
3. [**LongTerm**][Near FIFO]: Changes for default scheduler to do FIFO queuing according to the Pod creation time instead of the Pod last queued time by scheduler. - **Long term to be on AKS and [small chances to starve later jobs](https://github.com/kubernetes/kubernetes/issues/71486#issuecomment-445567835) and cannot tolerate job interleave such as task retry**
Tracked in https://github.com/kubernetes/kubernetes/issues/83834 and fixed by https://github.com/kubernetes/kubernetes/pull/83835.
4. [Infeasible][Weak FIFO]: Leverage existing default scheduler FIFO queue - **Not work: Very weak FIFO**
5. [Infeasible][Weak FIFO]: Leverage existing default scheduler Priority queue - **Not work: Very weak FIFO**

**Tests**:
Setup scheduler with No HA and No Restart.
- [x] Priority queue + Decreasing Pod Priority: **Works: near FIFO** (<5% cut in line)
- [x] Priority queue + Same Pod Priority: **Not work** (>40% cut in line)
- [x] Priority queue + Same Pod Priority + Using Pod creation time to sort: **Works: near FIFO** (<5% cut in line)
- [x] FIFO queue: **Not work** (>40% cut in line)
image
image
image
image

abuccts:
Add pod priority to support FIFO job scheduling.

Closes #3704.

### DIFF
--- a/src/rest-server/src/middlewares/v2/hived.js
+++ b/src/rest-server/src/middlewares/v2/hived.js
@@ -24,8 +24,8 @@ const vcConfig = require('@pai/config/vc');
 const convertPriority = (priorityClass) => {
   // TODO: make it a cluster-wise config
   const priorityMap = {
-    prod: 1000,
-    test: 100,
+    prod: 100,
+    test: 10,
   };
   return priorityClass in priorityMap ? priorityMap[priorityClass] : null;
 };
--- a/src/rest-server/src/middlewares/v2/hived.js
+++ b/src/rest-server/src/middlewares/v2/hived.js
@@ -23,7 +23,7 @@ const vcConfig = require('@pai/config/vc');
 
 const convertPriority = (priorityClass) => {
   // TODO: make it a cluster-wise config
-  // allowed range: [-1, 126]
+  // allowed range: [-1, 126], default priority 0
   const priorityMap = {
     prod: 100,
     test: 10,
--- a/src/rest-server/src/middlewares/v2/hived.js
+++ b/src/rest-server/src/middlewares/v2/hived.js
@@ -23,6 +23,7 @@ const vcConfig = require('@pai/config/vc');
 
 const convertPriority = (priorityClass) => {
   // TODO: make it a cluster-wise config
+  // allowed range: [-1, 126]
   const priorityMap = {
     prod: 100,
     test: 10,
--- a/src/rest-server/src/models/v2/job/k8s.js
+++ b/src/rest-server/src/models/v2/job/k8s.js
@@ -290,7 +290,7 @@ const generateTaskRole = (taskRole, labels, config) => {
     jobPriority = Math.min(Math.max(jobPriority, -1), 126);
   }
   const jobCreationTime = Math.floor(new Date() / 1000) & (Math.pow(2, 23) - 1);
-  const podPriority = - ((126 - jobPriority) << 23 + jobCreationTime);
+  const podPriority = - (((126 - jobPriority) << 23) + jobCreationTime);
 
   const frameworkTaskRole = {
     name: convertName(taskRole),
--- a/src/rest-server/src/models/v2/job/k8s.js
+++ b/src/rest-server/src/models/v2/job/k8s.js
@@ -284,12 +284,13 @@ const generateTaskRole = (taskRole, labels, config) => {
   }
   // calculate pod priority
   // reference: https://github.com/microsoft/pai/issues/3704
-  let jobPriority = 127;
+  let jobPriority = 0;
   if (launcherConfig.enabledHived) {
-    jobPriority = 126 - parseInt(config.taskRoles[taskRole].hivedPodSpec.priority);
+    jobPriority = parseInt(config.taskRoles[taskRole].hivedPodSpec.priority);
+    jobPriority = Math.min(Math.max(jobPriority, -1), 126);
   }
   const jobCreationTime = Math.floor(new Date() / 1000) & (Math.pow(2, 23) - 1);
-  const podPriority = - (jobPriority << 23 + jobCreationTime);
+  const podPriority = - ((126 - jobPriority) << 23 + jobCreationTime);
 
   const frameworkTaskRole = {
     name: convertName(taskRole),
--- a/src/rest-server/src/models/v2/job/k8s.js
+++ b/src/rest-server/src/models/v2/job/k8s.js
@@ -282,6 +282,15 @@ const generateTaskRole = (taskRole, labels, config) => {
     gangAllocation = 'false';
     retryPolicy.fancyRetryPolicy = true;
   }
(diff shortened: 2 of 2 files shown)


src/rest-server/src/middlewares/v2/hived.js
```diff
@@ -24,8 +24,8 @@ const vcConfig = require('@pai/config/vc');
 const convertPriority = (priorityClass) => {
   // TODO: make it a cluster-wise config
   const priorityMap = {
-    prod: 1000,
-    test: 100,
+    prod: 100,
```
yqwang-ms: Comment our allowed range here:
[-1, 126]

abuccts: updated

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -282,6 +282,15 @@ const generateTaskRole = (taskRole, labels, config) => {
     gangAllocation = 'false';
     retryPolicy.fancyRetryPolicy = true;
   }
+  // calculate pod priority
+  // reference: https://github.com/microsoft/pai/issues/3704
+  let jobPriority = 127;
+  if (launcherConfig.enabledHived) {
+    jobPriority = 126 - parseInt(config.taskRoles[taskRole].hivedPodSpec.priority);
```
yqwang-ms: >priority [](start = 73, length = 8)

What if user does not specify hivedPodSpec.priority?

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -282,6 +282,15 @@ const generateTaskRole = (taskRole, labels, config) => {
     gangAllocation = 'false';
     retryPolicy.fancyRetryPolicy = true;
   }
+  // calculate pod priority
+  // reference: https://github.com/microsoft/pai/issues/3704
+  let jobPriority = 127;
```
yqwang-ms: >jobPriority [](start = 6, length = 11)

Can we enforce hivedPodSpec.priority in validation [-1, 126]

src/rest-server/src/models/v2/job/k8s.js
abuccts: default `hivedPodSpec.priority` is `null`, so default `jobPriority = 126 - parseInt(null)` equals to `126`.

src/rest-server/src/models/v2/job/k8s.js
abuccts: it shouldn't be validated here, can be done in hived spec validation.
but at present the priority class uses hard coded values and there's no such validation for the whole hived spec yet.

src/rest-server/src/models/v2/job/k8s.js
fanyangCS: isn't 126 too high?

yqwang-ms: Can we explictly specify to default as 0? Max=126 Min=-1 And doc it.

src/rest-server/src/models/v2/job/k8s.js
yqwang-ms: Ok, so for safety can we round overflowed jobPriority to be 126, and round undherflowed jobPriority to be -1?

yqwang-ms: >127 [](start = 20, length = 3)

If no hived enabled, let's treat the same as user not specify, i.e. default to jobPriority = 126 - 0

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -282,6 +282,15 @@ const generateTaskRole = (taskRole, labels, config) => {
...
+  if (launcherConfig.enabledHived) {
+    jobPriority = 126 - parseInt(config.taskRoles[taskRole].hivedPodSpec.priority);
+  }
+  const jobCreationTime = Math.floor(new Date() / 1000) & (Math.pow(2, 23) - 1);
```
yqwang-ms: >new Date() [](start = 37, length = 10)

Is new Date() returns msec instead of sec?

abuccts: yes

src/rest-server/src/models/v2/job/k8s.js
abuccts: updated

src/rest-server/src/models/v2/job/k8s.js
abuccts: updated comment in hived priority class

src/rest-server/src/models/v2/job/k8s.js
abuccts: yes

src/rest-server/src/models/v2/job/k8s.js
yqwang-ms: I mean orignal jobPriority to within  [-1, 126], i.e. round hivedPodSpec.priority within it :)

yqwang-ms: I suggest that you can first preprocess the hivedPodSpec.priority
1. If hived not enable, jobPriority = 0
2. else jobPriority = round (hivedPodSpec.priority)
Then reversejobPriority = 126 - jobPriority
Is it more clear?

PR #3744, opened 2019-10-17
Move GPU device plugin to service

abuccts:
Move GPU device plugin to service.

### DIFF
--- a/deployment/clusterObjectModel/test/data/configuration-template-generate/services-configuration.yaml
+++ b/deployment/clusterObjectModel/test/data/configuration-template-generate/services-configuration.yaml
@@ -89,6 +89,9 @@ drivers:
   # 418.56
   version: "384.111"
 
+device-plugin:
+  devices: []
+
 prometheus:
   # port for prometheus port
   port: 9091
--- a/examples/cluster-configuration/services-configuration.yaml
+++ b/examples/cluster-configuration/services-configuration.yaml
@@ -15,7 +15,7 @@
 # DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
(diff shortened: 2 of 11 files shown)


fanyangCS:
if there is a way to bypass some Plugins during deployment?

abuccts:
> if there is a way to bypass some Plugins during deployment?

we can add a configuration section, but the device plugin pod won't effect if it is deployed to a node without certain device

yqwang-ms:
We need to define whether **stop** PAI services should have impacts on running jobs. 
(Currently stop FC will have no impacts on exist running, but delete FC will delete running jobs)
Seems current stop for device plugin will impact?

abuccts:
> We need to define whether **stop** PAI services should have impacts on running jobs.
> (Currently stop FC will have no impacts on exist running, but delete FC will delete running jobs)
> Seems current stop for device plugin will impact?

moved to `delete.sh`

yqwang-ms:
> We need to define whether **stop** PAI services should have impacts on running jobs.
> (Currently stop FC will have no impacts on exist running, but delete FC will delete running jobs)
> Seems current stop for device plugin will impact?

@fanyangCS what is the expected behaiour for stop PAI services:

fanyangCS:
> > We need to define whether **stop** PAI services should have impacts on running jobs.
> > (Currently stop FC will have no impacts on exist running, but delete FC will delete running jobs)
> > Seems current stop for device plugin will impact?
> 
> @fanyangCS what is the expected behaiour for stop PAI services:

I think that is service dependent. For the plugin, it is also plugin dependent.

yqwang-ms:
> > > We need to define whether **stop** PAI services should have impacts on running jobs.
> > > (Currently stop FC will have no impacts on exist running, but delete FC will delete running jobs)
> > > Seems current stop for device plugin will impact?
> > 
> > 
> > @fanyangCS what is the expected behaiour for stop PAI services:
> 
> I think that is service dependent. For the plugin, it is also plugin dependent.

So If any PAI service stop.sh that cause running job killed, when user stop the PAI, all jobs will be impacted. I think all PAI services should reach a consensus for this if we do not want stop PAI to impact user jobs.

PR #3761, opened 2019-10-21
add job pod monitor

Binyang2014:


src/watchdog/src/watchdog.py
```diff
@@ -237,44 +244,59 @@ def parse_pod_item(pod, pai_pod_gauge, pai_container_gauge, pods_info):
...
-                    error_counter.labels(type="unexpected_container_state").inc()
-                    logger.error("unexpected state %s in container %s",
-                            json.dumps(state), container_name)
-                else:
-                    container_state = list(state.keys())[0].lower()
+        ready = False
 
-            pai_container_gauge.add_metric([service_name, pod_name, container_name,
-                namespace, container_state, host_ip, str(ready).lower()], 1)
+        if container_status.get("ready") is not None:
```
yqwang-ms: use Pod.Phase = Pending to judge this

Binyang2014: Sure, we use `Pod.Status.Phase` to judge this. Please refer to [https://github.com/microsoft/pai/blob/5ecde18ae3febb3e66f9c40cfccd85a483eca585/src/prometheus/deploy/alerting/jobs.rules#L29](https://github.com/microsoft/pai/blob/5ecde18ae3febb3e66f9c40cfccd85a483eca585/src/prometheus/deploy/alerting/jobs.rules#L29) and [https://github.com/microsoft/pai/blob/5ecde18ae3febb3e66f9c40cfccd85a483eca585/src/watchdog/src/watchdog.py#L224-L251](https://github.com/microsoft/pai/blob/5ecde18ae3febb3e66f9c40cfccd85a483eca585/src/watchdog/src/watchdog.py#L224-L251). Watchdog just update pod status to promethus. Then we can define promethus rules for alerts

---
In reply to: [337321537](https://github.com/microsoft/pai/pull/3761#discussion_r337321537) [](ancestors = 337321537)

src/watchdog/src/watchdog.py
```diff
@@ -185,35 +190,37 @@ def __init__(self, name, gpu):
...
-    status = pod["status"]
+    # generate pai_containers
+    if service_name is not None and status.get("containerStatuses") is not None:
+        container_statuses = status["containerStatuses"]
+        generate_container_metrics(
+            pai_container_gauge, service_name, pod_name, container_statuses, namespace, host_ip)
 
+
+def generate_pod_metrics(pai_pod_gauge, pai_job_pod_gauge, service_name, job_name, pod_name,
+                         host_ip, status, namespace):
```
xudifsd: Maybe add `assert service_name is not None and job_name is not None` to make sure caller passed expected parameter

src/watchdog/src/watchdog.py
```diff
@@ -185,35 +190,37 @@ def __init__(self, name, gpu):
...
-            limit_gpu = int(walk_json_field_safe(container, "resources", "limits",
-                    "nvidia.com/gpu") or 0)
-            used_gpu += max(req_gpu, limit_gpu)
-    pods_info[host_ip].append(PodInfo(pod_name, used_gpu))
-
     labels = pod["metadata"].get("labels")
-    if labels is None or "app" not in labels:
-        logger.info("unknown pod %s", pod["metadata"]["name"])
+
+    service_name = labels['app'] if labels is not None and 'app' in labels else None
```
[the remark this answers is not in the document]
Binyang2014: Thanks, will fix it

---
In reply to: [337399478](https://github.com/microsoft/pai/pull/3761#discussion_r337399478) [](ancestors = 337399478)

src/watchdog/src/watchdog.py
Binyang2014: Here we parse all pods running in k8s. Some pods such as nvidia-device-plugin will not contain app label and job label.

---
In reply to: [337403770](https://github.com/microsoft/pai/pull/3761#discussion_r337403770) [](ancestors = 337403770)

src/prometheus/deploy/alerting/jobs.rules
```diff
@@ -25,3 +25,9 @@ groups:
         for: 1h # only when it exceed 1 hour
         annotations:
           summary: "zombie container in {{$labels.instance}} detected"
+
+      - alert: PaiJobPending
+        expr: pai_job_pod_count{pod_scheduled="true",phase="pending"} > 0
```
yqwang-ms: Use Pod.Spec.NodeName to judge pod_bound, and use it here

src/prometheus/deploy/alerting/jobs.rules
```diff
@@ -25,3 +25,9 @@ groups:
...
+        for: 30m
```
yqwang-ms: Polling all pods every 3min is enough

src/watchdog/src/watchdog.py
```diff
@@ -244,12 +245,17 @@ def generate_pod_metrics(pai_pod_gauge, pai_job_pod_gauge, service_name, job_nam
                 error_counter.labels(type="unknown_pod_cond").inc()
                 logger.warning("unexpected condition %s in pod %s", cond_t, pod_name)
 
+    # used to judge if sechduler has bound pod to a certain node
+    pod_bound = "false"
+    if node_name is not None:
```
yqwang-ms: And is not empty?

src/watchdog/src/watchdog.py
xudifsd: No, you returned in caller in this case. Sorry, should be `assert service_name is not None or job_name is not None`

xudifsd: This function only generate metrics to job and service, the caller must know this, assert statement is to let caller know it, so I suggest adding this so in future, code reader or writer can know this by only looking assert

PR #3767, opened 2019-10-24
[FC]: Log persist and rotate

yqwang-ms:


src/log-manager/deploy/log-manager.yaml.template
```diff
@@ -58,6 +60,8 @@ spec:
           mountPath: /usr/local/pai/logs
         - name: pai-log-backup
           mountPath: /usr/local/pai/logs-backup
+        - name: frameworkcontroller-log
+          mountPath: /var/log/frameworkcontroller.log
```
wangdian: Seems no volumes defined?

yqwang-ms:
logrotate is not compatiable with fluentd, using docker rotate

PR #3771, opened 2019-10-25
[Rest Server] Decompress large framework

abuccts:
Decompress possible fields for large framework.

Reference: https://github.com/microsoft/frameworkcontroller/pull/44

### DIFF
--- a/src/rest-server/src/models/v2/job/k8s.js
+++ b/src/rest-server/src/models/v2/job/k8s.js
@@ -18,6 +18,7 @@
 
 // module dependencies
 const {Agent} = require('https');
+const zlib = require('zlib');
 const axios = require('axios');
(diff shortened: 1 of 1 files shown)


src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -107,6 +116,11 @@ const convertState = (state, exitCode, retryDelaySec) => {
 };
 
 const convertFrameworkSummary = (framework) => {
+  // check fields which may be compressed
+  if (framework.status.attemptStatus.taskRoleStatuses == null) {
```
yqwang-ms: Seems no need taskRoleStatuses for FrameworkSummary and job list? So you do not need to decompress it?

PR #3774, opened 2019-10-25
[Rest Server] Application token api

sunqinzheng:
# Changes
## Token format
- Application will be a JSON Web Token with an additional boolean field `application`
  - User / frontend service can decode the jwt and get if the token is an application token
- Since we will query `k8s secret` when authentication (check if it is revoked), the `admin` field in jwt is removed. The permission info will be retrieved dynamically from `k8s secret`
## Storage
- The signed jwt list will be stored in the `pai-user-token` namespace of `k8s secret`
- Each user's token list will be a k8s secret item
  - name: username's hex string
  - data: uuid -> token
- The namespace will be created (if not exist) when rest server starts
## Authentication error
- `Token is expired` error message is removed. 
- Only a `token is invalid` error message will be shown now.
## k8s secret model
- Since previous k8s secret interface is implemented for specific use, a new k8s secret interface (model) for general use is implemented.
## API routes
- GET /api/v1/token
  - Get list of available tokens (portal token + application token)
- POST /api/v1/token/application
  - Generate an application token
- DELETE /api/v1/token/:token
  - Revoke a token (portal token / application token)

### DIFF
--- a/src/rest-server/src/controllers/v2/token.js
+++ b/src/rest-server/src/controllers/v2/token.js
@@ -16,43 +16,42 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 // module dependencies
-const jwt = require('jsonwebtoken');
-const tokenConfig = require('@pai/config/token');
-const createError = require('@pai/utils/error');
-const userModel = require('@pai/models/v2/user');
 const querystring = require('querystring');
-
-function jwtSignPromise(userInfo, admin, expiration = '7d') {
-  return new Promise((res, rej) => {
-    jwt.sign({
-      username: userInfo.username,
-      admin: admin,
-    }, tokenConfig.secret, {expiresIn: expiration}, (signError, token) => {
-      signError ? rej(signError) : res(token);
-    });
-  });
-}
+const userModel = require('@pai/models/v2/user');
+const groupModel = require('@pai/models/v2/group');
+const tokenModel = require('@pai/models/token');
+const createError = require('@pai/utils/error');
+const {encrypt} = require('@pai/utils/manager/user/user');
 
 /**
  * Get the token.
  */
 const get = async (req, res, next) => {
+  const username = req.body.username;
+  const password = req.body.password;
+  const hash = await encrypt(username, password);
+  let userItem;
   try {
-    const username = req.body.username;
-    const userInfo = await userModel.getUser(username);
-    const admin = await userModel.checkAdmin(username);
-    const token = await jwtSignPromise(userInfo, admin, tokenConfig.tokenExpireTime);
-    return res.status(200).json({
-      user: userInfo.username,
-      token: token,
-      admin: admin,
-      hasGitHubPAT: userInfo.extension.hasOwnProperty('githubPAT')&& Boolean(userInfo.extension['githubPAT']),
-    });
+    userItem = await userModel.getUser(username);
   } catch (error) {
     if (error.status && error.status === 404) {
-      return next(createError('Bad Request', 'NoUserError', `User ${req.params.username} is not found.`));
+      return next(createError('Bad Request', 'NoUserError', `User ${req.body.username} is not found.`));
     }
-    return next(createError.unknown(error));
+  }
+  if (hash !== userItem['password']) {
+    return next(createError('Bad Request', 'IncorrectPasswordError', 'Password is incorrect.'));
+  }
+  try {
+    const admin = await groupModel.getGroupsAdmin(userItem.grouplist);
+    const token = await tokenModel.create(username);
+    return res.status(200).json({
+      user: userItem.username,
+      token: token,
+      admin: admin,
+      hasGitHubPAT: userItem.extension.githubPAT,
+    });
+  } catch (err) {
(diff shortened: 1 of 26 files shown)


src/rest-server/src/controllers/v2/token.js
```diff
@@ -16,43 +16,42 @@
...
+  try {
+    const admin = await groupModel.getGroupsAdmin(userItem.grouplist);
+    const token = await tokenModel.create(username);
     return res.status(200).json({
-      user: userInfo.username,
+      user: userItem.username,
       token: token,
       admin: admin,
-      hasGitHubPAT: userInfo.extension.hasOwnProperty('githubPAT')&& Boolean(userInfo.extension['githubPAT']),
+      hasGitHubPAT: userItem.extension.githubPAT,
```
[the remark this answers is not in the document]
sunqinzheng: I don't know the current state of github pat feature.
I just keep the original logic and simplify the code.

src/rest-server/src/middlewares/token.js
```diff
@@ -15,45 +15,53 @@
...
+  if (/^Basic$/i.test(scheme)) {
+    // application token
+    const raw = Buffer.from(credentials, 'base64').toString();
+    const [username, password] = raw.split(':');
+    const token = await tokenModel.verify(password);
+    if (token.username === username && token.application) {
+      return token;
+    } else {
+      throw new Error('Invalid authorization header');
     }
```
Binyang2014: Not quiet understand for this line. In which situation, the HTTP authentication header will be "Basic credentials"

sunqinzheng: Other platforms ([github](https://developer.github.com/v3/auth/#basic-authentication), [vso](https://docs.microsoft.com/en-us/azure/devops/organizations/accounts/use-personal-access-tokens-to-authenticate?view=azure-devops#q-how-can-i-use-a-pat-in-my-code)) uses basic authentication for their pats (`BASIC username:pat_string`).
I provide a similar authentication option here if user is more used to them.

Binyang2014: Thanks, got it

---
In reply to: [339970596](https://github.com/microsoft/pai/pull/3774#discussion_r339970596) [](ancestors = 339970596)

src/rest-server/src/models/k8s-secret.js
```diff
@@ -0,0 +1,212 @@
...
+  if (apiserver.ca) {
+    config.httpsAgent = new Agent({ca: apiserver.ca});
+  }
+  if (apiserver.token) {
+    config.headers['Authorization'] = `Bearer ${apiserver.token}`;
+  }
+  return axios.create(config);
+};
+
+const serialize = (object) => {
```
ydye: Encode data to base64？

src/rest-server/src/models/k8s-secret.js
```diff
@@ -0,0 +1,212 @@
...
+
+  const result = {};
+  for (const [key, val] of Object.entries(object)) {
+    result[key] = Buffer.from(val).toString('base64');
+  }
+  return result;
+};
+
+
+const deserialize = (object) => {
```
ydye: decode data from base64

src/rest-server/src/models/k8s-secret.js
```diff
@@ -0,0 +1,212 @@
...
+  const builder = [];
+  for (const [key, val] of Object.entries(labelSelector)) {
+    builder.push(`${key}=${val}`);
+  }
+  return builder.join(',');
+};
+
+const createNamespace = async (namespace) => {
+  const url = new URL(`/api/v1/namespaces/`, apiserver.uri).href;
+  const config = {
```
ydye: Is it deplicate with initClient

src/rest-server/src/models/token.js
```diff
@@ -0,0 +1,140 @@
...
+      secret,
+      expiration != null ? {expiresIn: expiration} : {},
+      (signError, token) => {
+        signError ? rej(signError) : res(token);
+      }
+    );
+  });
+};
+
+const purge = (data) => {
```
ydye: Delete invalid token?

src/rest-server/src/models/k8s-secret.js
sunqinzheng: It's a namespace api with different base uri.
Maybe it should be in another file like `k8s-util.js` or `k8s-namespace.js`.
I think we can reorganize the k8s api file structure later.

src/rest-server/src/models/k8s-secret.js
sunqinzheng: It's an inner function and I think it won't be used by other components.
I prefer using a shorter function name.

sunqinzheng: I will add some comments on them to improve code readability

yiyione:
## Token
- [ ] Login and get the browser token from cookies
- [ ] Decode the jwt token and the data should not conclude admin field.
- [ ] Check if the token is stored in k8s secret's pai-user-token namespace as the structure described before.
- [ ] Revoke the browser token through rest api
- [ ] View job list through web portal, and "invalid token" alert message will be popped out.
## Application Token
- [ ] Tested in unit test "userToken.js"

Issue #3772, opened 2019-10-25
Normal user should see memory/CPU consumption of the VC


Binyang2014:
**What would you like to be added**:
Show mem/CPU consumption of the VC which he/she can use for normal user.
**Why is this needed**:
User should know the whole resource usage about the VC. Not just GPU usage
**Components that may involve changes**:
Webportal

**Possible UX design**
Just like admin page, but user can only see the VCs he/she can use.
![image](https://user-images.githubusercontent.com/9415966/67545402-85165b00-f72b-11e9-8891-4a50bcf5b522.png)

Binyang2014:
@scarlett2018 Any comments about this feature or UX design?

yangou1988:
@Binyang2014 Let's add the utilization and detail (Memory, CPU, GPU)  to the normal user's VC. and hide the button: view all and column: action.

Binyang2014:
@scarlett2018 @fanyangCS @yangou1988 PR: #3784 will add this feature. But previous home page will show if VC is dedicated for normal user, and display different colors: https://github.com/microsoft/pai/blob/2f060ca59da85672238ccb897f6b35943713b135/src/webportal/src/app/home/home/virtual-cluster-list.jsx#L60 After this change, such info will missing. Do we need to keep this info?

scarlett2018:
@Binyang2014 - any visual example? compare the 2 effects..

Binyang2014:
![image](https://user-images.githubusercontent.com/9415966/67847753-58ed4680-fb3e-11e9-9bcd-2ff06cd8cb55.png)
Just like this

PR #3784, opened 2019-10-30
Show CPU/Mem usage in home page for normal user

Binyang2014:
Please refer to issue #3772 . We want to normal user see CPU/Mem usage info


Binyang2014:
image
Current home page view for normal user

PR #3791, opened 2019-11-01
Change query storage configs/servers rest-apis to GET

wangdian:
Change query storage configs/servers rest-apis to GET

### DIFF
--- a/src/rest-server/src/controllers/v2/storage.js
+++ b/src/rest-server/src/controllers/v2/storage.js
@@ -18,6 +18,7 @@
 // module dependencies
 const createError = require('@pai/utils/error');
 const storageModel = require('@pai/models/v2/storage');
+const {isArray, isEmpty} = require('lodash');
 
 const getStorageServer = async (req, res, next) => {
   try {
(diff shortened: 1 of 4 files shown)


src/webportal/src/app/job-submission/utils/conn.js
```diff
@@ -115,38 +116,40 @@ export async function listUserStorageConfigs(user) {
...
 export async function fetchStorageServers(serverNames) {
-  const bodyData = {};
-  bodyData.names = serverNames;
+  if (isEmpty(serverNames)) {
+    return [];
+  }
+
   const storageServers = await fetchWrapper(
-    `${config.restServerUri}/api/v2/storage/servers`,
+    `${config.restServerUri}/api/v2/storage/server?${querystring.stringify({
```
debuggy: server or servers? If it returns an array, maybe servers is better?

src/rest-server/src/controllers/v2/storage.js
```diff
@@ -31,7 +32,11 @@ const getStorageServer = async (req, res, next) => {
 
 const getStorageServers = async (req, res, next) => {
   try {
-    const names = req.body.names;
+    const names = isEmpty(req.query.names)
```
debuggy: As the API doc says, the query names will always be an array. So if the query is a string instead of a one element array, it will be a bad request. 

Or the API doc should add an additional description that query accepts both array and string format.

wangdian: Http Get cannot guarantee the parameter is array. Adding multiple query example in DOC per discussion.

src/webportal/src/app/job-submission/utils/conn.js
wangdian: Using /storage/server and /storage/config similar to user and group

PR #3793, opened 2019-11-04
[Error Spec] fix typo

Binyang2014:


Binyang2014:
> Since 248 is already unknown, why it is known to be PAIRuntimeInitContainerUnknownError?
> Suggest to remove it and leverage 256 is enough.

Checked the mail-chain, this exitCode is added by this PR #3695. The reason to add this code is we want to make exit reason more specific, since we can know it's a init container bug.

Without this change, we just know it's a runtime bug, and need to check log to figure out. From PR #3695, we know sometimes pod may contains previous logs. It make us a little confuse about which part cause such error. Add this error type will make debugging much easier.

Anyway, this type of error just provide extra stage info: `initializing`. If we can get such info from somewhere, it's fine to remove it.

yqwang-ms:
It is easy to figure out which container's error. If we only have init container log, it must be init container error, otherwise, it must be app container error.

Binyang2014:
> It is easy to figure out which container's error. If we only have init container log, it must be init container error, otherwise, it must be app container error.

Currently, that's true. But I think report the error stage to user/dev is important too. (less confusing and easy to know why). Currently, the error spec format is:
```yaml
- code: 256
  phrase: PAIRuntimeExitAbnormally
  issuer: PAI_RUNTIME
  causer: PAI_RUNTIME
  type: PLATFORM_FAILURE
  stage: UNKNOWN
  behavior: UNKNOWN
  reaction: RETRY_TO_MAX
  reason: "PAI Runtime exit abnormally with undefined exitcode, it may have bugs"
  repro:
  - "PAI Runtime exits with exitcode 1"
  solution:
  - "Contact PAI Dev to fix PAI Runtime bugs"
```

How about changing the stage to a  dynamic type and let runtime report current stage, such as initializing, running..., (if it could report the stage)?
Just a proposal, will not do this in this PR

yqwang-ms:
Spec only contains **static** fields, soI think you can report "which container" in a new field inside the runtime **Dynamic** ExitInfo, such as caughtSite?

yqwang-ms:
I suggest we first remove the code, because PAIRuntimeInitContainerUnknownError is too general and more confusing than PAIRuntimeExitAbnormally

Binyang2014:
> I suggest we first remove the code, because PAIRuntimeInitContainerUnknownError is too general and more confusing than PAIRuntimeExitAbnormally

Fine updated

src/kube-runtime/src/init
```diff
@@ -66,7 +66,7 @@ function exit_handler()
       ;;
     *)
       echo "Unknown exit code, platform error"
-      exit 248
+      exit 1
```
yqwang-ms: Just exit the orignal exit code, it is already comparable

Binyang2014: We don't check if the exitCode is duplicated with some user predefined code here. Exit 1 is safety.

yqwang-ms: You already checked all predefined code by:
  pattern:
    runtimeContainerPatterns:
    - exitCode: XXX
So it is safe to return, otherwise, you may have inconsistent with our spec

yqwang-ms: It is same safe as if your bash process crashed/killed before exit 1. So exit 1 is unnecessary and overwrite real raw comparable exit code

PR #3794, opened 2019-11-05
force to flush after user command finished

Binyang2014:
Tested in vnext bed. Maybe many factors affect the IO cache behaviors.  But it works currently.


src/kube-runtime/src/runtime.d/runtime_entry.sh
```diff
@@ -96,6 +96,10 @@ echo "[INFO] USER COMMAND START"
 USER_PID=$!
 
 wait ${USER_PID}
+
+# flush cache to disk
+sync
```
abuccts: have you tested both stdout & stderr, w/ or w/o output?

Binyang2014: stdout & stderr tested.
with/without output tested.

src/kube-runtime/src/runtime.d/runtime_entry.sh
```diff
@@ -96,6 +96,10 @@ echo "[INFO] USER COMMAND START"
 USER_PID=$!
 
 wait ${USER_PID}
+
+# flush cache to disk
```
abuccts: `# synchronize cached writes to disk` ?

PR #3797, opened 2019-11-05
Fix tensorboard precommand bug

### LINKED ISSUE #3796: tensorboard could not add precommand after toggle it
Repro steps:

1. In 16 bed, toggle the tensorboard option

image

2. open yaml editor and the tensorboard precommand is missing

image

3. compare with correct yaml:

image

debuggy:
In original code: 

```javascript  
if (!jobData.containData) {
    return;
}
```
If job data.containData is false,  the tensorboard command will not be prefixed.

closes #3796

### DIFF
--- a/src/webportal/src/app/job-submission/utils/utils.js
+++ b/src/webportal/src/app/job-submission/utils/utils.js
@@ -142,7 +142,6 @@ export function getPortFromUrl(url) {
 }
 
 function addPreCommandsToProtocolTaskRoles(protocol, preCommands) {
-  addTensorBoardCommandsToProtocolTaskRoles(protocol);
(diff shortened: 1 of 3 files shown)


src/webportal/src/app/job-submission/utils/utils.js
```diff
@@ -183,6 +182,8 @@ function addTensorBoardCommandsToProtocolTaskRoles(protocol) {
 }
 
 export async function populateProtocolWithDataCli(user, protocol, jobData) {
```
Binyang2014: Change this function name? It not only inject data cli into the command. It also used to inject the tensorflow cli.

debuggy: done

PR #3804, opened 2019-11-06
[Web Portal] User profile page

sunqinzheng:
# User profile page
## Entries
- Top navbar - username's dropdown
- Sidebar 'My Profile' item
### Header
- Group info
- Email
- Edit profile / password dialog if basic authentication is used
### Token list
- Generate application token
- Revoke token
### Storage list
# Others
- Refine rest api unauthorized error handling


src/webportal/src/app/user/user-logout/user-logout.component.js
```diff
@@ -36,4 +36,23 @@ const userLogout = (origin = window.location.href) => {
...
+  if (!origin) {
+    window.location.replace('/index.html');
+  } else {
+    window.location.replace(
+      `/index.html?${querystring.stringify({ from: origin })}`,
+    );
+  }
+};
+
+module.exports = { userLogout, clearToken };
```
abuccts: revoke current browser token in `userLogout`? otherwise, every login/logout will create one new token

sunqinzheng: All exists tokens will be verified when rest server handles token operations (login, create, revoke) and the expired tokens will be cleaned. So don't worry about that there will be too many tokens stored in k8s secret.

Besides, the idea is great. As the release date is approaching, I prefer to merge this pull request first and I will implement the logic that revoke tokens when user logout / change password in another pull request.

yiyione:
## Entries
- [x] Should be able to open the user profile page through top navbar dropdown / sidebar button
## Token
- [x] Create an application token
- [ ] Should be able to view job list using the application token through rest api
- [ ] Revoke the application token
- [ ] Should not be able to view job list
- [ ] Revoke current browser token and user should be logged out
## User's information
- [ ] Header
  - Group info
  - Email
- [ ] Virtual clusters
  - Should display the same component as home page
- [ ] Storage list
## Edit user profile (basic authentication only)
- [ ] Should be hided when auth method is OIDC
- [ ] Edit profile (email)
- [ ] Edit password

PR #3811, opened 2019-11-08
[webportal] bug fix

Binyang2014:
Fix dedicated vc display


Binyang2014:
fixed in #3804 close this

PR #3812, opened 2019-11-08
[Rest Server] Update hived config validation

abuccts:
Update hived config validation.


src/rest-server/src/middlewares/v2/hived.js
```diff
@@ -72,7 +74,11 @@ const hivedValidate = (protocolObj) => {
         );
       }
 
-      const affinityGroupName = taskRoleConfig.affinityGroupName;
+      let affinityGroupName = taskRoleConfig.affinityGroupName;
+      // put the gang scheduling job into one affinity group by default
+      if (affinityGroupName == null && gangAllocation) {
+        affinityGroupName = 'default';
+      }
```
yqwang-ms: Should use job name as affinityGroupName

PR #3826, opened 2019-11-12
K8S managed NFS+SMB storage

wangdian:
Storage-manager
Create k8s managed NFS + SMB storage server.
Integrated with team-wise storage.


deployment/paiLibrary/paiService/service_management_refresh.py
```diff
@@ -51,7 +51,7 @@ def refresh_all_label(self):
         self.logger.info("Begin to refresh all the nodes' labels")
         machinelist = self.cluster_object_model['layout']['machine-list']
 
-        labels = ['pai-master', 'pai-worker', 'no-drivers', 'no-nodeexporter']
+        labels = ['pai-master', 'pai-worker', 'pai-storage', 'no-drivers', 'no-nodeexporter']
```
hzy46: Should we limit the number of "pai-storage" labelled machines <= 1? If users mislabelled multiple nodes as "pai-storage", there might be a data inconsistence after service restarts.

src/storage-manager/deploy/storage-manager.yaml.template
```diff
@@ -0,0 +1,95 @@
...
+        ports:
+        - containerPort: 445
+          hostPort: 445
+          name: sm-smbport
+        - containerPort: 2049
+          hostPort: 2049
+          name: sm-nfsport
+      volumes:
+        - name: pai-storage
+          hostPath:
```
hzy46: Since it is a hostPath, should we require that machine labelled as "pai-master" should not be labelled as "pai-storage"? Users' data and models are sometimes in a huge size, thus there is some risk if we deploy it to master.

src/storage-manager/deploy/scripts/entrypoint.sh
```diff
@@ -0,0 +1,76 @@
...
+  echo "join domain ----------"
+  net ads join -U "$DOMAINUSER"%"$DOMAINPWD"
+  echo "domain info ----------"
+  net ads info
+fi
+
+# create folders
+echo "create folders ----------"
+mkdir -p /share/pai/data
+mkdir -p /share/pai/users
```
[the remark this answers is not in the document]
wangdian: Ok, will remove this.

deployment/paiLibrary/paiService/service_management_refresh.py
wangdian: This service can start multiple storage server to support different groups.

src/storage-manager/deploy/storage-manager.yaml.template
wangdian: Maybe not, if user only have one node to test, he may also use this feature.
Admin should take response for this.

src/storage-manager/README.md
```diff
@@ -0,0 +1,122 @@
+## Storage Manager
```
Binyang2014: Can you add some description about the purpose of storage manager? I'm not quite understand when should I deploy this service.

src/storage-manager/deploy/storage-manager-service.yaml
```diff
@@ -0,0 +1,34 @@
...
+# The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING
+# BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+# DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+apiVersion: v1
+kind: Service
```
Binyang2014: Sorry, I don't get the whole context. Why deploy this as a service? And who will access the storage manager?

wangdian: This service yaml file should be removed.
This is for teamwise storage integration, but this feature is removed.

src/storage-manager/README.md
wangdian: Done

src/storage-manager/deploy/storage-manager-service.yaml
wangdian: Done

PR #3831, opened 2019-11-12
Add api endpoint and webportal page of job retry history

### LINKED ISSUE #3788: job retry history design
## Webportal

- Make job retry number as a link in job detail page.

image

- When click the link, redirect to a new page with a job retry grid.

- Link each item to a detail page of attempts.

### `Job retry history health check`

Get the health status of job retry history endpoint

*Request*

```json
GET /api/v2/jobs/:frameworkName/jobRetries/healthz
```

*Response if succeeded*

```json
Status: 200
ok
```

*Response job history endpoint does not work*

```json
Status: 404
Not found
```
### `List job retries by job UUID`

List all job retry histories by job ID

*Request*

```json
GET /api/v2/jobs/:frameworkName/jobRetries
```

*Response if succeeded*

```json
Status: 200
{
  [<jobRetry schema>]
}
```

*Response if the job id is not found*

```json
Status: 404
The job ID is not found
```

Response if a server error occurred

Status: 500

{
  "code": "UnknownError",
  "message": "*Upstream error messages*"
}

### `Get specified job retry by retry index`

Get specified job retry by retry index

*Request*

```json
GET /api/v2/jobs/:frameworkName/jobRetries/:jobRetryIndex
```

*Response if succeeded*

```json
Status: 200
{
  <jobRetry schema>
}
```

*Response if the job retry is not found*

```json
Status: 404
The job retry is not found
```

Response if a server error occurred

Status: 500

{
  "code": "UnknownError",
  "message": "*Upstream error messages*"
}

debuggy:
1. add [job attempt api](https://github.com/microsoft/pai/blob/mintao%2Fjob-retry/docs/rest-server/API.md#get-apiv2jobsframeworknamejobattemptshealthz)

   Attempts include job retry history from elastic search and current job info from k8s api. The  attempt schema is converted from framework.

2. add page job-retry.html to webportal

image

closes #3788


yqwang-ms:
And Retry number is tooo small and user does not aware click it can go to retry history. Why not also add a link like "view job config"
image

yqwang-ms:
0. It would be better that, we can count each failure types on the grid page, such as user failure: 7, unknown failure 2, OOM failures: 5, etc?

1. Do we really need the API "Job retry history health check"?
Can we just check this by using PAI feature gate: job-history=true?

2. Query should be done by frameworkUID instead of frameworkName, and, pls also adjust the url to be:
`GET /api/v2/jobs/:frameworkUID/jobAttempts`

debuggy:
> 
> 
>     1. It would be better that, we can count each failure types on the grid page, such as user failure: 7, unknown failure 2, OOM failures: 5, etc?

1. From my point of view, this might be a overkill design, and the page would be more crowded and sophisticated (even now the page already has too much elements). However it could be decided by designer @yangou1988 if there is an elegant design.


>     2. Do we really need the API "Job retry history health check"?
>        Can we just check this by using PAI feature gate: job-history=true?

2. what do you mean about ```feature gate```, and what is the potential issue if we add health endpoint?

>     3. Query should be done by frameworkUID instead of frameworkName, and, pls also adjust the url to be:
>        `GET /api/v2/jobs/:frameworkUID/jobAttempts`

3. Currently all webportal job apis use framworkname as path params. If I use framework UID in url path, I have to firstly change job api to get UID from job detail. Or the job api should also use UID as path params?

debuggy:
> 
> 
> And Retry number is tooo small and user does not aware click it can go to retry history. Why not also add a link like "view job config"
> image

It is currently under design and discussion. Maybe it would be added later.

docs/rest-server/API.md
```diff
@@ -2889,6 +2889,172 @@ Status: 500
...
+```
+
+*Response if job attempts API not work*
+
+```json
+Status: 501
+Not healthy
+```
+
+### `GET /api/v2/jobs/:frameworkName/jobAttempts`
```
abuccts: previous api endpoint uses kebab case instead of camel case, e.g. `/api/v1/virtual-clusters`,
better to use `job-attempts` to keep consistent?

yqwang-ms:
Will we leverage history server for job list page? If not, pls create an issue to track it.

docs/rest-server/API.md
```diff
@@ -2889,6 +2889,172 @@ Status: 500
 }
 ```
 
+### `GET /api/v2/jobs/:frameworkName/jobAttempts/healthz`
+
```
yqwang-ms: Why need this?

Can we just check this by using PAI feature gate: job-history=true? Like this
[https://github.com/microsoft/pai/blob/35489463cb5df9c2da85c139c88d37d7afeffdef/src/fluentd/deploy/start.sh.template#L20](https://github.com/microsoft/pai/blob/35489463cb5df9c2da85c139c88d37d7afeffdef/src/fluentd/deploy/start.sh.template#L20)

docs/rest-server/API.md
yqwang-ms: >frameworkName [](start = 23, length = 13)

> Currently all webportal job apis use framworkname as path params. If I use framework UID in url path, I have to firstly change job api to get UID from job detail. Or the job api should also use UID as path params?

As framework of the same name can be deleted (such as evicted from APIServer due to etcd limited space), then created again, so just using framework name cannot locate a specific framework. But using framework UID can always locate it.
We can discuss in meeting how to handle it.
I prefer we support both query by frameworkname and frameworkUID. If frameworkname matches multiple frameworks, we just return the latest one.

yqwang-ms:
>   const {

Will we leverage history server for job list page? If not, pls create an issue to track it.



---
Refers to: src/webportal/src/app/job/job-view/fabric/JobList/Table.jsx:41 in 7b9c002. [](commit_id = 7b9c002ee7736e77e4ed371902fb08b9045f8203, deletion_comment = False)

docs/rest-server/API.md
debuggy: You mean query k8s API Server to get UID? Could this ensure the uniqueness?

Actually, if framework name has replication issue, we might firstly use UID in all job related APIs.

debuggy:
> 
> 
> > const {
> 
> Will we leverage history server for job list page? If not, pls create an issue to track it.
> 
> Refers to: src/webportal/src/app/job/job-view/fabric/JobList/Table.jsx:41 in [7b9c002](https://github.com/microsoft/pai/commit/7b9c002ee7736e77e4ed371902fb08b9045f8203). [](commit_id = [7b9c002](https://github.com/microsoft/pai/commit/7b9c002ee7736e77e4ed371902fb08b9045f8203), deletion_comment = False)

could you explain more about this comment, I did not get it

docs/rest-server/API.md
debuggy: You mean add a new ENV flag about this? From my point of view, the health check PAI is more flexible than configs, because it might be a scenario that the flag is true however elastic search or other prerequisites did not work for some rare reasons. I do not want this scenario to affect the main system works.

docs/rest-server/API.md
debuggy: I will change it to ```GET /api/v2/jobs/:frameworkName/job-attempts/:attemptIndex```

It seems like only virtual-clusters use dashes and path params like frameworkName, jobName use camel, however some exceptions:
1. username, like in DELETE /api/v2/user/:username
2. grouplist, like in PUT /api/v2/user/:username/grouplist
3. virtualcluster, like in PUT /api/v2/user/:username/virtualcluster

docs/rest-server/API.md
yqwang-ms: But if so, if healthz not work but the feature gate enabled, we should not just suppress it, we should alert it. Is alert also in this PR?

docs/rest-server/API.md
yqwang-ms: UID can ensure uniqueness.
Agree to use UID in all jobs api, but it is a breaking change that we would better smoothly migrate.

yqwang-ms:
> could you explain more about this comment, I did not get it

In current job list page, will you also query elastic search? If not, user cannot see very old jobs because they are evicted from APIServer.

debuggy:
> 
> 
> > could you explain more about this comment, I did not get it
> 
> In current job list page, will you also query elastic search? If not, user cannot see very old jobs because they are evicted from APIServer.

No, this PR did not change job list with elastic search. It will be a break change if add this feature, both in webportal and in rest api.

src/rest-server/src/utils/frameworkConverter.js
```diff
@@ -0,0 +1,233 @@
...
+  } catch (error) {
+    logger.warn('Get diagnostics info failed', error);
+    return diagnosticsSummary;
+  }
+
+  const summmaryInfo = diag.substring(0, matches.index + 'matched:'.length);
+  diagnosticsSummary = summmaryInfo + '\n' + yaml.safeDump(podCompletionStatus);
+
+  return diagnosticsSummary;
+};
```
Binyang2014: I remember there are nearly same code at https://github.com/microsoft/pai/blob/41465f33489701db4982b8c8e041383f45ce42db/src/rest-server/src/models/v2/job/k8s.js#L812-L847
Can we reuse this?

yqwang-ms:
> > > could you explain more about this comment, I did not get it
> > 
> > 
> > In current job list page, will you also query elastic search? If not, user cannot see very old jobs because they are evicted from APIServer.
> 
> No, this PR did not change job list with elastic search. It will be a break change if add this feature, both in webportal and in rest api.

So I mean "If not, pls create an issue to track it."

docs/rest-server/API.md
debuggy: So I will use UID in job retry first

src/rest-server/src/utils/frameworkConverter.js
yqwang-ms: Agree to reuse code as much as possible. And it also give user same experience on history and current page.

docs/rest-server/API.md
yqwang-ms: Diff disable and unexpected unhealthy

debuggy: image
image
image

docs/rest-server/API.md
abuccts: * params are js variables in express and the names aren't in endpoint actually
* no idea about exceptions

debuggy:
new retry detail design from @yangou1988 

image

debuggy:
> 
> 
> > > > could you explain more about this comment, I did not get it
> > > 
> > > 
> > > In current job list page, will you also query elastic search? If not, user cannot see very old jobs because they are evicted from APIServer.
> > 
> > 
> > No, this PR did not change job list with elastic search. It will be a break change if add this feature, both in webportal and in rest api.
> 
> So I mean "If not, pls create an issue to track it."

#3845

docs/rest-server/API.md
debuggy: done in updates

debuggy: done in updates

yqwang-ms:
> new retry detail design from @yangou1988
> 
> image

rename retry index to attempt id? since it is returned from attempts api

debuggy:
>     1. rename retry index to attempt id? since it is returned from attempts api

  1. Only retries show in this page, from a discussion of @scarlett2018 and @qfyin , so it's better to use retry index. But the api will return all attempts data, so there is a naming diff between frontend and backend.

>     2. also show the runtime and endtime? runtime -starttime is queuing time, endtime -runtime is job itself pure running duration

  2. Better to keep align with job detail. It might make user confusing if use additional info here.

>     3. also show the start time as an absolute time such as 2020.11.12 11:22, this help to debug with the timestamp inside his job log

  3. Here reuse the util method from job detail.

>     4. highlight the current attempt？

  4. Same as 1, there is no current attempt anymore.

debuggy:
latest UI:
image

src/rest-server/src/utils/frameworkConverter.js
debuggy: There are many detailed changes compared with previous code. Will do code refine and format in another PR, because this might change the behavior of current job API hence postpone release process.

debuggy:
Complement exit diagnostics message:

image

image

src/webportal/src/app/job/job-view/fabric/job-retry/job-retry-card.jsx
```diff
@@ -0,0 +1,313 @@
...
+              className={c(
+                FontClassNames.medium,
+                ColorClassNames.neutralSecondary,
+              )}
+              style={{ marginBottom: spacing.s1 }}
+            >
+              Exit Type:
+            </div>
+            <div className={c(FontClassNames.mediumPlus)}>
+              {`${jobRetry.exitType}`}
```
sunqinzheng: The template string is unnecessary.

src/webportal/src/app/job/job-view/fabric/job-retry/job-retry-card.jsx
```diff
@@ -0,0 +1,313 @@
...
+    return 'N/A';
+  }
+}
+
+const RetryCard = styled.div`
+  background: #f8f8f8;
+  box-shadow: rgba(0, 0, 0, 0.06) 0px 2px 4px, rgba(0, 0, 0, 0.05) 0px 0.5px 1px;
+`;
+
+const TaskRoleCard = styled.div`
```
sunqinzheng: The Card component has style and classname props, you can add your custom style there.
If the style is not reused, I think you don't need to create a new component with copied code here.

src/webportal/src/app/home/home/abnormal-job-list.jsx
```diff
@@ -70,9 +71,9 @@ const AbnormalJobList = ({ jobs, style }) => {
       onRender(job) {
         const { legacy, name, namespace, username } = job;
         const href = legacy
-          ? `/job-detail.html?jobName=${name}`
+          ? `/job-detail.html?jobname=${name}`
```
sunqinzheng: The camelCase `jobName` and lowercase `username` is defined in rest api. I think we need to keep the consistency here

PR #3841, opened 2019-11-13
[Rest Server] Filter pods without nodename and completed pods

abuccts:
Filter pods without nodename and completed pods in virtual cluster resources.

### DIFF
--- a/src/rest-server/src/models/v2/virtual-cluster/k8s.js
+++ b/src/rest-server/src/models/v2/virtual-cluster/k8s.js
@@ -55,17 +55,24 @@ const fetchNodes = async () => {
   });
 };
 
+const fetchPods = async () => {
+  const pods = await axios({
+    method: 'get',
+    url: `${apiserver.uri}/api/v1/pods?labelSelector=type=kube-launcher-task`,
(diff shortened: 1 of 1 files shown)


src/rest-server/src/models/v2/virtual-cluster/k8s.js
```diff
@@ -55,17 +55,24 @@ const fetchNodes = async () => {
...
 
+const fetchPods = async () => {
+  const pods = await axios({
+    method: 'get',
+    url: `${apiserver.uri}/api/v1/pods?labelSelector=type=kube-launcher-task`,
+    httpsAgent: apiserver.ca && new Agent({ca: apiserver.ca}),
+    headers: apiserver.token && {Authorization: `Bearer ${apiserver.token}`},
+  });
+  return pods.data.items.filter((pod) => {
+    return (pod.spec.nodeName != null && !(pod.status.phase === 'Succeeded' || pod.status.phase === 'Failed'));
```
yqwang-ms: != null and is not empty?

abuccts: update nodename not to be `''`, `null`, `undefined`, `0`, `NaN`, `false`

PR #3852, opened 2019-11-15
[Webportal] static content caching optimization

hzy46:
There are 2 main issues about static content (js, style css, font, svg, .., etc) in current webportal:

  - They are caching-disabled. The `max-age` in the response header for them is 0. Thus, the browser will always download these contents, which increases both the front-end loading time and the server's traffic.
  - Many js files are splitted into very small chunks (<1kb), which means they should also be transferred by new HTTP connection.

This PR sets a `max-age` for those contents,  use `contenthash` in webpack to ensure correct caching, 
and only js files larger than 30kb would be chunked.

Loading `submit.html` before and after optimization (tested for 5 times in each case):
   - Before optimization + First loading: 4.19s ~ 8.61s
   - Before optimization + Second loading: 4.21s~6.49s
   - After optimization   + First loading: 3.71s ~ 4.76s
   - After optimization   + Second loading: **0.65s~0.94s**
 
Since "second loading" would be the main use case, the user experience could be enhanced a lot by this optimization.

Chrome `Network` page for "Before optimization + Second loading":

before

Chrome `Network` page for "After optimization + Second loading":

after

### DIFF
--- a/src/webportal/config/webpack.common.js
+++ b/src/webportal/config/webpack.common.js
@@ -263,7 +263,7 @@ const config = (env, argv) => ({
       { from: 'src/assets/img/favicon.ico', to: 'favicon.ico' },
     ]),
     new MiniCssExtractPlugin({
-      filename: 'styles/[name].bundle.css',
+      filename: 'styles/[name].[contenthash].css',
     }),
     // required by ejs loader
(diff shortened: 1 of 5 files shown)


debuggy:
LGTM

Issue #3760, opened 2019-10-21
Pod stuck in Pending due to kubelet skip to create pod sandbox


yqwang-ms:
**Organization Name**:

**Short summary about the issue/question**:
https://github.com/kubernetes/kubernetes/issues/79398
https://github.com/kubernetes/kubernetes/pull/79451

Long term solution is (1.17 k8s) https://github.com/kubernetes/kubernetes/commit/3fac48f86ad0a2e417bcdb518c62cc3fbcf9ee1f

Short term workaround:
Change Pod restartPolicy=OnFailure for elastic job (TBD: and training job)

**Brief what process you are following**:

**How to reproduce it**:

**OpenPAI Environment**:
- OpenPAI version:
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Hardware (e.g. core number, memory size, storage size, GPU type etc.):
- Others:

**Anything else we need to know**:

related log:
```
kuberuntime_manager.go:427] No ready sandbox for pod "..." can be found. Need to start a new one
```

yqwang-ms:
@abuccts  pls help to do the change for elastic job :)

yqwang-ms:
let's hold on this first to test if 1.16.0 can be in this release

abuccts:
Tracked in #3866.

yqwang-ms:
Tracked in #3863.

PR #3856, opened 2019-11-15
[Rest Server] Update restart policy to avoid stuck pending pods

abuccts:
Update restart policy to avoid stuck pending pods #3760.

### DIFF
--- a/src/rest-server/src/models/v2/job/k8s.js
+++ b/src/rest-server/src/models/v2/job/k8s.js
@@ -319,6 +319,8 @@ const generateTaskRole = (taskRole, labels, config) => {
         },
         spec: {
           privileged: false,
+          // workaround for pending pods due to no ready sandbox before the fix
+          // https://github.com/kubernetes/kubernetes/commit/3fac48f8
           restartPolicy: gangAllocation === 'true' ? 'Never' : 'OnFailure',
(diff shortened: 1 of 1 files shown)


src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -319,7 +319,7 @@ const generateTaskRole = (taskRole, labels, config) => {
         },
         spec: {
           privileged: false,
-          restartPolicy: 'Never',
+          restartPolicy: gangAllocation === 'true' ? 'Never' : 'OnFailure',
```
Binyang2014: Add comments for this? We should revert this change after we update k8s to 16.2 or above.

yqwang-ms:
User may not want to forever retry on failure. Could alert auto delete the pod if detected such issue? @Binyang2014

Binyang2014:
> User may not want to forever retry on failure. Could alert auto delete the pod if detected such issue? @Binyang2014

Maybe we can use alert-manager webhook to achieve this [web-hook](https://prometheus.io/docs/alerting/configuration/#webhook_config). Create another service receive such alert and do some actions.

For better experience, admin can config this webhook's reaction when alert is received.

For now, if we want to auto delete the pod. I think we should find a way to let admin know what will happen, and admin can turn off such feature if it causes other issues.

PR #3859, opened 2019-11-18
HiveD: intra-vc preemption after restart

zhypku:
If a restarted scheduler finds that the quota of a certain type of cell is insufficient, it will try to preempt (downgrade) pods with lower priority from the VC.
Tested on the 15 testbed.


yqwang-ms:
> 

Add an unit test for this issue?

---
Refers to: subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go:24 in db176df. [](commit_id = db176df94965cafa106096557bf7924e31a85620, deletion_comment = False)

PR #3861, opened 2019-11-18
HiveD intra-vc preemption for restart

zhypku:
If a restarted scheduler finds that the quota of a certain type of cell is insufficient, it will try to preempt (downgrade) pods with lower priority from the VC.


yqwang-ms:
> 					return nil, "", nil

This is another issue

---
Refers to: subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go:814 in 37d7bcc. [](commit_id = 37d7bcca1a204c7e37d278c0f4e6a4c74269c1d9, deletion_comment = False)

PR #3225, opened 2019-07-18
[Pure k8s] add scheduler validation in restserver

mzmssg:
1. Add deployment flag to enable/disable hived
2. Validate memory/cpu in hived mode
3. Add resource unit api to query cluster sku

PR #3867, opened 2019-11-19
[Rest Server] Update hived resource validation

abuccts:
Update hived resource validation.


mzmssg:
Maybe we can remove one from `memory`/`memoryMB` to uniform the variable name? 

@abuccts your opinion?

mzmssg:
Actually it's the same issue as #3595. A uniform name could resolve the problem.

abuccts:
> Actually it's the same issue as #3595. A uniform name could resolve the problem.

It's a **bug fix** instead of uniforming name.
In #3225, the source object only uses `memory` but reference uses both `memory` and `memoryMB`.

mzmssg:
Yeah, I know it's a bug. 
I just suggest to replace all `xxMemoryMB` with `xxMemory`, then variable name align with the source.

PR #3875, opened 2019-11-21
[runtime] [rest-server] Add teamwise_storage runtime plugin

Binyang2014:
After admin setup the storage config, user could mounted the storage by using this storage runtime plugin:
```yaml
extras:
  com.microsoft.pai.runtimeplugin:
    - plugin: teamwiase_storage
      parameters:
        storageConfigNames:
          - PAI_SAHRE  # storage config name provided by admin
```
If the parameter is empty, but teamwise_storage plugin is set in the protocol, the default storage config will be used.

Currently, we support following storage : NFS, Samba, Azure Blob and Azure file


src/rest-server/src/config/launcher.js
```diff
@@ -194,6 +196,7 @@ if (launcherType === 'yarn') {
   launcherConfig = {
     apiServerUri: apiserver.uri,
     apiVersion: 'frameworkcontroller.microsoft.com/v1',
+    restServerUri: process.env.REST_SERVER_URI,
```
abuccts: it's not a launcher config, put it in `src/config/index.js`

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -339,6 +339,25 @@ const generateTaskRole = (taskRole, labels, config) => {
...
+                  value: userToken,
+                },
+                {
+                  name: 'PAI_USER_NAME',
+                  value: labels.userName,
+                },
+                {
+                  name: 'PAI_JOB_NAME',
+                  value: labels.jobName,
+                },
```
abuccts: remove

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -339,6 +339,25 @@ const generateTaskRole = (taskRole, labels, config) => {
                   name: 'GANG_ALLOCATION',
                   value: gangAllocation,
                 },
+                // Pass user token to runtime to give runtime permission to call rest server
+                // Actually we should provide service token for kube-runtime and do not let
+                // runtime personate as a real user.
+                {
+                  name: 'PAI_USER_TOKEN',
```
abuccts: put it in `runtime-env.js`

src/rest-server/src/controllers/v2/job.js
```diff
@@ -43,6 +44,8 @@ const update = asyncHandler(async (req, res) => {
   const jobName = res.locals.protocol.name;
   const userName = req.user.username;
   const frameworkName = `${userName}~${jobName}`;
+  const userToken = req.headers.authorization;
```
abuccts: what if the token expires when pod starts

src/kube-runtime/src/init.d/initializer.py
```diff
@@ -8,46 +8,45 @@
...
+        LOGGER.info(line)
     proc.wait()
     if proc.returncode:
-        logger.error("failed to run %s, error code is %s", script_path, proc.returncode)
+        LOGGER.error("failed to run %s, error code is %s", script_path, proc.returncode)
+        raise Exception("Failed to run init script")
 
 # just a workaround, need to remove this function in the future
-def is_plugin_validate(jobConfig):
+def is_plugin_validate(jobconfig):
```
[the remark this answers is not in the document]
Binyang2014: I know, I think it's a bug. Before this change, I can not find the ref to `jobConfig`, this args is not used. And `jobconfig` is a global variable, so it can be reference here, Nothing wrong happens. 

I remove the global variable and change to `jobconfig` here

src/rest-server/src/config/launcher.js
Binyang2014: Sure

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: Will check this, maybe my testing bed is too old and not contain these envs

Binyang2014: I believe these envs only can be used in app container, not init container. For init container, we still need these envs

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: runtime-env is used for app container, not init container. We need the token in init container

src/kube-runtime/test/test_runtime.py
```diff
@@ -16,52 +17,258 @@
...
+
+#pylint: disable=wrong-import-position
+sys.path.append(os.path.join(os.path.dirname(os.path.abspath(__file__)), "../src/init.d"))
+sys.path.append(os.path.join(os.path.dirname(os.path.abspath(__file__)), "../src/plugins"))
 
 from initializer import init_plugins
+import teamwise_storage.init as storage_plugin
 
-package_directory_com = os.path.dirname(os.path.abspath(__file__))
+#pylint: enable=wrong-import-position
```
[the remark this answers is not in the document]
Binyang2014: Not easy to fix. To fix it, we need to create `__init__.py` under kube-runtime folder. But kube-runtime not a python module . And the folder name `init.d` contains a `dot`. Make it hard to import

src/kube-runtime/src/plugins/teamwise_storage/init.py
```diff
@@ -0,0 +1,185 @@
...
+import http
+import logging
+import os
+import sys
+
+import requests
+
+#pylint: disable=wrong-import-position
+sys.path.append(os.path.join(os.path.dirname(os.path.abspath(__file__)), ".."))
+from plugin_utils import plugin_init, inject_commands
```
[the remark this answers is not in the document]
Binyang2014: If we use relative Imports, seems we can make sure the script can be run at any folder. And will occur beyond-top-level-package-error in CI. Refer to: https://stackoverflow.com/questions/30669474/beyond-top-level-package-error-in-relative-import

src/kube-runtime/src/init.d/initializer.py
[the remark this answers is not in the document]
abuccts: you can change all names in this function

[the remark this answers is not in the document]
Binyang2014: Currently, I have removed the global variable of jobconfig. and no such lint error for this file 

---
In reply to: [349433452](https://github.com/microsoft/pai/pull/3875#discussion_r349433452) [](ancestors = 349433452)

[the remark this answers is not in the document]
abuccts: there're two words, which should be [separated by underscores to improve readability](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names)
you didn't get lint error because lint does not know there're two words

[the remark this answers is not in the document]
Binyang2014: ok, fine

---
In reply to: [349480043](https://github.com/microsoft/pai/pull/3875#discussion_r349480043) [](ancestors = 349480043)

src/rest-server/src/controllers/v2/job.js
Binyang2014: Then the init container will failed and return as platform error.
Currently, we don't have service account config so pass user token here. 
I think it's just a short-term solution, and current permission check avoid to create a generate token for runtime. We can fix this in future

---
In reply to: [348965232](https://github.com/microsoft/pai/pull/3875#discussion_r348965232) [](ancestors = 348965232)

src/rest-server/src/config/launcher.js
abuccts: pls delete here

Binyang2014: Fixed

---
In reply to: [349982348](https://github.com/microsoft/pai/pull/3875#discussion_r349982348) [](ancestors = 349982348)

PR #3879, opened 2019-11-22
[Kubespray] Playbooks to clean the env which installed OpenPAI before

ydye:


contrib/kubespray/apt-fix.yml
```diff
@@ -0,0 +1,19 @@
...
+  tasks:
+    - name: fix-apt-key
+      shell: apt-key adv --keyserver keyserver.ubuntu.com --recv-keys {keynumber}
+      args:
+        executable: /bin/bash
+
+    - name: remove cephlist
+      shell: rm -rf  /etc/apt/sources.list.d/ceph.list
+      args:
+        executable: /bin/bash
```
abuccts: it's only needed in specific machines, should not be included

PR #3907, opened 2019-11-26
Fix clone bug of jobname

debuggy:


src/webportal/src/app/job/job-view/fabric/job-detail/components/clone-button.jsx
```diff
@@ -28,7 +28,7 @@ const CloneButton = ({ rawJobConfig, namespace, jobName }) => {
       op: 'resubmit',
       type: 'job',
       user: namespace,
-      jobName: jobName,
+      jobname: jobName,
```
yqwang-ms: Seems in PAI codebase, there are still many places using jobName

debuggy: close this pr for further alignment

PR #3922, opened 2019-11-28
Fix clone job loading misbehavior

debuggy:


debuggy:
fixed in master branch #3923

PR #3933, opened 2019-11-29
[Rest Server] Specify image pull policy for app container

abuccts:
Specify image pull policy for app container.


yqwang-ms:
Why do this change? Default is already reasonable

PR #3936, opened 2019-11-29
[Rest Server] Check resource quota for gang scheduling only

abuccts:
Check resource quota for gang scheduling only.


src/rest-server/src/middlewares/v2/hived.js
```diff
@@ -186,7 +186,7 @@ const hivedValidate = (protocolObj) => {
     protocolObj.taskRoles[taskRole].hivedPodSpec = podSpec;
   }
   const maxGpuNumber = virtualCellCapacity[virtualCluster].resourcesTotal.gpu;
-  if (totalGpuNumber > maxGpuNumber) {
+  if (totalGpuNumber > maxGpuNumber && gangAllocation) {
```
yqwang-ms: Alsi check it is not oppo

PR #3943, opened 2019-12-02
[Rest Server] Add image pull secrets for private registry

abuccts:
Add image pull secrets for private registry.

Limitation: Do not support multiple users for same registry like `.docker/config.json`. Multiple registries in one job is fine.

### DIFF
--- a/src/rest-server/src/models/v2/job/k8s.js
+++ b/src/rest-server/src/models/v2/job/k8s.js
@@ -292,7 +292,7 @@ const convertFrameworkDetail = async (framework) => {
   return detail;
 };
 
-const generateTaskRole = (taskRole, labels, config, userToken) => {
+const generateTaskRole = (frameworkName, taskRole, labels, config, userToken) => {
   const ports = config.taskRoles[taskRole].resourcePerInstance.ports || {};
   for (let port of ['ssh', 'http']) {
     if (!(port in ports)) {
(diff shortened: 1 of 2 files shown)


src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -778,16 +878,19 @@ const put = async (frameworkName, config, rawConfig, userToken) => {
...
     } else {
       // do not await for delete
+      auths.length && deleteSecret(frameworkName);
       deletePriorityClass(frameworkName);
       throw error;
     }
   }
   if (response.status !== status('Created')) {
     // do not await for delete
+    auths.length && deleteSecret(frameworkName);
```
Binyang2014: Do we need to delete secret in watchdog to avoid potential leak?

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -681,6 +687,94 @@ const deletePriorityClass = async (frameworkName) => {
...
+    } else {
+      throw error;
+    }
+  }
+  if (response.status !== status('Created')) {
+    throw createError(response.status, 'UnknownError', response.data.message);
+  }
+};
+
+const patchSecret = async (frameworkName, frameworkUid) => {
```
yqwang-ms: patchSecretOwner?

src/rest-server/src/models/v2/job/k8s.js
yqwang-ms: better to have. only need to delete those who do not have owner.

src/rest-server/src/models/v2/job/k8s.js
abuccts: udpated

src/rest-server/src/models/v2/job/k8s.js
abuccts: there exist secrets which are created elsewhere

yqwang-ms: We can use **naming** to find job secrets created here

PR #3962, opened 2019-12-04
HiveD: record cell type in pod annotation

zhypku:
Record cell type name (instead of the level) in pod annotation. This enables us to shorten / lengthen cell chain.
This PR also adds some log enrichment.

### DIFF
--- a/subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
+++ b/subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
@@ -50,6 +50,8 @@ type HivedAlgorithm struct {
 	freeCellList map[CellChain]ChainCellList
 	// map each GPU type to all chains that contain this type
 	chains map[string][]CellChain
+	// map each level in a chain to the specific cell type name
+	cellTypes map[CellChain]map[CellLevel]api.CellType
 	// all affinity groups that have been allocated cells
 	allocatedAffinityGroups map[string]*AlgoAffinityGroup
 	// all reserved physical cells (VC -> reservation ID -> cells)
@@ -60,13 +62,14 @@ type HivedAlgorithm struct {
 
 // NewHivedAlgorithm initializes a HivedAlgorithm from the config file
 func NewHivedAlgorithm(sConfig *api.Config) *HivedAlgorithm {
-	pcl, gpuNums, gpuTypeToChain, nonReservedVcl, reservedVcl, reservedPc := ParseConfig(sConfig)
+	pcl, gpuNums, gpuTypeToChain, cellLevelToType, nonReservedVcl, reservedVcl, reservedPc := ParseConfig(sConfig)
 	h := &HivedAlgorithm{
 		vcSchedulers:            make(map[api.VirtualClusterName]intraVCScheduler),
 		opportunisticSchedulers: map[CellChain]*topologyAwareScheduler{},
 		fullCellList:            pcl,
 		freeCellList:            make(map[CellChain]ChainCellList),
 		chains:                  gpuTypeToChain,
+		cellTypes:               cellLevelToType,
 		allocatedAffinityGroups: make(map[string]*AlgoAffinityGroup),
 		reservedCells:           reservedPc,
 	}
@@ -127,7 +130,15 @@ func (h *HivedAlgorithm) Schedule(pod *core.Pod, suggestedNodes []string) intern
 		podIndex = int32(len(group.allocatedPods[s.GpuNumber]))
 	}
 	return generatePodScheduleResult(
-		groupPhysicalPlacement, groupVirtualPlacement, s.GpuNumber, podIndex, newGroup, suggestedNodeSet, pod)
+		groupPhysicalPlacement,
+		groupVirtualPlacement,
+		h.cellTypes,
+		s.GpuNumber,
+		podIndex,
+		newGroup,
+		suggestedNodeSet,
+		s.VirtualCluster,
+		pod)
 }
 
 func (h *HivedAlgorithm) AddAllocatedPod(pod *core.Pod) {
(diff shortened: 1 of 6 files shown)


subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/api/types.go
```diff
@@ -110,9 +110,9 @@ type AffinityGroupMemberBindInfo struct {
 type PodPlacementInfo struct {
 	PhysicalNode       string  `yaml:"physicalNode"`
 	PhysicalGpuIndices []int32 `yaml:"physicalGpuIndices"`
-	// levels of the preassigned cells used by the pods. used to locate the virtual cells
+	// preassigned cell types used by the pods. used to locate the virtual cells
 	// when adding an allocated pod
-	PreassignedCellLevels []int32 `yaml:"preassignedCellLevels"`
+	PreassignedCellTypes []CellType `yaml:"preassignedCellTypes"`
 }
```
yqwang-ms: What is the behavior to recover previous bound Pods which only has field PreassignedCellLevels?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -737,50 +764,30 @@ func (h *HivedAlgorithm) findPhysicalGpuInChain(
...
 	if affinityGroupBindInfo == nil {
+		reason := "insufficient capacity in physical cluster"
+		if groupVirtualPlacement != nil {
+			reason = fmt.Sprintf("insufficient quota in VC %v", vc)
+		}
 		return internal.PodScheduleResult{
 			PodWaitInfo: &internal.PodWaitInfo{
 				// TODO: Enrich the Pod Waiting Reason.
-				Reason: "",
+				Reason: reason,
```
yqwang-ms: So remove this TODO

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -800,16 +807,19 @@ func generatePodScheduleResult(
 		// may cause the selected node to be excluded from the suggested nodes
 		// TODO: Keep selectedNode within suggestedNodeSet, so here should be Unreachable
 		if !suggestedNodeSet.Contains(selectedNode) && newGroup {
-			panic(fmt.Sprintf("[%v]: node %v picked by algorithm but not in K8S candidates",
-				internal.Key(pod), selectedNode))
+			return internal.PodScheduleResult{
+				PodWaitInfo: &internal.PodWaitInfo{
+					Reason: fmt.Sprintf("node %v picked by algorithm but not in K8S candidates", selectedNode),
+				},
+			}
```
yqwang-ms: Let's keep this still be panic.
Because after you keep selectedNode within suggestedNodeSet, so here should be Unreachable, and if it really reaches, it is exception (code bug).

yqwang-ms: I mean we keep allNodesInSuggested in getFewestOpporPhysicalCell, etc

---
In reply to: [353628498](https://github.com/microsoft/pai/pull/3962#discussion_r353628498) [](ancestors = 353628498)

zhypku: I mean, we can't do this. Some nodes may be excluded from the the suggested nodes only because there are other pods on it (we need to preempt them). If we only consider suggested nodes when scheduling, such preemption will never happen

---
In reply to: [353649575](https://github.com/microsoft/pai/pull/3962#discussion_r353649575) [](ancestors = 353649575,353628498)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -796,20 +789,35 @@ func generatePodScheduleResult(
...
+			return internal.PodScheduleResult{
+				PodWaitInfo: &internal.PodWaitInfo{
+					Reason: reason,
+				},
+			}
+		}
+		if selectedNode == "" {
+			return internal.PodScheduleResult{
+				PodWaitInfo: &internal.PodWaitInfo{
+					Reason: fmt.Sprintf("node %v picked by algorithm but not in K8S candidates", selectedNode),
```
yqwang-ms: >"node %v picked by algorithm but not in K8S candidates" [](start = 25, length = 55)

no node is picked here. Pls refine the msg :), such as:

insufficient capacity in physical cluster with respect to K8S candidates
/
insufficient quota in VC %v with respect to K8S candidates
?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -796,20 +789,35 @@ func generatePodScheduleResult(
...
+				reason = fmt.Sprintf("insufficient quota in VC %v", vc)
+			}
+			return internal.PodScheduleResult{
+				PodWaitInfo: &internal.PodWaitInfo{
+					Reason: reason,
+				},
+			}
+		}
+		if selectedNode == "" {
+			return internal.PodScheduleResult{
```
yqwang-ms: Still consider newGroup, so
  if selectedNode == "" && newGroup {

zhypku: if newGroup == false, selectedNode must be "". See collectPreemptionVictims

---
In reply to: [353695916](https://github.com/microsoft/pai/pull/3962#discussion_r353695916) [](ancestors = 353695916)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
zhypku: I used "insufficient capacity in physical cluster". "insufficient quota" seems to be not accurate because this case does not mean the user does not have quota :)

---
In reply to: [353695013](https://github.com/microsoft/pai/pull/3962#discussion_r353695013) [](ancestors = 353695013)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
yqwang-ms: So for line 876, it should be 

  if podGpuNum == currentGpuNum && (suggestedNodeSet.Contains(mbi.PodPlacements[currentPodIndex].PhysicalNode) || !newGroup) {

right?



---
In reply to: [354073449](https://github.com/microsoft/pai/pull/3962#discussion_r354073449) [](ancestors = 354073449,353695916)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
yqwang-ms: I mean can we do similar thing like:

   reason := "insufficient capacity in physical cluster with respect to K8S candidates"
   if groupVirtualPlacement != nil {
    reason = fmt.Sprintf("insufficient quota in VC %v with respect to K8S candidates", vc)
   }

---
In reply to: [354076707](https://github.com/microsoft/pai/pull/3962#discussion_r354076707) [](ancestors = 354076707,353695013)

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -796,20 +789,33 @@ func generatePodScheduleResult(
...
+		affinityGroupBindInfo, selectedNode, selectedGpuIndices := generateAffinityGroupBindInfo(
+			groupPhysicalPlacement, groupVirtualPlacement, cellLevelToType, currentGpuNum, currentPodIndex, newGroup, suggestedNodeSet)
+		var waitReason string
+		if affinityGroupBindInfo == nil {
+			waitReason = "insufficient capacity in physical cluster"
+			if groupVirtualPlacement != nil {
+				waitReason = fmt.Sprintf("insufficient quota in VC %v", vc)
+			}
+		}
+		if selectedNode == "" {
```
yqwang-ms: >if [](start = 2, length = 2)

else if

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -796,20 +789,33 @@ func generatePodScheduleResult(
...
+			waitReason = "cannot find a K8s candidate node in physical cluster"
```
yqwang-ms: >in [](start = 50, length = 2)

within

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -796,20 +789,33 @@ func generatePodScheduleResult(
...
+			if groupVirtualPlacement != nil {
+				waitReason = fmt.Sprintf("cannot find a K8s candidate node with VC %v's quota", vc)
```
yqwang-ms: >with [](start = 63, length = 4)

within

PR #3971, opened 2019-12-05
[Rest Server] Fix duplicate affinity group name issue

abuccts:
Fix duplicate affinity group name issue.

### DIFF
--- a/src/rest-server/src/middlewares/v2/hived.js
+++ b/src/rest-server/src/middlewares/v2/hived.js
@@ -33,7 +33,7 @@ const convertPriority = (priorityClass='test') => {
   return priorityClass in priorityMap ? priorityMap[priorityClass] : null;
 };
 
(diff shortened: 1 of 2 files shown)


src/rest-server/src/middlewares/v2/hived.js
```diff
@@ -162,14 +162,14 @@ const hivedValidate = (protocolObj) => {
 
       const affinityGroupName = hivedConfig.taskRoles[taskRole].affinityGroupName;
       podSpec.affinityGroup = affinityGroupName ? {
-        name: protocolObj.name + '/' + affinityGroupName,
+        name: `${username}/${protocolObj.name}/${affinityGroupName}`,
```
yqwang-ms: Use the same way for this, i.e. ${username}~${protocolObj.name}

PR #3980, opened 2019-12-06
Fix typos

abuccts:
Fix typos by [misspell](https://github.com/client9/misspell).

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -28,3 +28,18 @@ jobs:
     - name: Lint
       run: |
         python src/utilities/doc_checker.py .
+
+  spelling:
+    name: Spelling check
+    runs-on: ubuntu-16.04
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v1
+    - name: Install dependencies
+      run: |
+        curl -L https://git.io/misspell | sudo bash -s -- -b /bin
+    - name: Check spelling
+      run: |
+        rm -rf ./subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/vendor/
+        misspell -error .
--- a/RELEASE_NOTE.md
+++ b/RELEASE_NOTE.md
@@ -48,7 +48,7 @@ For more details about this release, please refer to [detailed release note](htt
 - New alerts for unhealthy GPUs, currently including following alerts #2209:
 - Admin could know all running jobs on a node. #2197
 - Filter supports in Job List View. #302
-- Hold the Env for failed jobs which are casued by user error. #2272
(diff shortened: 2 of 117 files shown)


yqwang-ms:
It should be an option (default disable) instead of enforcement. For example, it may wrongly rename some variables, such as trunc to tune which causes bugs.

yqwang-ms:
> What problems does it have?

Unlike the other projects, this doesn't know what a "word" is. There may be more false positives and false negatives due to this. On the other hand, it sometimes catches things others don't.

Either way, please file bugs and we'll fix them!

Since it operates in parallel to make corrections, it can be non-obvious to determine exactly what word was corrected.

It's making mistakes. How can I debug?

Run using -debug flag on the file you want. It should then print what word it is trying to correct. Then file a bug describing the problem. Thanks!

Why is it making mistakes or missing items in golang files?

The matching function is case-sensitive, so variable names that are multiple worlds either in all-upper or all-lower case sometimes can cause false positives. For instance a variable named bodyreader could trigger a false positive since yrea is in the middle that could be corrected to year. Other problems happen if the variable name uses a English contraction that should use an apostrophe. The best way of fixing this is to use the Effective Go naming conventions and use camelCase for variable names. You can check your code using golint

Binyang2014:
Not sure what's the right process for spell check. 

One option is add plugin in the IDE and we can check the spell by our own. For `vs code`, we can use [Code Spell Checker](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker)

Another option is we can trigger the spell checker for each PR, but only check the modified files. It should not blocking the submission, just give some warnings. And the PR author can do the decision.
It will not cause the lint always failed, if we involved some special variable names by purpose. 

More strict option is provide a config for the spell checker. And run the spell checker for all files. If some false negative cases occur, let service owner modify the config file to pass by this case. If the spell checker seldom makes mistakes, it maybe a good choice. Chromium seems use this way: [Chromium spell checker](https://www.chromium.org/developers/how-tos/editing-the-spell-checking-dictionaries)

abuccts:
> It should be an option (default disable) instead of enforcement. For example, it may wrongly rename some variables, such as trunc to tune which causes bugs.

What does *it* mean?
1. For the GitHub Action, the spelling check only does check and won't do auto correction, it will be run in CI but won't change anything in files.
2. For this PR, I have reviewed all fixs, although recall may be low, the precision is 100%.

abuccts:
> Make it just warning and optional better to have thing

What does *warning* and *optional* mean?

The GitHub Action is not required in merging PR, and you can always add words to exceptions, or ignore this Action.
At least for now, the tool could find 200+ typos with 0 false positive. It's just added to remind there're typos.

PR #3986, opened 2019-12-09
Binyli/job ssh

Binyang2014:
Remove ssh plugin when gangAllocation is set to false.
Tested in int bed with following job config
```yaml
extras:
  gangAllocation: false
  com.microsoft.pai.runtimeplugin:
    - plugin: ssh
      parameters:
        jobssh: true
```


src/kube-runtime/test/test_runtime.py
```diff
@@ -24,6 +24,7 @@
 import logging
 import logging.config
 
+import initializer
 from initializer import init_plugins
```
abuccts: duplicate

Binyang2014:
Remove ssh plugin when gangAllocation is set to false.
Tested in int bed with following job config
```yaml
extras:
  gangAllocation: false
  com.microsoft.pai.runtimeplugin:
    - plugin: ssh
      parameters:
        jobssh: true
```

PR #3996, opened 2019-12-10
[Docs] OpenAPI docs for RESTful API

abuccts:
Add OpenAPI docs for RESTful API.


debuggy:
How to differ api/v1 and api/v2? And where should be a new endpoint added, v1/xxx or v2/xxx?

abuccts:
> the group name better in path params than request body

1. I met the same error on [online Swagger editor](https://editor.swagger.io/).
But it's not an error in latest Swagger UI or [redoc](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/microsoft/pai/xiongyf/openapi/src/rest-server/docs/swagger.yaml).
2. It should be in API refactors. This PR just adds openapi spec.

abuccts:
> How to differ api/v1 and api/v2? And where should be a new endpoint added, v1/xxx or v2/xxx?

For jobs endpoint, v1 means old json format config, v2 means protocol yaml config.
For others, I do not know neither. Not sure why groups, storage, etc are v2, this PR just migrated from md docs to openapi spec.

PR #4003, opened 2019-12-10
[HiveD] fix pod hanging after reconfiguration

zhypku:
If a pod's placement has been invalid (e.g., due to reconfiguration), we will retrieve its placement from other pods' annotation and then insist the binding decision.


yqwang-ms:
Have you end to end test it?
BTW, add unit test here?

subprojects/GOPATH/src/github.com/microsoft/hivedscheduler/pkg/algorithm/hived_algorithm.go
```diff
@@ -1026,6 +1039,24 @@ func collectPreemptionVictims(
...
+				for _, mbi := range info.AffinityGroupBindInfo {
+					if gpuNum == int32(len(mbi.PodPlacements[0].PhysicalGpuIndices)) {
+						return mbi.PodPlacements[podIndex], info.CellChain
+					}
+				}
+			}
+		}
+	}
+	panic(fmt.Sprintf("No allocated pod in an allocated group %v", group.name))
+}
```
yqwang-ms: also Sprintf current pod info:gpuNum int32, podIndex int32

PR #4006, opened 2019-12-11
Chx/storage plugin

Sissie1122:
Add storage plugin


src/kube-runtime/src/plugins/storage/init.py
```diff
@@ -0,0 +1,63 @@
...
+import argparse
+import yaml
+
+from plugin_utils import plugin_init, inject_commands
+
+logger = logging.getLogger(__name__)
+CLUSTER_ALIAS = 'cluster_alias'
+USER_NAME = os.environ.get("PAI_USER_NAME")
+PAI_REST_SERVER_URI = os.environ.get("PAI_REST_SERVER_URI")
+USER_TOKEN = os.environ.get("PAI_USER_TOKEN").split()[-1]
```
mydmdm: please add comment 
`# $PAI_USER_TOKEN = "Bearer <token>"`

src/kube-runtime/src/plugins/storage/init.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+if __name__ == "__main__":
+    [parameters, pre_script, post_script] = plugin_init()
+
+    if parameters is not None:
+        sdk_version = parameters.get('sdkBranch', 'master')
```
mydmdm: Since some of our updates are in the SE branch, I think we could let the default be `pai-for-edu` temporarily

src/kube-runtime/src/plugins/storage/init.py
```diff
@@ -0,0 +1,63 @@
...
+        install_uri = '-e "git+https://github.com/Microsoft/pai@{}#egg=openpaisdk&subdirectory=contrib/python-sdk"'.format(sdk_version)
+        container_sync_space = parameters.get('syncSpace', None)
```
mydmdm: We could use a default path such as `/persistent`

src/kube-runtime/src/plugins/storage/init.py
```diff
@@ -0,0 +1,63 @@
...
+        assert container_sync_space, 'Must specify a directory to save on persistent storage'
+        #TODO: check DB for storage path
+        storage_path_prefix = f"pai://{CLUSTER_ALIAS}/0"
+        pre_commands = [
+            # install openpaisdk
+            f'python -m pip install {install_uri}',
+            # add cluster
+            f'pai add-cluster --cluster-alias {CLUSTER_ALIAS} --pai-uri {PAI_REST_SERVER_URI} --user {USER_NAME} --token {USER_TOKEN}',
```
mydmdm: Let's add a todo *pai add-cluster* could receiver `--rest <rest-server-uri>` (we could let the `api/v1` API to give cluster uri) (to deal with the case that there are multiple rest servers)

src/kube-runtime/src/plugins/storage/init.py
```diff
@@ -0,0 +1,63 @@
...
+            # download 
+            f'pai copy {storage_path_prefix}/$PAI_USER_NAME/$PAI_JOB_NAME/$PAI_CURRENT_TASK_ROLE_NAME/$PAI_CURRENT_TASK_ROLE_CURRENT_TASK_INDEX {container_sync_space}'
```
mydmdm: please check with https://github.com/microsoft/pai/blob/pai-for-edu/docs/pai-for-edu.md#job-persistence
here we only need to download the `inputs` subdirectory
`pai copy {storage_path_prefix}/$PAI_USER_NAME/$PAI_JOB_NAME/inputs {container_sync_space}/inputs`

src/kube-runtime/src/plugins/storage/init.py
```diff
@@ -0,0 +1,63 @@
...
+        ]
+        post_commands = [
+            # delete file 
+            f'pai delete {storage_path_prefix}/$PAI_USER_NAME/$PAI_JOB_NAME/$PAI_CURRENT_TASK_ROLE_NAME/$PAI_CURRENT_TASK_ROLE_CURRENT_TASK_INDEX',
+            # upload file
+            f'pai copy {container_sync_space} {storage_path_prefix}/$PAI_USER_NAME/$PAI_JOB_NAME/$PAI_CURRENT_TASK_ROLE_NAME/$PAI_CURRENT_TASK_ROLE_CURRENT_TASK_INDEX']
```
mydmdm: also we only upload `/persistent/outputs`
```bash
pai copy {container_sync_space}/outputs {storage_path_prefix}/$PAI_USER_NAME/$PAI_JOB_NAME/$PAI_CURRENT_TASK_ROLE_NAME/$PAI_CURRENT_TASK_ROLE_CURRENT_TASK_INDEX
```

src/kube-runtime/src/plugins/storage/init.py
Sissie1122: How about 'retry' case? Do we need to download the ```outputs``` if it has older checkpoints?

PR #4007, opened 2019-12-11
[Rest Server] Expose preemption status in scheduler

abuccts:
Expose preemption status in scheduler.

### DIFF
--- a/src/rest-server/deploy/rest-server.yaml.template
+++ b/src/rest-server/deploy/rest-server.yaml.template
@@ -68,6 +68,8 @@ spec:
 {% else %}
           value: defaultscheduler
 {% endif %}
+        - name: HIVED_WEBSERVICE_URI
+          value: {{ cluster_cfg['hivedscheduler']['webservice'] }}
         - name: LOG_MANAGER_PORT
           value: "{{ cluster_cfg['log-manager']['port'] }}"
 {%- endif %}
--- a/src/rest-server/src/models/v2/job/k8s.js
+++ b/src/rest-server/src/models/v2/job/k8s.js
@@ -224,6 +224,8 @@ const convertTaskDetail = async (taskStatus, ports, userName, jobName, taskRoleN
     ...launcherConfig.enabledHived && {
       hived: {
         affinityGroupName,
+        lazyPreempted: null,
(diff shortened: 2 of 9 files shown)


src/rest-server/deploy/rest-server.yaml.template
```diff
@@ -68,6 +68,8 @@ spec:
 {% else %}
           value: defaultscheduler
 {% endif %}
+        - name: HIVED_WEBSERVICE_URI
+          value: {{ cluster_cfg['hivedscheduler']['webservice'] }}
```
yqwang-ms: Is this ready?

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -288,6 +301,26 @@ const convertFrameworkDetail = async (framework) => {
       ),
     };
   }
+
+  if (launcherConfig.enabledHived) {
+    const affinityGroups = {};
+    (await axios({
+      method: 'get',
+      url: `${launcherConfig.hivedWebserviceUri}/v1/inspect/affinitygroups/`,
+      headers: launcherConfig.requestHeaders,
```
yqwang-ms: Why need to get all groups for this single job?
Should use /v1/inspect/affinitygroups/affinityGroupName

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -288,6 +301,26 @@ const convertFrameworkDetail = async (framework) => {
       ),
     };
   }
+
+  if (launcherConfig.enabledHived) {
+    const affinityGroups = {};
+    (await axios({
+      method: 'get',
```
yqwang-ms: If this get http call failed, log it, and set lazyPreemption to unknown, instead of lazyPreemption=false or =true.

yqwang-ms: Currently, null lazyPreemptionStatus has 2 meaning, one is lazyPreemption=false, another is lazyPreemption=unknown. I meant that, you need to differenate it, so you may need to introduce another field to task status.

---
In reply to: [356571019](https://github.com/microsoft/pai/pull/4007#discussion_r356571019) [](ancestors = 356571019)

PR #4019, opened 2019-12-16
[Prometheus ] Fix prometheus high io issue

Binyang2014:
Prometheus cause high io usage because some folders owner is root and permission is 755. But Prometheus default user is nobody:nogroup.
Prometheus service can not change the previous folder, and stuck at compacting logs. So high disk io occur.
Fix this issue by change the folder owner to nobody:nogroup

### DIFF
--- a/src/prometheus/deploy/prometheus-deployment.yaml.template
+++ b/src/prometheus/deploy/prometheus-deployment.yaml.template
@@ -42,11 +42,13 @@ spec:
       hostNetwork: true
       initContainers:
       - name: init
-        image: bash:4
+        image: busybox:1.31.1
+        securityContext:
+          runAsUser: 0
         volumeMounts:
         - name: prometheus-data
           mountPath: /prometheus-data
-        command: ["chmod", "777", "/prometheus-data"] # newly create dir have permission 755, which makes prometheus container unable to write
+        command: ["chown", "-R", "nobody:nogroup", "/prometheus-data"] # prometheus run as nobody:nogroup, refer to: https://github.com/prometheus/prometheus/pull/2859
       containers:
       - name: prometheus
         image: {{ cluster_cfg["cluster"]["docker-registry"]["prefix"] }}prometheus:{{ cluster_cfg["cluster"]["docker-registry"]["tag"] }}


src/prometheus/deploy/prometheus-deployment.yaml.template
```diff
@@ -42,11 +42,13 @@ spec:
...
       - name: init
-        image: bash:4
+        image: busybox:1.31.1
+        securityContext:
+          runAsUser: 0
         volumeMounts:
         - name: prometheus-data
           mountPath: /prometheus-data
-        command: ["chmod", "777", "/prometheus-data"] # newly create dir have permission 755, which makes prometheus container unable to write
+        command: ["chown", "-R", "nobody:nogroup", "/prometheus-data"] # prometheus run as nobody:nogroup, refer to: https://github.com/prometheus/prometheus/pull/2859
```
fanyangCS: do you have the permission to run "chown -R"?

Binyang2014: Sure please refer to:
https://github.com/microsoft/pai/blob/125fc65fb4dcb167888db7584e5e198d13006fd6/src/prometheus/deploy/prometheus-deployment.yaml.template#L46-L47
Init container will run as root user, and I tested in INT bed

fanyangCS: what if the path exists and it is readonly to the owner itself?

Binyang2014: If so, initContainer will failed, and k8s will try to restart the Prometheus according to restartPolicy.
Refer to: https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#detailed-behavior. Service will not start at the point.

fanyangCS: how about just add a write permission by default?

Binyang2014: It will also give Prometheus permission to modify previous data. And the problem will also solved. But 777 permission seems too general, and will leave prometheus data folder with files which own by different users (root and nobody).

PR #4032, opened 2019-12-18
Enable https for k8s-dashboard

Binyang2014:
After this change, admin need to use kube-config or k8s service-account token to login dashboard.
And we need to make sure we use `https` in layout.yaml
```yaml
kubernetes:
  api-servers-url: https://ip:6443
  dashboard-url: https://ip:8443
```


deployment/clusterObjectModel/test/data/configuration-template-generate/layout.yaml
```diff
@@ -19,7 +19,7 @@
 #TODO
 kubernetes:
   api-servers-url: http://10.0.0.1:8080
-  dashboard-host: 10.0.0.1
```
ydye: dashboard-ip-address?

Binyang2014: dashboard-url already used in current layout.yaml. Please refer to https://github.com/microsoft/pai/blob/3adbcddada8fa95844e065843f4ed3009edac967/deployment/quick-start/layout.yaml.template#L23-L24

I'm not sure why there is a field call `dashboard-host`, this filed is not documented

src/k8s-dashboard/deploy/start.sh.template
```diff
@@ -20,8 +20,15 @@
 pushd $(dirname "$0") > /dev/null
 
 {% if cluster_cfg['cluster']['common']['k8s-rbac'] != 'false' %}
+mkdir -p certs
+cp {{ cluster_cfg["pylon"]["ssl"]["crt_path"] }} certs/tls.cert
```
ydye: Do we need this key and cert? Because I found you added the option ```auto-generate-certificates``` and it seems dashboard will generated a selfsign cert?

Binyang2014: To access k8s-dashboard directly without kubectl proxy, we need set certs file. And `auto-generate-certificates` only used as a fallback.

According to https://github.com/kubernetes/dashboard/blob/master/docs/user/installation.md#recommended-setup
```
Custom certificates have to be stored in a secret named kubernetes-dashboard-certs in the 
same namespace as Kubernetes Dashboard. Assuming that you have tls.crt and tls.key files 
stored under $HOME/certs directory, you should create secret with contents of these files:

kubectl create secret generic kubernetes-dashboard-certs --from-file=$HOME/certs -n 
kubernetes-dashboard

For Dashboard to pickup the certificates, you must pass arguments --tls-cert-file=/tls.crt and
 --tls-key-file=/tls.key to the container.
```
k8s dashboard will read the certs under /certs folder

src/k8s-dashboard/deploy/k8s-dashboard.yaml.template
```diff
@@ -38,11 +38,14 @@ spec:
...
       - name: kubernetes-dashboard-openpai
         image: k8s.gcr.io/kubernetes-dashboard-amd64:v1.10.1
-        command:
-          - /dashboard
-          - --insecure-bind-address=0.0.0.0
         args:
-          - --insecure-port=9090
+          - --auto-generate-certificates
+          - --port={{ cluster_cfg['k8s-dashboard']['dashboard-port'] }}
+          - --tls-cert-file=/tls.crt
```
ydye: If set tls cert path as the value ```/tls.crt```, dashboard will find it in ```/certs/tls.crt```? I'm very confused about this, can you remove the option ```--auto-generate-certificates``` and try again. I suspect the crt and key is generated by k8s-dashboard
'

PR #4036, opened 2019-12-19
[Cluster Object Model] Generate necessary service config based on cluster type

ydye:
Design 3 parse type: yarn, k8s, common.
If the cluster-type is yarn, the configuration of yarn and common will be generated.
If the cluster-type is k8s, the configuration of k8s and common will be genereated.


ydye:
> #3437 uses `config/service.yaml` and `config/service.k8s.yaml` to differ different cluster types.
> If you add a new field in `config/service.yaml`, you need to revert changes in #3437 first.
> 
> The name `openpai_parser_type` in `config/service.yaml` is also confusing, could you use another name like `service_type`?

I wil change it in anther PR

ydye:
> #3437 uses `config/service.yaml` and `config/service.k8s.yaml` to differ different cluster types.
> If you add a new field in `config/service.yaml`, you need to revert changes in #3437 first.
> 
> The name `openpai_parser_type` in `config/service.yaml` is also confusing, could you use another name like `service_type`?

#4038 Please review this PR first.

src/rest-server/deploy/stop.sh
```diff
@@ -27,7 +27,7 @@ if kubectl get configmap | grep -q "auth-configuration"; then
     kubectl delete configmap auth-configuration || exit $?
 fi
 
-if kubectl get configmap | grep -q "job-exit-spec-configuration"; then
+if kubectl get configmap | grep -q " job-exit-spec-configuration"; then
```
abuccts: why add a space here?

ydye: k8s-job-exit-spec-configuration will detected by this grep command

ydye:
> `openpai_parser_type` is confusing, maybe change to another name, e.g. `service-type`?

Done

PR #4037, opened 2019-12-19
[Rest Server] kubernetes pods/nodes list api

sunqinzheng:
# Changes
## Kubernetes request model
- All components now use the axios client generated by `src/model/kubernetes.js`
- Use k8s official nodejs client lib to initialize the kubernetes config
- Developer can use a single `.kubeconfig` file to connect the remote cluster
  - Pass the file path to `K8S_KUBECONFIG_PATH` environment variable
## Kubernetes pods/nodes list api
- Remove old kubernetes proxy
- Add two new apis to fetch the nodes/pods list
  - Admin permission neeeded
- Documents and test cases
## Web portal service page
- Rewrite the fetch(ajax) function to improve performance
## Miscs
- Hived request bug fix (https://github.com/microsoft/pai/pull/4037/commits/ec2be715339a28174543173b74b5d2d62a67460c)


src/rest-server/docs/swagger.yaml
```diff
@@ -1680,6 +1682,54 @@ paths:
...
+      security:
+      - bearerAuth: []
+      responses:
+        200:
+          description: Succeeded
+          content:
+            application/json:
+              example:
+              - Please refer to Kubernetes API doc
+              - https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#list-node-v1-core
```
abuccts: use description instead?
```yaml
schema:
  type: object
  description: |
    line1
    line2
```

sunqinzheng: I tried the description field, but I found that the description will be hided by redoc if schema's type is `object`.
图片

And if the type is not specified, the response's description will be rendered in a collapsed div and sample response will be null.
图片

I think it may be more straightfoward if the hints are in example field.
图片

abuccts: well, seems a redoc bug that does not render `schema.description`, SwaggerUI works fine

Binyang2014:
Storage part looks fine to me

src/rest-server/src/config/kubernetes.js
```diff
@@ -41,20 +51,32 @@ if (process.env.RBAC_IN_CLUSTER === 'false') {
...
-    apiserverConfig.ca = readFileSync(caPath, 'utf8');
+  if (K8S_APISERVER_CA_FILE) {
+    k8s.Config.SERVICEACCOUNT_CA_PATH = K8S_APISERVER_CA_FILE;
+  }
+  if (K8S_APISERVER_TOKEN_FILE) {
     // Will be a string since http header can only receive a string.
-    apiserverConfig.token = readFileSync(tokenPath, 'utf8');
+    k8s.Config.SERVICEACCOUNT_TOKEN_PATH = K8S_APISERVER_TOKEN_FILE;
+  }
+  try {
```
ydye: I think in a pod, you should load the cluster config in the recommanded way. https://github.com/kubernetes-client/javascript/blob/d262aff9acc43efda602142a587baa0183c4970a/examples/in-cluster.js

src/rest-server/src/config/kubernetes.js
```diff
@@ -41,20 +51,32 @@ if (process.env.RBAC_IN_CLUSTER === 'false') {
...
+    const kc = new k8s.KubeConfig();
+    if (K8S_KUBECONFIG_PATH) {
+      kc.loadFromFile(K8S_KUBECONFIG_PATH);
```
ydye: In webportal, I think there won't be a kubeconfig. So adding the env here is for local test?

src/rest-server/src/config/kubernetes.js
```diff
@@ -41,20 +51,32 @@ if (process.env.RBAC_IN_CLUSTER === 'false') {
...
+    } else {
+      kc.loadFromDefault();
```
ydye: I think it should be loadFromCluster

src/rest-server/src/config/kubernetes.js
```diff
@@ -41,20 +51,32 @@ if (process.env.RBAC_IN_CLUSTER === 'false') {
...
+    }
+    const cluster = kc.getCurrentCluster();
+    const user = kc.getCurrentUser();
+    apiserverConfig.uri = cluster.server;
+    if (cluster.caFile) {
+      apiserverConfig.ca = readFileSync(cluster.caFile);
```
ydye: What's relative between this line and the line 77

src/rest-server/src/models/token.js
```diff
@@ -17,16 +17,15 @@
...
 const k8sSecret = require('@pai/models/k8s-secret');
+const k8sModel = require('@pai/models/kubernetes');
 
 const namespace = process.env.PAI_TOKEN_NAMESPACE || 'pai-user-token';
 
 // create namespace if not exists
 if (process.env.NODE_ENV !== 'test') {
-  logger.info(`Create token secret namespace (${namespace}) if not exist`);
-  k8sSecret.createNamespace(namespace);
+  k8sModel.createNamespace(namespace);
```
ydye: What's the behavior, if the namespace exists?

ydye:
BTW, do we need so complicated logic to init a config? I think the function loadFromYaml is enough

src/rest-server/src/config/kubernetes.js
```diff
@@ -41,20 +51,32 @@ if (process.env.RBAC_IN_CLUSTER === 'false') {
...
-  const tokenPath = K8S_APISERVER_TOKEN_FILE || path.join(root, 'token');
-  const host = process.env.KUBERNETES_SERVICE_HOST;
-  const port = process.env.KUBERNETES_SERVICE_PORT;
-  apiserverConfig.uri = `https://${host}:${port}`;
-
-  try {
-    // Will be a buffer since SSL context can receive a buffer.
-    apiserverConfig.ca = readFileSync(caPath, 'utf8');
+  if (K8S_APISERVER_CA_FILE) {
+    k8s.Config.SERVICEACCOUNT_CA_PATH = K8S_APISERVER_CA_FILE;
```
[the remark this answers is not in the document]
sunqinzheng: https://github.com/kubernetes-client/javascript/blob/da9f3d872bdebaebf37fe22f089b2a1c655fe591/src/config.ts#L433

src/rest-server/src/config/kubernetes.js
sunqinzheng: https://github.com/kubernetes-client/javascript/blob/da9f3d872bdebaebf37fe22f089b2a1c655fe591/src/config.ts#L317

src/rest-server/src/config/kubernetes.js
sunqinzheng: Yes

src/rest-server/src/config/kubernetes.js
sunqinzheng: https://github.com/kubernetes-client/javascript/blob/da9f3d872bdebaebf37fe22f089b2a1c655fe591/src/config.ts#L317

src/rest-server/src/config/kubernetes.js
sunqinzheng: Sorry, I forgot to remove those redundant lines. Thanks.

src/rest-server/src/config/kubernetes.js
sunqinzheng: Yes

sunqinzheng:
> 
> 
> BTW, do we need so complicated logic to init a config? I think the function loadFromYaml is enough

1. Sorry, I can not find the `loadFromYaml` function. Could you provide a link?

2. I think the current logic is not that complicated. Only two official libraries' method are called. And other changes are used to retrieve information from kubeConfig structure (referenced from godaddy's kubernetes client's implementation) and to be compatible with old env vars

PR #4040, opened 2019-12-19
[Hivedscheduler] Create every VC for every vc

ydye:


yqwang-ms:
Thanks @ydye, we incorporated this into https://github.com/microsoft/pai/pull/4041

PR #4043, opened 2019-12-20
[runtime] fix official docker image check

Binyang2014:


src/kube-runtime/src/init.d/image_checker.py
```diff
@@ -37,6 +37,13 @@ def _is_docker_hub_uri(uri):
     return False
 
 
+def _get_docker_repository_name(image_name):
+    pathes = image_name.split("/")
```
abuccts: paths

src/kube-runtime/test/test_image_checker.py
```diff
@@ -94,6 +94,21 @@ def test_docker_hub_uri(self):
         for uri in invalid_docker_hub_uris:
             self.assertFalse(image_checker._is_docker_hub_uri(uri))
 
+    def test_get_docker_repository_name(self):
+        official_image_names = [
+            "ubuntu", "python:latest", "golang:1.12.6-alpine"
+        ]
+        third_party_image_name = [
```
abuccts: names

src/kube-runtime/src/init.d/image_checker.py
Binyang2014: fixed

src/kube-runtime/test/test_image_checker.py
Binyang2014: fixed

PR #4047, opened 2019-12-20
[ErrorSpec] make pattern case insensitive

Binyang2014:


src/k8s-job-exit-spec/config/k8s-job-exit-spec.yaml
```diff
@@ -811,7 +811,7 @@ spec:
   pattern:
     runtimeContainerPatterns:
     - userLogRegex: "(?msi)tensorflow.*ResourceExhaustedError.*OOM.*"
-    - userLogRegex: "ran out of memory trying to allocate"
+    - userLogRegex: "(?i)ran out of memory trying to allocate"
```
yqwang-ms: Use (?msi)?

Binyang2014: Maybe (?i) is enough? According to https://www.regular-expressions.info/modifiers.html and https://stackoverflow.com/questions/27680097/what-does-ms-in-regex-mean/27680233 (?s) is use to make `.` match the line breaks. And (?m) makes `^` and `$` match the beginning and end of a line

yqwang-ms: Depends on you. haha

Issue #4014, opened 2019-12-13
P1 Candidate - a failed runtime plugin shouldn't fail the entire runtime


fanyangCS:
Please avoid apt-get or any external resource fetching, if possible.

Binyang2014:
How about use init container to prepare binary package for user and don't install any package in user container. (For ssh, I think we can do this by building go sserver: https://github.com/gliderlabs/ssh or other ways)  This will separate runtime plugin and user commands and avoid wired issues.

Binyang2014:
For `runtime plugin shouldn't fail entire runtime`. Maybe we can change the runtime protocol and add failurePolicy for plugins. The failurePolicy can be ignore or fail. The default value is Fail
```yaml
extras:
  com.microsoft.pai.runtimeplugin:
    - plugin: ssh
      parameters:
        jobssh: boolean
      failurePolicy: Ignore/Fail
```

PR #4048, opened 2019-12-20
[Runtime] add failurePolicy to runtime plugin

Binyang2014:
Refer to issue #4014 add failurePolicy for runtime plugin.
User can make runtime plugin doesn't fail the entire runtime by setting `failurePolicy: ignore`

The default failurePolicy is `fail`


src/kube-runtime/src/plugins/plugin_utils.py
```diff
@@ -16,36 +16,41 @@
...
-
-    return [parameters, args.pre_script, args.post_script]
-
-def inject_commands(commands, script):
-    if commands is not None and len(commands) > 0:
-        new_commands = [x+"\n" for x in commands]
-        with open(script, 'a+') as f:
-            f.writelines(new_commands)
-
+    plugin_config = yaml.load(args.plugin_config, Loader=yaml.SafeLoader)
```
abuccts: unify to one of `safe_load` and `load(Loader=yaml.SafeLoader)`?

src/kube-runtime/src/plugins/plugin_utils.py
```diff
@@ -16,36 +16,41 @@
...
+from common.utils import init_logger
+
+LOGGER = logging.getLogger(__name__)
+
+
+class PluginHelper:  #pylint: disable=too-few-public-methods
+    def __init__(self, plugin_config: dict):
+        self._parameters = plugin_config.get("parameters", "")
+        self._plugin_name = plugin_config.get("plugin", "")
+        self._failure_policy = plugin_config.get("failurePolicy", "fail")
```
abuccts: is there a validation for this field?

src/kube-runtime/src/plugins/plugin_utils.py
Binyang2014: fixed

src/kube-runtime/src/plugins/plugin_utils.py
Binyang2014: If this field is not `ignore`, we will treat it as `fail`.  Refer to https://github.com/microsoft/pai/pull/4048/files#diff-33c702285a3718b9aae88340cbe60538R38. Not validate it here, just trust rest-server will pass valid config to runtime.
Invalid config may fail the runtime and job.

PR #4055, opened 2019-12-23
remove grafana chart

Binyang2014:
updated dashboard, remove `total nodes/worker nodes/allocated gpu/free gpus/allocated gpus`
image

Add dashboard back & hide `view all` for k8s VC
image


fanyangCS:
Since the dashboard is only visible to admin, maybe you should put in under “Administration“.

Issue #3935, opened 2019-11-29
P0 - Job Name, UID and Description


yqwang-ms:
**If RestServer uses Job UID instead of Job Name as Job Key to serve query:**
**Pros:**
1. Job Name can be very long (store in annotation as Job Description)
2. History job (of the same name) can be in job list and job detail page: https://github.com/microsoft/pai/issues/3845

**Cons:**
1. RestServer may need to cache the mapping from uid to name, so that it can query APIServer efficently by naming, otherwise, it has to iterate all frameworks. (Or use UID as framework label)
2. Maybe many changes in backend

### Proposal-1:
Job **Name** to submit **idempotently**, 
Job **UID** to query **uniquely**,
Job **Description** to attach metadata **arbitrarily**.
UID generated by K8S

1. Add a new field in PAI Job Spec called `description`, which can be any string in any reasonable length (<10k), and RestServer stores it into k8s framework annotation.
2. If user specified job name (he wants idempotence), then RestServer uses this job **name** as k8s framework name **to submit**, but RestServer **still** uses k8s framework **uid** as this job key **serve query** (may still can use name to serve active job query).

3. If user did not specify job name (he does not care idempotence, like Aether), then RestServer uses **empty name** as k8s framework name **to submit** (k8s will auto generate it if `metadata.generateName` is set) , and RestServer uses k8s framework **uid** as this job key **serve query** (may still can use name to serve active job query).

**Example:**
```
  POST /jobs/  
      --> If Request's JOB_NAME is not empty
         --> Response includes its JOB_NAME and JOB_UID (Always the same if it is not GC)
      --> Else
         --> Response includes its JOB_NAME and JOB_UID (Always different)

  GET /jobs/{JOB_UID}
      --> Can query all jobs, both in history and active
      --> Response includes its JOB_NAME (Always the same)
      --> Useful for query both in history and active, such as webportal, etc
```
active jobs: jobs in k8s apiserver
history jobs: jobs only in elasticsearch

**TBD:**
```
  POST /activejobs/  
  GET /alljobs/{JOB_UID}
  GET /activejobs/{JOB_NAME}
      --> Only can query active jobs, i.e. are not GC to history
      --> Response includes its JOB_UID (Always the same if it is not GC)
      --> Useful for just check existence for stateless job submitter:
            If !(GET JOB_NAME)
               Prepare Externals: Cleanup previous intermediate data 
               POST JOB_NAME
            WATCH JOB_NAME
```

### Proposal-2:
[...]

PR #4069, opened 2019-12-25
[Rest Server] Update job name encoding method

abuccts:
Update job name encoding method, use md5 hash instead.
Query job by k8s label selector.

Resolve job name related issues in #3935.


yqwang-ms:
why not directly use md5 name to lookup, it is more efficient than label?
And label may be replaced by annotation in future, let's avoid to use it to look up.

yqwang-ms:
And we haven't even release pure k8s image, compatibility is not so important than doing the right thing.

PR #4086, opened 2019-12-30
[Storage-Manager] move apt-get to docker image

Binyang2014:
Tested in intx. This config works fine for samba and nfs

### DIFF
--- a/src/storage-manager/build/storage-manager.dockerfile
+++ b/src/storage-manager/build/storage-manager.dockerfile
@@ -31,7 +31,7 @@ RUN apt-get -y update && \
       libpam-winbind \
       libnss-winbind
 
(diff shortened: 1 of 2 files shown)


src/storage-manager/build/storage-manager.dockerfile
```diff
@@ -26,7 +26,16 @@ RUN apt-get -y update && \
       nfs-common \
       netbase \
       nfs-kernel-server \
-      kmod
+      kmod \
+      winbind \
+      libpam-winbind \
+      libnss-winbind
+
+# Use defualt config for Kerberos
```
abuccts: typo

Binyang2014: fixed

PR #4087, opened 2019-12-30
Remove the obsolete step of update docker tag

liuc-shaiic:
There's no "docker tag" in services-configuration.yaml, and this step should be removed.


abuccts:
Hi @liuc-shaiic, thanks for your contribution, but there is a `tag` field for Docker registry:
https://github.com/microsoft/pai/blob/120179950934692bd304ce5f38a54f0ecba673eb/examples/cluster-configuration/services-configuration.yaml#L45-L60

Maybe you can update `docker-tag` to `docker-registry.tag`.

PR #4088, opened 2019-12-31
[Storage Manager] persist storage logs 

Binyang2014:


src/storage-manager/deploy/storage-manager.yaml.template
```diff
@@ -68,6 +68,8 @@ spec:
         - name: host-modules
           mountPath: /lib/modules
```
abuccts: what is `/ilb/modules` used for?

src/storage-manager/deploy/conf/smb.conf.template
```diff
@@ -49,7 +49,8 @@
     encrypt passwords = Yes
     restrict anonymous = 2
     log level = 2
-    log file = /var/log/samba/samba.log
+    log file = /var/log/storage/samba.log
+    max log size = 10000
```
abuccts: 10MB is too small

src/storage-manager/deploy/conf/smb.conf.template
```diff
@@ -49,7 +49,8 @@
     encrypt passwords = Yes
     restrict anonymous = 2
     log level = 2
-    log file = /var/log/samba/samba.log
+    log file = /var/log/storage/samba.log
```
abuccts:
```suggestion
    log file = /var/log/storage/%m.log
```

src/storage-manager/deploy/storage-manager.yaml.template
Binyang2014: If think it used for:
https://github.com/microsoft/pai/blob/120179950934692bd304ce5f38a54f0ecba673eb/src/storage-manager/deploy/scripts/entrypoint.sh#L46-L47

src/storage-manager/deploy/conf/smb.conf.template
Binyang2014: Change to 100MB fixed

src/storage-manager/deploy/conf/smb.conf.template
Binyang2014: fixed

src/storage-manager/deploy/storage-manager.yaml.template
Binyang2014: Refer to https://stackoverflow.com/questions/33013539/docker-loading-kernel-modules

src/storage-manager/deploy/conf/smb.conf.template
abuccts: could you increase to 2GB?

abuccts: and change `log level` to 3

Binyang2014: fixed

PR #4093, opened 2020-01-02
Add support for AMD GPU

### LINKED ISSUE #4090: support different GPU vendor in one cluster
We decide to support this feature only when HiveD is enabled.

* We assume one VC only includes one GPU vendor (AMD or NVIDIA).
* When rest server is preparing the env, it will prepare both for AMD and NVIDIA
* k8s cluster will deploy both nvidia plugin and amd plugin, only one plugin will function when deploying to a GPU node

abuccts:
Add support for AMD GPU:
* add AMD GPU device plugin
* support AMD GPU job submission in hived scheduler
* add AMD ROCM deep learning examples

Fixes #4090.

### DIFF
--- a/src/rest-server/src/models/v2/job/k8s.js
+++ b/src/rest-server/src/models/v2/job/k8s.js
@@ -578,7 +578,7 @@ const generateTaskRole = (frameworkName, taskRole, labels, config, storageConfig
         },
       },
       {
-        name: 'HIP_VISIBLE_DEVICES',
+        name: 'PAI_AMD_VISIBLE_DEVICES',
         valueFrom: {
           fieldRef: {
             fieldPath: `metadata.annotations['hivedscheduler.microsoft.com/pod-gpu-isolation']`,
--- a/src/rest-server/src/models/v2/job/k8s.js
+++ b/src/rest-server/src/models/v2/job/k8s.js
@@ -576,7 +576,16 @@ const generateTaskRole = (frameworkName, taskRole, labels, config, storageConfig
             fieldPath: `metadata.annotations['hivedscheduler.microsoft.com/pod-gpu-isolation']`,
           },
         },
-      });
+      },
+      {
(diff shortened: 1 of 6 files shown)


src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -576,7 +576,16 @@ const generateTaskRole = (frameworkName, taskRole, labels, config, storageConfig
             fieldPath: `metadata.annotations['hivedscheduler.microsoft.com/pod-gpu-isolation']`,
           },
         },
-      });
+      },
+      {
+        name: 'PAI_AMD_VISIBLE_DEVICES',
```
yqwang-ms: Where is the code to consume PAI_AMD_VISIBLE_DEVICES? I.e. your AMD runtime code

abuccts: pls check [here](https://github.com/abuccts/rocm-container-runtime/blob/master/rocm-container-runtime), since drivers and docker are out of the scope of pai, it's not suitable to maintain such a runtime which may have driver/docker issue

yqwang-ms: OK we can discuss this in scrum.
BTW you also need to update PAI_AMD_VISIBLE_DEVICES in your repo

fanyangCS: @abuccts, how do you deploy the AMD runtime to the bed?

abuccts: > BTW you also need to update PAI_AMD_VISIBLE_DEVICES in your repo

I have made it configurable, the runtime needs a config file like nvidia-container-runtime

> how do you deploy the AMD runtime to the bed?

just copy the binary and config to `/usr/bin` and `/etc`, we could use ansible playbook to copy those two files like installing nvidia runtime

PR #4106, opened 2020-01-06
[Rest Server] Remove job name length limitation

### LINKED ISSUE #4101: lift the length limitation of job name
In v0.16, we assume job name cannot be longer that 64 char. we decide the remove this constraints in v0.17.
In the system, a job name actually consists of user name and job name, separated by a "\~": username\~realjobname. In the new design, we do the following:
1. keep username in a label of a pod, this preserves the compatibility of a job REST API that query job by username.
2. keep realjobname in pod annotation, which doesn't have length limitation
3. hash jobname using md5, and use the hashed md5 string as framework's name, which cannot be longer than 63.

abuccts:
Move jobname from framework labels to annotations.

Fixes #4101.

### DIFF
--- a/src/rest-server/src/models/v2/job/k8s.js
+++ b/src/rest-server/src/models/v2/job/k8s.js
@@ -75,9 +75,9 @@ const encodeName = (name) => {
   }
 };
 
-const decodeName = (name, labels) => {
-  if (labels && labels.jobName) {
-    return labels.jobName;
+const decodeName = (name, annotations) => {
+  if (annotations && annotations.jobName) {
+    return annotations.jobName;
   } else {
     // framework name has not been encoded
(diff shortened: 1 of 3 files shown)


src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -594,8 +595,12 @@ const generateFrameworkDescription = (frameworkName, virtualCluster, config, raw
     kind: 'Framework',
     metadata: {
       name: encodeName(frameworkName),
-      labels: frameworkLabels,
+      labels: {
+        userName: frameworkLabels.userName,
```
yqwang-ms: So rename var frameworkLabels, or just remove it

PR #4132, opened 2020-01-13
Upgrade deprecated k8s apiversion

abuccts:
Upgrade deprecated k8s apiversion "extensions/v1beta1" to "apps/v1".
Reference: https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

### DIFF
--- a/src/watchdog/test/data/no_condtion_pod.json
+++ b/src/watchdog/test/data/no_condtion_pod.json
@@ -18,7 +18,7 @@
                 },
                 "ownerReferences": [
                     {
-                        "apiVersion": "extensions/v1beta1",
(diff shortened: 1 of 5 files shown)


src/watchdog/test/data/no_condtion_pod.json
```diff
@@ -18,7 +18,7 @@
                 },
                 "ownerReferences": [
                     {
-                        "apiVersion": "extensions/v1beta1",
```
ydye: It seems the spec.selector.matchLabels is missing in this ds.

abuccts: it's `ownerReferences`, not daemon set spec

PR #4133, opened 2020-01-13
[Runtime, Rest-server] Pass secret from frontend to runtime

Binyang2014:
Pass secret from frontend to runtime.
**Changes:**
- Create k8s secret obj and mount it for init container
- User command will not concatenate with `&&`, change to concatenate with `\n`. Runtime will write user command to a file.

**Tested:**
- job w/ & w/o secret
- job w/ & w/o private docker auth

### DIFF
--- a/src/kube-runtime/src/init
+++ b/src/kube-runtime/src/init
@@ -89,7 +89,7 @@ PAI_INIT_DIR=${PAI_WORK_DIR}/init.d
 PAI_RUNTIME_DIR=${PAI_WORK_DIR}/runtime.d
 
 PAI_LOG_DIR=${PAI_WORK_DIR}/logs/${FC_POD_UID}
-PAI_SERCRET_FILE=${PAI_WORK_DIR}/secret
+PAI_SERCRET_DIR=${PAI_WORK_DIR}/secrets
 
 # Move previous logs to another folder. Notice: exclude for init.log, new log will append to previous log file
 LOG_FILES=$(find $PAI_LOG_DIR -maxdepth 1 -type f ! -name "init.log")
@@ -177,7 +177,7 @@ python ${PAI_INIT_DIR}/image_checker.py ${PAI_RUNTIME_DIR}/job_config.yaml
 # write user commands to user.sh
 # priority=100
 CHILD_PROCESS="RENDER_USER_COMMAND"
-python ${PAI_INIT_DIR}/user_command_render.py ${USER_CMD} ${PAI_SERCRET_FILE} ${PAI_RUNTIME_DIR}/user.sh
+python ${PAI_INIT_DIR}/user_command_renderer.py ${PAI_SERCRET_DIR}/secrets.yaml ${PAI_RUNTIME_DIR}/user.sh
 
 # for debug
 echo -e "finished entry\nuser.sh has:"
--- a/src/kube-runtime/src/init
+++ b/src/kube-runtime/src/init
@@ -89,6 +89,7 @@ PAI_INIT_DIR=${PAI_WORK_DIR}/init.d
 PAI_RUNTIME_DIR=${PAI_WORK_DIR}/runtime.d
 
 PAI_LOG_DIR=${PAI_WORK_DIR}/logs/${FC_POD_UID}
+PAI_SERCRET_FILE=${PAI_WORK_DIR}/secret
 
 # Move previous logs to another folder. Notice: exclude for init.log, new log will append to previous log file
 LOG_FILES=$(find $PAI_LOG_DIR -maxdepth 1 -type f ! -name "init.log")
@@ -175,7 +176,8 @@ python ${PAI_INIT_DIR}/image_checker.py ${PAI_RUNTIME_DIR}/job_config.yaml
 
 # write user commands to user.sh
 # priority=100
-echo "${USER_CMD}" >> ${PAI_RUNTIME_DIR}/user.sh
+CHILD_PROCESS="RENDER_USER_COMMAND"
+python ${PAI_INIT_DIR}/user_command_render.py ${USER_CMD} ${PAI_SERCRET_FILE} ${PAI_RUNTIME_DIR}/user.sh
 
 # for debug
 echo -e "finished entry\nuser.sh has:"
--- a/src/kube-runtime/src/init
+++ b/src/kube-runtime/src/init
@@ -180,7 +180,5 @@ CHILD_PROCESS="RENDER_USER_COMMAND"
 python ${PAI_INIT_DIR}/user_command_renderer.py ${PAI_SECRET_DIR}/secrets.yaml ${PAI_RUNTIME_DIR}/user.sh
 
 # for debug
-echo -e "finished entry\nuser.sh has:"
-cat ${PAI_RUNTIME_DIR}/user.sh
 echo -e "\nruntime_env.sh has:"
 cat ${PAI_RUNTIME_DIR}/runtime_env.sh
--- a/src/kube-runtime/src/init
+++ b/src/kube-runtime/src/init
@@ -89,7 +89,7 @@ PAI_INIT_DIR=${PAI_WORK_DIR}/init.d
 PAI_RUNTIME_DIR=${PAI_WORK_DIR}/runtime.d
 
 PAI_LOG_DIR=${PAI_WORK_DIR}/logs/${FC_POD_UID}
-PAI_SERCRET_DIR=${PAI_WORK_DIR}/secrets
+PAI_SECRET_DIR=${PAI_WORK_DIR}/secrets
 
 # Move previous logs to another folder. Notice: exclude for init.log, new log will append to previous log file
 LOG_FILES=$(find $PAI_LOG_DIR -maxdepth 1 -type f ! -name "init.log")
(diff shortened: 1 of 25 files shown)


src/kube-runtime/src/init
```diff
@@ -89,6 +89,7 @@ PAI_INIT_DIR=${PAI_WORK_DIR}/init.d
 PAI_RUNTIME_DIR=${PAI_WORK_DIR}/runtime.d
 
 PAI_LOG_DIR=${PAI_WORK_DIR}/logs/${FC_POD_UID}
+PAI_SERCRET_DIR=${PAI_WORK_DIR}/secrets
```
abuccts: typo

src/kube-runtime/src/init
```diff
@@ -175,7 +176,8 @@ python ${PAI_INIT_DIR}/image_checker.py ${PAI_RUNTIME_DIR}/job_config.yaml
 
 # write user commands to user.sh
 # priority=100
-echo "${USER_CMD}" >> ${PAI_RUNTIME_DIR}/user.sh
+CHILD_PROCESS="RENDER_USER_COMMAND"
+python ${PAI_INIT_DIR}/user_command_renderer.py ${PAI_SERCRET_DIR}/secrets.yaml ${PAI_RUNTIME_DIR}/user.sh
```
abuccts: typo

src/kube-runtime/src/init.d/user_command_renderer.py
```diff
@@ -0,0 +1,90 @@
...
+    converted_obj = {}
+    if isinstance(obj, list):
+        for i, value in enumerate(obj):
+            converted_obj[str(i)] = value
+    elif isinstance(obj, dict):
+        for key, value in obj.items():
+            converted_obj[key] = _convert_to_dict(value)
+    else:
+        converted_obj = obj
+    return converted_obj
```
abuccts: what is this function for

src/kube-runtime/src/init.d/user_command_renderer.py
```diff
@@ -0,0 +1,90 @@
...
+    if not os.path.isfile(args.secret_file):
+        secrets = None
+    else:
+        with open(args.secret_file) as f:
+            secrets = yaml.safe_load(f.read())
+
+    user_command = os.getenv("USER_CMD")
+    rendered_user_command = _render_user_command(user_command, secrets)
+    _output_user_command(rendered_user_command, args.output_file)
+    logging.info("User command already rendered and output to %s",
```
abuccts: outputted

src/kube-runtime/test/render_test_secrets.yaml
```diff
@@ -0,0 +1,9 @@
+time: 10
+---
+time:
+  - 10
+  - 20
+---
+time:
+  time1: 10
+  time2: 20
```
abuccts: could you add some complex cases, like nested objects and arrays

src/kube-runtime/test/test_render_user_command.py
```diff
@@ -0,0 +1,66 @@
...
+class TestUserCommandRender(unittest.TestCase):
+    def setUp(self):
+        try:
+            os.chdir(PACKAGE_DIRECTORY_COM)
+        except Exception:  #pylint: disable=broad-except
+            pass
+
+    def test_user_command_render(self):
+        user_commands = [
+            "sleep <% $secrets.time %>", "sleep <% $secrets.time[0] %>",
```
abuccts: what about `$secrets.time.0`

src/kube-runtime/test/test_render_user_command.py
```diff
@@ -0,0 +1,66 @@
...
+            user_commands[0], secrets[0])
+        self.assertEqual(res, "sleep 10")
+
+        res = user_command_renderer._render_user_command(
+            user_commands[1], secrets[1])
+        self.assertEqual(res, "sleep 10")
+
+        res = user_command_renderer._render_user_command(
+            user_commands[2], secrets[2])
+        self.assertEqual(res, "sleep 10")
```
abuccts: could you use a for loop for this

src/rest-server/src/models/kubernetes.js
```diff
@@ -85,10 +97,50 @@ const getPods = async (options = {}) => {
...
+    logger.warn('Failed to create secret, err is: ', err);
+    throw err;
+  }
+};
+
+const deleteSecret = async (namespace, name) => {
+  const client = getClient();
+  const url = `/api/v1/namespaces/${namespace}/secrets/${name}`;
+  await client.delete(url);
+  logger.debug(`delete secret ${name} successfully`);
```
abuccts: do not need to add successful logs

src/rest-server/src/models/kubernetes.js
```diff
@@ -85,10 +97,50 @@ const getPods = async (options = {}) => {
...
+};
+
+const patchSecret = async (namespace, name, data) => {
+  const client = getClient();
+  const url = `/api/v1/namespaces/${namespace}/secrets/${name}`;
+  await client.patch(url, data, patchOption);
+  logger.debug(`Patch secret ${name} successfully`);
+};
```
abuccts: why do not catch errors in delete and patch but check errors in create

src/rest-server/src/models/kubernetes.js
```diff
@@ -85,10 +97,50 @@ const getPods = async (options = {}) => {
...
+      type: type,
+    });
+    if (response.status !== status('Created')) {
+      logger.warn('Failed to create secret');
+      throw createError(response.status, 'UnknownError', response.data.message);
+    }
+    logger.debug(`create secret ${name} successfully`);
+  } catch (err) {
+    logger.warn('Failed to create secret, err is: ', err);
+    throw err;
```
abuccts: pls check response body when axios response is not 20x, do not throw error directly here

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -723,75 +735,84 @@ const createSecret = async (frameworkName, auths) => {
...
-    throw createError(response.status, 'UnknownError', response.data.message);
+};
+
+const deleteDockerSecret = async (frameworkName) => {
+  try {
+    await k8sModel
+      .getClient()
+      .deleteSecret('default', `${encodeName(frameworkName)}-regcred`);
+  } catch (error) {
+    logger.warn('Failed to delete protocol secret', error);
```
abuccts: remove protocol

src/kube-runtime/src/init.d/user_command_renderer.py
Binyang2014: According to issue: https://github.com/defunkt/pystache/issues/196
pystache not support syntax such as `array.0`, we need to convert array to dict first, then we can access array item.

src/rest-server/src/models/kubernetes.js
Binyang2014: Fixed by using axios interceptors

src/rest-server/src/models/kubernetes.js
```diff
@@ -5,8 +5,25 @@ const axios = require('axios');
...
+const createError = require('@pai/utils/error');
 const logger = require('@pai/config/logger');
 
+const patchOption = {
+  headers: {
+    'Content-Type': 'application/merge-patch+json',
+  },
+};
+
+const rethrowResponseError = (error) => {
```
sunqinzheng: I think the error handler could be enabled to each k8s requests (could be added to `getClient` function).

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -868,7 +903,10 @@ const put = async (frameworkName, config, rawConfig) => {
   const auths = Object.values(config.prerequisites.dockerimage)
     .filter((dockerimage) => dockerimage.auth != null)
     .map((dockerimage) => dockerimage.auth);
-  auths.length && await createSecret(frameworkName, auths);
+  auths.length && await createDockerSecret(frameworkName, auths);
```
sunqinzheng: I think we need a more complete exception catch logic. We should store each k8s objects we created. And try to delete all created objects if error happened

src/rest-server/src/models/kubernetes.js
Binyang2014: As offline discuss, will leave GC to watch dog

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: As offline discuss, will leave GC to watch dog

src/rest-server/src/models/kubernetes.js
sunqinzheng: Ok, I was talking about another problem about axios.
Axios will throw errors with wrong stack trace and error message only conclude status code. And only this two field will be logged to console if the error is not handled.
I think we need to modify the error message and add some request info into the message (at least the request url). Another options is to log the failed request info into console and keep the original error object.

By the way, I think it is not a good idea to throw a express http error here.
The model file may be used by other models and utilities which may need the orignial axios error object with all information.

Binyang2014: OK, I have changed the error handling logic. Modify the error message and add some request info into the message. Don't throw express http error here

PR #4156, opened 2020-01-20
[Runtime] Add image checker in init container

Binyang2014:
modify:
1. support image with auth
2. support azure container registry


src/kube-runtime/test/test_image_checker.py
```diff
@@ -52,69 +61,145 @@ def wrapper(self, *args, **kwargs):
...
+        image_info = image_checker._get_normalized_image_info()
+        add_official_registry_v2_response(image_info,
+                                          options={"image_not_found": True})
+        self.assertFalse(image_checker.is_docker_image_accessible())
+
+    @prepare_image_check("docker_image_acr_registry.yaml")
+    @responses.activate
+    def test_acr_image(self):
+        image_checker = ImageChecker(self.job_config, self.secret)
+        image_info = image_checker._get_normalized_image_info()
```
abuccts: why not add the common lines to decorator

Binyang2014: fixed

---
In reply to: [368789410](https://github.com/microsoft/pai/pull/4156#discussion_r368789410) [](ancestors = 368789410)

PR #4157, opened 2020-01-20
[Rest Server] Refactor storage by leveraging persistent volumes

abuccts:
Refactor storage by leveraging kubernetes pv and pvc.

TODO:

- [x] backward compatibility
- [x] samba and azure storage tests
- [x] documents


src/rest-server/src/models/v2/group.js
```diff
@@ -100,26 +100,25 @@ const getGroupsVCs = async (grouplist) => {
...
+const getStorageConfigsWithGroupInfo = async (groupItems, filterDefault=false) => {
   let storageConfigs = new Set();
   for (const groupItem of groupItems) {
     if (groupItem.extension && groupItem.extension.acls) {
       if (groupItem.extension.acls.storageConfigs) {
-        storageConfigs = new Set([...storageConfigs, ...groupItem.extension.acls.storageConfigs]);
+        if (filterDefault) {
+          storageConfigs.add(groupItem.extension.acls.storageConfigs[0]);
+        } else {
+          storageConfigs = new Set([...storageConfigs, ...groupItem.extension.acls.storageConfigs]);
```
Binyang2014: Use 
```js
groupItem.extension.acls.storageConfigs.forEach(item=>storageConfigs.add(item))
```
Instead of create a set each loop?

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -573,6 +570,25 @@ const generateTaskRole = (frameworkName, taskRole, jobInfo, frameworkEnvList, co
       name: `${encodeName(frameworkName)}-regcred`,
     });
   }
+  // add storages
```
Binyang2014: what if `teamwise_storage` in runtime plugin? Can it work?

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -573,6 +570,25 @@ const generateTaskRole = (frameworkName, taskRole, jobInfo, frameworkEnvList, co
...
+  if ('extras' in config && config.extras.storages) {
+    for (let storage of config.extras.storages) {
```
Binyang2014: And what if storage name is missing. Will use default storage?

src/rest-server/docs/swagger.yaml
```diff
@@ -1317,165 +1317,31 @@ paths:
                   $ref: '#/components/responses/NoJobConfigError/content/application~1json/examples/NoJobConfigError'
         500:
           $ref: '#/components/responses/UnknownError'
-  /api/v2/storage/server/{storage}:
```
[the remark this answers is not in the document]
abuccts: I don't know neither, previous storage uses v2 and just kept the same

src/rest-server/docs/swagger.yaml
```diff
@@ -1317,165 +1317,31 @@ paths:
...
-        description: filter storage server with names, default name empty will be ignored
-        schema:
-          type: string
       responses:
         200:
           description: Succeeded
           content:
             application/json:
               schema:
-                type: array
```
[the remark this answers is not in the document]
abuccts: It's recommended to return object always, because array cannot be extended if you want to add more information in the future, here's [a discussion](https://softwareengineering.stackexchange.com/a/304638)

src/rest-server/src/models/v2/group.js
abuccts: It exists in previous code, will update

src/rest-server/src/models/v2/job/k8s.js
abuccts: if old config `teamwise_storage` is used, runtime plugin can mount the storage as long as the plugin hasn't been deleted
if new config is used, rest server will add pod spec and runtime plugin can be skipped

src/rest-server/docs/swagger.yaml
```diff
@@ -1871,52 +1618,59 @@ components:
...
+        volumeName:
           type: string
         type:
           type: string
+          enum:
+          - nfs
+          - samba
+          - azureFile
+          - azureBlob
+          - other
```
mydmdm: what is the difference of `other` and `unknown`

src/rest-server/docs/swagger.yaml
```diff
@@ -1871,52 +1618,59 @@ components:
...
-              mountPoint:
-                type: string
-              path:
-                type: string
-              server:
-                type: string
-              permission:
-                type: string
-      required: [name, default, servers, mountInfos]
+      required: [name, share, volumeName]
```
mydmdm: `type` is not `required`?

src/rest-server/src/models/v2/group.js
abuccts: updated

src/rest-server/src/models/v2/job/k8s.js
abuccts: I thought "default storage" is only used in old config `teamwise_storage`, and current backward compatibility has supported it. Do we still need "default storage" in new design?

Some users do not need storage, and adding default storage may break their job when there're network issues etc. with storage; For other users, they should now the storage explicitly if they need to use.

src/rest-server/docs/swagger.yaml
abuccts: in current implementation, there're three three types of persistent volume: `nfs`, `azureFile`, `flexVolume`. `flexVolume` includes `samba`, `azureBlob`, and **`other`**. Other types of persistent volume (e.g. `azureDisk`, `cephFS` etc.) are treated as **`unknown`**.

just to distinguish different k8s pv, both `other` and `unknown` do not have extra server information in our api, but could still be used in jobs.
we can also merge them into one type in api.

src/rest-server/docs/swagger.yaml
abuccts: yes, it is. I have updated swagger docs

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: But I saw you remove `storage_config` when generate task_role. Runtime need this config, and need to read k8s secret to get storage info. After this change, seems `teamwise_plugin`will not work. (We don't have `storage secret` in k8s, and runtime can't get config through ENV)

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: I mean do we support spec such as:
```yaml
extras:
  storages: default
```

Let customer does't care about the storage, and just use the default one.
If `storage` field is missing, it means user doesn't want storage.

Since `NNI` has such require, and it doesn't have permission to get every groups storage config. Add the `default` option will let third-party tools leverage the storage feature. And `runtime plugin` can be deprecated.

src/rest-server/docs/swagger.yaml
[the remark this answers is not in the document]
Binyang2014: OK... Hope we can provide a process about how to promote a API from `beta` to `release`. And when to deprecated old API....

src/rest-server/docs/swagger.yaml
```diff
@@ -1871,52 +1618,90 @@ components:
...
-      required: [name, default, servers, mountInfos]
+            required: [shareName]
+          - type: object
+            description: azureBlob type
+            properties:
+              containerName:
+                type: string
+              accountName:
+                type: string
+              accountKey:
```
mydmdm: Since there are some concerns about exposing `accountKey` to end user, consider pass `shared access signature token` instead? (both of them may be optional, and let admin to choose what to do?)

abuccts: will update according to [different types of secrets](https://github.com/Azure/kubernetes-volume-drivers/blob/master/flexvolume/blobfuse/README.md#1-create-a-secret-which-stores-azure-storage-account) used by blobfuse flexvolume

this api only exposes information in back-end persistent volume, we could add an option to show data in secrets or not

src/rest-server/src/models/v2/job/k8s.js
abuccts: updated, `teamwise_storage` config will be converted to `storages` config

src/rest-server/src/models/v2/job/k8s.js
abuccts: updated, set `storages: []` to use default storages

src/rest-server/docs/swagger.yaml
abuccts: updated

src/rest-server/docs/swagger.yaml
mydmdm: got it. thanks, and I think it is good idea to keep `other` and `unknown`.

src/rest-server/src/models/v2/job/k8s.js
mydmdm: > Since `NNI` has such require, and it doesn't have permission to get every groups storage config. Add the `default` option will let third-party tools leverage the storage feature. And `runtime plugin` can be deprecated.

A basic question, how (who) to specify a storage as default (for user)?

src/rest-server/docs/swagger.yaml
mydmdm: I have no more concerns here, resolved

src/rest-server/src/models/v2/job/k8s.js
abuccts: storage is configured for each group (not user, user belongs to one/multi group)

**admin** specifies default storage when he configured storage for groups (create in deployment config or update through group api)

mydmdm:
Could we mark **deprecated** to the `storageConfig, storageServer` API and the `teamwise_storage` plugin, considering 
- they works only in internal release and for a short while, 
- we actually want users to use the new API and k8s pv/pvc to mount

src/rest-server/src/models/v2/job/k8s.js
mydmdm: Got it, thanks. Thus *admin* could assign a storage as *default* for a group, right? 
1. If a user belongs to multiple groups, do we allow multiple default storages (one for each group)?
1. If we want to leverage the default storage (e.g. in Jupyter extension, VSC extension), could the sdk be aware which storage is default? (if we could, we could reduce user's effort to select)

abuccts: yes, 1. yes, 2, yes, currently the first storage (in array) in group config is default storage

`/api/storages` will return storage list which current user has permission to, we could also add a `default` query in this api to return default storage list for current uesr.

abuccts:
> Could we mark **to-be-deprecated** to the `storageConfig, storageServer` API and the `teamwise_storage` plugin, considering
> 
> * they are only released internally and for a short while,
> * we actually want users to use the new API and k8s pv/pvc to mount

yes, those apis have been removed from api docs in this PR, @Binyang2014 will delete runtime plugin after this PR merged.

user will only see docs on new api and use new, existing code uses old api won't be effected.

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -877,8 +893,54 @@ const put = async (frameworkName, config, rawConfig) => {
...
+          config.extras.storages =
+            plugin.parameters.storageConfigNames.map((name) => {
+              return {name};
+            });
+        } else {
+          config.extras.storages = [];
+        }
+      }
+    }
+  }
```
Binyang2014: Can we delete the `teamwise_storage` in config?
If `config.extras.storages` is set. Just delete the `teamwise_storage` plugin.
If `config.extras.storages` is not set, but `teamwise_storage` is here, convert to `extras.storage` and delete the `teamwise_stirage` in config.
May need to notice user the config is changed.

In this way, we will decouple the `rest-server` and `runtime` for storage part.

src/rest-server/src/models/v2/job/k8s.js
abuccts: added `default` query in `/api/storages`

src/rest-server/src/models/v2/job/k8s.js
abuccts: 1. it's not recommended to modify user's config
2. rest server is not aware of `com.microsoft.pai.runtimeplugin` field, which is validated, parsed and used in runtime

The current solution is just a workaround for backward compatibility, modifying this config needs more complex validation and parsing for `com.microsoft.pai.runtimeplugin` field.

To decouple storage part, why not just delete the storage plugin code in runtime simply.
If those code may be reused in the future, just rename the plugin name and there won't exist conflicts.

abuccts:
> This change only keep `get` and `list` API.
> Do we need add some doc about how to `create` or `delete` storage?

pls refer to [this document](https://github.com/microsoft/pai/blob/xiongyf%2Frefactor-storage/docs/setup-persistent-volumes-on-pai.md#assign-storage-to-pai-groups)

> And previous tools such as [storagectl](https://github.com/microsoft/pai/blob/master/contrib/storage_plugin/storagectl.py) will not work.

yes, I have added a deprecated notice for it

> Maybe we need to fix this tool to let admin add storage for certain group. Otherwise, we need to modify configMap manually.

admin could use paictl or group api to create/update/delete storage for a given group, as described in [this document](https://github.com/microsoft/pai/blob/xiongyf%2Frefactor-storage/docs/setup-persistent-volumes-on-pai.md#assign-storage-to-pai-groups)

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: That's make sense.

abuccts:
Test case:

1. follow [docs](https://github.com/Microsoft/pai/blob/master/docs/setup-persistent-volumes-on-pai.md) to setup nfs/samba/azurefile/azureblob persistent volume separately
2. check whether they're able to be mounted in PAI.

PR #4164, opened 2020-01-23
Use Postgresql as an alternative storage for job history

hzy46:
#2850 

How to use

The following service config should be set to enable internal storage, postgresql, and set rest-server to read job history from database.

```yaml
cluster:
  common:
    job-history: "true"
    job-attempt-storage-engine: "db"

internal-storage:
  root-path: /paiInternal
  enable: true

postgresql:
  enable: true
```

## What will happen if the database crashes

In such case, the health check of job attempt api will not pass. User won't be able to click it.

image

In backend side, `fluentd` will try to send chunk to postgresql forever. Once the postgresql db is restarted, the job attempt history will be recovered.

## Test

0. Check if the internal-storage and postgresql db can be deployed 

- [x] Check if the job history works normally
   - Submit a special job, which will fail after 60 seconds: `sleep 60s; echo failed at $(date); exit 233`, and set a retry numbers for it.
   - Expect: all the job retry history will be recorded normally.

- [x] Check the behavior when the database is down
   - Delete the database service by `./paictl.py service stop -m postgresql`
   - Submit a job with retries, now the user will not be able to see the retry history
   - Start the database service by `./paictl.py service start -m  postgresql`
   - Expect: all the job retry history will be recorded normally after a few minutes

- [x] Kill the database process to check if it can recover
   - find the database process in docker, kill it.
   - check if it can recover itself
   - check if the job history server can become normal within a few minutes

- [x] Bulk test: submit a job with 1000 instances, stop the postgres db from time to time.


src/fluentd/deploy/fluentd-config.yaml.template
```diff
@@ -100,14 +100,6 @@ data:
       remove_keys log,time,stream,logSeverity,logPid,logSite
     </filter>
 
-    # Just for debugging
-    <filter fc-objectsnapshots>
-      @type stdout
-      <format>
-        @type json
-      </format>
-    </filter>
```
yqwang-ms: Why remote it?

src/fluentd/deploy/fluentd-config.yaml.template
```diff
@@ -143,4 +135,26 @@ data:
           overflow_action block
         </buffer>
       </store>
+{% if (cluster_cfg['internal-storage']['enable']) and (cluster_cfg['postgresql']['enable']) %}
+      <store>
```
yqwang-ms: Also and IF for elasticsearch?

src/fluentd/deploy/fluentd-config.yaml.template
hzy46: Actually this log makes my debugging harder since `fc-objectsnapshots` is very huge, which makes other userful logs difficult to see. If it is necessary, I can revert this change.

src/fluentd/deploy/fluentd-config.yaml.template
hzy46: Done

src/fluentd/deploy/fluentd-config.yaml.template
yqwang-ms: If so let's just comment it out, 

---
In reply to: [391375514](https://github.com/microsoft/pai/pull/4164#discussion_r391375514) [](ancestors = 391375514)

hzy46: > If so let's just comment it out,
> 
> In reply to: [391375514](https://github.com/microsoft/pai/pull/4164#discussion_r391375514) [](ancestors = 391375514)

OK, done.

src/fluentd/build/fluentd.k8s.dockerfile
```diff
@@ -19,10 +19,24 @@ FROM fluent/fluentd:v1.7-1
...
-        sudo build-base ruby-dev \
+        sudo build-base ruby-dev make gcc libc-dev postgresql-dev git \
+ && apk add --no-cache --update libpq \
  && sudo gem install fluent-plugin-elasticsearch \
  && sudo gem install fluent-plugin-concat \
- && sudo gem sources --clear-all \
+ && sudo gem install rake bundler pg
+
+# Build fluent-plugin-pgjson from scratch
+# Original fluent-plugin-pgjson is from https://github.com/fluent-plugins-nursery/fluent-plugin-pgjson
```
yqwang-ms: Comment why we need build our own

src/fluentd/deploy/fluentd-config.yaml.template
```diff
@@ -143,4 +144,27 @@ data:
           overflow_action block
         </buffer>
       </store>
+{% endif %}
+{% if (cluster_cfg['internal-storage']['enable']) and (cluster_cfg['postgresql']['enable']) %}
```
yqwang-ms: seems you can use if else...

yqwang-ms: Better use something like:
job-attempt-storage-engine: "elasticsearch"  # set it to "db" to use database backend


---
In reply to: [391421409](https://github.com/microsoft/pai/pull/4164#discussion_r391421409) [](ancestors = 391421409)

src/fluentd/build/fluentd.k8s.dockerfile
hzy46: Done.

src/fluentd/deploy/fluentd-config.yaml.template
hzy46: use cluster_cfg['cluster']['common']['job-attempt-storage-engine']

src/rest-server/src/utils/postgresUtil.js
```diff
@@ -0,0 +1,40 @@
...
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+
+const {Sequelize} = require('sequelize');
+const launcherConfig = require('@pai/config/launcher');
+
+if (launcherConfig.sqlConnectionString !== 'unset') {
+  const sequelize = new Sequelize(
+    process.env.SQL_CONNECTION_STR,
```
debuggy: the env variable might be null

PR #4167, opened 2020-02-04
[Rest-server] Notify user the default group externalname when login failed.

ydye:
Current behavior
In OIDC mode, if the user is not in the configured group, an alert frame will be poped.

image

And after user click `ok` button, the page will return to the index page

image


ydye:
> Could you return a http response and show the error message in webportal instead of showing raw api error?

Hi @abuccts 

That needs to write a simple page and directly logic. And what's more, other error message should be considered, for example, when you failed to submit a task. So it will be a new feature not a fix.
Before the new feature, we can treat this PR as a quick fix.

@scarlett2018 @fanyangCS  What's your opinion?

abuccts:
> That needs to write a simple page and directly logic. And what's more, other error message should be considered, for example, when you failed to submit a task. So it will be a new feature not a fix.
Before the new feature, we can treat this PR as a quick fix.

All errors in webportal are shown in alert popup except for AAD login, which shows a raw api error.
You just need to return the http response, in webportal check http status and alert the text if error occurs.

ydye:
> > That needs to write a simple page and directly logic. And what's more, other error message should be considered, for example, when you failed to submit a task. So it will be a new feature not a fix.
> > Before the new feature, we can treat this PR as a quick fix.
> 
> All errors in webportal are shown in alert popup except for AAD login, which shows a raw api error.
> You just need to return the http response, in webportal check http status and alert the text if error occurs.

Can you give me an example? Cause I didn't touch it before.

abuccts:
> Can you give me an example of alert popup from our source code? Cause I didn't touch it before.

just `alert(string);` https://github.com/microsoft/pai/search?l=JavaScript&p=1&q=alert

src/rest-server/src/controllers/v2/user.js
```diff
@@ -107,6 +107,10 @@ const createUserIfUserNotExist = async (req, res, next) => {
         if (authConfig.groupConfig.groupDataSource === 'ms-graph') {
           forbiddenMessage = forbiddenMessage + `Please contact your admin, and join the AAD group named [ ${authConfig.groupConfig.defaultGroup.externalName} ].`;
         }
+        return res.redirect(req.returnBackURI + '?'+ querystring.stringify({
```
abuccts: should throw error here and handle all kinds of errors in [`getAAD`](https://github.com/microsoft/pai/blob/master/src/rest-server/src/controllers/v2/token.js#L76)

src/rest-server/src/controllers/v2/user.js
```diff
@@ -103,7 +103,15 @@ const createUserIfUserNotExist = async (req, res, next) => {
...
       if (grouplist && grouplist.length === 0) {
-        return next(createError('Forbidden', 'ForbiddenUserError', `User ${userData.username} is not in configured groups.`));
+        let forbiddenMessage = `User ${userData.username} is not in configured groups.`;
+        if (authConfig.groupConfig.groupDataSource === 'ms-graph') {
+          forbiddenMessage = forbiddenMessage + `Please contact your admin, and join the AAD group named [ ${authConfig.groupConfig.defaultGroup.externalName} ].`;
+        }
+        return res.redirect(req.returnBackURI + '?'+ querystring.stringify({
+          from: fromURI,
+          loginFailMsg: forbiddenMessage
+        }));
```
abuccts: should throw error here and handle all kinds of errors in [`getAAD`](https://github.com/microsoft/pai/blob/master/src/rest-server/src/controllers/v2/token.js#L76)

src/webportal/src/app/home/index.jsx
```diff
@@ -56,6 +56,10 @@ if (config.authnMethod === 'OIDC') {
   }
 }
 
+if (query.has('loginFailMsg')) {
```
abuccts: should check the login failure right after line 41 and redirect to index page if login failed

src/webportal/src/app/home/index.jsx
```diff
@@ -56,6 +56,10 @@ if (config.authnMethod === 'OIDC') {
...
+  alert(query.has('loginFailMsg'));
```
abuccts: `query.has` -> `query.get`

src/rest-server/src/controllers/v2/user.js
abuccts: `from` field seems useless here because it's better to redirect user to index page instead of `fromURI`?

ydye: It seems redirecting user to the origin page will be better？

src/webportal/src/app/home/index.jsx
ydye: fixed

src/webportal/src/app/home/index.jsx
ydye: I think return to the source page not index page will be better.

src/rest-server/src/controllers/v2/user.js
ydye: fixed

src/rest-server/src/controllers/v2/token.js
```diff
@@ -58,7 +59,13 @@ const get = async (req, res, next) => {
 /**
  *  Get the token in AAD Mode.
  */
-const getAAD = async (req, res, next) => {
+const getAAD = async (err, req, res, next) => {
+  logger.info(JSON.stringify(err));
```
abuccts: `err` may be empty, should move to `if (err) {}` block
and better `logger.warn`

PR #4168, opened 2020-02-05
Watchdog code refactor & update

Binyang2014:
1. Refactor watchdog with golang
2. Add garbage collector, will auto delete invalid priority classes and secrets created by rest-server


yqwang-ms:
Could we split it to 2 components, one is only for exporter (i.e. metrics collector), another is only for take repair actions (i.e. the watchdog)

yqwang-ms:
And do we really need to do such big changes, to add the garbage collector feature?

Binyang2014:
> Could we split it to 2 components, one is only for exporter (i.e. metrics collector), another is only for take repair actions (i.e. the watchdog)

Split it to 2 components need more changes. Such as deployment change and other components (such as alert rules) rely on watchdog metrics may need to change also. 

And for metrics collector and gc, they all need to query API server and has some common logic.  I think put then into same project is acceptable if neither of them is too heavy.

Binyang2014:
> And do we really need to do such big changes, to add the garbage collector feature?

Actually, the change not too big. I don't change the basic logic of the previous code, just do some refactor work. (Previous  watchdog put all code into one file and make it not easy to maintain... ). Most changes are go dependencies....

src/watchdog/GOPATH/src/github.com/microsoft/watchdog/Gopkg.toml
```diff
@@ -0,0 +1,20 @@
...
+  name = "github.com/prometheus/client_golang"
+  version = "1.1.0"
+
+[[constraint]]
+  name = "github.com/stretchr/testify"
+  version = "1.4.0"
+
+[[constraint]]
+  name = "github.com/microsoft/frameworkcontroller"
+  version = "0.5.1"
```
yqwang-ms: Seems you do not have to depend on frameworkcontroller?

yqwang-ms:
Seems you have changed the whole watchdog codes. Where is previous watchdog code? Will you delete them? Will you also run them?

---
In reply to: [582705262](https://github.com/microsoft/pai/pull/4168#issuecomment-582705262) [](ancestors = 582705262)

src/watchdog/GOPATH/src/github.com/microsoft/watchdog/Gopkg.toml
Binyang2014: I want to list frameworks here: https://github.com/microsoft/pai/blob/218d27517f92bc556fb283b3a8ecfeb7f3a401ff/src/watchdog/GOPATH/src/github.com/microsoft/watchdog/pkg/watchdog/k8s_client.go#L98

For GC, need to known if framework exist, then decide whether to delete the priorityClass or secret. I'm not sure if there is a better way.

Binyang2014:
> Seems you have changed the whole watchdog codes. Where is previous watchdog code? Will you delete them? Will you also run them?
> 
> In reply to: [582705262](https://github.com/microsoft/pai/pull/4168#issuecomment-582705262) [](ancestors = 582705262)

Delete them in this change, and it will not run anymore. Here is the previous code: https://github.com/microsoft/pai/blob/master/src/watchdog/src/watchdog.py

src/watchdog/GOPATH/src/github.com/microsoft/watchdog/Gopkg.toml
yqwang-ms: Let's try to use dynamic client as k8s GC controller

Binyang2014: Thanks, change it to dynamic client.

src/watchdog/GOPATH/src/github.com/microsoft/watchdog/pkg/watchdog/exporter.go
```diff
@@ -0,0 +1,60 @@
...
+
+// Exporter used to export collected metrics
+type Exporter struct {
+	collector *PromMetricCollector
+}
+
+// NewExporter create exporter instance
+func NewExporter(c *PromMetricCollector) (*Exporter, error) {
+	if c == nil {
+		return nil, errors.New("Invalid collector")
```
abuccts: did you use tab for indents? the widths (6-space tab then 8-space tab) look strange here

abuccts: seems a GitHub bug

src/watchdog/GOPATH/src/github.com/microsoft/watchdog/cmd/watchdog/main.go
```diff
@@ -0,0 +1,78 @@
...
+	gcInterval         int
+	port               int
+)
+
+func init() {
+	klog.InitFlags(nil)
+	flag.Set("v", "2")
+
+	flag.IntVar(&collectionInterval, "collection-interval", 180, "Interval between two collections (seconds)")
+	flag.IntVar(&gcInterval, "gc-interval", 60, "Interval between two garbage collections (minutes)")
```
yqwang-ms: Do we need to collect (call APIServer) every 3mins? Is it too frequent?

Binyang2014: By default it will collect every 3mins, This changed by this PR: https://github.com/microsoft/pai/pull/3761, before that, we collect metrics every 30 seconds...

src/watchdog/GOPATH/src/github.com/microsoft/watchdog/Gopkg.toml
```diff
@@ -0,0 +1,17 @@
+[prune]
+  go-tests = true
+  non-go=true
+  unused-packages = true
+
+[[constraint]]
+  name = "k8s.io/client-go"
+  version = "kubernetes-1.14.2"
```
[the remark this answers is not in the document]
Binyang2014: Bump k8s-client version to 1.15.6. Seem as kubespray

PR #4174, opened 2020-02-10
update pai version

Binyang2014:


Binyang2014:
> what about `version/K8S.VERSION`?

I'm not sure about this file. Do we still need this? Since customer may deploy k8s by their own, this file seems meaningless  .

PR #4175, opened 2020-02-10
update master branch version

Binyang2014:


version/PAI.VERSION
```diff
@@ -1 +1 @@
-v0.14.0
+v0.0.0
```
abuccts: why change to 0.0.0?

Binyang2014: Just want to show this is a dev version. For each release, we need to change this to the correct version number.
Previous, `v0.14.0` is confused. 

I want to change the version to something like `v0.0.0-dev` or `development ` but cause Jenkins  failed...

PR #4180, opened 2020-02-10
[paictl] Notify user to deploy k8s with kubespray.

ydye:


deployment/clusterCmd.py
```diff
@@ -38,6 +38,14 @@
 
 class ClusterCmd():
     def register(self, parser):
+
+        logger.warning("Kubernetes deployment in paictl maybe be deprecated in the future!")
```
fanyangCS: "maybe be deprecated" --> "will be deprecated"

deployment/clusterCmd.py
```diff
@@ -38,6 +38,14 @@
...
+        logger.warning("We highly recommend deploying kubernetes with kubespray.")
+        logger.warning("If you wanna deploy k8s with kubespray, please refer to https://github.com/microsoft/pai/tree/master/contrib/kubespray")
+        logger.warning("paictl cluster command will sleep here for 10 seconds.")
+        logger.warning("If you wanna quit this process, please type ctrl + c")
+        time.sleep(10)
```
fanyangCS: Instead of sleep 10 second, I suggest to explicitly ask user to input "YES" to proceed to the next step. Default to "NO".

ydye: This will have an influne on Jenkins. If we add additional input, we will have to change jenkins script.

deployment/clusterCmd.py
ydye: Done

deployment/clusterCmd.py
fanyangCS: could you change the Jenkins accordingly?

abuccts: @ydye you could change lines [here](https://github.com/microsoft/pai/blob/master/tests/jenkins/stage_deploy.sh#L45-L52) or [there](https://github.com/microsoft/pai/blob/master/tests/jenkins/stage_clean.sh#L30-L34), just add `echo "Y" | python ...`

ydye: Done

tests/jenkins/stage_clean.sh
```diff
@@ -31,7 +31,11 @@ cd /root/pai
 echo -e "Y\npai\n" | python paictl.py service delete
 
 # 2. cleanup k8s
-echo "Y" | python paictl.py cluster k8s-clean -p /cluster-configuration -f
+python paictl.py cluster k8s-clean -p /cluster-configuration -f << EOF_PAI_CLUSTER
+Y
+Y
+EOF_PAI_CLUSTER
```
abuccts: `yes Y | python ...` is enough

abuccts: sorry, should be `(yes Y || true) | python ...`

PR #4199, opened 2020-02-13
Add version in webportal

debuggy:
same as #4190


src/webportal/src/app/layout/components/sidebar.jsx
```diff
@@ -116,6 +116,10 @@ const Sidebar = ({ className, style }) => {
 
   const plugins = config.PAI_PLUGINS;
   const isAdmin = cookies.get('admin') === 'true';
+  const versionTag =
+    window.PAI_COMMIT_VERSION === ''
+      ? `${window.PAI_VERSION}`
+      : `${window.PAI_VERSION} ${window.PAI_COMMIT_VERSION}`;
```
abuccts: seems the condition is redundant
```js
`${window.PAI_VERSION} ${window.PAI_COMMIT_VERSION}`.trim()
```

PR #4205, opened 2020-02-13
[Build] Build image depends on build mode

ydye:
```
# Build all image
./pai_build.py build -c /path/to/config [-m all]

# Build k8s image
./pai_build.py build -c /path/to/config -m k8s

# Build yarn image
./pai_build.py build -c /path/to/config -m yarn
```

```
# push all image
./pai_build.py push -c /path/to/config [-m all]

# push k8s image
./pai_build.py push -c /path/to/config -m k8s

# push yarn image
./pai_build.py push -c /path/to/config -m yarn
```


build/core/build_center.py
```diff
@@ -65,7 +66,14 @@ def construct_graph(self):
             service_name = service_name[-2] if len(service_name) > 1 else None
             for file_name in file_list:
                 if file_name.endswith(".dockerfile"):
-                    self.graph.add_image_to_service(str(os.path.splitext(file_name)[0]), service_name)
+                    passed = False
+                    if self.task_type == 'all':
+                        passed = True
+                    elif file_name.endswith(".common.dockerfile") or file_name.endswith(
+                            ".{0}.dockerfile".format(self.task_type)):
```
abuccts: why not determine by `service_type` in config?

build/pai_build.py
```diff
@@ -39,11 +39,11 @@ def load_build_config(config_dir):
     return configModel
 
 def build_service(args,config_model):
-    pai_build = build_center.BuildCenter(config_model,args.service)
+    pai_build = build_center.BuildCenter(config_model,args.service,args.mode)
```
abuccts: one space after comma `,`

docs/zh_CN/zookeeper/zookeeper.md
```diff
@@ -22,7 +22,7 @@ Cluster-configuration.
 
 * Manual build
 
-refer to [dockerfile](../../../src/zookeeper/build/zookeeper.dockerfile).
+refer to [dockerfile](../../../src/zookeeper/build/zookeeper.yarn.dockerfile).
```
abuccts: better to use `service_type` in config, otherwise, you need to rename files again if there're three types of modes and the service is used by two of them in the future

build/core/build_center.py
ydye: Build should be seperated from deploy. And some service don't have image (/build folder). And some services and its build folder are not in the same path or their name is different (yarn-exporter and hadoop-resource-manager).

build/pai_build.py
ydye: Done

Issue #4211, opened 2020-02-17
Add offline apt package cache in kube runtime


hzy46:
Runtime plugins use `apt update` and `apt install ...` to install some additional packages. If the network connection from pai cluster to the apt repository is poor, `apt install` may take a long period of time or even fail.

Since we use kube runtime to initialize the environment before job container, it is possible to use the runtime container as a "cache" of some frequently-needed packages. Here is one possible solution in detail:

  -  When we build runtime container, add all `.deb` files relied by certain package and certain os version. It could be achieved by the following commands:
  ```bash
  apt-get update
  apt-get -y install --print-uris package-name | cut -d\' -f2 | grep http:// > apturls
  wget -i <path-to-apturls-file> 
  ```
 - In every runtime plugin, if it needs some package, use the pre-downloaded `.deb` files to install it.

PR #4222, opened 2020-02-19
[Edu Version] Add offline apt package cache in kube runtime

hzy46:
To resolve: https://github.com/microsoft/pai/issues/4211

**How to use:**

  **1. Add packages you want to store cache for in the file `package_cache_info`**:

  ```
  # group_name, os, packages(space-gapped), precommands
  # "#" can be used for comments
  ssh,ubuntu16.04,openssh-client openssh-server,
  ssh,ubuntu18.04,openssh-client openssh-server,
  nfs,ubuntu16.04,nfs-common,
  nfs,ubuntu18.04,nfs-common,
  ```

  The first column is `group_name`. One group can contain multiple packages. The second column stands for the OS type. Currently only `ubuntu16.04` and `ubuntu18.04` are supported. The third column is the packages you want to add for the group. The last column is the precommands, which will be executed before gathering packages and it can be left empty.

  **2. In `init.py` of each plugin:**

  ```python
from plugins.plugin_utils import try_to_install_by_cache
command = [
    try_to_install_by_cache('<group_name>') + ' || { <install packages if cache is not found>; }',
    "<other commands>",
]
  ```

`try_to_install_by_cache('<group_name>')` will generate a script to install all packages of a certain group name. It guarantees:

  - If it returns 0, all the packages are installed successfully.
  - If it has a non-zero exit code, the package installation has failed. Reasons could be that the required cache is  not found or other internal problems. In such case, plugin should fallback to `apt-get`, `yum` or other cmdlines to install the packages.

Here is an example for the `ssh` plugin:

```python
command = [
    try_to_install_by_cache('ssh') + ' || { apt-get update; apt-get install -y openssh-client openssh-server; }',
    '.....' 
]
```

  **3. Add an environmental variable `ENABLE_PACKAGE_CACHE=true` when you build the kube-runtime image**:

  ```bash
  ENABLE_PACKAGE_CACHE=true ./build/pai_build.py build -s kube-runtime -c <pai-config>
  ./build/pai_build.py push -i kube-runtime -c <pai-config>
  ```

  By default package cache is disabled to save build time.

**Optimization (TBD)**

1. Use hash of package url to save storage space: Packages with the same url can be saved together.


src/kube-runtime/src/plugins/plugin_utils.py
```diff
@@ -67,6 +69,19 @@ def plugin_init():
...
+    target_folder = os.path.join(PAI_WORK_DIR, 'package_cache')
+    exists_group_names = os.listdir(source_folder)
+    needed_group_names = [name for name in exists_group_names if name.startswith(group_name + '-')]
+    for name in needed_group_names:
+        name_source_folder = os.path.join(source_folder, name)
+        name_target_folder = os.path.join(target_folder, name)
+        if not(os.path.exists(name_target_folder)):  # avoid duplicate copy
+            shutil.copytree(name_source_folder, name_target_folder)
+    return '/bin/bash {}/runtime.d/install_group.sh '.format(PAI_WORK_DIR) + group_name
+
```
mydmdm: I think it would the below behavior would be better
```python
def try_to_install_by_cache(group_name: string, failed_cmds: list):
    # no change, copy caches
    cached_cmd = '/bin/bash {}/runtime.d/install_group.sh '.format(PAI_WORK_DIR) + group_name
    return '%s || {%s}' % (cached_cmd, ';'.join(failed_cmds))
```
and the example usage is 
```bash
cmds = [
    ...
    try_to_install_by_cache('ssh', ['apt-get update', 'apt-get install -y openssh-client openssh-server']),
]
```

mydmdm:
Would you please add this conversion to a README in somewhere?

> To resolve: #4211
> 
> **How to use:**
> 
> **1. Add packages you want to store cache for in the file `package_cache_info`**:
> 
> ```
> # group_name, os, packages(space-gapped), precommands
> # "#" can be used for comments
> ssh,ubuntu16.04,openssh-client openssh-server,
> ssh,ubuntu18.04,openssh-client openssh-server,
> nfs,ubuntu16.04,nfs-common,
> nfs,ubuntu18.04,nfs-common,
> ```
> 
> The first column is `group_name`. One group can contain multiple packages. The second column stands for the OS type. Currently only `ubuntu16.04` and `ubuntu18.04` are supported. The third column is the packages you want to add for the group. The last column is the precommands, which will be executed before gathering packages and it can be left empty.
> 
> **2. In `init.py` of each plugin:**
> 
> ```python
> from plugins.plugin_utils import try_to_install_by_cache
> command = [
>   try_to_install_by_cache('<group_name>') + ' || { <install packages if cache is not found>; }',
>   "<other commands>",
> ]
> ```
> 
> `try_to_install_by_cache('<group_name>')` will generate a script to install all packages of a certain group name. It guarantees:
> 
> * If it returns 0, all the packages are installed successfully.
> * If it has a non-zero exit code, the package installation has failed. Reasons could be that the required cache is  not found or other internal problems. In such case, plugin should fallback to `apt-get`, `yum` or other cmdlines to install the packages.
> 
> Here is an example for the `ssh` plugin:
> 
> ```python
> command = [
>     try_to_install_by_cache('ssh') + ' || { apt-get update; apt-get install -y openssh-client openssh-server; }',
>     '.....' 
> ]
> ```
> 
> **3. Add an environmental variable `ENABLE_PACKAGE_CACHE=true` when you build the kube-runtime image**:
> 
> ```shell
> ENABLE_PACKAGE_CACHE=true ./build/pai_build.py build -s kube-runtime -c <pai-config>
> ./build/pai_build.py push -i kube-runtime -c <pai-config>
> ```
> 
> By default package cache is disabled to save build time.
> 
> **Optimization (TBD)**
> 
> 1. Use hash of package url to save storage space: Packages with the same url can be saved together.

src/kube-runtime/src/plugins/plugin_utils.py
hzy46: will fix this in a different PR

hzy46:
Close this to use https://github.com/microsoft/pai/pull/4226 instead.

PR #4226, opened 2020-02-20
Add offline apt package cache in kube runtime 

hzy46:
To resolve: https://github.com/microsoft/pai/issues/4211

Docs: https://github.com/microsoft/pai/blob/zhiyuhe/package_cache/src/kube-runtime/package_cache.md

### DIFF
--- a/src/kube-runtime/build/kube-runtime.k8s.dockerfile
+++ b/src/kube-runtime/build/kube-runtime.k8s.dockerfile
@@ -27,7 +27,27 @@ COPY GOPATH/src/github.com/microsoft/runtime/ ${PROJECT_DIR}
 RUN ${PROJECT_DIR}/build/runtime/go-build.sh && \
   mv ${PROJECT_DIR}/dist/runtime/ ${INSTALL_DIR}
 
+FROM ubuntu:16.04 as ubuntu_16_04_cache
+
+WORKDIR /src
+COPY src/package_cache ./
+RUN chmod -R +x ./
+RUN /bin/bash ubuntu_build.sh package_cache_info ubuntu16.04
+
+
+FROM ubuntu:18.04 as ubuntu_18_04_cache
+
+WORKDIR /src
+COPY src/package_cache ./
+RUN chmod -R +x ./
+RUN /bin/bash ubuntu_build.sh package_cache_info ubuntu18.04
+
+
 FROM python:3.7-alpine
+
+RUN mkdir -p /opt/package_cache
+COPY --from=ubuntu_16_04_cache /package_cache /opt/package_cache/
+COPY --from=ubuntu_18_04_cache /package_cache /opt/package_cache/
 
 RUN pip install kubernetes pyyaml requests jinja2 pystache
 
--- a/src/kube-runtime/build/kube-runtime.k8s.dockerfile
+++ b/src/kube-runtime/build/kube-runtime.k8s.dockerfile
@@ -15,17 +15,7 @@
 # DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-
-FROM golang:1.12.6-alpine as builder
-
-ENV PROJECT_DIR=${GOPATH}/src/github.com/microsoft/runtime
-ENV INSTALL_DIR=/opt/kube-runtime
-
-RUN apk update && apk add --no-cache bash && \
-  mkdir -p ${PROJECT_DIR} ${INSTALL_DIR}
-COPY GOPATH/src/github.com/microsoft/runtime/ ${PROJECT_DIR}
-RUN ${PROJECT_DIR}/build/runtime/go-build.sh && \
-  mv ${PROJECT_DIR}/dist/runtime/ ${INSTALL_DIR}
+# Package Cache Data Layer Starts
 
 FROM ubuntu:16.04 as ubuntu_16_04_cache
 
@@ -42,6 +32,18 @@ COPY src/package_cache ./
 RUN chmod -R +x ./
 RUN /bin/bash ubuntu_build.sh package_cache_info ubuntu18.04
 
+# Package Cache Data Layer Ends
+
+FROM golang:1.12.6-alpine as builder
+
+ENV PROJECT_DIR=${GOPATH}/src/github.com/microsoft/runtime
+ENV INSTALL_DIR=/opt/kube-runtime
+
+RUN apk update && apk add --no-cache bash && \
+  mkdir -p ${PROJECT_DIR} ${INSTALL_DIR}
+COPY GOPATH/src/github.com/microsoft/runtime/ ${PROJECT_DIR}
+RUN ${PROJECT_DIR}/build/runtime/go-build.sh && \
+  mv ${PROJECT_DIR}/dist/runtime/ ${INSTALL_DIR}
 
 FROM python:3.7-alpine
 
--- a/src/kube-runtime/package_cache.md
+++ b/src/kube-runtime/package_cache.md
@@ -1,0 +1,51 @@
+## Package Cache
+
+Some runtime plugins use `apt update` and `apt install ...` to install packages. If the network connection from pai cluster to the apt repository is poor, `apt install` may take a long period of time or fail.
+
+Since we use kube runtime to initialize the environment before job container, it is possible to use the runtime container as a "cache" of some frequently-needed packages.
+
+## How to enable cache for your plugin
+
+**1. Add packages you want to store cache for in the file [`package_cache_info`](src/package_cache/package_cache_info)**:
+
+```
+# group_name, os, packages(space-gapped)
+# "#" can be used for comments
+ssh,ubuntu16.04,openssh-client openssh-server
+ssh,ubuntu18.04,openssh-client openssh-server
+nfs,ubuntu16.04,nfs-common
+nfs,ubuntu18.04,nfs-common
+```
+
+The first column is `group_name`. One group can contain multiple packages. The second column stands for the OS type. Currently only `ubuntu16.04` and `ubuntu18.04` are supported. The third column is the packages you want to add for the group. The last column is the precommands, which will be executed before gathering packages and it can be left empty.
+
+**2. In `init.py` of each plugin:**
+
+```python
+from plugins.plugin_utils import try_to_install_by_cache
+command = [
+try_to_install_by_cache('<group_name>', fallback_cmds=[
+    '<fallback commands>',
+]),
+```
+
+`try_to_install_by_cache(group_name, fallback_cmds)` will generate a script to install all packages of a certain group name. It guarantees:
+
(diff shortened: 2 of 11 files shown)


src/kube-runtime/src/runtime.d/install_group.sh
```diff
@@ -0,0 +1,73 @@
...
+# THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING
+# BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+# DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+PAI_WORK_DIR=/usr/local/pai
+CACHE_ROOT_DIR=${PAI_WORK_DIR}/package_cache
+
+ubuntu_is_successfully_installed(){
```
Binyang2014: Use `function` key works here?

src/kube-runtime/src/runtime.d/install_group.sh
```diff
@@ -0,0 +1,73 @@
...
+
+if cat /etc/issue | grep "Ubuntu 16.04" &> /dev/null ; then
+  os='ubuntu16.04'
+elif cat /etc/issue | grep "Ubuntu 18.04" &> /dev/null ; then
+  os='ubuntu18.04'
+else
+  echo "[package_cache] This os doesn't support package cache!"
+  exit 1
+fi
+if [ -d $CACHE_ROOT_DIR"/${name}-${os}" ]; then
```
Binyang2014: Do we need to check arch such as `x64` or `i386` here?

src/kube-runtime/build/kube-runtime.k8s.dockerfile
```diff
@@ -27,8 +27,28 @@ COPY GOPATH/src/github.com/microsoft/runtime/ ${PROJECT_DIR}
...
+
+
+FROM ubuntu:18.04 as ubuntu_18_04_cache
+
+WORKDIR /src
+COPY src/package_cache ./
+RUN chmod -R +x ./
+RUN /bin/bash ubuntu_build.sh package_cache_info ubuntu18.04
+
+
```
Binyang2014: Please move these to the top of this file. And add some comments such as:
`*****************Data Cache Layer************`

src/kube-runtime/src/runtime.d/install_group.sh
hzy46: A 64-bit system cannot use a 32-bit docker image, and vice versa. Since OpenPAI is based on 64-bit os, I think there is no need to check the system arch here.

hzy46: > Do we need to check arch such as `x64` or `i386` here?

A 64-bit system cannot use a 32-bit docker image, and vice versa. Since OpenPAI is based on 64-bit os, I think there is no need to check the system arch here.

src/kube-runtime/src/package_cache/ubuntu_build.sh
```diff
@@ -0,0 +1,72 @@
...
+      cat /dry_run_log  | grep Conf | cut -d " " -f 2 > ./order
+      if [ $? -ne 0 ]; then
+        echo 'There is an error during package collection.'
+        exit 1
+      fi
+    fi
+  fi
+done < $package_cache_info  
+
+apt-get -y install wget
```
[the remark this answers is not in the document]
hzy46: If I move it to the dockerfile, `ubuntu_build.sh` will break into two files, which seems too detailed.

src/kube-runtime/src/runtime.d/install_group.sh
mydmdm: We may support 64bit only now, but 
- it is better to clarify it in the documents (e.g. Only `ubuntu16. 04` and `ubuntu18. 04` with 64bit arch are supported)
- have a check here and alert a warning if 32bit arch (rarely?) here

fanyangCS: it is a good idea to check the environment. 
maybe we can fallback to online apt-get when 32bit is required?

src/kube-runtime/build/kube-runtime.k8s.dockerfile
hzy46: Done

src/kube-runtime/src/runtime.d/install_group.sh
hzy46: Done.

src/kube-runtime/src/plugins/plugin_utils.py
```diff
@@ -54,3 +56,20 @@ def plugin_init():
...
+    target_folder = os.path.join(PAI_WORK_DIR, 'package_cache')
+    exists_group_names = os.listdir(source_folder)
+    needed_group_names = [name for name in exists_group_names if name.startswith(group_name + '-')]
+    for name in needed_group_names:
+        name_source_folder = os.path.join(source_folder, name)
+        name_target_folder = os.path.join(target_folder, name)
+        if not(os.path.exists(name_target_folder)):  # avoid duplicate copy
+            shutil.copytree(name_source_folder, name_target_folder)
+    cached_cmd = '/bin/bash {}/runtime.d/install_group.sh '.format(PAI_WORK_DIR) + group_name
+    return '%s || { %s; }' % (cached_cmd, ';'.join(fallback_cmds))
```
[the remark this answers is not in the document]
hzy46: Done.

src/kube-runtime/src/runtime.d/install_group.sh
hzy46: Done. If a job runs on a 32-bit ubuntu system and the required packages are not installed, kube-runtime will use `apt-get` to install them.

src/kube-runtime/src/runtime.d/install_group.sh
```diff
@@ -0,0 +1,78 @@
...
+# THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING
+# BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+# DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+PAI_WORK_DIR=/usr/local/pai
+CACHE_ROOT_DIR=${PAI_WORK_DIR}/package_cache
+
+function ubuntu_is_successfully_installed(){
```
Binyang2014: Change name to `is_installed`?

src/kube-runtime/package_cache.md
```diff
@@ -0,0 +1,52 @@
+## Package Cache
+
+Some runtime plugins use `apt update` and `apt install ...` to install packages. If the network connection from pai cluster to the apt repository is poor, `apt install` may take a long period of time or fail.
```
Binyang2014: Add notice for we only support `x64` package here?

src/kube-runtime/src/runtime.d/install_group.sh
hzy46: Done.

src/kube-runtime/package_cache.md
hzy46: Done.

src/kube-runtime/src/runtime.d/install_group.sh
```diff
@@ -0,0 +1,78 @@
...
+  done
+  return 0
+}
+
+if [ $# -ne 1 ]; then
+  echo "Usage: bash -x install_group.sh <group_name>"
+  exit 1
+else
+  name=$1
+fi
```
abuccts: do we need to group packages by plugin name? some plugins may require the same package, then it will be cached multiple times in current implementation

```
try_to_install_by_cache(["package1", "package2"])
bash install_package.sh package1 package2
```
will this be better? and you can also remove `fallback_cmds` parameter

src/kube-runtime/src/plugins/plugin_utils.py
```diff
@@ -54,3 +56,20 @@ def plugin_init():
...
     return [plugin_config, args.pre_script, args.post_script]
+
+
+PAI_WORK_DIR = "/usr/local/pai"
+
+
+def try_to_install_by_cache(group_name: str, fallback_cmds: list):
+    source_folder = "/opt/package_cache"
+    target_folder = os.path.join(PAI_WORK_DIR, "package_cache")
+    exists_group_names = os.listdir(source_folder)
```
abuccts: pls check dir exists here to fix unit test

src/kube-runtime/src/runtime.d/install_group.sh
hzy46: This will not work since we must use the group name instead of multiple package names. If package names are used, the dependency order of these packages should also be calculated, which is not implemented.

src/kube-runtime/src/plugins/plugin_utils.py
hzy46: Done. But the unit test should also be fixed because the commands are sightly different.

src/kube-runtime/src/runtime.d/install_group.sh
Binyang2014: Actually I mean change `ubuntu_is_successfully_installed` to `is_installed` or `is_ubuntu_package_installed`. `ubuntu_is_successfully_installed` let me thought you want to install Ubuntu, not package...

hzy46: > Actually I mean change `ubuntu_is_successfully_installed` to `is_installed` or `is_ubuntu_package_installed`. `ubuntu_is_successfully_installed` let me thought you want to install Ubuntu, not package...

lol.

hzy46:
Test cases:

1. submit a job using `ubuntu:16.04`; enable ssh and a storage for it; make sure SSH and storage is usable; make sure there is no `apt-get` command in generated logs

2. submit a job using `ubuntu:15.10`; enable ssh and a storage for it; make sure SSH and storage is usable; In this case, package cache should fall back to `apt-get`

Binyang2014:
> Test cases:
> 
> 1. submit a job using `ubuntu:16.04`; enable ssh and a storage for it; make sure SSH and storage is usable; make sure there is no `apt-get` command in generated logs
> 2. submit a job using `ubuntu:15.10`; enable ssh and a storage for it; make sure SSH and storage is usable; In this case, package cache should fall back to `apt-get`

Test finished

PR #4243, opened 2020-03-03
[Runtime] Skip teamwise-storage and fix UT

Binyang2014:


src/kube-runtime/src/plugins/teamwise_storage/init.py
```diff
@@ -30,6 +30,15 @@
 
 
 def main():
+    '''
+    Teamwise plugin is deperated. Keep this piece of code since we may reuse them to
```
abuccts: deprecated

src/kube-runtime/src/plugins/teamwise_storage/init.py
```diff
@@ -30,6 +30,15 @@
...
+    support user defined storage.
+    '''
+
+    LOGGER.warning("This plugin is deperated, will ignore this plugin")
```
abuccts: deprecated

src/kube-runtime/src/plugins/teamwise_storage/init.py
Binyang2014: fixed

src/kube-runtime/src/plugins/teamwise_storage/init.py
Binyang2014: fixed

PR #4255, opened 2020-03-04
[quick-start] change default version to v0.17.0

ydye:


contrib/kubespray/example/config.yml
```diff
@@ -1,7 +1,7 @@
 user: forexample
 password: forexample
 branch-name: master
-docker-image-tag: quick-start
```
Binyang2014: Can we change it to latest release version? Or use `<% latest-release %>` to refer the latest version.

ydye: Done

Issue #4252, opened 2020-03-04
Failed to load service page when pod number is huge


Binyang2014:
When there are many jobs running on OpenPAI. The service pages will stuck due to web-portal want to fetch all pods, and we get  `net::ERR_CONNECTION_RESET 200 (OK)`

![image](https://user-images.githubusercontent.com/9415966/75862992-bd951680-5e3a-11ea-8bac-cc832a571bb2.png)

For pai service, it will has `app` label, such as `app: rest-server` , we should use this to filter pai-service

Binyang2014:
Fixed by #4260

PR #4260, opened 2020-03-06
support labelSelector in restAPI

Binyang2014:


src/rest-server/src/models/kubernetes/kubernetes.js
```diff
@@ -95,15 +96,12 @@ const getNodes = async () => {
...
   const client = getClient();
-  const labelSelectorStr = encodeSelector(labelSelector, negativeLabelSelector);
+
   const requestOptions = {};
-  if (labelSelectorStr) {
-    requestOptions.params = {
-      labelSelector: labelSelectorStr,
-    };
-  }
+  requestOptions.params = cloneDeep(options);
```
abuccts: in es6, you could use built-in `Object.assign()`

src/rest-server/src/models/kubernetes/kubernetes.js
```diff
@@ -95,15 +96,12 @@ const getNodes = async () => {
...
+  const {namespace} = options;
   const client = getClient();
-  const labelSelectorStr = encodeSelector(labelSelector, negativeLabelSelector);
+
   const requestOptions = {};
-  if (labelSelectorStr) {
-    requestOptions.params = {
-      labelSelector: labelSelectorStr,
-    };
-  }
```
abuccts: seems you can use `options` object directly if you do not need to generate `labelSelectorStr` anymore

src/rest-server/src/models/kubernetes/kubernetes.js
```diff
@@ -99,14 +98,11 @@ const getPods = async (options = {}) => {
   const {namespace} = options;
   const client = getClient();
 
-  const requestOptions = {};
-  requestOptions.params = cloneDeep(options);
-  delete requestOptions.params.namespace;
```
abuccts: do you need to `delete options.namespace`?

Binyang2014: k8s API server will ignore unknown query parameter. But I don't find any doc for this. Remove `options.namespace` is a more safety way.

I don't want to change the input object, will clone a new object and remote `namespace` property.

Binyang2014:
Closes #4252

PR #4275, opened 2020-03-10
[watchdog] Remove go dependency source code

Binyang2014:
Remove source code under `vendor` folder.
Using `go modules` instead of `go dep`

According to doc: https://github.com/golang/go/wiki/Modules#is-gosum-a-lock-file-why-does-gosum-include-information-for-module-versions-i-am-no-longer-using
`go.mod` will 100% reproduce the build and `go.sum` is used to validate the download modules.

Tested for build process, deployed in INT bed. Seems all works fine


yqwang-ms:
Can we directly depend on explict version? They are 0.0.0....
	k8s.io/client-go v0.0.0-20191114101336-8cba805ad12d
	k8s.io/utils v0.0.0-20200229041039-0a110f9eb7ab // indirect

Binyang2014:
> Can we directly depend on explict version? They are 0.0.0....
> k8s.io/client-go v0.0.0-20191114101336-8cba805ad12d
> k8s.io/utils v0.0.0-20200229041039-0a110f9eb7ab // indirect

This file is generated automatically by `go get k8s.io/client-go@kubernetes-1.15.6` 
According to https://github.com/golang/go/wiki/Modules#can-a-module-consume-a-package-that-has-not-opted-in-to-modules
If repository doesn't opt in to modules,  go module will use a `Pseudo-versions` with timestamp and commit hash.

PR #4289, opened 2020-03-16
[webportal] show account name for Azure-file and Azure-blob in webportal

Binyang2014:
image

Sample output

### DIFF
--- a/src/webportal/src/app/user/fabric/user-profile/storage-list.jsx
+++ b/src/webportal/src/app/user/fabric/user-profile/storage-list.jsx
@@ -45,10 +45,10 @@ function getStorageServerUri(server) {
     case 'samba':
       return `smb://${data.address}/${normalizePath(data.rootPath)}`;
     case 'azurefile':
-      fileAddress = `${data.accountName}.file.core.windows.net`;
+      fileAddress = `${data.accountName}.<AZURE_STORAGE_DNS_SUFFIX>`;
       return `smb://${fileAddress}/${normalizePath(data.fileShare)}`;
     case 'azureblob':
-      blobAddress = `${data.accountName}.blob.core.windows.net`;
+      blobAddress = `${data.accountName}.<AZURE_STORAGE_DNS_SUFFIX>`;
       return `http://${blobAddress}/${normalizePath(data.containerName)}`;
     case 'hdfs':
       return `hdfs://${data.namenode}:${data.port}`;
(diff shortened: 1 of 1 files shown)


src/webportal/src/app/user/fabric/user-profile/storage-list.jsx
```diff
@@ -37,17 +37,18 @@ function normalizePath(path) {
 
 function getStorageServerUri(server) {
   let blobAddress;
+  let fileAddress;
```
abuccts: define when those variables are used and change to `const`?

src/webportal/src/app/user/fabric/user-profile/storage-list.jsx
```diff
@@ -37,17 +37,18 @@ function normalizePath(path) {
...
       return `smb://${data.address}/${normalizePath(data.rootPath)}`;
     case 'azurefile':
-      return `smb://${data.dataStore}/${normalizePath(data.fileShare)}`;
+      fileAddress = `${data.accountName}.file.core.windows.net`;
+      return `smb://${fileAddress}/${normalizePath(data.fileShare)}`;
     case 'azureblob':
-      blobAddress =
-        data.dataStore || `${data.accountName}.blob.core.windows.net`;
+      blobAddress = `${data.accountName}.blob.core.windows.net`;
       return `http://${blobAddress}/${normalizePath(data.containerName)}`;
```
abuccts: where will this address be used? for blobfuse and strorage explorer, two parameters storage account and container are needed

Binyang2014: Just want to show `accountName` and `containerName` here. Any recommend representation? `accountName:containerName/path` ?

src/webportal/src/app/user/fabric/user-profile/storage-list.jsx
Binyang2014: Can not define variables in `switch-case` scope. Need a new code block. Define the variables outside for convenient.

src/webportal/src/app/user/fabric/user-profile/storage-list.jsx
Binyang2014: Keep this format and change the `blob.core.windows.net` to `<AZURE_STORAGE_DNS_SUFFIX>` since different region has different suffix. And we just want to show `account name` and `container name` here

src/webportal/src/app/user/fabric/user-profile/storage-list.jsx
Binyang2014: Sorry, can not define variables in case/default clauses because of lint issue: https://eslint.org/docs/rules/no-case-declarations

PR #4296, opened 2020-03-18
Support submit job from local storage

debuggy:


src/webportal/src/app/job-submission/job-submission-page.jsx
```diff
@@ -298,7 +299,27 @@ export const JobSubmissionPage = ({
...
+      jobInfo.name = generateJobName(jobInfo.name);
+      if (get(jobConfig, 'extras.submitFrom')) {
+        delete jobConfig.extras.submitFrom;
+      }
+      setJobProtocol(new JobProtocol(jobConfig));
+      setJobTaskRoles(taskRoles);
+      setParameters(parameters);
+      setJobInformation(jobInfo);
+      setExtras(extras);
+      setLoading(false);
```
Binyang2014: This piece of code is same with above. Can we wrap them to a function?

debuggy: I will refact this piece of code in next pr to resolve yaml modification issue. I'd rather seperate the two behaviors right now to keep the logic clear.

PR #4301, opened 2020-03-19
[Rest Server] Add launched time for job

### LINKED ISSUE #3249: How to get the actual duration of the job，the appLaunchedTime have value while the job is still in waiting.
**Organization Name**: Advantech

**Short summary about the issue/question**: We want to calculate the actual duration time of a job, from the start of the job to the end of the job. We need to exclude the waiting time. At first we use the createdTime, and later we changed to use the appLaunchedTime. But when the job is still in waiting status, the appLaunchedTime appears. 

**Brief what process you are following**:

**How to reproduce it**:

**OpenPAI Environment**: Cluster
- OpenPAI version: 0.10.1
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Hardware (e.g. core number, memory size, storage size, GPU type etc.):
- Others:

### LINKED ISSUE #4295: Job API not return correct appLaunchedTime
Refer to https://github.com/microsoft/pai/blob/338f5a3e01c8c5fce9f81177d7de3cf154c304c9/src/rest-server/src/models/v2/job/k8s.js#L254

Currently `appLaunchedTime ` equal to `creationTimestamp`. So it will include waiting time. We need to return correct `appLaunchedTime`. And it's better to return `launchTime` for each task. Then we can monitor the actually resource utilization for job and task.

abuccts:
Add launched time for job.

Closes #4295.
Resolves #3249

### DIFF
--- a/src/rest-server/src/models/v2/job/k8s.js
+++ b/src/rest-server/src/models/v2/job/k8s.js
@@ -219,12 +219,13 @@ const convertFrameworkDetail = async (framework) => {
   if (!framework.status) {
     framework.status = mockFrameworkStatus();
   }
+  const attemptStatus = framework.status.attemptStatus;
   // check fields which may be compressed
-  if (framework.status.attemptStatus.taskRoleStatuses == null) {
-    framework.status.attemptStatus.taskRoleStatuses = decompressField(framework.status.attemptStatus.taskRoleStatusesCompressed);
+  if (attemptStatus.taskRoleStatuses == null) {
+    attemptStatus.taskRoleStatuses = decompressField(attemptStatus.taskRoleStatusesCompressed);
   }
 
-  const completionStatus = framework.status.attemptStatus.completionStatus;
+  const completionStatus = attemptStatus.completionStatus;
   const diagnostics = completionStatus ? completionStatus.diagnostics : null;
   const exitDiagnostics = generateExitDiagnostics(diagnostics);
   const detail = {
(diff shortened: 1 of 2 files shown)


src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -248,11 +249,11 @@ const convertFrameworkDetail = async (framework) => {
       retryDelayTime: framework.status.retryPolicyStatus.retryDelaySec,
       createdTime: new Date(framework.metadata.creationTimestamp).getTime(),
       completedTime: new Date(framework.status.completionTime).getTime(),
-      appId: framework.status.attemptStatus.instanceUID,
+      appId: attemptStatus.instanceUID,
       appProgress: completionStatus ? 1 : 0,
       appTrackingUrl: '',
-      appLaunchedTime: new Date(framework.metadata.creationTimestamp).getTime(),
-      appCompletedTime: new Date(framework.status.completionTime).getTime(),
+      appLaunchedTime: new Date(attemptStatus.runTime || attemptStatus.completionTime).getTime(),
```
yqwang-ms: Also update related doc for appLaunchedTime.

For a job attempt:
WholeDuration = CompletionTime - StartTime
RunningDuration = CompletionTime - LaunchedTime
WaitingDuration = WholeDuration - RunningDuration

For the whole job which may have multiple attempts:
WholeDuration = CompletionTime - StartTime
RunningTime = Sum_all_attempts (CompletionTime - LaunchedTime)
WaitingDuration = WholeDuration - RunningDuration

abuccts:
Test cases:

* submit a `sleep 1s` job, check whether `appLaunchedTime` equals to `appCompletedTime`
* submit a `sleep 10m` job, check whether `appLaunchedTime` equals to `appCompletedTime - 10m`

PR #4302, opened 2020-03-19
delete elasticsearch in rest-server

hzy46:


contrib/kubespray/quick-start/services-configuration.yaml.template
```diff
@@ -55,6 +55,11 @@ webportal:
     uri: https://gerhut.github.io/store/marketplace/plugin.js
   server-port: 9286
 
+internal-storage:
```
yqwang-ms: What if internal-storage=true but postgresql=false?

How about remove internal-storage flag, and just add a field:
internal-storage: postgresql

hzy46: It is ok for internal-storage=true and postgresql=false. Since the two components are separate services, it is not convenient to use the config `internal-storage: postgresql`.

PR #4333, opened 2020-03-27
[Rest Server] fix fetch job status failed

### LINKED ISSUE #4330: rest-server reports: fetch job status failed: podCompletionStatus.containers is not iterable
How to reproduce:

1. Setup a hived vc which has more cpu capacity than ground truth. For example, you have a node with 10 CPUs, but configure the hived as it has 20 CPUs.
2. Submit a job to that vc, which takes all CPUs of the vc capacity.

Then try to access the submitted job:

image

It cannot be stopped:

image

Binyang2014:
fix #4330

### DIFF
--- a/src/rest-server/src/models/v2/job/k8s.js
+++ b/src/rest-server/src/models/v2/job/k8s.js
@@ -1093,7 +1093,7 @@ const generateExitDiagnostics = (diag) => {
 };
 
 const extractRuntimeOutput = (podCompletionStatus) => {
-  if (_.isEmpty(podCompletionStatus)) {
+  if (!podCompletionStatus || !Array.isArray(podCompletionStatus.containers)) {
     return null;
   }
 


src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -1093,7 +1093,10 @@ const generateExitDiagnostics = (diag) => {
 };
 
 const extractRuntimeOutput = (podCompletionStatus) => {
-  if (_.isEmpty(podCompletionStatus)) {
+  if (
+    _.isEmpty(podCompletionStatus) ||
```
abuccts: what do you want to check for `podCompletionStatus`, isn't `if (podCompletionStatus)` enough

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -1093,7 +1093,10 @@ const generateExitDiagnostics = (diag) => {
...
+    !_.isArray(podCompletionStatus.containers)
```
abuccts: why not use built-in `Array.isArray` in es6

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: Just want to check for `null` or `undefined`. Will change to use simple `if` sentence

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: Good to known that

Binyang2014: Fixed

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: Fixed

PR #4355, opened 2020-04-02
refine rest swagger and codes

mydmdm:


src/rest-server/docs/swagger.yaml
```diff
@@ -1,53 +1,45 @@
 openapi: 3.0.1
 info:
   title: OpenPAI RESTful API
-  description: 'Open Platform for AI RESTful API docs.'
+  description: Open Platform for AI RESTful API docs.
   license:
     name: MIT License
-    url: https://github.com/microsoft/pai/blob/master/LICENSE
+    url: "https://github.com/microsoft/pai/blob/master/LICENSE"
```
abuccts, line 7: we use single quotes `'` instead of `"` in `rest-server/` and `webportal/`, better to keep consistent

mydmdm: this style is preferred by MS swagger, however, I would convert it for consistency

mydmdm: Done

src/rest-server/docs/swagger.yaml
```diff
@@ -541,478 +725,417 @@ paths:
...
+        '404':
+          $ref: '#/components/responses/NoGroupError'
+        '500':
           $ref: '#/components/responses/UnknownError'
-  /api/v2/group/{group}:
     delete:
       tags:
-      - group
+        - group
       summary: Deelete a group in the system.
```
yiyione, line 853: typo.

PR #4358, opened 2020-04-02
[log-manager] Fix return too much log issue

### LINKED ISSUE #4356: truncate user log to 16KB in fronend
truncate user log to prevent log line too lone

Binyang2014:
fix #4356

### DIFF
--- a/src/log-manager/src/nginx/nginx.conf
+++ b/src/log-manager/src/nginx/nginx.conf
@@ -58,7 +58,7 @@ server {
       errcheck = " || echo No such file!"
       if (ngx.var[1] == "tail")
       then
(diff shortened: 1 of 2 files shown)


fanyangCS:
1. is it possible to provide some hint to indicate that this line has been truncated? e.g., "... (truncated)"
2. I assume user can still see the full log from the full log view, right?

Binyang2014:
> 1. is it possible to provide some hint to indicate that this line has been truncated? e.g., "... (truncated)"
> 2. I assume user can still see the full log from the full log view, right?

paste the ui
image

For 1. Here is the hint: last 16304 bytes
For 2. Sure user can use browse log folder to get the full log

PR #4360, opened 2020-04-03
Disable `default` vc option in edit user dialog

hzy46:
image


debuggy:
Will this default cluster select by default?

hzy46:
> Will this default cluster select by default?

It will be selected by default because every user has access to it.

PR #4362, opened 2020-04-03
[Rest Server] Change gpuType(s) to skuType(s)

abuccts:
Change gpuType(s) to skuType(s).


yqwang-ms:
Can we still support gpuType for backward compatibility? But do not mention it in our protocal and doc?

PR #4363, opened 2020-04-03
[Docs] Update job and virtual cluster api docs

abuccts:
Update job and virtual cluster api docs.

### DIFF
--- a/src/rest-server/docs/swagger.yaml
+++ b/src/rest-server/docs/swagger.yaml
@@ -420,6 +420,8 @@ paths:
       summary: Get the list of virtual clusters.
       description: Get the list of virtual clusters.
       operationId: v2getVirtualClusters
+      security:
+        - bearerAuth: []
       responses:
         '200':
           description: Succeeded
@@ -438,6 +440,8 @@ paths:
       summary: Get virtual cluster status in the system.
       description: Get virtual cluster status in the system.
       operationId: v2getVirtualCluster
+      security:
+        - bearerAuth: []
       parameters:
         - $ref: '#/components/parameters/vc'
       responses:
@@ -458,6 +462,8 @@ paths:
       summary: Get sku types in the virtual cluster.
       description: Get sku types in the virtual cluster.
       operationId: v2getVirtualClusterSkuTypes
+      security:
+        - bearerAuth: []
       parameters:
         - $ref: '#/components/parameters/vc'
       responses:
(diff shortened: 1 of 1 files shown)


src/rest-server/docs/swagger.yaml
```diff
@@ -419,14 +419,16 @@ paths:
         - virtual cluster
       summary: Get the list of virtual clusters.
       description: Get the list of virtual clusters.
-      operationId: getVirtualClusters
+      operationId: v2getVirtualClusters
```
mydmdm, line 422: In other opeations we skip `v2` because this file is for v2 only. Let's keep insistence.

src/rest-server/docs/swagger.yaml
```diff
@@ -588,26 +620,71 @@ paths:
...
+          content:
+            text/yaml:
+              schema:
+                $ref: '#/components/schemas/JobProtocol'
+        '404':
+          description: NoJobError or NoJobConfigError
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Response'
```
mydmdm, line 641: let's use `oneOf` schema to define the errors clearly, like 
```yaml
        '404':
          description: NoJobError or NoJobConfigError
          content:
            application/json:
              schema:
                oneOf:
                  - $ref: '#/components/response/NoJobError'
                  - $ref: '#/components/response/NoJobConfigError'
```

src/rest-server/docs/swagger.yaml
```diff
@@ -815,29 +892,338 @@ components:
...
-      description: TODO
-      properties: {}
     JobProtocol:
       type: object
-      description: TODO
-    JobSummary:
-      type: object
+      description: openpai job protocol
       properties:
+        protocolVersion:
```
mydmdm, line 899: how to make sure this definition is synchronzied with openpai-protocol repo?

src/rest-server/docs/swagger.yaml
abuccts: updated

src/rest-server/docs/swagger.yaml
abuccts: there's no such way currently, maybe we can use [`$ref`](https://swagger.io/docs/specification/using-ref/) after we migrate protocol repo to openapi spec

mydmdm: yes. and I think openapi spec is clear and strong enough to describe the protocol

PR #4371, opened 2020-04-07
[Restserver] Refine group api

ydye:
- Add leacy and deprecated comment for API to update singal value.
- Bind group api with /api/v2/groups
- Add api put /api/v2/groups to update a group's value.

### DIFF
--- a/src/rest-server/src/controllers/v2/group.js
+++ b/src/rest-server/src/controllers/v2/group.js
@@ -88,16 +88,22 @@ const updateGroup = async (req, res, next) => {
   try {
     if (req.user.admin) {
       let groupInfo = await groupModel.getGroup(groupname);
-      if (req.body.description !== '') {
-        groupInfo['description'] = req.body.description;
-      }
-      if (req.body.externalName !== '') {
-        groupInfo['externalName'] = req.body.externalName;
-      }
-      if (Object.keys(req.body.extension) > 0) {
-        for (let [key, value] of Object.entries(req.body.extension)) {
-          groupInfo['extension'][key] = value;
+      if (req.body.patchExtension) {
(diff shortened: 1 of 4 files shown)


src/rest-server/src/controllers/v2/group.js
```diff
@@ -83,6 +83,54 @@ const createGroup = async (req, res, next) => {
...
+  try {
+    if (req.user.admin) {
+      let groupInfo = await groupModel.getGroup(groupname);
+      if (req.body.description !== '') {
+        groupInfo['description'] = req.body.description;
+      }
+      if (req.body.externalName !== '') {
+        groupInfo['externalName'] = req.body.externalName;
+      }
+      if (Object.keys(req.body.extension) > 0) {
```
mydmdm, line 97: So we could only set an `extension` item to `null` or empty, but could not delete the key, right?

mydmdm: actually here we are mixing the use of `put` (replacing, in other fields) and `patch` (modifying partially, in extensions). Refer to [here](https://stackoverflow.com/questions/24241893/should-i-use-patch-or-put-in-my-rest-api). We should unify the way we using `put` to update a map-like properties. Here are some options:
- use `put` to replace all dict
- use `put` like `patch` for dict objects, but clarify in documents
- add a parameter (e.g. `patchExtension=<true|false>`)

ydye: > So we could only set an `extension` item to `null` or empty, but could not delete the key, right?

We can left it as ```{}```

ydye: > * patchExtension
Has added a parameter ```patchExtension```. Its default value is ```true```

PR #4376, opened 2020-04-09
Add job attempt schema in api doc

debuggy:


src/rest-server/docs/swagger.yaml
```diff
@@ -934,7 +934,7 @@ paths:
           $ref: '#/components/responses/NoJobError'
         '501':
           $ref: '#/components/responses/UnknownError'
-  '/api/v2/jobs/{user}~{job}/job-attempts/{attempt}':
+  '/api/v2/jobs/{user}~{job}/job-attempts/{attemptIndex}':
```
mydmdm, line 937: let's also rename `attempt` in parameters

debuggy: done

PR #4378, opened 2020-04-10
Update runtime and errorSpec

### LINKED ISSUE #4314: User job failed due to ECC error and failed to retry
When nvidia GPU has ECC error. The user job running on it will failed.
Since the exit reason is `Segmentation fault`. 

According to errorSpec: https://github.com/microsoft/pai/blob/e30100559ae1b0cbdd827c21547984d27198d455/src/k8s-job-exit-spec/config/k8s-job-exit-spec.yaml#L1048-L1064

The jobs failed because of `Segmentation fault` will not retry.

Consider fix this problem by:
1. For runtime, when return exit code, combine the user log with running env. Then provide reasonable retry suggestion to FC.
2. Maybe we should monitor the device-health to avoid schedule job to such device.

Binyang2014:
Bump runtime version to v0.1.1
Update errorSpec to handle ecc error

Fix #4314


src/k8s-job-exit-spec/config/k8s-job-exit-spec.yaml
```diff
@@ -926,21 +926,24 @@ spec:
...
+  - "Initialize tensorflow streamExecutor in GPU with ECC error"
   solution:
-  - "Correct the user code"
+  - "Wait result from next retry"
+  - "Contact Cluster Admin"
   pattern:
     runtimeContainerPatterns:
-    - userLogRegex: "(?i)core dumped"
+    - userLogRegex: "(?i)CUDA_ERROR_ECC_UNCORRECTABLE: uncorrectable ECC error encountered"
+    - gpuInfo:
```
yqwang-ms, line 945: Do we still need ContainerCudaUncorrectableECCError?

src/k8s-job-exit-spec/config/k8s-job-exit-spec.yaml
```diff
@@ -926,21 +926,24 @@ spec:
...
-  type: USER_FAILURE
+  causer: PAI_HW
+  type: PLATFORM_FAILURE
   stage: RUNNING
-  behavior: PERMANENT
-  reaction: NEVER_RETRY
-  reason: "Container failed due to engine core dump"
+  behavior: TRANSIENT
+  reaction: RETRY_TO_MAX
+  reason: "Container failed due GPU ecc error"
```
yqwang-ms, line 936: >iner failed due GPU ec [](start = 16, length = 22)

may be due to?

Binyang2014: Fixed

src/k8s-job-exit-spec/config/k8s-job-exit-spec.yaml
Binyang2014: Thanks, I found there already a pattern to match `CUDA_ERROR_ECC_UNCORRECTABLE`
```yaml

- code: 223
  phrase: ContainerCudaUncorrectableECCError
  issuer: PAI_RUNTIME
  causer: PAI_HW
  type: PLATFORM_FAILURE
  stage: RUNNING
  behavior: TRANSIENT
  reaction: ALWAYS_RETRY
  reason: "Container failed due to GPU uncorrectable ECC error"
  repro:
  - "Run tensorflow on a GPU which has ECC error"
  solution:
  - "Wait result from next retry"
  - "Contact Cluster Admin"
  pattern:
    runtimeContainerPatterns:
    - userLogRegex: "(?msi)CUDA_ERROR_ECC_UNCORRECTABLE.*"
```
Change name to `ContainerMayFailDueToGpuDeviceEccError` and remove `userLogRegex`

In some case ecc error with just cause segment fault/core dump without any log

src/k8s-job-exit-spec/config/k8s-job-exit-spec.yaml
```diff
@@ -926,21 +926,23 @@ spec:
...
+  causer: PAI_HW
+  type: PLATFORM_FAILURE
   stage: RUNNING
-  behavior: PERMANENT
-  reaction: NEVER_RETRY
-  reason: "Container failed due to engine core dump"
+  behavior: TRANSIENT
+  reaction: RETRY_TO_MAX
+  reason: "Container failed due to GPU ecc error"
   repro:
```
yqwang-ms, line 937: reason: "Container failed may due to GPU ecc error" ?

PR #4379, opened 2020-04-12
Refine user api

ydye:


src/rest-server/src/controllers/v2/user.js
```diff
@@ -386,6 +407,23 @@ const updateUserEmail = async (req, res, next) => {
...
 
+const updateAdminPermissionInternal = async (user, admin) => {
+  const groupInfo = await groupModel.getListGroup(user.grouplist);
+  const existed = groupModel.getAdminWithGroupInfo(groupInfo);
+  let newGroupList = [];
+  if (!existed && admin) {
+    // non-admin -> admin, add into adminGroup
+    newGroupList = [...user.grouplist, authConfig.groupConfig.adminGroup.groupname];
+  } else if (existed && !admin) {
+    for (const groupItem of groupInfo) {
```
mydmdm, line 418: add a comment 
```js
// admin -> non-admin, remove from all (internal?) adminGroup 
```

src/rest-server/src/config/v2/user.js
```diff
@@ -77,14 +77,102 @@ const userCreateInputSchema = Joi.object().keys({
...
+        otherwise: Joi.object({
+          email: Joi.required(),
+          newPassword: Joi.required(),
+        }),
+      }
+    ),
+});
+
+// define the input schema for the 'update user' api in oidc mode
+const oidcUserUpdateInputSchema = Joi.object().keys({
```
mydmdm, line 152: actually this is for `admin user` only, right? if so, let's rename it to `oidcAdminUserUpdateInputSchema`

PR #4386, opened 2020-04-14
Refine gpu chart

### LINKED ISSUE #4327: The description for X-axis is not clear and not visual optimized
there are user feedback, that the current X-axis is confusing. 

We could add one link to describe the chart and make the X-axis label simple.
i.e. description: 
The chart is to show the distribution of nodes of available GPU in nodes. 
X axis represents nodes where the number of available GPUs is X
Y axis represents the amount of the nodes

X-axis' label can be simplified to 1, 2, 3

When there are no nodes with available GPU, we should not keep the chart empty, which seems weird, we should add one nodes there saying: "No node has a GPU available.". 

image

debuggy:
closes #4327

image

### DIFF
--- a/src/webportal/src/app/home/home/gpu-chart.jsx
+++ b/src/webportal/src/app/home/home/gpu-chart.jsx
@@ -111,7 +111,9 @@ const GpuChart = ({ style, gpuPerNode, virtualClusters, userInfo }) => {
         type: 'bar',
         groups: [['shared', 'dedicated']],
         labels: {
-          format: x => (x === 0 ? '' : x),
+          format: {
+            shared: x => x
+          },
         },
         empty: { label: { text: 'No available GPU nodes now' } },
       },
--- a/src/webportal/src/app/home/home/gpu-chart.jsx
+++ b/src/webportal/src/app/home/home/gpu-chart.jsx
@@ -41,6 +41,7 @@ import {
   BREAKPOINT1,
   BREAKPOINT2,
 } from './util';
+import { TooltipIcon } from '../../job-submission/components/controls/tooltip-icon';
 
 const GpuChart = ({ style, gpuPerNode, virtualClusters, userInfo }) => {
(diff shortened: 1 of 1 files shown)


scarlett2018:
Adding @qfyin and @sunqinzheng for code review, I'm okey with the design, thanks =).

debuggy:
update: 
image

debuggy:
@scarlett could you help to rename the axis label to be more clear to users?

scarlett2018:
@debuggy - how do you think the proposal by @yangou1988 ?
https://github.com/microsoft/pai/issues/4327#issuecomment-612902332

debuggy:
> 
> 
> @debuggy - how do you think the proposal by @yangou1988 ?
> [#4327 (comment)](https://github.com/microsoft/pai/issues/4327#issuecomment-612902332)

For me, the ```Number of available GPUs``` is kind of confusing. However I am ok if you both think reasonable.

scarlett2018:
> > @debuggy - how do you think the proposal by @yangou1988 ?
> > [#4327 (comment)](https://github.com/microsoft/pai/issues/4327#issuecomment-612902332)
> 
> For me, the `Number of available GPUs` is kind of confusing. However I am ok if you both think reasonable. @scarlett2018 @yangou1988

@debuggy - how about "Number of available GPUs on the node"? accurate enough? is that too long?

debuggy:
> 
> 
> > > @debuggy - how do you think the proposal by @yangou1988 ?
> > > [#4327 (comment)](https://github.com/microsoft/pai/issues/4327#issuecomment-612902332)
> > 
> > 
> > For me, the `Number of available GPUs` is kind of confusing. However I am ok if you both think reasonable. @scarlett2018 @yangou1988
> 
> @debuggy - how about "Number of available GPUs on the node"? accurate enough? is that too long?

Better for me, here is the update:
image

PR #4400, opened 2020-04-17
[Rest Server] Fix log path length

### LINKED ISSUE #4391: Cannot submit jobs with very long job name
How to reproduce:

1. Submit a job with a very long job name.

2. The job will keep in `WAITING` status.

3. `kubectl describe po` returns:

image

image

abuccts:
Fix log path length, set job name length limit to 255.

Closes #4391.


src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -636,6 +637,7 @@ const generateFrameworkDescription = (frameworkName, virtualCluster, config, raw
       labels: {
         userName: jobInfo.userName,
         virtualCluster: jobInfo.virtualCluster,
+        logPathInfix: jobInfo.logPathInfix,
```
Binyang2014, line 640: Can put this into annotations? Maybe we don't need to filter framework according to `logPathInfix`

PR #4418, opened 2020-04-21
[Rest Server] Add rate limit for RESTful API

abuccts:
Add rate limit for RESTful API, defaults to:
* 600 all api requests per ip every 1 min
* 60 list job requests per user every 1 min
* 60 submit job requests per user every 1 hour


Binyang2014:
May need to create another PR to pai-0.18.y branch

PR #4420, opened 2020-04-22
update api version to 2.0.0

mydmdm:
Our current API version (now `0.16.0` is set by release version). However, it is confusing since we prefix APIs by `v2`. I want to change the API version to `2.0.0`, which will be decoupled with the release version. 

Then the policy of version updating could be 
- hot fixes (e.g. fixing typo, correcting formatting) will update `patch` version (e.g. `2.0.0` to `2.0.1`)
- non-breaking API changes (e.g. adding new api, supporting new params in an existing api) will update `minor` version (e.g. `2.0.0` to `2.1.0`)
- big breaking changes will change the `major` version (e.g. `2.0.0` to `3.0.0`)

### DIFF
--- a/src/rest-server/docs/swagger.yaml
+++ b/src/rest-server/docs/swagger.yaml
@@ -5,7 +5,7 @@ info:
   license:
     name: MIT License
     url: 'https://github.com/microsoft/pai/blob/master/LICENSE'
-  version: 0.16.0
+  version: 2.0.0
 externalDocs:
   description: Find out more about OpenPAI
   url: 'https://github.com/microsoft/pai'


yqwang-ms:
What is the relation between api version and protocolVersion in job spec? 
Can we align and doc them?

fanyangCS:
> What is the relation between api version and protocolVersion in job spec?
> Can we align and doc them?

protocol version is for job spec. API version is for REST APIs in rest server.

yqwang-ms:
Why job spec is not a part of our API?

mydmdm:
> Why job spec is not a part of our API?

The job spec is loosely coupled with REST API. For example, our current API version is `2.0.0` and job protocol version is `2`. Now our swagger looks like
```yaml
info:
  title: OpenPAI RESTful API
  version: 2.0.0
paths:
  /api/v2/jobs:
    post:
      requestBody:
        description: Job protocol
        content:
          text/yaml:
            schema:
              $ref: '#/components/schemas/JobProtocol'
```
Actually the component `JobProtocol` would be `JobProtocolV2` and it could refer to a external yaml file (e.g. to job protocol repo, as far as we describe the job protocol in `OpenAPI` specificaiton.

If we move forward to a newer protocol version (e.g. `3`), then the API would also evolve to support like (e.g. to `2.1.0`)
```yaml
info:
  title: OpenPAI RESTful API
  version: 2.1.0
paths:
  /api/v2/jobs:
    post:
      requestBody:
        description: Job protocol
        content:
          text/yaml:
            schema:
              oneOf:
                - $ref: '<protocol-repo>/v2/schema.yaml#/components/schemas/JobProtocol'
                - $ref: '<protocol-repo>/v3/schema.yaml#/components/schemas/JobProtocol'
```
The cluster could support both `v2` and `v3` job specs, or only support one of them. It could tell user by `get /api/v2/cluster/info` API.

yqwang-ms:
This looks like job spec is 3rd party stardard, and not a part of PAI. 
If so, why name the repo as **OpenPAI** Protocol
https://github.com/microsoft/openpai-protocol

yqwang-ms:
Offline synced with Fan, got the story, but still better to doc it since it is too different with common APIs (interfaces).

PR #4444, opened 2020-04-24
[Rest Server] Rename reserved cells to pinned cells

### LINKED ISSUE #4410: Rename reserved cells to pinned cells
In next release, remember to Rename reserved cells to pinned cells in PAI repro, after 
https://github.com/microsoft/hivedscheduler/pull/14
is merged.

**Why is this needed**:

abuccts:
Rename reserved cells to pinned cells.

Closes #4410.


yqwang-ms:
Depends on a new hived image change

PR #4449, opened 2020-04-26
[Hotfix] Update nullable fields in job api (#4437)

abuccts:
Update nullable fields in job api.

### DIFF
--- a/src/rest-server/docs/swagger.yaml
+++ b/src/rest-server/docs/swagger.yaml
@@ -1565,7 +1565,7 @@ components:
             description: job name
           username:
             type: string
-            description: user name
+            description: user name or "unknown"
           state:
             type: string
             description: general job state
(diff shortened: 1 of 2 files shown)


src/rest-server/docs/swagger.yaml
```diff
@@ -1533,22 +1533,11 @@ components:
...
-                  type: object
-                  additionalProperties:
-                    type: object
-                    properties:
-                      gpuType:
-                        type: string
-                      reservationId:
-                        type: string
-                      affinityGroupName:
-                        type: string
```
fanyangCS, line 1551: could you only hide reservationId and affinityGroupName? I think we should rename gpuType to skuType and keep it here.

abuccts: updated

PR #4460, opened 2020-04-27
Fix job attempt schema

### LINKED ISSUE #4425: Some job history API response not match schema in swagger
During the test, I found some jobs history response not match the schema in [swagger](https://github.com/microsoft/pai/blob/pai-0.18.y/src/rest-server/docs/swagger.yaml)

1. get /api/v2/jobs/{user}~{job}/job-attempts
error message:
```js
{
    keyword: 'type',
    dataPath: '[0].exitDiagnostics.runtime',
    schemaPath: '#/items/properties/exitDiagnostics/properties/runtime/type',
    params: { type: 'object' },
    message: 'should be object'
}
// Got "runtime": null
```

2. get /api/v2/jobs/{user}~{job}/job-attempts/{attemptIndex}
```js
{
    keyword: 'type',
    dataPath: '.exitDiagnostics.runtime',
    schemaPath: '#/properties/exitDiagnostics/properties/runtime/type',
    params: { type: 'object' },
    message: 'should be object'
}
// Got "runtime": null
```

### LINKED ISSUE #4434: Need document about nullable fields in API response
About the API response, we have schema in  [swagger](https://github.com/microsoft/pai/blob/pai-0.18.y/src/rest-server/docs/swagger.yaml),
Some fields in the response (e.g. "retryDelayTime" in Job details, "runtime" in Job history) could be null, but the schema didn't have that information.
I think we need some document to tell user which field in the API response could be null.

Related issues: #4424 #4425

debuggy:
closes #4425 
closes #4434

### DIFF
--- a/src/rest-server/docs/swagger.yaml
+++ b/src/rest-server/docs/swagger.yaml
@@ -2088,13 +2088,17 @@ components:
               type: string
             runtime:
               type: object
+              nullable: true
               properties:
                 exitCode:
                   type: integer
+                  nullable: true
                 originUserExitCode:
                   type: integer
+                  nullable: true
                 errorLogs:
                   type: object
+                  nullable: true
                   properties:
                     user:
                       type: string
@@ -2159,10 +2163,13 @@ components:
               properties:
                 exitCode:
                   type: integer
+                  nullable: true
                 originUserExitCode:
                   type: integer
+                  nullable: true
                 errorLogs:
                   type: object
+                  nullable: true
                   properties:
                     user:
                       type: string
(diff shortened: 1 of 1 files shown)


src/rest-server/docs/swagger.yaml
```diff
@@ -2070,14 +2070,19 @@ components:
...
+          nullable: true
         exitType:
           type: string
+          nullable: true
         exitDiagnostics:
           type: object
+          nullable: true
           properties:
             diagnosticsSummary:
               type: string
```
yiyione, line 2088: The `runtime` object in next line is also nullable.

debuggy: done

PR #4462, opened 2020-04-27
Refine job attempt schema

debuggy:
#4460

### DIFF
--- a/src/rest-server/docs/swagger.yaml
+++ b/src/rest-server/docs/swagger.yaml
@@ -2108,7 +2108,7 @@ components:
               type: string
             runtime:
               type: object
-              nullalbe: true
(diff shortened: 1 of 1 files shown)


src/rest-server/docs/swagger.yaml
```diff
@@ -2090,14 +2090,19 @@ components:
...
+          nullable: true
         exitType:
           type: string
+          nullable: true
         exitDiagnostics:
           type: object
+          nullable: true
           properties:
             diagnosticsSummary:
               type: string
```
yiyione, line 2108: The `runtime` object in next line is also nullable.

debuggy: done

Issue #4483, opened 2020-05-05
function `pod_is_ready_or_not` sometimes reports error


hzy46:
During installation,  when we wait for amd/nvidia plugin's readiness, the following errors happen from time to time.

error 1:
```
TypeError: not all arguments converted during string formatting
Call stack:
File "/root/pai/contrib/kubespray/script/openpai-generator.py", line 250, in <module>
main()
File "/root/pai/contrib/kubespray/script/openpai-generator.py", line 221, in main
wait_nvidia_device_plugin_ready()
File "/root/pai/contrib/kubespray/script/openpai-generator.py", line 153, in wait_nvidia_device_plugin_ready
while pod_is_ready_or_not("name", "nvidia-device-plugin-ds", "Nvidia-Device-Plugin") != True:
File "/root/pai/contrib/kubespray/script/openpai-generator.py", line 118, in pod_is_ready_or_not
logger.info("Ready: {1}",format(ready))
Message: 'Ready: {1}'
Arguments: ('16',)
2020-05-01 03:48:19,002 [INFO] - openpai-generator.py:154 : Nvidia-Device-Plugin is not ready yet. Please wait for a moment!
```

error 2:
```
Traceback (most recent call last):
File "/root/pai/contrib/kubespray/script/openpai-generator.py", line 250, in <module>
main()
File "/root/pai/contrib/kubespray/script/openpai-generator.py", line 222, in main
wait_amd_device_plugin_ready()
File "/root/pai/contrib/kubespray/script/openpai-generator.py", line 163, in wait_amd_device_plugin_ready
while pod_is_ready_or_not("name", "amdgpu-dp-ds", "AMD-Device-Plugin") != True:
File "/root/pai/contrib/kubespray/script/openpai-generator.py", line 109, in pod_is_ready_or_not
for container in pod.status.container_statuses:
TypeError: 'NoneType' object is not iterable 
```

PR #4487, opened 2020-05-06
Openpai-Generator issue fix.

ydye:


ydye:
https://github.com/microsoft/pai/issues/4483

PR #4493, opened 2020-05-06
[pylon] support nginx ssl in basic mode.  

ydye:
When ssl is configured in pylon configuration, ssl server in nginx will be enabled. After this PR, pylon will support ssl in basic mode.

### DIFF
--- a/src/pylon/deploy/pylon.yaml.template
+++ b/src/pylon/deploy/pylon.yaml.template
@@ -37,7 +37,7 @@ spec:
           volumeMounts:
             - mountPath: /pylon-config
               name: pylon-configuration
-{% if cluster_cfg['authentication']['OIDC'] %}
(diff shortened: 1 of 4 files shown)


src/pylon/deploy/pylon.yaml.template
```diff
@@ -51,9 +51,9 @@ spec:
               value: {{ cluster_cfg['layout']['kubernetes']['api-servers-url'] }}
             - name: CLUSTER_TYPE
               value: cluster_cfg["cluster"]["common"]["cluster-type"]
-{% if cluster_cfg['authentication']['OIDC'] %}
-            - name: AUTH_MODE
-              value: OIDC
+{% if 'ssl' in cluster_cfg['pylon'] %}
+            - name: SSL_ENABLE
+              value: true
```
hzy46, line 56: This should be "true".

PR #4510, opened 2020-05-11
[Rest-server] Using hash function to generate port number

### LINKED ISSUE #4384: Change to use Hash(podUid, portName, portIndex) to calculate the port number, avoid always retry
Currently, task port number is assigned by rest-server.

In some situation, the port number will conflict with other job.
Then it will cause job retry util conflict resolved. (since the retried job may be scheduled to the same node, port number may conflict again)

Current solution:
Runtime/Restserver use `Hash(podUid, portName, portIndex)`  to generate port number, podUid as the seed. For distributed job, since one task can get other tasks' podUid, portName and portIndex from `framework`, it can calculate other tasks' port number independently.

If port number conflict, job failed. New pods are created with different UID. Then calculate new port number.

Binyang2014:
Fix #4384
To avoid always retry due to port conflict, change to use: 
```
(int(md5(podUid + taskPortName + portIndexForPortWithSameName)[0:12] ,16) +
 int(md5(podUid + taskPortName + portIndexForPortWithSameName)[12:24] ,16) +
 int(md5(podUid + taskPortName + portIndexForPortWithSameName)[24:32] ,16)) % 
 (globalPortEnd - globalPortStart) + globalPortStart
```
to generate port number.

Where `globalPortEnd` and `globalPortStart` can be configured by admin.

Notice:
This method is `Best Efforts` to avoid port conflict. Since we use hash function to generate port the possibility for port conflict is small.

We can not guarantee:
The ports in different jobs which running on same node will not conflict. (Same as previous)
The ports which belongs to one job but in different tasks will not conflict. (Different with previous)
All allocated ports in a single task will not conflict. (Different with previous)


src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -55,6 +55,10 @@ exitSpecList.forEach((val) => {
   exitSpecMap[val.code] = val;
 });
 
+// schedule ports range is [20000, 40000)
+const portStart = 20000;
+const portEnd = 40000;
```
fanyangCS, line 60: could you make these two configuration? i.e., allows an admin/operator to specify the range during deployment.

Binyang2014: Fixed

yqwang-ms: Pls limit the Ephemeral Port Range used by k8s (and other daemons), and rand outside the limit.

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -180,10 +181,22 @@ const convertFrameworkSummary = (framework) => {
...
+  if (ports && taskStatus.attemptStatus.podUID) {
     const randomPorts = JSON.parse(ports);
-    for (let port of Object.keys(randomPorts)) {
-      containerPorts[port] = randomPorts[port].start + taskStatus.index * randomPorts[port].count;
+    if (randomPorts.ports) {
+      for (let port of Object.keys(randomPorts.ports)) {
+        const portNums = [...Array(randomPorts.ports[port].count).keys()].map((index) => {
+          const rawString = taskStatus.attemptStatus.podUID + port + index;
+          return parseInt(crypto.createHash('md5').update(rawString)
+            .digest('hex').substring(0, 12), 16) % (randomPorts.portEnd - randomPorts.portStart) + randomPorts.portStart;
```
yqwang-ms, line 191: Why not just parseInt(crypto.createHash('md5').update(rawString).digest('hex'), 64) ?
This will not lost more hash info.

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -180,10 +181,22 @@ const convertFrameworkSummary = (framework) => {
...
-  if (ports) {
+  if (ports && taskStatus.attemptStatus.podUID) {
     const randomPorts = JSON.parse(ports);
-    for (let port of Object.keys(randomPorts)) {
-      containerPorts[port] = randomPorts[port].start + taskStatus.index * randomPorts[port].count;
+    if (randomPorts.ports) {
+      for (let port of Object.keys(randomPorts.ports)) {
+        const portNums = [...Array(randomPorts.ports[port].count).keys()].map((index) => {
+          const rawString = taskStatus.attemptStatus.podUID + port + index;
+          return parseInt(crypto.createHash('md5').update(rawString)
```
yqwang-ms, line 190: Add landmark, so it would be const rawString = [ + taskStatus.attemptStatus.podUID +] + [ + port +] + [ + index +];

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: It's not rest-server part. Synced with yifan. For k8s or other daemons, we need to run command on the host to limit the port range. May need to change it in deployment part

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: It's because for js the int range is `253 - 1`. Refer to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER.

For runtime part, the `number` don't have such limit. If don't truncate the string, it will cause different port number in rest-server and runtime.

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: Sorry, don't get your point.... With/Without landmark seems does't impact anything

yqwang-ms: If no mark, port 12 + index 3 map to the same rawString 123 as port 1 + index 23.
But if use mark, the result rawString will be different, such as [12][3] and [1][23].
Let's keep the source info as more as possible, so the final port will be more rand,.

yqwang-ms:
Refine your PR description, such as int(md5(podUid + podPortName + podPortIndex)[:12] ,16) % (all_job_port_end - all_job_port_start) + all_job_port_start
and explian each var, such as all_job_port_start, all_job_port_end

yqwang-ms:
Refine description.
Using hash cannot ensure "All allocated ports in a single task will not conflict."
You can just make the possibility is very small.

src/rest-server/src/models/v2/job/k8s.js
yqwang-ms: Fold the md5 to within range

Binyang2014: Fixed

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: Fixed

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -180,10 +181,26 @@ const convertFrameworkSummary = (framework) => {
...
+  };
+  if (ports && taskStatus.attemptStatus.podUID) {
     const randomPorts = JSON.parse(ports);
-    for (let port of Object.keys(randomPorts)) {
-      containerPorts[port] = randomPorts[port].start + taskStatus.index * randomPorts[port].count;
+    if (randomPorts.ports) {
+      for (let port of Object.keys(randomPorts.ports)) {
+        const portNums = [...Array(randomPorts.ports[port].count).keys()].map((index) => {
+          const rawString = `[${taskStatus.attemptStatus.podUID}][${port}][${index}]`;
+          return hashFunc(rawString) % (randomPorts.portEnd - randomPorts.portStart) + randomPorts.portStart;
```
yqwang-ms, line 195: Will hashFunc overflow then become negative? If so, make use % in js is unsigned mod operation

Binyang2014: It will not overflow. The max int in JS is `2^53 - 1`. For the hash func. It's `48 bit int + 48 bit int + 32 bit int` The max result is a `49 bit int`

Binyang2014: It will not overflow. The max int in JS is `2^53 - 1`. For the hash func. It's `48 bit int + 48 bit int + 32 bit int` The max result is a `49 bit int`

src/rest-server/deploy/rest-server.yaml.template
```diff
@@ -93,6 +93,10 @@ spec:
           value: "{{ cluster_cfg['rest-server']['rate-limit-list-job-per-min'] }}"
         - name: RATE_LIMIT_SUBMIT_JOB_PER_HOUR
           value: "{{ cluster_cfg['rest-server']['rate-limit-submit-job-per-hour'] }}"
+        - name: SCHEDULE_PORT_START
```
yqwang-ms, line 96: Doc and comment the impact after change this config.
Such as will it impact existing running jobs, and how

Binyang2014: Sure, the rest-server config doc seems out of date. Will sync with @abuccts about how to update it.
And it will not impact running job, the backward compatibility code is added: https://github.com/microsoft/pai/blob/8e57e41e3748d01bed0e8eb76d5572868edc3e39/src/rest-server/src/models/v2/job/k8s.js#L200-L203

Binyang2014: Sure, will do it

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -180,10 +181,26 @@ const convertFrameworkSummary = (framework) => {
...
+        });
```
yqwang-ms, line 196: Bring your github pr description to here comment or doc it somewhere.

src/rest-server/src/models/v2/job/k8s.js
yqwang-ms, line 195: >portEnd [](start = 52, length = 7)

Can we also rename portEnd and portStart in code? it is also too confusing

src/openpai-runtime/build/openpai-runtime.k8s.dockerfile
```diff
@@ -15,5 +15,5 @@
 # DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-FROM openpairuntime/openpai-runtime:v0.1.1
+FROM openpairuntime/openpai-runtime:latest
```
yqwang-ms, line 18: >openpairuntime/openpai-runtime:latest [](start = 5, length = 37)

Why use latest instead of a fixed version? If so, how we define a PAI version, it cannot be fixed as the runtime is not fixed here

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -180,10 +181,26 @@ const convertFrameworkSummary = (framework) => {
 const convertTaskDetail = async (taskStatus, ports, logPathPrefix) => {
   // get container ports
   const containerPorts = {};
-  if (ports) {
+  const hashFunc = (str) => {
+    const hexStr = crypto.createHash('md5').update(str).digest('hex');
+    return parseInt(hexStr.substring(0, 12), 16) +
+      parseInt(hexStr.substring(12, 24), 16) + parseInt(hexStr.substring(24, 32), 16);
```
yqwang-ms, line 187: >, 32) [](start = 75, length = 5)

align with runtime, do not limit the string end.

src/openpai-runtime/build/openpai-runtime.k8s.dockerfile
Binyang2014: It's a temporary change. I will use fixed before PAI next release.  In my plan, runtime may bring more features in the next release, will change this line after new runtime release is published

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: Fine. Although md5 is used fixed 128 bit, I think it's same

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: Fine

src/rest-server/src/models/v2/job/k8s.js
Binyang2014: Use it to pass portStart and portEnd to runtime. The point is rest-server decide the port range and pass these information to runtime. So we don't need to maintain same config in two places.
I could change the name

src/openpai-runtime/build/openpai-runtime.k8s.dockerfile
yqwang-ms: But even in master branch or by yourself debugging/testing, it is also hard for you to figure current version when sth bad happens.

Binyang2014: In my point of view. The `latest` means the latest code version. I will update the runtime docker image after each commit. So if something bad happens, the bug is in the latest code. (I am looking for a way to build docker image automatically through docker hub after each commit).

For `debug/testing` purpose, I could use another tag, but after runtime code commit, it's better to use `latest` tag in PAI repo.

Maybe the process can be PAI-master->runtime:latest in devolving phrase.
                                           PAI-release-branch->runtime:release in Testing phrase

fanyangCS: It seems to me we need to update the build process. Why do we build the openpai-runtime in this repo, but not in https://github.com/microsoft/openpai-runtime?

yqwang-ms: The build pipeline can fail.
Anyway, we should make sure the **master** branch code is also reproducible at any time given a fixed PAI git version.

Binyang2014: @fanyangCS 
Build the openpai-runtime in this repo will give customer the ability to customize the registry mirror, and avoid pulling image failure.

@yqwang-ms 
Agree should make **master** branch be reproducible. Maybe we can use git `submodules` to solve this issue.

Another reason for this issue is runtime and rest-server are tightly coupled, there is no clear interface, and change one may break the other. Maybe give clear interface will solve this issue.

And for this case, current implement also tightly coupled, if we can make runtime be more flexible. Then we can implement this feature in PAI repo actually.
Such as provide rest-server change and a runtime plugin in PAI side.
When pai build runtime image, refer the `openpairuntime/openpai-runtime` image, copy plugin to certain folder. Then pai will have a customized runtime.

For `openpai-runtime` repo just keep common plugins and main logic.

src/rest-server/config/rest-server.md
```diff
@@ -31,66 +32,23 @@ rest-server:
...
+| rest-server.jwt-secret                 | com["rest-server"]["jwt-secret"]                 | cluster_cfg["rest-server"]["jwt-secret"]                 | String    |
+| rest-server.jwt-expire-time            | com["rest-server"]["jwt-expire-time"]            | cluster_cfg["rest-server"]["jwt-expire-time"]            | String    |
+| rest-server.default-pai-admin-username | com["rest-server"]["default-pai-admin-username"] | cluster_cfg["rest-server"]["default-pai-admin-username"] | String    |
+| rest-server.default-pai-admin-password | com["rest-server"]["default-pai-admin-password"] | cluster_cfg["rest-server"]["default-pai-admin-password"] | String    |
+| schedule-port-start                    | com["rest-server"]["schedule-port-start"]        | cluster_cfg["rest-server"]["schedule-port-start"]        | Int       |
+| schedule-port-end                      | com["rest-server"]["schedule-port-end"]          | cluster_cfg["rest-server"]["schedule-port-end"]          | Int       |
+
+## Notice
+For config `schedule-port-start` and `schedule-port-end`. If there are running jobs, change these config will cause rest server return incorrect port number.
```
yqwang-ms, line 53: Add how to config it

src/openpai-runtime/build/openpai-runtime.k8s.dockerfile
yqwang-ms: Wait for runtime release, here will change

PR #4523, opened 2020-05-13
[Rest Server] Support read only storage

### LINKED ISSUE #4503: Cannot set up a read-only storage in OpenPAI
How to reproduce:

1. Set up a ***ReadOnlyMany*** PV and corresponding PVC by:

```yaml
apiVersion: v1                                                    
kind: PersistentVolume                                            
metadata:                                                         
  name: pv-azureblob-classdata                                    
  labels:                                                         
     name: pv-azureblob-classdata                                 
spec:                                                             
  capacity:                                                       
    storage: 10Gi                                                 
  accessModes:                                                    
    - ReadOnlyMany                                                
  persistentVolumeReclaimPolicy: Retain                           
  flexVolume:                                                     
    driver: "azure/blobfuse"                                      
    secretRef:                                                    
      name: storage-hzyeurope                                     
    options:                                                      
      container: classdata                                        
      tmppath: /tmp/blobfuse                                      
      mountoptions: "--file-cache-timeout-in-seconds=120"         
---                                                               
apiVersion: v1                                                    
kind: PersistentVolumeClaim                                       
metadata:                                                         
  name: classdata                                                 
spec:                                                             
  accessModes:                                                    
    - ReadOnlyMany                                                
  storageClassName: ""                                            
  resources:                                                      
    requests:                                                     
      storage: 10Gi                                               
  selector:                                                       
    matchLabels:                                                  
      name: pv-azureblob-classdata                                
```

2. In webportal, we can see its access mode is RW:
image

3. We can write to the storage in container.

abuccts:
Support read only storage.

Closes #4503.

### DIFF
--- a/src/rest-server/src/models/v2/storage.js
+++ b/src/rest-server/src/models/v2/storage.js
@@ -60,12 +60,14 @@ const convertVolumeDetail = async (pvc) => {
       server: pv.spec.nfs.server,
       path: pv.spec.nfs.path,
     };
+    storage.readOnly = (pv.spec.nfs.readOnly === true);
     storage.mountOptions = pv.spec.mountOptions;
   } else if (pv.spec.azureFile) {
     storage.type = 'azureFile';
     storage.data = {
       shareName: pv.spec.azureFile.shareName,
     };
+    storage.readOnly = (pv.spec.azureFile.readOnly === true);
     storage.secretName = pv.spec.azureFile.secretName;
   } else if (pv.spec.flexVolume) {
     if (pv.spec.flexVolume.driver === 'azure/blobfuse') {
@@ -82,6 +84,7 @@ const convertVolumeDetail = async (pvc) => {
       storage.type = 'other';
       storage.data = {};
     }
+    storage.readOnly = (pv.spec.flexVolume.readOnly === true);
     if (pv.spec.flexVolume.secretRef) {
       storage.secretName = pv.spec.flexVolume.secretRef.name;
     }
@@ -91,6 +94,7 @@ const convertVolumeDetail = async (pvc) => {
   } else {
     storage.type = 'unknown';
     storage.data = {};
+    storage.readOnly = false;
   }
 
   if (storage.secretName) {
--- a/src/rest-server/docs/swagger.yaml
+++ b/src/rest-server/docs/swagger.yaml
@@ -2019,6 +2019,8 @@ components:
             - type: object
               description: other/unknown type
               properties: {}
+        readOnly:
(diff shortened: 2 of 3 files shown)


src/rest-server/src/models/v2/storage.js
```diff
@@ -82,6 +84,7 @@ const convertVolumeDetail = async (pvc) => {
       storage.type = 'other';
       storage.data = {};
     }
+    storage.readOnly = (pv.spec.flexVolume.readOnly === true);
```
Binyang2014, line 87: Seems we need to set PV as `readOnly` when we want to use readOnly feature.
But from the k8s API doc. The PV readOnly has different meaning. 
Refer to 
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#flexpersistentvolumesource-v1-core `ReadOnly here will force the ReadOnly setting in VolumeMounts.` So it used to mount subpath as readOnly?

My suggestion is not storage readOnly in k8s object unless it has seem meaning. How about store it in PAI group object? If so, admin can config different read/write permission for same PVC.

abuccts: > ReadOnly here will force the ReadOnly setting in VolumeMounts

it's the same meaning, only k8s does not respect the field as said in docs yet, `spec.PersistentVolume.Spec.FlexVolume.ReadOnly` is never used when mounting currently

>  How about store it in PAI group object? If so, admin can config different read/write permission for same PVC.

it will conflict with pv spec if k8s respects the field in the future. in current implementation, admin can also configuration different storage permissions for different groups

Binyang2014: Actually, my concern is we really need to set is: `pod.spec.volumes.persistentVolumeClaim.readOnly` .

But now we use `spec.PersistentVolume.Spec.<StorageType>.ReadOnly`. 

It's two different settings and we can't make sure they will have same meaning for all storage type.
Seems only has a config for `pod.spec.volumes.persistentVolumeClaim.readOnly` is more make sense and less confuse.

abuccts: `Pod.Spec.Volumes.PersistentVolumeClaim.ReadOnly` is set at
https://github.com/microsoft/pai/blob/60ae3b54acc0b52ca9a77a8eb72914f25fac2810/src/rest-server/src/models/v2/job/k8s.js#L567-L573

`PersistentVolume.Spec.<PersistentVolumeSource>.ReadOnly` is used to tell whether the PV should be mounted as read only as k8s docs described

PR #4527, opened 2020-05-13
Transfer relative path to absolute

ydye:


contrib/kubespray/script/k8s-generator.py
```diff
@@ -71,6 +71,13 @@ def generate_template_file(template_file_path, output_path, map_table):
     write_generated_file(output_path, generated_template)
 
 
+def path_handler(path):
+    if path[0] == '~':
```
mydmdm, line 75: it could be `os.path.abspath(os.path.expanduser(path))` because of `os.path.expanduser` will return unchanged path if the path does not begin with a tilde

ydye: Done

hzy46:
If we use a relative path in quick-start-kubespray.sh, like:
```bash
bash quick-start-kubespray.sh -m ./master.csv -w ./worker.csv -c ./config
```

The script still fails because there is a dir change: https://github.com/microsoft/pai/blob/a87f5520f56746984f23b4c0f40ac980e6cece34/contrib/kubespray/script/configuration.sh#L21

PR #4537, opened 2020-05-14
kubespray issue fix

ydye:


contrib/kubespray/requirement.sh
```diff
@@ -35,6 +35,7 @@ then
   echo "Pass: Cluster meets the requirements"
 else
   echo "Faild: Please check the output, and modify the cluster setting to meet the requirement"
+  exit $ret_code_check
```
hzy46, line 38: Add `rm -rf ${HOME}/pai-pre-check/` ?

ydye: Added

contrib/kubespray/roles/requirement/devbox/tasks/ubuntu.yml
```diff
@@ -37,5 +37,39 @@
   when:
     - dev_box_docker_version_replace is version('1.10.0', '<=')
 
+- name: "Dev-box 3.1 Check whether the vm can raw.githubusercontent.com"
+  raw: ping raw.githubusercontent.com -c 3
```
hzy46, line 41: Can we use a `wget` to download some file from this site? like `wget https://raw.githubusercontent.com/microsoft/pai/master/README.md`. Because sometimes we can ping it, but cannot access it.

ydye: change to curl

hzy46:
If the answer is "Y" and the network plugin is not set in config, the script will report an error:

image

ydye:
https://github.com/ansible/ansible/pull/9839 @hzy46

PR #4543, opened 2020-05-18
remove outdated job submission doc

hzy46:


hzy46:
Close this since we need to delete all docs in `docs/`

PR #4546, opened 2020-05-18
limit PAI worker daemon port range

Binyang2014:
refer to issue #4384.
Work host daemon may use port range which overlap PAI job port range. Change the work node port range to avoid conflict.


contrib/kubespray/roles/port/defaults/main.yml
```diff
@@ -0,0 +1 @@
+host_daemon_port_range: "40000 60000"
```
yqwang-ms, line 1: Current job ports range also needs to exclue low range ports, such as 80, 86, etc.
So, how about change it to 0 19999, so the available job ports range is 20000 to 65535?

Pls also check PAI other services ports, that them should be within  0 19999

contrib/kubespray/example/config.yml
```diff
@@ -67,4 +67,9 @@ docker_image_tag: <% latest-release %>
...
 # openpai_kubespray_extra_var:
 #   kay: value
-#   key: value
\ No newline at end of file
+#   key: value
+
+#######################################################################
+#                     host daemon port setting                        #
+#######################################################################
+# host_daemon_port_range: "40000 60000"
```
ydye, line 75: Although it works, I suggest to seprate the value into 2 parts and both of the have a default value. Because setting range in 1 vars is rarely seen

contrib/kubespray/roles/port/defaults/main.yml
Binyang2014: Sure, will change this value.
Checked pai services. 
hived webservice-port is 30096 by default: https://github.com/microsoft/pai/blob/bdf7f49902ebf613ff90c79ad19ae21c8cd358d9/src/hivedscheduler/config/hivedscheduler.yaml#L21

Other services are using ports within 0,19999

yqwang-ms: If so, could u pls also change hived to be 10096?

contrib/kubespray/example/config.yml
Binyang2014: fixed

contrib/kubespray/roles/port/defaults/main.yml
Binyang2014: Fixed @abuccts , please help to check

contrib/kubespray/example/config.yml
abuccts: > Although it works, I suggest to seprate the value into 2 parts and both of the have a default value. Because setting range in 1 vars is rarely seen

this is exactly what systemd does, it's one config entry

src/hivedscheduler/config/hivedscheduler.yaml
```diff
@@ -18,5 +18,5 @@
 
 service_type: "k8s"
 
-webservice-port: 30096
+webservice-port: 10096
```
abuccts, line 21: do not need to change, scheduler always runs on master node

contrib/kubespray/roles/port/tasks/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+- name: "Set host daemon port range"
+  sysctl:
+    name: net.ipv4.ip_local_port_range
+    value: "{{ host_daemon_port_start }} {{ host_daemon_port_end }}"
```
abuccts, line 5: if you edit `/etc/sysctl.conf` directly, do you need to restart network for the change to take effect by `/etc/rc.d/init.d/network restart`?

contrib/kubespray/boot.sh
```diff
@@ -56,6 +56,8 @@ ansible-playbook -i ${HOME}/pai-deploy/cluster-cfg/hosts.yml drivers-install.yml
 
 ansible-playbook -i ${HOME}/pai-deploy/cluster-cfg/hosts.yml docker-runtime-setup.yml || exit $?
 
+ansible-playbook -i ${HOME}/pai-deploy/cluster-cfg/hosts.yml set-host-daemon-port-range.yml || exit $?
```
abuccts, line 59: do you make sure the `net.ipv4.ip_local_port_range` is only configured on worker nodes?

contrib/kubespray/boot.sh
```diff
@@ -56,6 +56,8 @@ ansible-playbook -i ${HOME}/pai-deploy/cluster-cfg/hosts.yml drivers-install.yml
 
 ansible-playbook -i ${HOME}/pai-deploy/cluster-cfg/hosts.yml docker-runtime-setup.yml || exit $?
```
abuccts, line 57: same for docker runtime configuration, since master node may not have gpu, changing default runtime will break running containers

contrib/kubespray/roles/port/defaults/main.yml
abuccts: you do not need to check ports for services which only run on master nodes

contrib/kubespray/roles/port/tasks/main.yml
Binyang2014: ansible will run `/sbin/sysctl -p` automatically. Tested in bed, it take effect. Refer to https://docs.ansible.com/ansible/latest/modules/sysctl_module.html

contrib/kubespray/boot.sh
Binyang2014: Sure, only run on worker node

contrib/kubespray/boot.sh
Binyang2014: Tested, it only run on worker node.

src/hivedscheduler/config/hivedscheduler.yaml
Binyang2014: Sure, it only runs on master node. But limit it to 1024-19999 for consistency.

contrib/kubespray/example/config.yml
Binyang2014: Anyway, all has same effect. Separate  config is more clear for me

contrib/kubespray/roles/port/defaults/main.yml
Binyang2014: Refer to https://kubernetes.io/docs/concepts/services-networking/service/#nodeport, k8s service node port range is: ` 30000-32767. This change will break CI.

If it only run on master node, will revert this change

contrib/kubespray/preinstall.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+while getopts "c:" opt; do
+  case $opt in
+    c)
+      CLUSTER_CONFIG=$OPTARG
+      ;;
+    \?)
```
abuccts, line 8:
```suggestion
    *)
```

contrib/kubespray/roles/port/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+host_daemon_port_start: 1024
+host_daemon_port_end: 19999
```
abuccts, line 2: I found this is not a good range, for example, job exporter uses 9102 port on worker nodes, if kubelet occupies this port during k8s deployment, job exporter won't be started anymore.

It's better to avoid port conflicts with system/pai services, that's why the default value starts from 32768.

yqwang-ms: Let's reduce the ports range of non-job ports, so job ports range can be larger and less possible to conflict

yqwang-ms: May be we can tune this to be 10000-19999?

abuccts: another choice is using `net.ipv4.ip_local_reserved_ports` instead, you can refer to [IP Variables](https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt) for details. you can also reuse the job port range for this config.

---

when calling `connect()` or `bind()` with port number 0, OS will choose port which which meets all of the following:
1. in `net.ipv4.ip_local_port_range`
2. not in `net.ipv4.ip_local_reserved_ports`
3. not **currently** in use

yqwang-ms: But each port need to specify in ip_local_reserved_ports, and we have too many services. A range is easy to maintain?

abuccts: 1. let `net.ipv4.ip_local_port_range` uses default range 32768~60999
2. we set job port range in `net.ipv4.ip_local_reserved_ports`, e.g. 20000~40000
3. services use static ports, no need to reserve, kubelet will be assigned a free port between 40000~60999 when binding to port 0

yqwang-ms: What I meant is ip_local_reserved_ports is not a range.

Binyang2014: How about just set `net.ipv4.ip_local_port_range` to 40000~60000. And we assume:
PAI-service use port within 1024-19999.
PAI-job use port within 20000-39999
Random port is: 40000-60000

For PAI service, we need to make sure the port number doesn't conflict with system service.
Random port and PAI job port will not conflict according to this config.

yqwang-ms: Agree. But lets enlarge PAI-job ports range, such as 
PAI-service use port within 1024-**9999** (given we only have very few services)
PAI-job use port within 10000-39999
Random port is: 40000-**65535**

yqwang-ms: @abuccts just saw ip_local_reserved_ports  can also specify range, like 1-39999

abuccts: > What I meant is ip_local_reserved_ports is not a range.

it is a *list of comma separated ranges*, so it can be one range, or comma separated multiple ranges

> How about just set net.ipv4.ip_local_port_range to 40000~60000

it's OK

> PAI-job use port within 10000-39999, Random port is: 40000-65535

kubelet also uses static ports 10249 and 10250
to enlarge job ports, it's better to use 20000~50000

yqwang-ms: Random port will also impacts jobs as they may use.
So lets make Random port and PAI-job port both large.
If you do not want to change kubelet ports, we can use range:
PAI-service use port within 1024-15000 (given we only have very few services)
PAI-job use port within 15000-39999
Random port is: 40000-65535

contrib/kubespray/preinstall.sh
Binyang2014: According to https://sookocheff.com/post/bash/parsing-bash-script-arguments-with-shopts/, it works

contrib/kubespray/roles/port/defaults/main.yml
Binyang2014: Fine for me, will change random port to `40000-65535`

abuccts: seems fine

PR #4553, opened 2020-05-21
Update architecture pic in readme

debuggy:
https://github.com/microsoft/pai/tree/mintao/arch-pic#


hzy46:
Please don't delete arch_v2.png because system_architecture.md needs it.

PR #4556, opened 2020-05-22
Cherry pick to master - Update PAI version

hzy46:


version/K8S.VERSION
```diff
@@ -1 +1 @@
-v0.17.0-1.9.9
\ No newline at end of file
+v1.0.0-1.9.9
```
Binyang2014, line 1: Maybe we can remove this file, I remember this is for yarn version, and doesn't need for now.

hzy46: Remove it in the next release? I'm not sure if the removal will cause any problem.

Binyang2014: Sure, fine for me

PR #4558, opened 2020-05-25
[storage] add write permission for dir under home folder

Binyang2014:


fanyangCS:
shall we deploy it as a hotfix to 1.0?

Binyang2014:
> shall we deploy it as a hotfix to 1.0?

With sync with @hzy46  about this

PR #4561, opened 2020-05-25
Configurable urls for contents in raw.githubusercontent.com

hzy46:


contrib/kubespray/example/config.yml
```diff
@@ -74,3 +74,10 @@ docker_image_tag: <% latest-release %>
 #######################################################################
 # host_daemon_port_start: 40000
 # host_daemon_port_end: 65535
+
+
+# Change the following urls if you cannot access them
+# nvidia_device_plugin_yaml_url: "https://raw.githubusercontent.com/NVIDIA/k8s-device-plugin/1.0.0-beta4/nvidia-device-plugin.yml"
+# amd_device_plugin_yaml_url: "https://raw.githubusercontent.com/RadeonOpenCompute/k8s-device-plugin/master/k8s-ds-amdgpu-dp.yaml"
+# rdma_config_yaml_url: "https://raw.githubusercontent.com/Mellanox/k8s-rdma-sriov-dev-plugin/master/example/hca/rdma-hca-node-config.yaml"
+# rdma_device_plugin_yaml_url: "https://raw.githubusercontent.com/Mellanox/k8s-rdma-sriov-dev-plugin/master/example/device-plugin.yaml"
```
abuccts, line 83: device plugin is a service, why couple with k8s deployment?

contrib/kubespray/quick-start/services-configuration.yaml.template
```diff
@@ -228,3 +228,18 @@ authentication:
...
+  {% endif %}
+  {% if "amd_device_plugin_yaml_url" in env["cfg"] %}
+  amd_device_plugin_yaml_url: "{{ env['cfg']['amd_device_plugin_yaml_url'] }}"
+  {% endif %}
+  {% if "rdma_config_yaml_url" in env["cfg"] %}
+  rdma_config_yaml_url: "{{ env['cfg']['rdma_config_yaml_url'] }}"
+  {% endif %}
+  {% if "rdma_device_plugin_yaml_url" in env["cfg"] %}
+  rdma_device_plugin_yaml_url: "{{ env['cfg']['rdma_device_plugin_yaml_url'] }}"
+  {% endif %}
```
abuccts, line 245: these lines will never be used in this template

contrib/kubespray/example/config.yml
hzy46: It is the configuration for the quick start script, not only for k8s deployment.

contrib/kubespray/quick-start/services-configuration.yaml.template
hzy46: If you configure `nvidia_device_plugin_yaml_url` in config.yml, they will be used

hzy46:
Close to use #4563

PR #4563, opened 2020-05-25
[Device Plugin] Update device plugin yaml download script

abuccts:
Update device plugin yaml download script, use svn client instead.


abuccts:
@hzy46 pls help to check whether it works in your network

hzy46:
> @hzy46 pls help to check whether it works in your network
> 
> resolve #4561, will fallback to git submodules if this does not work

It works. I will close #4561.

PR #4591, opened 2020-06-01
[Runtime] Fix private registry issue

Binyang2014:
Bump runtime version to `v0.1.1-hp1`


Binyang2014:
Version wrong, close it

Issue #4586, opened 2020-05-29
Feedback v1.0.0 Private Registry Use Problem


xulzee:
Hello,
   I checked the help document about how to use private registry, but I found that when I used to use v0.14.0, I could use 
`192.168.1.106:5000/cheng Tong/ pytorch:1.3-cuda10.0-cudnn7-runtime-dali `
But in v1.0.0, I need to fill in all username, passwd and registryuri in auth, but my private registry does not need password.

Is there any way to leave some options blank?

abuccts:
what error did you get if you only specify `uri: 192.168.1.106:5000/...` without `auth` field in job config?

xulzee:
I broke my system (added a node), so I can't give you detailed logs. =，=

The approximate output is,  **can't find 192.168.1.106:5000/... in docker.index.io**

xulzee:
The error is 
```
[Exit Trigger Info]

ExitTriggerMessage: FailedTaskCount 1 has reached MinFailedTaskCount 1 in the TaskRole
ExitTriggerTaskRole: taskrole
ExitTriggerTaskIndex: 0

--------------------------------------------------------------------------------

[Exit Spec]

code: 254
phrase: ContainerImageNotExists
issuer: PAI_RUNTIME
causer: PAI_RUNTIME
type: USER_FAILURE
stage: LAUNCHING
behavior: PERMANENT
reaction: NEVER_RETRY
reason: 'Pull access denied, repository does not exist or may require authentication'
repro:
  - Submit job with incorrect docker image uri
solution:
  - Correct docker image uri in job protocol


--------------------------------------------------------------------------------

[Exit Diagnostics]

Pod failed: PodPattern matched:
containers:
  - name: init
    message: >

      + python /usr/local/pai/init.d/image_checker.py
      /usr/local/pai/runtime.d/job_config.yaml
      /usr/local/pai/secrets/secrets.yaml

      2020-05-30 14:10:33,234 - INFO - image_checker.py:236 - get job config
      from /usr/local/pai/runtime.d/job_config.yaml

      2020-05-30 14:10:33,239 - INFO - image_checker.py:246 - Start checking
      docker image

      2020-05-30 14:10:33,242 - DEBUG - connectionpool.py:959 - Starting new
      HTTPS connection (1): index.docker.io:443

      2020-05-30 14:10:34,563 - DEBUG - connectionpool.py:437 -
      https://index.docker.io:443 "HEAD /v2/ HTTP/1.1" 401 0

[...]

PR #4592, opened 2020-06-01
Fix private image pull issue

Binyang2014:
fix: https://github.com/microsoft/pai/issues/4586


Binyang2014:
Fix: https://github.com/microsoft/pai/issues/4586

PR #4595, opened 2020-06-02
Fix setuptools version in dev-box

hzy46:
According to https://setuptools.readthedocs.io/en/latest/python%202%20sunset.html, Python2 is not supported by setuptools. This PR pins setuptools version to lower than 45.


hzy46:
> I suggest to fix the version, i.e., == 45, instead of < 45

Sure. It is better to have a fixed number. I changed it to `44.1.0` because setuptools>= 45 is not recommended for python2

PR #4597, opened 2020-06-02
Change to use hostNetwork for rest-server

Binyang2014:
To avoid CNI issue, change to use host network


src/rest-server/deploy/rest-server.yaml.template
```diff
@@ -29,7 +29,7 @@ spec:
       labels:
         app: rest-server
     spec:
-      hostNetwork: false
+      hostNetwork: true
```
abuccts, line 32: could you fix the cni issue instead? it's not a good practice to use `hostNetwork`

src/rest-server/deploy/rest-server.yaml.template
```diff
@@ -144,7 +146,7 @@ spec:
 {% endif %}
         ports:
         - name: rest-server
-          containerPort: 8080
+          containerPort: 9186
```
abuccts, line 149: why specify `ports` if you use `hostNetwork`

src/rest-server/deploy/rest-server.yaml.template
Binyang2014: CNI issue is not easy to fix and we don't have a solution yet. Waiting for @ydye investigation.

src/rest-server/deploy/rest-server.yaml.template
Binyang2014: Will remove this

Binyang2014:
> won't it affect other services? e.g. pylon

Currently, it only impact the service which need to access outside network. Seems other service works fine. I think we can put the PR here for deploy reference. Not hurry to merge.

Binyang2014:
Close since we can remove CNI to fix this issue

PR #4620, opened 2020-06-14
[Rest Server] Add examples to swagger

yiyione:


mydmdm:
Let's follow the actions of [Semantic versioning](https://semver.org/spec/v2.0.0.html) from now, which uses `major.minor.patch` as versioning and 
>[ Any modifications MUST be released as a new version.](https://semver.org/spec/v2.0.0.html#spec-item-3)

Since only examples are added in this PR, and I think you may change the version to `2.0.1` like 
```yaml
info:
  title: OpenPAI RESTful API
  description: >
    Open Platform for AI RESTful API docs.
    Version 2.0.1: add more examples and fix typos
  license:
    name: MIT License
    url: 'https://github.com/microsoft/pai/blob/master/LICENSE'
  version: 2.0.1
```
you may diff current version with the version in `pai-0.18-y` for a precise versioning info.

PR #4634, opened 2020-06-19
pgjson: multi-threading, one conneciton for each thread

suiguoxin:
Approach:
- Create one connection for each thread, use thread local variable to store it. 

Exp:
- In comparison to single-thread version, multi-threading can speed up more than 100 times.


suiguoxin:
Included in https://github.com/microsoft/pai/pull/4716

PR #4652, opened 2020-06-28
[Rest Server] Fix default parameter of list storages API

yiyione:
Fix call `/api/v2/storages?default=false` will even only get default storages.

### DIFF
--- a/src/rest-server/src/controllers/v2/storage.js
+++ b/src/rest-server/src/controllers/v2/storage.js
@@ -1,19 +1,5 @@
-// Copyright (c) Microsoft Corporation
-// All rights reserved.
-//
-// MIT License
-//
-// Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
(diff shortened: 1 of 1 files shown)


src/rest-server/src/controllers/v2/storage.js
```diff
@@ -22,7 +8,7 @@ const storage = require('@pai/models/v2/storage');
 
 const list = asyncHandler(async (req, res) => {
   const userName = req.user.username;
-  const filterDefault = Boolean(req.query.default);
+  const filterDefault = (req.query.default === 'true');
```
abuccts, line 11: the original endpoint is to use `/api/v2/storages?default&...`

yiyione: @abuccts I think the change will break `/api/v2/storages?default&...`, is there anyone still use this query ?

abuccts: no

PR #4672, opened 2020-07-06
[Doc] Add steps to remove weave-net

Binyang2014:


ydye:
Is it possible to write an ansible role to do this work?

Binyang2014:
> Is it possible to write an ansible role to do this work?

I am not sure. The `kubectl` need to run on dev-box, Seems we need to add `localhost` in the `inventory`.
The `kubectl` command only need to run once. It's not time consuming. And admin should have basic knowledge about k8s.

ydye:
> > Is it possible to write an ansible role to do this work?
> 
> I am not sure. The `kubectl` need to run on dev-box, Seems we need to add `localhost` in the `inventory`.
> The `kubectl` command only need to run once. It's not time consuming. And admin should have basic knowledge about k8s.

Agree with you about the point that admin should have knowledge of k8s.

docs/manual/cluster-admin/installation-faqs-and-troubleshooting.md
```diff
@@ -60,18 +60,38 @@ To remove the network plugin, you could use following `ansible-playbook`:
...
+    
+    - name: restart docker
+      shell: systemctl restart docker
       args:
         executable: /bin/bash
 ```
 
-After this step, if your pod still can not access internet, please change the pod spec to use `hostNetwork`.
+After this step you need to change the `coredns` to fix dns resolution issue.
+Please run `kubectl edit cm coredns -n kube-system -o yaml`, change `.:53 {` to `.:9053`
```
abuccts, line 92: what is ` {`

Binyang2014: Thanks, it's typo. Fixed

PR #4683, opened 2020-07-08
Mintao/fix exitspec

debuggy:


src/webportal/src/app/job/job-view/fabric/job-detail/components/task-role-container-list.jsx
```diff
@@ -450,7 +454,7 @@ export default class TaskRoleContainerList extends React.Component {
 
                 // diagnostics
                 const diag = item.containerExitDiagnostics;
-                if (diag) {
+                if (!isNil(diag)) {
```
Binyang2014, line 457: When the Diagnostics info will be null

debuggy: when the job is just submitted, there is chance that diag is null?

PR #4692, opened 2020-07-09
[webportal] Fix permission UI of read-only storage

### LINKED ISSUE #4688: Read-only storage's permission is still `RW`
How to reproduce:

1. Set up a read-only storage by setting `PersistentVolume.Spec.<PersistentVolumeSource>.ReadOnly=true`:

2. In webportal, the permission for this new storage is `RW`:
image

yiyione:
fix #4688

After fix:
fix-readonly-1
fix-readonly-2


hzy46:
Can we use `RO` instead of `R`? It is easier to understand `RO` because Kubernetes and Linux also uses it.

yiyione:
> Can we use `RO` instead of `R`? It is easier to understand `RO` because Kubernetes and Linux also uses it.

Done.

PR #4702, opened 2020-07-13
[log-manager] reconfigure log-manager

Binyang2014:
1. Change log-manager QoS class to Burstable
2. Auto rotate the log file which size more than 300M
3. The rotated log will keep no more than 30 days

### DIFF
--- a/src/log-manager/src/logrotate/logrotate.conf
+++ b/src/log-manager/src/logrotate/logrotate.conf
@@ -1,12 +1,15 @@
-/usr/local/pai/logs/*/*/*/*/*.* {
+/usr/local/pai/logs/*/*/*/*/*.log
+/usr/local/pai/logs/*/*/*/*/*.stdout
+/usr/local/pai/logs/*/*/*/*/*.stderr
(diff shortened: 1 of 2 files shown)


src/log-manager/src/logrotate/logrotate.conf
```diff
@@ -1,12 +1,15 @@
-/usr/local/pai/logs/*/*/*/*/*.* {
+/usr/local/pai/logs/*/*/*/*/*.log
+/usr/local/pai/logs/*/*/*/*/*.stdout
+/usr/local/pai/logs/*/*/*/*/*.stderr
+/usr/local/pai/logs/*/*/*/*/*.all
+{
     nomail
-    olddir /usr/local/pai/logs-backup
     weekly
```
abuccts, line 7: what is "weekly" for?

Binyang2014: Weekly means log file will be rotated weekly. So if there is a log file which exist more than one week, it will be rotated

Binyang2014: Change it to monthly for long running job

PR #4727, opened 2020-07-16
Change api time return null instead of 0

debuggy:


yqwang-ms:
Create a common function for this?
@abuccts to comment

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -167,8 +167,8 @@ const convertFrameworkSummary = (framework) => {
       resource: 0,
     },
     retryDelayTime: framework.status.retryPolicyStatus.retryDelaySec,
```
yqwang-ms, line 169: Also fix job history part?

yqwang-ms:
why so many other commits? Could you compact to 1 commit and force push?

debuggy:
> 
> 
> why so many other commits? Could you compact to 1 commit and force push?

because I merge commits from fix-time-error-release branch, it will be squashed when merged

debuggy:
> Could you compact to 1 commit and force push?

You could only take care of file changes in all commits, because in this pr only files changed matters. And I will squash all commits into one before merge into master, it equals to ```compact to 1 commit and force push```

PR #4735, opened 2020-07-17
Azure Easy Deployment Based On AKS-Engine

ydye:


contrib/azure/azure.py
```diff
@@ -0,0 +1,362 @@
+#!/usr/bin/env python
+
```
abuccts, line 2: license?

ydye: added

ydye:
> rename `contrib/azure` to `contrib/aks-engine`?

Done

PR #4757, opened 2020-07-22
[Rest Server] Update hived cell number calculation

abuccts:
Update hived cell number calculation and refine error messages.


src/rest-server/src/middlewares/v2/hived.js
```diff
@@ -216,16 +217,24 @@ const hivedValidate = async (protocolObj, username) => {
...
+    const {gpu = 0, cpu, memoryMB} = protocolObj.taskRoles[taskRole].resourcePerInstance;
+    if (resourcePerCell.gpu === 0 && gpu > 0 ||
+        resourcePerCell.cpu === 0 && cpu > 0 ||
+        resourcePerCell.memory === 0 && memoryMB > 0) {
       throw createError(
         'Bad Request',
         'InvalidProtocolError',
-        `Taskrole ${taskRole} requests ${gpu} GPU, ${cpu} CPU, ${memoryMB}MB memory; ` +
-        `sku allows ${resourcePerCell.gpu} GPU, ${resourcePerCell.cpu} CPU, ${resourcePerCell.memory}MB memory per cell.`
+        `Taskrole ${taskRole} requests (${gpu} GPU, ${cpu} CPU, ${memoryMB}MB memory), ` +
```
scarlett2018, line 227: "Minimum SKU" limitation is confusing. May we create a doc to introduce why we introduce "minimum sku" limit for users reference?

src/rest-server/src/middlewares/v2/hived.js
```diff
@@ -234,7 +243,7 @@ const hivedValidate = async (protocolObj, username) => {
     throw createError(
       'Bad Request',
       'InvalidProtocolError',
-      `Exceed ${cellQuota} GPU quota in ${virtualCluster} VC.`
+      `Job requests ${requestCellNumber} sku, violate ${cellQuota} sku [${cellUnitsStr}] quota in ${virtualCluster} VC.`
```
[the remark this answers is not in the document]
abuccts: maximum and quota are duplicate, updated

src/rest-server/src/middlewares/v2/hived.js
abuccts: removed

scarlett2018: @abuccts - after removed that limit, the sentence seems not providing any actions to the user? we shall provide an actionable follow up to the user, so that they can fix the error by themselves.

abuccts: `Taskrole worker requests (1 GPU, 4 GPU, 4096MB memory), but SKU does not configure GPU` means user couldn't request GPU, the config file should contain 0 GPU.

scarlett2018: Taskrole worker requests 1 _GPU_, but there are SKU does not configure _GPU_. Please contact admin if the taskrole need _GPU_ resources.

PR #4759, opened 2020-07-22
New RestServer Architecture: RestServer -> DB -> ApiServer; P0 Items

hzy46:
# Database Controller Test Cases

  - [End-to-end Test](#test-jobs)
  - [Path Test](#path-test)
  - [Upgrade Test](#upgrade-test)
  - [Stress Test](#stress-test)

### Test Jobs

- [x] A job with simple output

    Case: job command `echo success`

    Expect: Succeed, output `success`.

- [x]  Stop a job

    Case: job command `sleep 1h`; then stop this job

    Expect: The job can be stopped.

-  [x] Job with retries

    Case: job command: `exit 1`; set max retry to be 10

    Expect: The job retry history can be viewed.

- [x] Job with docker secret

    Case: Submit a job with docker auth info

    Expect: The job can run successfully.

- [ ] Job with priority class

    Case: Submit a job with priority class

    Expect: The job priority class is correct.

- [x] Job with secret

    Case: Submit a job with secret info

    Expect: The secret info is successfully written into job.

- [ ] Bulk job

    Case: Submit a job with 1000+ instances.

    Expect: The job can run successfully.

- [ ] Bulk job with retries

    Case: Submit a job with 1000+ instances, and 50+ retries.

    Expect: The job can run successfully. The job retry history can be viewed.

### Test in Certain Case

- [x] Database failure

    Case: submit a job; shutdown the database; try to submit another job; start the database

    Expect: the first job is not affected; the second job cannot be submitted.

- [x] Database controller failure

    Case: submit a job; shutdown the database controller; try to submit another job; start the database

    Expect: the first job is not affected; the second job cannot be submitted.

- [x] Framework controller failure

    Case: shutdown the framework controller; try to submit a job; start the framework controller

    Expect: the job can be submitted, but in `WAITING` status. After the framework controller is started, its state will turn to `RUNNING`.

#### Rest-server checks conflicts

- [x] Case: Submit a job twice

    Expect: The second job submission is not allowed.

- [x] Case: Stop a non-existed job

    Expect: The operation is not allowed.

#### (Write merger) Add/Update FR (If not equal, override and mark !requestSynced, else no-op)

- [x] Environment: Write-merger doesn't forward request; Shutdown database poller;

    Case: Fake two same framework requests to write-merger.

    Expect: The second request will be ignored.

#### (Write merger) If recovery mode, use the FR from API server

- [x] Environment: Turn on recovery mode

    Case: Create a framework directly in API server;

    Expect: The framework is inserted in database.

- [x] Environment: Turn off recovery mode

    Case: Create a framework directly in API server;

    Expect: The framework can't be inserted in database, write merger reports 404 error, and watcher keeps trying to synchroinze it.

#### (Write merger) If not equal, override and mark as !requestSynced

- [x] Environment: Shutdown database poller

    Case: Submit a job; Then change its requestGeneration in API server.

    Expect: The framework is marked as `requestSynced=false`.

#### (Database poller) If API server 404 error, mock a delete Framework

- [x] Case: Submit A job; Then shutdown watcher after the job starts; Wait until the job succeeded; Stop poller; Start watcher; Stop watcher; Delete this framework manually; Start Poller

    Expect:

    After watcher is restarted, the job is marked as `state=Completed` and `requestSynced=true`;
    After poller is restarted, it can mock a delete framework event to write merger.
    Finally, the job will be marked as `apiServerDeleted=true`.

## Upgrade Test

- [ ] Upgrade from `v1.0.y`

    Drop the database; Deploy a `v1.0.y` bed, submit some jobs (must include: one running job, one completed job with retry history, and one completed job without retry history). Then upgrade it. Make sure the job information is correct, and running jobs are not affected.

- [ ] Upgrade from `v1.1.y`

    Drop the database; Deploy a `v1.1.y` bed, submit some jobs (must include: one running job, one completed job with retry history, and one completed job without retry history). Then upgrade it. Make sure the job information is correct, and running jobs are not affected.

## Stress Test

- [ ] Submit 100000+ jobs in 1 hour. It should be handled properly.


hzy46:
Closed: try to solve a CI problem.

Issue #4694, opened 2020-07-10
User job write too much logs will cause disk pressure


Binyang2014:
PAI will keep user job log under /var/log/pai.
If user job write too much logs, it will cause machine disk pressure.

We need to:
1. Make the log path configurable, then we can store user log into a large disk
2. Investigate how to kill such offence job

fanyangCS:
relate to https://github.com/microsoft/pai/issues/3765 and https://github.com/microsoft/pai/issues/3340

fanyangCS:
@binyang2014, please firstly make sure the OpenPAI service pods are of higher QoS class than job pods. In some case the service pods get evicted.

Binyang2014:
> @Binyang2014, please firstly make sure the OpenPAI service pods are of higher QoS class than job pods. In some case the service pods get evicted.

We may need to mark these pods as critical to achieve this: https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/

Binyang2014:
Checked the QoS class. Currently, job-exporter, node-exporter qos are `Burstable`, log-manager qos class is `BestEffort`  and pod user job qos class is `Guaranteed`.
For k8s node eviction, `BestEffort` will be first evicted. `Guaranteed` and `BestEffort`  pods whose usage is beneath requests are evicted last. (https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/?spm=a2c65.11461447.0.0.6a497eafo7oGQp#evicting-end-user-pods)

For `Guaranteed` and `BestEffort`  pods which resource usage is not exceed their requests, will ordered by pod priority.


For this case, the resource is disk, and all pod don't claim the requests for the disk. So the eviction order will rank by pod priority, then resource usage.
We can get the eviction order from the log:
```
eviction manager: must evict pod(s) to reclaim ephemeral-storage
eviction_manager.go:362] eviction manager: pods ranked for eviction: user-job-pod, job-exporter-zgf4j_default(e9e0a4a5-660a-493b-ac4b-a95f8977867a),
nginx-proxy-prodk80bg000012_kube-system(6d20234d7a8eda76fb23d52d6f743b77),
log-manager-ds-9tflc_default(67956c0d-ba89-4aa9-a14e-ed2fbbcd915e),
k8s-host-device-plugin-daemonset-kqh67_kube-system(8853e257-8f09-46ca-b372-2632cb94eea5),
[...]

PR #4775, opened 2020-07-31
make pai daemon with high priority

Binyang2014:
Assign PAI daemon with high priority. Make sure PAI service will not be evicted before user pod.
According to: https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/#evicting-end-user-pods
The kubelet ranks Pods for eviction first by resource, then by priority.
We already changed pai daemon service QoS class to Burstable and make sure it will not exceed the limit resource.

### DIFF
--- a/src/cluster-configuration/deploy/priority-class.yaml
+++ b/src/cluster-configuration/deploy/priority-class.yaml
@@ -1,0 +1,7 @@
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: pai-daemon-service-priority
+value: 1000000
+globalDefault: false
+description: "This priority class should be used for PAI daemon service only."
--- a/src/cluster-configuration/deploy/priority-class.yaml
+++ b/src/cluster-configuration/deploy/priority-class.yaml
@@ -1,7 +1,7 @@
 apiVersion: scheduling.k8s.io/v1
 kind: PriorityClass
 metadata:
-  name: pai-daemon-service-priority
+  name: pai-daemon-priority
 value: 1000000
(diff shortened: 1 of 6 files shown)


src/cluster-configuration/deploy/priority-class.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: pai-daemon-service-priority
+value: 1000000
+globalDefault: false
+description: "This priority class should be used for PAI daemon service only."
```
abuccts, line 7: could you use the built in priority `system-node-critical`?

Binyang2014: This priority is limited to `kube-system` namespace. We can use `system-node-critical`priority in other namespace after 1.1.7 version.

Refer to
https://github.com/kubernetes/kubernetes/issues/60596#issuecomment-559537077

src/cluster-configuration/deploy/priority-class.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: pai-daemon-service-priority
```
abuccts, line 4: daemon and service are duplicate, pls keep only one term

Binyang2014: fixed

Binyang2014:
related with: https://github.com/microsoft/pai/issues/4694

PR #4782, opened 2020-08-04
refine exit spec

Binyang2014:
For regex, `s` means `single line mode` which will treat all logs as a single line.
So if pattern contains `.*` at the end of pattern, it will match all logs to the end of log file.
This will cause matching string in exitInfo contains too much logs and confuse end user.

Change to only use multi-line mode in this case


Binyang2014:
related to https://github.com/microsoft/openpai-runtime/issues/18

yqwang-ms:
change msi -> mi for all places?

Issue #4777, opened 2020-08-03
Cannot find the stdout/stderr log of a long run job


zheng-ningxin:
**Organization Name**: MS internal usage

**Short summary about the issue/question**:
Hi, I met a problem when I used the OpenPAI to run the training jobs, and I think it may confuse other users, so I raise this issue.
I submitted a task, after the task is completed, I want to check the stdout of this task, unexpectedly found that the output of the task was empty. 

![image](https://user-images.githubusercontent.com/49771382/89157452-21b13780-d59f-11ea-83a1-61191c32ca3a.png)

Later, it was found that the output of the task was compressed into the log folder, and the website page no longer displayed the output of the task.
![image](https://user-images.githubusercontent.com/49771382/89157857-c3388900-d59f-11ea-88a2-5287fb315956.png)

I think this is confusing for users, so I raise this issue.
**Brief what process you are following**:

**How to reproduce it**:

**OpenPAI Environment**:
- OpenPAI version:
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Hardware (e.g. core number, memory size, storage size, GPU type etc.):
- Others:

**Anything else we need to know**:

fanyangCS:
@zheng-ningxin , please share us the job url. 
@debuggy , this seems to be a bug?

zheng-ningxin:
Hi~ the url is at: https://ne.openpai.org/job-detail.html?username=nizhen&jobName=nizhen___mobilenet_v2_l1_07_0001

fanyangCS:
@Binyang2014 , pls take a look.

Binyang2014:
This is caused by log-rotate . Current log-rotate config is:
```
{
    nomail
    weekly
    rotate 4
    compress
    missingok
    notifempty
    copytruncate
    maxsize 300M
    maxage 30
}
```
It will auto compress the log if size is more than 300M or exist more than one week. This work start at 7.28, which is one-week ago.
So log is rotated.

fanyangCS:
> This is caused by log-rotate . Current log-rotate config is:
> 
> ```
> {
>     nomail
>     weekly
>     rotate 4
>     compress
>     missingok
>     notifempty
>     copytruncate
>     maxsize 300M
>     maxage 30
> }
> ```
> 
> It will auto compress the log if size is more than 300M or exist more than one week. This work start at 7.28, which is one-week ago.
> So log is rotated.

[...]

PR #4792, opened 2020-08-07
Change logrotate config

Binyang2014:
We only rotate the log if log size is more than 100MB. Max keep 2 logs, if user still write log, the old log will be dropped.
log-rotate will run every 10 mins

### DIFF
--- a/src/log-manager/src/logrotate/logrotate.conf
+++ b/src/log-manager/src/logrotate/logrotate.conf
@@ -9,6 +9,6 @@
     missingok
     notifempty
     copytruncate
-    size 100M
+    size 256M
     maxage 30
 }
--- a/src/log-manager/src/logrotate/logrotate.conf
+++ b/src/log-manager/src/logrotate/logrotate.conf
@@ -4,12 +4,11 @@
 /usr/local/pai/logs/*/*/*/*/*.all
 {
     nomail
-    weekly
-    rotate 4
-    compress
+    rotate 1
+    nocompress
     missingok
     notifempty
     copytruncate
-    maxsize 300M
+    size 100M
     maxage 30
 }
--- a/src/webportal/src/app/job/job-view/fabric/job-detail/conn.js
+++ b/src/webportal/src/app/job/job-view/fabric/job-detail/conn.js
@@ -176,7 +176,7 @@ export async function getContainerLog(logUrl) {
     text: null,
   };
   const res = await fetch(logUrl);
-  const text = await res.text();
+  var text = await res.text();
   if (!res.ok) {
     throw new Error(res.statusText);
   }
@@ -222,6 +222,13 @@ export async function getContainerLog(logUrl) {
       throw new Error(`Log not available`);
     }
   } else if (config.logType === 'log-manager') {
+    if (text.length() === 0) {
+      const rotatedLogUrl = logUrl + '.1';
+      const rotatedLogRes = await fetch(rotatedLogUrl);
+      if (rotatedLogRes.ok) {
+        text = await rotatedLogRes.text();
+      }
+    }
     ret.text = text;
     ret.fullLogLink = logUrl.replace('/tail/', '/full/');
     return ret;
--- a/src/webportal/src/app/job/job-view/fabric/job-detail/conn.js
+++ b/src/webportal/src/app/job/job-view/fabric/job-detail/conn.js
@@ -222,11 +222,12 @@ export async function getContainerLog(logUrl) {
       throw new Error(`Log not available`);
     }
   } else if (config.logType === 'log-manager') {
-    if (text.length() === 0) {
+    if (text.length === 0) {
       const rotatedLogUrl = logUrl + '.1';
       const rotatedLogRes = await fetch(rotatedLogUrl);
-      if (rotatedLogRes.ok) {
-        text = await rotatedLogRes.text();
+      const rotatedText = await rotatedLogRes.text();
+      if (rotatedLogRes.ok && rotatedText !== 'No such file!') {
+        text = rotatedLogRes;
       }
     }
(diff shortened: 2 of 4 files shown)


Binyang2014:
fix #4777

src/log-manager/src/logrotate/logrotate.conf
```diff
@@ -4,12 +4,11 @@
...
-    weekly
-    rotate 4
-    compress
+    rotate 1
+    nocompress
     missingok
     notifempty
     copytruncate
-    maxsize 300M
+    size 100M
```
yqwang-ms, line 12: So user log can only keep [100M, 200M]? Which is too small.
Let's keep 300M here

src/webportal/src/app/job/job-view/fabric/job-detail/conn.js
```diff
@@ -222,6 +222,14 @@ export async function getContainerLog(logUrl) {
       throw new Error(`Log not available`);
     }
   } else if (config.logType === 'log-manager') {
+    if (text.length === 0) {
```
yqwang-ms, line 225: Did you concat these 2 files? This may be better to help user, as anyway backend store 2 files.

Binyang2014: Front-end only return 16MB logs to end-user. And we already provide a link to log folder, user can find the full logs when he/she need more information.
Concat the two logs will cause fron-end read too much data and slow down the website.
Here we only look the rotated log when origin log  file is empty.

src/log-manager/src/logrotate/logrotate.conf
Binyang2014: You mean [300M, 600M]? or [150M, 300M]. I think keep 600MB logs it too large, I'm OK to change to it [150M,300M]

src/webportal/src/app/job/job-view/fabric/job-detail/conn.js
yqwang-ms: If so, you should first concat (may not need real concat) then extract first latest 15M.
Otherwise, if the first file is 1B, and 2nd is 100MB, you can only show the 1B to user

src/log-manager/src/logrotate/logrotate.conf
yqwang-ms: How about [256, 512]?

src/webportal/src/app/job/job-view/fabric/job-detail/conn.js
Binyang2014: log-rotate doesn't promise not loss log. Concat the two logs may cause strange log string. We'd better give user some hint when the log is rotated.
```
copytruncate
Truncate the original log file in place after creating a copy, instead of moving the old log file and optionally creating a new one. It can be used when some program cannot be told to close its logfile and thus might continue writing (appending) to the previous log file forever. Note that there is a very small time slice between copying the file and truncating it, so some logging data might be lost. When this option is used, the create option will have no effect, as the old log file stays in place.
```

src/log-manager/src/logrotate/logrotate.conf
Binyang2014: 256M seems fine. I will raise my concern here

In current config, if log file not exceed `size`. The log file will be keep forever. So if we set size to `256M`, and many jobs scheduled to this node write much log. The max log size will be 256M * number_of_jobs_scheduled _to_this_node (even this job is deleted by FC). That's the reason I don't want to keep too much user log here.

After we make a final log collection solution, this will be solved. But for now, still has such issue.

src/webportal/src/app/job/job-view/fabric/job-detail/conn.js
yqwang-ms: Maybe insert one line below is better than just suddenly can only see 1B 1st file:

1st file content 
--------log is rotated, may be lost during this--------
2nd file content

src/log-manager/src/logrotate/logrotate.conf
yqwang-ms: If so why previous 300MB is fine but current 256MB is not?

Binyang2014: Before, we use `weekly` rotate. Logs exists more than one week will be rotated. And rotated logs exists more than one month will be dropped.
And we compress the rotated logs before.

yqwang-ms: BTW, does the log forever kept? Why not delete the log after its job is deleted in K8s or DB (our next version)? Because after that, user cannot access the job anyway.

src/webportal/src/app/job/job-view/fabric/job-detail/conn.js
Binyang2014: OK, will change to this

src/log-manager/src/logrotate/logrotate.conf
yqwang-ms: So, change to rotate by size, we also need to support a way to delete log in this PR

Binyang2014: ...It's not a easy fix and need more efforts. I think this PR is for quick fix and I want to redesign the log collection mechanism in next release.

src/webportal/src/app/job/job-view/fabric/job-detail/conn.js
Binyang2014: fixed

src/log-manager/src/logrotate/logrotate.conf
Binyang2014: fixed

PR #4800, opened 2020-08-10
Update vc and group experience for user

yiyione:


scarlett2018:
@yiyione may you book a quick demo with us? cc @qfyin @yangou1988

debuggy (approved):

PR #4844, opened 2020-08-26
Fix job list order by column

yiyione:


yqwang-ms (approved):

src/webportal/src/app/job/job-view/fabric/JobList/Ordering.js
```diff
@@ -31,7 +74,11 @@ export default class Ordering {
       query = 'totalTaskNumber';
     } else if (field === 'gpuCount') {
       query = 'totalGpuNumber';
+    } else if (field === 'duration') {
+      // rest-server don't receive duration order will use submissionTime
+      return { order: `submissionTime,${descending ? 'ASC' : 'DESC'}` };
```
hzy46, line 79: For duration, the logic for running job and completed job is different.

For running job, duration is `now date - submission time`;

For completed job, duration is `completed time - submission time`.

I suggest not to calculate duration ordering in backend. In front-end, we can disable duration sorting.

yiyione: Done.

hzy46 (approved):

Issue #4789, opened 2020-08-06
GPU utilization monitoring & handling


suiguoxin:
- [x] Alert-manager: Kill low-gpu-utilization jobs, tag abnormal jobs
    - [x] add virtual cluster info in `job-exporter`
    - [x] config monitor rules in `prometheus`
    - [x] send action request through webhook 
    - [x] `job-handler`: deal with webhook request & redirect to `RestServer`
    - [x] realize customized SMTP service in alert-handler, send alert email to user when possible, change email template to ejs.
- [x] Job tags:
    - [x] DB: job-tag table
    - [x] RestServer: 
        - getJobList : filter by tag 
        - getJobDetails : with tag info
        - tag : put / delete
- [x] Cordon node  with k8s API when GPU GCC Error

PR #4853, opened 2020-08-28
Job tagging

suiguoxin:
This PR is for part of issue #4789:

- [ ] Alert-manager: Kill low-gpu-utilization jobs, tag abnormal jobs
    - [ ] add virtual cluster info in `job-exporter`
    - [ ] config monitor rules in `prometheus`
    - [ ] send action request through webhook 
    - [ ] `job-handler`: deal with webhook request & redirect to `RestServer`
- [x] Job tags:
    - [x] DB: job-tag table
    - [x] RestServer:  Refer to [swagger](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/suiguoxin/pai/job-tag/src/rest-server/docs/swagger.yaml#operation/addTag) for the new API.
        - getJobList : filter by tag 
        - getJobDetails : with tag info
        - tag : put / delete
- [ ] WebPortal abnormal jobs: refactor with tag filter
- [ ] Cordon node  with k8s API when GPU GCC Error


src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -134,6 +136,10 @@ const convertFrameworkSummary = (framework) => {
     state: framework.state,
     subState: framework.subState,
     executionType: framework.executionType.toUpperCase(),
+    tags: framework.tags.reduce(
+      (arr, curr) => (arr.includes(curr.name) ? arr : [...arr, curr.name]),
```
hzy46, line 140: I notice primary key `uid=hash(frameworkName, tagName)` in the tag table, so we don't need to remove duplicates here.

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -267,9 +273,14 @@ const convertFrameworkDetail = async (framework) => {
   const completionStatus = attemptStatus.completionStatus;
   const diagnostics = completionStatus ? completionStatus.diagnostics : null;
   const exitDiagnostics = generateExitDiagnostics(diagnostics);
+
   const detail = {
     debugId: framework.metadata.name,
     name: jobName,
+    tags: tags.reduce(
```
hzy46, line 280: The same as above.

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -878,19 +889,44 @@ const getConfigSecretDef = (frameworkName, secrets) => {
...
+    const frameworkNames = await databaseModel.Tag.findAll({
+      attributes: ['frameworkName'],
+      where: tagsFilter,
+      group: ['tag.frameworkName'],
+      having: Sequelize.where(
+        Sequelize.fn('count', Sequelize.col('tag.frameworkName')),
+        tagsFilter.name.length,
+      ),
+    });
+    filters.name = frameworkNames.reduce(
```
hzy46, line 911: The same as above.

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -1142,6 +1186,63 @@ const getSshInfo = async (frameworkName) => {
...
+    where: { name: encodeName(frameworkName) },
+  });
+
+  if (framework) {
+    // add tag
+    const data = await databaseModel.Tag.findOrCreate({
+      where: {
+        frameworkName: encodeName(frameworkName),
+        name: tag,
+        uid: encodeName(frameworkName + tag),
```
hzy46, line 1201: `encodeName(frameworkName + "#%" + tag)` will be better?

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -1142,6 +1186,63 @@ const getSshInfo = async (frameworkName) => {
...
+  const framework = await databaseModel.Framework.findOne({
+    where: { name: encodeName(frameworkName) },
+  });
+
+  if (framework) {
+    // add tag
+    const data = await databaseModel.Tag.findOrCreate({
+      where: {
+        frameworkName: encodeName(frameworkName),
+        name: tag,
```
hzy46, line 1200: Add a sep char like `frameworkName + "#%" + tag`?

hzy46, line 1200: Check the tag. it shouldn't have `,`

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -878,19 +889,44 @@ const getConfigSecretDef = (frameworkName, secrets) => {
...
   offset,
   limit,
   withTotalCount,
 ) => {
   let frameworks;
   let totalCount;
+
+  if (Object.keys(tagsFilter).length !== 0) {
+    const frameworkNames = await databaseModel.Tag.findAll({
+      attributes: ['frameworkName'],
```
hzy46, line 903: add other filter and order & limit here?

src/rest-server/src/models/v2/job/k8s.js
suiguoxin: > I notice primary key `uid=hash(frameworkName, tagName)` in the tag table, so we don't need to remove duplicates here.

Removed.

src/rest-server/src/models/v2/job/k8s.js
suiguoxin: > The same as above.

Fixed

src/rest-server/src/models/v2/job/k8s.js
suiguoxin: > The same as above.

No such problem here.

src/rest-server/src/models/v2/job/k8s.js
suiguoxin: > `encodeName(frameworkName + "#%" + tag)` will be better?

Using encodeName(frameworkName + "+" + tag) to be consistent with the encoding algo of framework_history uid.

src/rest-server/src/models/v2/job/k8s.js
suiguoxin: > Check the tag. it shouldn't have `,`

Added a middleware here : https://github.com/microsoft/pai/pull/4853/files#diff-1e4d6d8193e8e8f5095fd5e5102c0a22

src/rest-server/src/models/v2/job/k8s.js
suiguoxin: > add other filter and order & limit & offset here? if there are many framework with this tag, querying them will cause memory/CPU issues.

Merged these two sql sentences into one to avoid data transfer between DB and RestServer.

hzy46 (approved):

suiguoxin:
Duplicated by https://github.com/microsoft/pai/pull/4924

PR #4855, opened 2020-08-30
[quick-start] Update hived config for quick-start

Binyang2014:
When qos-switch set to true, PAI daemon will requests addition resources.
Take PAI service resources into account when calculate hived config

### DIFF
--- a/contrib/kubespray/example/config.yml
+++ b/contrib/kubespray/example/config.yml
@@ -18,6 +18,11 @@ docker_image_tag: <% latest-release %>
 # docker_registry_username: exampleuser
 # docker_registry_password: examplepasswd
 
+#####################################
+# OpenPAI's daemon service qos config. #
+#####################################
+# qos-switch: "true"
+
 ###########################################################################################
 #                         Pre-check setting                                               #
 # By default, we assume your gpu environment is nvidia. So your runtime should be nvidia. #
--- a/contrib/kubespray/script/openpai-generator.py
+++ b/contrib/kubespray/script/openpai-generator.py
@@ -9,6 +9,7 @@ from kubernetes import client, config
 from kubernetes.utils import parse_quantity
 from kubernetes.client.rest import ApiException
 from pprint import pprint
+import re
 import sys
 import time
 
@@ -153,13 +154,54 @@ def get_kubernetes_node_info_from_API():
     return ret
 
 
+def get_pai_daemon_resource_request(cfg):
+    if "qos-switch" not in cfg or cfg["qos-switch"] == "false":
+        logger.info("Ignore calculate pai daemon resource usage since qos-switch set to false")
+
+    pai_daemon_services = ["node-exporter", "job-exporter", "log-manager"]
+    ret = {
+        "cpu-resource": 0,
+        "mem-resource": 0
+    }
+    pai_source_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), "../../src")
+
+    # {%- if cluster_cfg['cluster']['common']['qos-switch'] == "true" %}
+    start_match = r"{%-?\s*if\s*cluster_cfg\['cluster'\]\['common'\]\['qos-switch'\][^}]+%}"
+    end_match = r"{%-?\s*endif\s*%}"  # {%- end %}
+    str_match = f"{start_match}(.*?){end_match}"
+    regex = re.compile(str_match, flags=re.DOTALL)
+
+    for pai_daemon in pai_daemon_services:
+        deploy_template_path = os.path.join(pai_source_path, "{0}/deploy/{0}.yaml.template".format(pai_daemon))
+        if os.path.exists(deploy_template_path):
+            template = read_template(deploy_template_path)
+            match = re.search(template, regex)
+            if not match:
+                logger.warning("Could not find resource request for service %s", pai_daemon)
+                continue
+            resources = yaml.load(match.group(1), yaml.SafeLoader)["resources"]
+            if "requests" in resources:
+                ret["cpu-resource"] += int(parse_quantity(resources["requests"]["cpu"])) \
+                     if "cpu" in resources["requests"] else 0
+                ret["mem-resource"] += int(parse_quantity(resources["requests"]["memory"]) / 1024 / 1024) \
+                    if "memory" in resources["requests"] else 0
+            elif "limits" in resources:
+                ret["cpu-resource"] += int(parse_quantity(resources["limits"]["cpu"])) \
+                    if "cpu" in resources["limits"] else 0
+                ret["limits"] += int(parse_quantity(resources["limits"]["memory"]) / 1024 / 1024) \
+                    if "memory" in resources["limits"] else 0
+        else:
+            logger.warning("Could not find resource request for PAI daemon %s", pai_daemon)
+    return ret
+
+
 def wait_nvidia_device_plugin_ready(total_time=3600):
     while pod_is_ready_or_not("name", "nvidia-device-plugin-ds", "Nvidia-Device-Plugin") != True:
         logger.info("Nvidia-Device-Plugin is not ready yet. Please wait for a moment!")
         time.sleep(10)
         total_time = total_time - 10
         if total_time < 0:
-            logger.error("An issue occure when starting up Nvidia-Device-Plugin")
+            logger.error("An issue occur when starting up Nvidia-Device-Plugin")
             sys.exit(1)
 
 
@@ -173,7 +215,7 @@ def wait_amd_device_plugin_ready(total_time=3600):
             sys.exit(1)
 
 
-def hived_config_prepare(worker_dict, node_resource_dict):
+def hived_config_prepare(worker_dict, node_resource_dict, pai_daemon_resource_dict):
     hived_config = dict()
     hived_config["nodelist"] = []
 
@@ -188,8 +230,8 @@ def hived_config_prepare(worker_dict, node_resource_dict):
             logger.error("Allocatable GPU number in {0} is 0, current quick start script does not allow.".format(key))
             logger.error("Please remove {0} from your workerlist, or check if the device plugin is running healthy on the node.".format(key))
             sys.exit(1)
-        min_cpu = min(min_cpu, node_resource_dict[key]["cpu-resource"])
-        min_mem = min(min_mem, node_resource_dict[key]["mem-resource"])
+        min_cpu = min(min_cpu, node_resource_dict[key]["cpu-resource"] - pai_daemon_resource_dict["cpu-resource"])
+        min_mem = min(min_mem, node_resource_dict[key]["mem-resource"] - pai_daemon_resource_dict["mem-resource"])
         min_gpu = min(min_gpu, node_resource_dict[key]["gpu-resource"])
         hived_config["nodelist"].append(key)
     if not hived_config["nodelist"]:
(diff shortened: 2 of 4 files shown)


contrib/kubespray/example/config.yml
```diff
@@ -18,6 +18,14 @@ docker_image_tag: <% latest-release %>
 # docker_registry_username: exampleuser
 # docker_registry_password: examplepasswd
 
+################################################################
+# OpenPAI's daemon qos config.                                 #
+# Bt default, the QoS class for PAI daemon is BestEffort.      #
```
yqwang-ms, line 23: Typo

contrib/kubespray/script/openpai-generator.py
```diff
@@ -153,13 +154,51 @@ def get_kubernetes_node_info_from_API():
     return ret
 
 
+def get_pai_daemon_resource_request(cfg):
+    ret = {
+        "cpu-resource": 0,
+        "mem-resource": 0
+    }
+    if "qos-switch" not in cfg or cfg["qos-switch"] == "false":
```
yqwang-ms, line 162: Have you checked all PAI services, including 3rd party daemons are all without resource request specified if the qos-switch is false?
image
image

contrib/kubespray/script/openpai-generator.py
```diff
@@ -153,13 +154,51 @@ def get_kubernetes_node_info_from_API():
...
+        logger.info("Ignore calculate pai daemon resource usage since qos-switch set to false")
+        return ret
+
+    pai_daemon_services = ["node-exporter", "job-exporter", "log-manager"]
```
yqwang-ms, line 166: This should include all pai service

contrib/kubespray/example/config.yml
Binyang2014: fixed

contrib/kubespray/script/openpai-generator.py
```diff
@@ -153,13 +154,51 @@ def get_kubernetes_node_info_from_API():
...
+        if os.path.exists(deploy_template_path):
+            template = read_template(deploy_template_path)
+            match = regex.search(template)
+            if not match:
+                logger.warning("Could not find resource request for service %s", pai_daemon)
+                continue
+            resources = yaml.load(match.group(1), yaml.SafeLoader)["resources"]
+            if "requests" in resources:
+                ret["cpu-resource"] += parse_quantity(resources["requests"].get("cpu", 0))
+                ret["mem-resource"] += parse_quantity(resources["requests"].get("memory", 0)) / 1024 / 1024
```
yqwang-ms, line 186: Can we use math.ceil to an int?

contrib/kubespray/script/openpai-generator.py
```diff
@@ -153,13 +154,51 @@ def get_kubernetes_node_info_from_API():
...
+            elif "limits" in resources:
+                ret["cpu-resource"] += parse_quantity(resources["limits"].get("cpu", 0))
+                ret["mem-resource"] += parse_quantity(resources["limits"].get("memory", 0)) / 1024 / 1024
```
yqwang-ms, line 189: Can we use math.ceil to an int?

contrib/kubespray/script/openpai-generator.py
Binyang2014: When we generate hived config, we will change it to int.
Refer: https://github.com/microsoft/pai/blob/364aee769a421771ab76284d55fd7039767d3f20/contrib/kubespray/script/openpai-generator.py#L200-L201

contrib/kubespray/script/openpai-generator.py
Binyang2014: That's all PAI services which will deployed in worker node. Others are 3rd-party services and will calculated by:
https://github.com/microsoft/pai/blob/364aee769a421771ab76284d55fd7039767d3f20/contrib/kubespray/script/openpai-generator.py#L145-L148

`dshuttle` is not released, we don't include it

contrib/kubespray/script/openpai-generator.py
Binyang2014: We starts to deploy PAI services after all third party services are ready. QoS switch doesn't affect the 3rd-party services.
And 3rd-party resource requests is calculated by: https://github.com/microsoft/pai/blob/364aee769a421771ab76284d55fd7039767d3f20/contrib/kubespray/script/openpai-generator.py#L145-L149

Do not need to recalculate here

contrib/kubespray/script/openpai-generator.py
yqwang-ms: ok

contrib/kubespray/script/openpai-generator.py
```diff
@@ -188,8 +227,8 @@ def hived_config_prepare(worker_dict, node_resource_dict):
             logger.error("Allocatable GPU number in {0} is 0, current quick start script does not allow.".format(key))
             logger.error("Please remove {0} from your workerlist, or check if the device plugin is running healthy on the node.".format(key))
             sys.exit(1)
-        min_cpu = min(min_cpu, node_resource_dict[key]["cpu-resource"])
-        min_mem = min(min_mem, node_resource_dict[key]["mem-resource"])
+        min_cpu = min(min_cpu, node_resource_dict[key]["cpu-resource"] - pai_daemon_resource_dict["cpu-resource"])
```
yqwang-ms, line 230: Could we leave a buffer (such as reserve 1CPU 1GB?), and also minus it here? So that even if the whole caluculate is not precise, it will not cause previous issue.

contrib/kubespray/script/openpai-generator.py
yqwang-ms: Are you sure all the 3rd-party pod resource are deducted from the node.status.allocatable?

contrib/kubespray/script/openpai-generator.py
yqwang-ms: How we mantain this list? How new service owner knows he should add this service here?
Why not just check all services on worker, for both PAI and 3rd party? In case they specified resource

Binyang2014: The reason is:  At this point, the service not deployed in worker. So we need to parse source code to get the resource.

contrib/kubespray/script/openpai-generator.py
Binyang2014: Sure,  3rd-party services already boot-up at this point. If the service not ready, installation will failed.

contrib/kubespray/script/openpai-generator.py
yqwang-ms: Deploy storage manager only on master instead of worker on int bed.

contrib/kubespray/script/openpai-generator.py
yqwang-ms: 3rd-party services allocated resource is in node allocatable, but not node free. See https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/

contrib/kubespray/script/openpai-generator.py
Binyang2014: Fxied

contrib/kubespray/script/openpai-generator.py
Binyang2014: Fxied

contrib/kubespray/script/openpai-generator.py
Binyang2014: Will do it later

contrib/kubespray/script/openpai-generator.py
```diff
@@ -173,32 +268,40 @@ def wait_amd_device_plugin_ready(total_time=3600):
...
+        if node_resource_free[key]["gpu-resource"] == 0:
             logger.error("Allocatable GPU number in {0} is 0, current quick start script does not allow.".format(key))
             logger.error("Please remove {0} from your workerlist, or check if the device plugin is running healthy on the node.".format(key))
             sys.exit(1)
-        min_cpu = min(min_cpu, node_resource_dict[key]["cpu-resource"])
-        min_mem = min(min_mem, node_resource_dict[key]["mem-resource"])
-        min_gpu = min(min_gpu, node_resource_dict[key]["gpu-resource"])
+        reserved_cpu = min(node_resource_allocatable[key]["cpu-resource"] * Decimal(PAI_RESERVE_RESOURCE_PERCENTAGE), PAI_MAX_RESERVE_CPU_PER_NODE)
+        reserved_mem = min(node_resource_allocatable[key]["mem-resource"] * Decimal(PAI_RESERVE_RESOURCE_PERCENTAGE), PAI_MAX_RESERVE_MEMORY_PER_NODE)
+        min_cpu = min(min_cpu, node_resource_free[key]["cpu-resource"] - pai_daemon_resource_dict["cpu-resource"] - reserved_cpu)
```
yqwang-ms, line 291: > min_cpu = min(min_cpu, node_resource_free[key]["cpu-resource"] - pai_daemon_resource_dict["cpu-resource"] - reserved_cpu) [](start = 7, length = 122)

This can be negative, could we use below to avoid this:
```
reserved_cpu = min((node_resource_free[key]["cpu-resource"] - pai_daemon_resource_dict["cpu-resource"]) * Decimal(PAI_RESERVE_RESOURCE_PERCENTAGE), PAI_MAX_RESERVE_CPU_PER_NODE)
```

contrib/kubespray/script/openpai-generator.py
```diff
@@ -173,32 +268,40 @@ def wait_amd_device_plugin_ready(total_time=3600):
...
         hived_config["nodelist"].append(key)
     if not hived_config["nodelist"]:
         logger.error("No worker node is detected.")
         sys.exit(1)
 
     hived_config["min-gpu"] = min_gpu
-    hived_config["unit-cpu"] = int( min_cpu / min_gpu )
-    hived_config["unit-mem"] = int( min_mem / min_gpu )
+    hived_config["unit-cpu"] = math.ceil( min_cpu / min_gpu )
+    hived_config["unit-mem"] = math.ceil( min_mem / min_gpu )
```
yqwang-ms, line 304: We should use floor here, so user request are always within free.

yqwang-ms: See step1 of https://github.com/microsoft/hivedscheduler/blob/master/doc/user-manual.md#config-quickstart

Binyang2014: my wrong, fixed

contrib/kubespray/script/openpai-generator.py
Binyang2014: Prefer exit in this case:
https://github.com/microsoft/pai/blob/35abe26c60d55f512295a54f76c90c9d3c66cb27/contrib/kubespray/script/openpai-generator.py#L293-L297

The resource is too small to run a single job.

contrib/kubespray/script/openpai-generator.py
```diff
@@ -292,16 +291,18 @@ def hived_config_prepare(worker_dict, node_resource_dict, pai_daemon_resource_di
         min_mem = min(min_mem, node_resource_free[key]["mem-resource"] - pai_daemon_resource_dict["mem-resource"] - reserved_mem)
         min_gpu = min(min_gpu, node_resource_free[key]["gpu-resource"])
         if min_cpu <= 0 or min_mem <= 0:
-            logger.error("The node resource is not satisfy minmal requests, we need at least cpu: %s, mem: %sMB", min_cpu, min_mem)
+            logger.error("The node resource is not satisfy minmal requests, requests cpu: %s, mem: %sMB",
```
yqwang-ms, line 294: Log more statistic such as allocatable, reserved etc to help user understand?

yqwang-ms (approved):

contrib/kubespray/script/openpai-generator.py
Binyang2014: Done

PR #4868, opened 2020-09-02
Add HiveD scheduler config adapter for CA

abuccts:
Add HiveD scheduler config adapter for Cluster Autoscaler.


fanyangCS:
maybe adapter is a more appropriate name?

contrib/hived-config-patcher/src/patcher.py
```diff
@@ -0,0 +1,102 @@
...
+        def itoa(n):
+            return chr(n + ord("0")) if (0 <= n <= 9) else chr(n - 10 + ord("A"))
+        base, str36 = 36, ""
+        while num > 0:
+            str36 += itoa(num % base)
+            num = num // base
+        return str36[::-1].zfill(6)
+
+    def pad_fake_nodes(self, nodes):
+        return nodes[:self.max_nodes] + self.__fake_nodes[len(nodes):]
```
yqwang-ms, line 56: Log warning if k8s nodes more than max_nodes?

contrib/hived-config-patcher/src/patcher.py
```diff
@@ -0,0 +1,102 @@
...
+class HiveDConfigPatcher(object):
+    def __init__(self, min_nodes=1, max_nodes=2**31-1, hived_config_file=None):
+        logging.basicConfig(level=logging.INFO, format="%(asctime)s [%(levelname)s] %(message)s")
+        self.min_nodes = min_nodes
+        self.max_nodes = max_nodes
+        self.hived_config_file = hived_config_file
+        self.__init_hived_config()
+        self.__init_kube_client()
+        self.__nodes = []
+        self.__fake_nodes = ["fake{}".format(self.base36(i)) for i in range(max_nodes)]
```
yqwang-ms, line 20: Why not just i instead of base36(i), for easy read and parse

contrib/hived-config-patcher/src/patcher.py
```diff
@@ -0,0 +1,102 @@
...
+    parser.add_argument("--max-nodes", type=int, help="maximum node number for CA")
+    parser.add_argument("--hived-config-file", help="hived configuration file path")
+    args = parser.parse_args()
+
+    patcher = HiveDConfigPatcher(
+        max_nodes=args.max_nodes,
+        hived_config_file=args.hived_config_file)
+    while True:
+        try:
+            patcher.list_nodes()
```
yqwang-ms, line 98: How about directly watch without first list?

contrib/hived-config-patcher/src/patcher.py
```diff
@@ -0,0 +1,102 @@
...
+        logging.info("List nodes: %s.", self.__nodes)
+        self.update_hived_config()
+
+    def watch_nodes(self):
+        w = watch.Watch()
+        for event in w.stream(self.__client.list_node, _request_timeout=30):
+            if event["type"] == "ADDED":
+                logging.info("Watch event: %s node %s.", event["type"], event["object"].metadata.name)
+                if event["object"].metadata.name not in self.__nodes:
+                    self.__nodes.append(event["object"].metadata.name)
```
yqwang-ms, line 70: Make __nodes as a map instead of array, better for dedup here

contrib/hived-config-patcher/src/patcher.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def list_nodes(self):
+        ret = self.__client.list_node(watch=False)
+        self.__nodes = [item.metadata.name for item in ret.items]
+        logging.info("List nodes: %s.", self.__nodes)
+        self.update_hived_config()
+
+    def watch_nodes(self):
+        w = watch.Watch()
+        for event in w.stream(self.__client.list_node, _request_timeout=30):
```
yqwang-ms, line 66: What is the _request_timeout?
If no events during the timeout, will the for loop exit?
Shall we increase it?

contrib/hived-config-patcher/ds.yaml
```diff
@@ -0,0 +1,46 @@
...
+      hostPID: false
+      containers:
+      - name: app
+        image: openpai/hived-config-patcher
+        imagePullPolicy: Always
+        command: ["python"]
+        args: [
+          "./patcher.py",
+          "--max-nodes={{cfg['openpai_worker_vmss']['max_vm_count']}}",
+          "--hived-config-file=/hived-config/hivedscheduler.yaml"
```
yqwang-ms, line 37: Can the local file change be propagated to ConfigMap?

contrib/hived-config-patcher/src/patcher.py
```diff
@@ -0,0 +1,102 @@
...
+                logging.info("Watch event: %s node %s.", event["type"], event["object"].metadata.name)
+                if event["object"].metadata.name in self.__nodes:
+                    self.__nodes.remove(event["object"].metadata.name)
+                    self.update_hived_config()
+            else:
+                continue
+
+    def update_hived_config(self):
+        all_nodes = self.pad_fake_nodes(sorted(self.__nodes))
+        self.__hived_config["physicalCluster"]["physicalCells"][0]["cellChildren"] = \
```
yqwang-ms, line 82: Write back to file?

contrib/hived-config-patcher/src/patcher.py
```diff
@@ -0,0 +1,102 @@
...
+            sys.exit(1)
+        if ("physicalCluster" not in self.__hived_config or
+                "physicalCells" not in self.__hived_config["physicalCluster"] or
+                len(self.__hived_config["physicalCluster"]["physicalCells"]) == 0):
+            logging.error("Invalid hived config, exit.")
+            sys.exit(1)
+
+    def __init_kube_client(self):
+        try:
+            config.load_kube_config("../config.log")
```
yqwang-ms, line 40: What's this? You can use in cluster config

contrib/hived-config-patcher/src/patcher.py
abuccts: added

contrib/hived-config-patcher/src/patcher.py
abuccts: to keep consistent with vmss naming format, [six digits base 36](https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-instance-ids#scale-set-vm-computer-name)

contrib/hived-config-patcher/src/patcher.py
abuccts: if there're `n` nodes initially, watch will have `n` ADDED events and configmap will be updated `n` times.
list is used to merge those updates.

contrib/hived-config-patcher/src/patcher.py
abuccts: updated to hash set, but the check is still preserved to avoid duplicate configmap update

contrib/hived-config-patcher/ds.yaml
abuccts: no, added `patch_namespaced_config_map` to update configmap

contrib/hived-config-patcher/src/patcher.py
abuccts: will patch configmap

contrib/hived-config-patcher/src/patcher.py
abuccts: updated, it was for local debug

contrib/hived-config-patcher/src/patcher.py
abuccts: > What is the _request_timeout?

It's used in single `list_node` api, [reference](https://github.com/kubernetes-client/python-base/blob/789de6a60dea9442ee47601a788d92c653ea69e9/watch/watch.py#L122).

> If no events during the timeout, will the for loop exit?

No, there's another `timeout_seconds` parameter.

> Shall we increase it?

I think it's fine, [the example](https://github.com/kubernetes-client/python#examples) uses 60.

contrib/hived-config-patcher/src/patcher.py
yqwang-ms: ok, but seems no benifts to keep this consistent

contrib/hived-config-patcher/src/patcher.py
yqwang-ms: So let's make timeout_seconds  longer ?
See https://github.com/kubernetes-client/python/issues/124

contrib/hived-config-patcher/src/patcher.py
```diff
@@ -0,0 +1,131 @@
...
+import yaml
+import logging
+import argparse
+from kubernetes import client, config, watch
+
+
+class HiveDConfigPatcher(object):
+    def __init__(
+            self,
+            min_nodes=None,
```
yqwang-ms, line 14: >min_nodes [](start = 12, length = 9)

Seems we do not use min_nodes

yqwang-ms (approved):
:shipit:

contrib/hived-config-patcher/src/patcher.py
abuccts: it's an attribute of CA like max_nodes, included here in case we need it in the future

PR #4888, opened 2020-09-09
Support job archive

### LINKED ISSUE #4453: Support delete/archive jobs
This ticket is not for the current release but for a future support.

**Scenario**
As an OpenPAI user, I could permanent delete (clean up) some of my jobs to get rid of not in use jobs. After deletion, no one can restore this jobs. User need to confirm before the deletion.

As an Administrator, I could able to archive some bad job to clean up the view for all the users. After archive, the jobs are disappeared from job list and search views, but accessible from the job url (as archived read only job). Admin and user may or may not able to restore this jobs (to be decides, comments are welcomed here).

hzy46:
to resolve #4453

Support job archive (only `Completed` jobs can be archived, cannot unarchive job):

- Add rest-server interface:
  - PUT /jobs/{user}~{job}/archive
    - Admin can archive any job; non-admin user can only archive his own job.
    - Only completed job can be archived. If job is not a completed job, return 400 Bad Request.
    - If the job cannot be found, return 404 not found.
    - If job is successfully archived, return 202 Accepted.
    - If other error happens, return 500.
- Add write-merger interface
  - PUT /archiveFramework/:frameworkName
    - Only completed job can be archived. If job is not a completed job, return 400 Bad Request.
    - If the job is not found, return 404 not found. 
    - If job is successfully archived, return 200 OK.
    - If other error happens, return 500.
- Change rest-server list job api: 
    - support to include archived jobs or not. Default is not to include archived jobs

### DIFF
--- a/src/database-controller/src/write-merger/handler.js
+++ b/src/database-controller/src/write-merger/handler.js
@@ -279,7 +279,6 @@ async function putFrameworkRequest(req, res, next) {
   }
 }
 
-
 async function archiveFramework(req, res, next) {
   // The handler to handle framework archive
   // Only completed job can be archived. If job is not a completed job, return 400 Bad Request.
@@ -293,9 +292,7 @@ async function archiveFramework(req, res, next) {
     }
     await lock.acquire(frameworkName, async () => {
       const framework = await databaseModel.Framework.findOne({
-        attributes: [
-          'subState',
-        ],
+        attributes: ['subState'],
         where: { name: frameworkName },
       });
       if (!framework) {
@@ -306,7 +303,7 @@ async function archiveFramework(req, res, next) {
       } else {
         // If the old framework exists and is completed, try to set archived=true
         await databaseModel.Framework.update(
-          {archived: true},
+          { archived: true },
           { where: { name: frameworkName } },
         );
       }
--- a/src/database-controller/src/write-merger/handler.js
+++ b/src/database-controller/src/write-merger/handler.js
@@ -279,9 +279,48 @@ async function putFrameworkRequest(req, res, next) {
   }
 }
 
+
+async function archiveFramework(req, res, next) {
+  // The handler to handle framework archive
+  // Only completed job can be archived. If job is not a completed job, return 400 Bad Request.
+  // If the job is not found, return 404 not found.
+  // If job is successfully archived, return 200 OK.
+  // If other error happens, return 500.
+  try {
+    const frameworkName = req.params.frameworkName;
+    if (!frameworkName) {
+      return next(createError(400, 'Cannot find framework name.'));
+    }
+    await lock.acquire(frameworkName, async () => {
+      const framework = await databaseModel.Framework.findOne({
+        attributes: [
+          'subState',
+        ],
+        where: { name: frameworkName },
+      });
+      if (!framework) {
+        // if the old framework doesn't exist, report a 404 error
+        throw createError(404, `Cannot find framework ${frameworkName}.`);
+      } else if (framework.subState !== 'Completed') {
+        throw createError(400, `Framework ${frameworkName} is not completed.`);
+      } else {
+        // If the old framework exists and is completed, try to set archived=true
+        await databaseModel.Framework.update(
+          {archived: true},
+          { where: { name: frameworkName } },
+        );
+      }
+    });
+    res.status(200).json({ message: 'ok' });
+  } catch (err) {
+    return next(err);
+  }
+}
+
 module.exports = {
   ping: ping,
   putFrameworkRequest: putFrameworkRequest,
   patchFrameworkRequest: patchFrameworkRequest,
   postWatchEvents: postWatchEvents,
+  archiveFramework: archiveFramework,
 };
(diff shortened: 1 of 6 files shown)


src/database-controller/src/write-merger/handler.js
```diff
@@ -279,9 +279,45 @@ async function putFrameworkRequest(req, res, next) {
...
+        where: { name: frameworkName },
+      });
+      if (!framework) {
+        // if the old framework doesn't exist, report a 404 error
+        throw createError(404, `Cannot find framework ${frameworkName}.`);
+      } else if (framework.subState !== 'Completed') {
+        throw createError(400, `Framework ${frameworkName} is not completed.`);
+      } else {
+        // If the old framework exists and is completed, try to set archived=true
+        await databaseModel.Framework.update(
```
yqwang-ms, line 305: The word 'archive' is not accurate, prefer 'hide' as it only stills stores in DB, and occupy the job name.

BTW, if another user submit a job naming conflict with a 'hide' job, the user has no idea, as he cannot get the conflict 'hide' job by default.

So, could we do not introduce this feature?
Intead, I prefer 'archive' to be:
1. Directly delete the job in DB, and move it to other backlog storage or even just do not backup
2. Split the API and webportal view between jobs in DB and jobs in the backlog storage.

@fanyangCS  to more comment on this

yqwang-ms (changes requested):
The feature breaking API consistency

fanyangCS:
if a job has been archived, does it appear in the history server?

hzy46:
> if a job has been archived, does it appear in the history server?

What does the term `history server` refers to? Since `v1.2.0`, all frameworks are stored in the table `frameworks`. Archive a framework means set `archived=true` in this table. The archived jobs are excluded from job listing by default.

There is another table named `framework_history`. But it only stores retried framework object snapshots. Frameworks with no history retries don't have records in the `framework_history` table. And this table is not related to job archive.

fanyangCS:
I feel the definition of "archive" might be a bit complicated to end user. How about we simply call it delete? A deleted job won't appear in history server or job list view.

And we can keep all the deleted jobs in a separate table or separate DB. User can view their deleted jobs but cannot restore or further delete it (for auditing issue).

Admin can delete rows in the table.

yqwang-ms:
> if a job has been archived, does it appear in the history server?

After DB, there is no history job concept, only job previous retry attempts are stored in history job table.
For my previous comment "backlog storage" can be the "history job table", but this PR still keeps the archived job in "current job table" which occupy the job namespace.

> I feel the definition of "archive" might be a bit complicated to end user. How about we simply call it delete? A deleted job won't appear in history server or job list view.
And we can keep all the deleted jobs in a separate table or separate DB. User can view their deleted jobs but cannot restore or further delete it (for auditing issue).
Admin can delete rows in the table.

Agree with this. But seems the orignal requirement from user, is that:
> As an Administrator, I could **able to archive some bad job to clean up the view for all the users**. After archive, the jobs are disappeared from job list and search views, but accessible from the job url (as archived read only job). Admin and user may or may not able to restore this jobs (to be decides, comments are welcomed here).

Looks like a authz requirement that Admin want to kill and hide some bad jobs, so that only admin can see it again.

Anyway, we need to figure out the requirement and a detailed design review before this PR.

yqwang-ms:
BTW, seems what they really want is job label given the perf is not a problem anymore, such as only filter the prod jobs (non-test jobs):

> We need admin user can delete some testing jobs. On developing some features in our platform based on OpenPAI, we need get the job list. If admin user cannot delete those useless jobs, the list of jobs that we get by API will be so long. And it seems that too many jobs will impact the performance of OpenPAI.

src/database-controller/src/write-merger/handler.js
fanyangCS: > > if a job has been archived, does it appear in the history server?
> 
> After DB, there is no history job concept, only job previous retry attempts are stored in history job table.
> For my previous comment "backlog storage" can be the "history job table", but this PR still keeps the archived job in "current job table" which occupy the job namespace.
> 
> > I feel the definition of "archive" might be a bit complicated to end user. How about we simply call it delete? A deleted job won't appear in history server or job list view.
> > And we can keep all the deleted jobs in a separate table or separate DB. User can view their deleted jobs but cannot restore or further delete it (for auditing issue).
> > Admin can delete rows in the table.
> 
> Agree with this. But seems the orignal requirement from user, is that:
> 
> > As an Administrator, I could **able to archive some bad job to clean up the view for all the users**. After archive, the jobs are disappeared from job list and search views, but accessible from the job url (as archived read only job). Admin and user may or may not able to restore this jobs (to be decides, comments are welcomed here).
> 
> Looks like a authz requirement that Admin want to kill and hide some bad jobs, so that only admin can see it again.
> 
> Anyway, we need to figure out the requirement and a detailed design review before this PR.

We can explicitly provide a web page to show the deleted jobs. We need a new API to retrieve deleted jobs.

yqwang-ms: Yep. But this looks like not traditional 'archive' feature (move old ones to backlog to improve serving perf), now I am feeling our goal (and user requirement) is not clear, we would better to first figure out them. As my latest comment, job label may be enough for them.

hzy46:
After discussion with @mydmdm , I think we may use job tag #4853 to resolve this problem.

PR #4923, opened 2020-09-21
[cherry-pick] make internal storage depend on cluster configuration

hzy46:


hzy46:
close this because CI seems have bug.

PR #4939, opened 2020-09-29
[display event] add event watcher in database controller

hzy46:
database size control strategy:

  - In event watcher, check the disk usage in the beginning. If the disk usage > 80%, stop the event watcher and exit with a non-zero code.
  - The disk check also happens  every 60s. If the disk usage > 80%, stop the event watcher and exit with a non-zero code.
  - `60s` and `80%` are configurable.

Problem found:
  - If a job generates too many events, it will affect all other jobs.

### DIFF
--- a/src/database-controller/config/database-controller.yaml
+++ b/src/database-controller/config/database-controller.yaml
@@ -27,7 +27,7 @@ write-merger-max-heap-mb: 2048
 # db poller
 # Polling interval of database poller. Default value is 120.
 db-poller-interval-second: 120
-# Max connection number to database in write merger.
+# Max connection number to database in poller.
 db-poller-max-db-connection: 10
 # Max rpc concurrency for db poller
 db-poller-max-rpc-concurrency: 50
@@ -39,3 +39,13 @@ db-poller-max-heap-mb: 4096
 framework-watcher-max-rpc-concurrency: 150
 # Framework watcher max heap size in MB
 framework-watcher-max-heap-mb: 8192
+
+# cluster event watcher
+# Max rpc concurrency for cluster event watcher
+cluster-event-watcher-max-rpc-concurrency: 40
+# cluster event watcher max heap size in MB
+cluster-event-watcher-max-heap-mb: 2048
+# Max connection number to database in cluster event watcher.
+cluster-event-max-db-connection: 40
+# Max disk usage in internal storage for cluster event watcher
+cluster-event-watcher-max-disk-usage-percent: 80
--- a/src/database-controller/src/watcher/cluster-event/index.js
+++ b/src/database-controller/src/watcher/cluster-event/index.js
@@ -55,7 +55,7 @@ async function synchronizeEvent(eventType, apiObject) {
 
 const eventHandler = (eventType, apiObject) => {
   /*
-    framework name-based lock + always retry
+    event uid-based lock + always retry
   */
   const receivedTs = new Date().getTime();
   const involvedObjKind = apiObject.involvedObject.kind;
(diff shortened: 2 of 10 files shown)


src/database-controller/src/watcher/cluster-event/index.js
```diff
@@ -0,0 +1,139 @@
...
+      );
+    });
+  } else {
+    logger.info(
+      `Cluster Event type=${eventType} receivedTs=${receivedTs} uid=${uid} involvedObjKind=${involvedObjKind} involvedObjName=${involvedObjName} received but ignored.`,
+    );
+  }
+};
+
+async function assertDiskUsageHealthy() {
```
yqwang-ms, line 86: Better to limit the quota per job and global, and does not impact the critical path.

src/database-controller/src/watcher/cluster-event/index.js
```diff
@@ -0,0 +1,139 @@
...
+    message: apiObject.message,
+    firstTimestamp: apiObject.firstTimestamp,
+    lastTimestamp: apiObject.lastTimestamp,
+    count: apiObject.count,
+    sourceComponent: _.get(apiObject, 'source.component', null),
+    sourceHost: _.get(apiObject, 'source.host', null),
+    event: JSON.stringify(apiObject),
+  };
+
+  databaseModel.FrameworkEvent.upsert(obj, { where: { uid: uid } });
```
yqwang-ms, line 53: Update existing uid, instead of add a new one?

hzy46: yes

src/database-controller/config/database-controller.yaml
```diff
@@ -39,3 +39,13 @@ db-poller-max-heap-mb: 4096
...
+
+# cluster event watcher
+# Max rpc concurrency for cluster event watcher
+cluster-event-watcher-max-rpc-concurrency: 40
+# cluster event watcher max heap size in MB
+cluster-event-watcher-max-heap-mb: 2048
+# Max connection number to database in cluster event watcher.
+cluster-event-max-db-connection: 40
+# Max disk usage in internal storage for cluster event watcher
+cluster-event-watcher-max-disk-usage-percent: 80
```
yqwang-ms, line 51: Also limit for history ? Why not move non-critical things to another DB server?

src/database-controller/src/watcher/cluster-event/index.js
hzy46: Recorded in #4953 . We can solve this problem in the future.

src/database-controller/config/database-controller.yaml
hzy46: Recorded in #4953 . We can solve this problem in the future.

yqwang-ms (approved):

hzy46:
Event Watcher Test Cases:

1. Test: the event watcher works properly 

Submit a job that will be always in waiting status (e.g. use a lot of resource). Then check if there is any event about "failed scheduling" on the job event page after a few minutes.

2. Test: the event watcher can handle a large number of events.

Submit a job with 2000+ tasks. After a few minutesm check the event page can work properly.

3. Test: the event watcher will exit if too much disk size is used.

Go to internal storage to see the existing file usage:

```
kubectl exec -it `kubectl get po  | grep internal-storage | awk '{print $1}' ` bash
df -h
```

Please notice the usage of loop device `/paiInternal/storage`.

Create a big file under `/paiInternal/storage` and make its usage larger than 80%.

After a few minutes, confirm that 1. there is a NodeFilesystemUsage alert shown on webportal 2. the event watcher should exit automatically.

Remove the big file. After a few minutes, confirm that: 1. there is no more NodeFilesystemUsage alert 2. the event watcher should work properly, and we can see events of new jobs on webportal.

PR #4959, opened 2020-10-14
[Rest Server] Update pod priority calculation

abuccts:
Update pod priority calculation:
* add one bit in FIFO bits
* decrease priority value every 16s

Known issue: FIFO will be broken every ~8.5 years when mod result meets 0.

### DIFF
--- a/contrib/aks-engine/templates/k8s.json.j2
+++ b/contrib/aks-engine/templates/k8s.json.j2
@@ -31,7 +31,7 @@
                     ],
                     "config": {
                         "scan-interval": "1m",
-                        "expendable-pods-priority-cutoff": "-2130706431"
+                        "expendable-pods-priority-cutoff": "-2147483647"
                     }
 {% else %}
                     "enabled": false
--- a/contrib/aks-engine/templates/k8s.json.j2
+++ b/contrib/aks-engine/templates/k8s.json.j2
@@ -31,7 +31,7 @@
                     ],
                     "config": {
                         "scan-interval": "1m",
-                        "expendable-pods-priority-cutoff": "-1200000000"
+                        "expendable-pods-priority-cutoff": "-2130706431"
                     }
 {% else %}
                     "enabled": false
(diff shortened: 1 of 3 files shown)


yqwang-ms (dismissed):

src/rest-server/src/middlewares/v2/hived.js
```diff
@@ -26,12 +26,12 @@ const { hivedWebserviceUri } = require('@pai/config/launcher');
...
 const convertPriority = (priorityClass = 'test') => {
   // TODO: make it a cluster-wise config
-  // allowed range: [-1, 126], default priority 0
+  // allowed range: [-1, 126], default priority 10
   const priorityMap = {
     crit: 120,
     prod: 100,
     test: 10,
-    oppo: -1,
+    oppo: 0,
```
yqwang-ms, line 34: But hived still use -1 as oppo, is that ok?

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -1099,8 +1099,8 @@ const put = async (frameworkName, config, rawConfig) => {
       jobPriority = Math.min(Math.max(jobPriority, -1), 126);
     }
     const jobCreationTime =
-      Math.floor(submissionTime / 1000) & (Math.pow(2, 23) - 1);
-    const podPriority = -(((126 - jobPriority) << 23) + jobCreationTime);
+      Math.floor(submissionTime / 16000) & (Math.pow(2, 24) - 1);
```
yqwang-ms, line 1102: track an issue for when the FIFO will be fliped/

contrib/aks-engine/templates/k8s.json.j2
```diff
@@ -31,7 +31,7 @@
                     ],
                     "config": {
                         "scan-interval": "1m",
-                        "expendable-pods-priority-cutoff": "-1200000000"
+                        "expendable-pods-priority-cutoff": "-2147483648"
```
yqwang-ms, line 34: Should be -2147483647? So we reserved -2147483648 as a priority which CA can ignore?

yqwang-ms (approved):

PR #4996, opened 2020-10-21
fix nested config default val issue

suiguoxin:


src/alert-manager/config/alert_manager.py
```diff
@@ -1,6 +1,15 @@
...
 import copy
+from flatten_dict import flatten, unflatten
+
+def soft_update_dict(dict_0, dict_1):
+    """
+    keep original key in nested dict if not present in the new dict
+    """
+    res = flatten(dict_0)
+    res.update(flatten(dict_1))
+    return unflatten(res)
```
Binyang2014, line 12: How about use this solution: https://stackoverflow.com/questions/3232943/update-value-of-a-nested-dictionary-of-varying-depth instead of introduce a new lib

contrib/kubespray/quick-start/services-configuration.yaml.template
```diff
@@ -166,10 +166,10 @@ authentication:
 
 # if you want to enable alert-handler actions, uncomment following lines and fill the right values.
 # alert-manager:
-#   port: 9093 # do not change this if you do not want to change the port alert-manager is listening on
+#   port: 9093 # optional, do not write this if you do not want to change the port alert-manager is listening on
```
Binyang2014, line 169: Use `modify`? Same as  lines

src/alert-manager/config/alert_manager.py
suiguoxin: > How about use this solution: https://stackoverflow.com/questions/3232943/update-value-of-a-nested-dictionary-of-varying-depth instead of introduce a new lib

Updated, thanks

contrib/kubespray/quick-start/services-configuration.yaml.template
suiguoxin: > Use `modify`? Same as lines

Updated, thanks

src/alert-manager/config/alert_manager.py
```diff
@@ -1,6 +1,18 @@
 #!/usr/bin/env python
 
 import copy
+import collections
+
+def soft_update_dict(d, u):
```
Binyang2014, line 6: change to another name? such as `update_nested_dict(default, update)`

Binyang2014 (approved):

PR #5020, opened 2020-10-28
Fix Promise Issues in Alert handler

suiguoxin:


src/alert-manager/src/alert-handler/controllers/node.js
```diff
@@ -22,47 +22,57 @@ const logger = require('@alert-handler/common/logger');
...
-            logger.error(err);
-            res.status(500).json({
-              message: `alert-handler failed to cordon node ${nodeName}`,
-            });
-          },
-        );
-    }
-  });
+  // extract nodes to cordon
+  const nodeNames = req.body.alerts.reduce(
```
Binyang2014, line 47: Why using `reduce`? If you just want to get nodes name, how about using:
```js
alerts.filter(alert => alert.status === 'firing' && 'node_name' in alert.labels)
      .map(alert.labels.node_name)
```

src/alert-manager/src/alert-handler/emails/general-templates/subject.ejs
```diff
@@ -1,7 +1 @@
-OpenPAI Alerts : <%= groupLabels.alertname %>,
-<% if (alerts.filter( element => element.status =="firing").length > 0) { %>
-<%= alerts.filter( element => element.status =="firing").length %> Firing
-<% } %>
-<% if (alerts.filter( element => element.status =="resolved").length > 0) { %>
```
Binyang2014, line 5: Why remove resolve?

src/alert-manager/src/alert-handler/controllers/job.js
```diff
@@ -15,93 +15,113 @@
...
+  logger.info(`alert-handler will tag these jobs: ${jobNames}`);
 
-  logger.info('alert-handler successfully tag jobs.');
-  res.status(200).json({
-    message: 'alert-handler successfully tag jobs.',
-  });
+  // tag all these jobs
+  Promise.all(
+    jobNames.map(
+      async (jobName) => await tagJob(jobName, req.params.tag, req.token),
```
Binyang2014, line 110: tagJob already an async function, seems doesn't need async/await here

src/alert-manager/src/alert-handler/controllers/mail.js
```diff
@@ -130,73 +59,117 @@ const sendEmailToAdmin = async (req, res) => {
...
-    res.status(500).json({
-      message: `Send email encourted Unknown Error`,
+    await Promise.all(
+      alerts.map(async (alert) => {
+        const userName = await getUserNameByJobName(
+          alert.labels.job_name,
+          req.token,
+        );
+        if (userName && userName in alertsGrouped) {
+          alertsGrouped[userName].push(alert);
```
Binyang2014, line 126: map function will run parrallel, does push function is thread-safe?
Can you get all names first then do some group action?

Binyang2014 (approved):

PR #5029, opened 2020-10-29
[Doc] introduce how to use summary in customized alerts

suiguoxin:


suiguoxin:
When users define `alerts`, they are not aware of how they will handle the alerts.
So this approach **involves more complexity** for admin users.

PR #5032, opened 2020-10-30
update users's own profile: support AAD mode; update extension

suiguoxin:
This PR is for release v1.4. Needed for multi-cluster feature.

update users own profile API:
- support AAD mode
- suppport `extension` update
- Some code is refactored: `checkAdmin` should be a common logic in token check, while `checkSelf` should be controller specific since it relies on the request body.

### DIFF
--- a/src/rest-server/src/routes/v2/user.js
+++ b/src/rest-server/src/routes/v2/user.js
@@ -56,6 +56,7 @@ if (authnConfig.authnMethod === 'basic') {
     /** Create /api/v2/users */
     .post(
       token.checkNotApplication,
+      token.checkAdmin,
       param.validate(userInputSchema.userCreateInputSchema),
       userController.createUser,
     );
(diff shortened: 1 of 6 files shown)


hzy46 (approved):

src/rest-server/src/routes/v2/user.js
```diff
@@ -65,16 +66,18 @@ if (authnConfig.authnMethod === 'basic') {
     /** Put /api/v2/users */
     .put(
       token.checkNotApplication,
+      token.checkAdmin,
       param.validate(userInputSchema.basicAdminUserUpdateInputSchema),
```
abuccts, line 70: change the order to keep consistent?

suiguoxin: > change the order to keep consistent?

`Auth check` should have a higher level of priority than `content check`. `checkAdmin` is independent of the input schema, so I think it should be placed before.

suiguoxin: > change the order to keep consistent?

Order changed to keep the original logic.

abuccts (approved):

PR #5039, opened 2020-11-02
Use url-join to be more compatible for env config

debuggy:


src/webportal/src/app/job/job-view/fabric/task-attempt/task-attempt-dialog.jsx
```diff
@@ -1,75 +0,0 @@
-// Copyright (c) Microsoft Corporation
```
hzy46, line 1: Why this file is deleted?

debuggy: It is an out-dated file, which designed to open a dialog. And the new detail will open a new page instead of a dialog, so it is deleted

hzy46 (approved):

PR #5046, opened 2020-11-03
Add permission and new API in log-manager

Binyang2014:
- Add permission check in log-manage
- Add log-manager API to retrieve log


src/log-manager/config/log-manager.yaml
```diff
@@ -18,3 +18,7 @@
 service_type: "k8s"
 
 port: 9103
+admin_name: "admin"
+admin_password: "admin"
+jwt_secret: "jwt_secret"
+token_expired_time: 120
```
hzy46, line 24: Does it means `token_expired_second`? I think `token_expired_time` is a bit confusing for admin. He/she may not know the real meaning of it.

src/log-manager/config/log-manager.yaml
```diff
@@ -18,3 +18,7 @@
 service_type: "k8s"
 
 port: 9103
+admin_name: "admin"
```
hzy46, line 21: Please add this change to https://github.com/microsoft/pai/blob/master/contrib/kubespray/quick-start/services-configuration.yaml.template#L239

src/log-manager/src/cleaner/entrypoint.sh
```diff
@@ -0,0 +1,43 @@
...
+# THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING
+# BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+# DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+set -o errexit
+set -o pipefail
+
+log_exist_time=30 # 30 day
```
hzy46, line 23: So a job's log will be removed after 30 days? Currently we don't clean user's job in database. If a user views his previous job after it completes for 30 days, and the job log is cleaned, what will happen?

Binyang2014: They can not retrieve the job log. 404 will return. If they want to persistent job log, we recommend customer use external storage, such as AzureBlob or NFS (It's next step, we don't implement yet).

hzy46: If applicable, I think it's better to remind users that their logs will be removed after 30 days in the webportal. It's strange that a previous existing log just disappears and report a 404 error.

Binyang2014: It's the same behavior with previous version. We could add doc for this. For webportal change, it's another enhance and we could open another issue to track

src/log-manager/config/log-manager.yaml
Binyang2014: Fxied

src/log-manager/config/log-manager.yaml
Binyang2014: Fixed

src/log-manager/src/cleaner/entrypoint.sh
hzy46: Agreed. Please open an issue for it.

Resolved this conversation.

hzy46 (approved):

PR #5057, opened 2020-11-05
Update docs for Cluster Autoscaler on AKS Engine

abuccts:
Update docs for Cluster Autoscaler on AKS Engine.


hzy46:
@abuccts Would you like to add this doc to the openpai handbook? I think we can add a question like `How to enable cluster autoscaler` in https://openpai.readthedocs.io/en/latest/manual/cluster-admin/installation-faqs-and-troubleshooting.html, and let users refer to `https://github.com/microsoft/pai/blob/master/contrib/aks-engine/readme.md` as an answer.

abuccts:
Currently there's no way to **enable** CA in an existing cluster, you need to use aks and deploy with aks-engine from scratch, so this is a standalone part under contrib. It's not applicable for most users, so it might not be suitable for faq?

hzy46:
Installation FAQ is not just FAQ for the quick-start installation. It could be any questions about the installation. I think it is OK to put this in the FAQ section and tell users who are interested to install the cluster from scratch and try this feature.

hzy46 (approved):

Issue #5021, opened 2020-10-28
alert handler stop-job notice not clear to end user


Binyang2014:
When enable `stop-job` and `email-user`. User only get email about this alert such as:
![image](https://user-images.githubusercontent.com/9415966/97398817-d224a280-1926-11eb-9b8e-197f557a5e70.png)

But doesn't get notice that the job will be stopped by alert manager.
This will confuse `end user`. When he found the job is stopped.

suiguoxin:
Fixed in #5064

PR #5064, opened 2020-11-09
Alert Email Template Refine

suiguoxin:
* add kill user job alert template #5021
* add troubleshooting link in general email template
* redefine `actions` schema in `customized-receivers` to facilitate parameters passing

Email Examples:
User:
image

Admin:
image


src/alert-manager/src/alert-handler/routes/actions.js
```diff
@@ -24,12 +24,12 @@ const router = express.Router();
 
 // email
 router
-  .route('/alert-handler/send-email-to-admin')
+  .route('/alert-handler/send-email-to-admin/:template*?')
```
Binyang2014, line 27: Change API to `/alert-handler/send-email-to-admin?template=template-name`?

Binyang2014:
Just wondering if we can put all `ejs` into separate folder, such as template folder and not build them into docker image (We can make them as a configmap and mount into container). So if customer want to change/add some templates, they just need to change some config.

contrib/kubespray/quick-start/services-configuration.yaml.template
```diff
@@ -183,16 +183,18 @@ authentication:
...
-#     - email-admin
-#     - email-user
-#     - stop-jobs
-#     - tag-jobs
-#     tags: 
-#     - 'stopped-by-alert-manager'
-
+#       # the email template for `email-admin` and `email-user `can be chosen from ['general-template', 'kill-low-efficiency-job-alert']
+#       # if no template specified, 'general-template' will be used.
+#       email-admin: No-Parameters-Required
```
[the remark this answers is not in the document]
suiguoxin: > Change to:
> 
> ```
> #       email-admin:  # No-Parameters-Required
> ```
> 
> So only need to delete first `#`

`No-Parameters-Required` is the value of key `email-admin`. Here we change the  value of `Actions` from a list to a dict. Simply `email-admin: ` doesn't conform to the yaml format. Users can assign whatever as the value if they don't need to specify the template.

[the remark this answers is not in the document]
Binyang2014: Can this work?
```yaml
 email-admin: {}
```

src/alert-manager/src/alert-handler/routes/actions.js
suiguoxin: > Change API to `/alert-handler/send-email-to-admin?template=template-name`?

Changed, thanks.

contrib/kubespray/quick-start/services-configuration.yaml.template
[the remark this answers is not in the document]
suiguoxin: > Can this work?
> 
> ```yaml
>  email-admin: {}
> ```

Modified, thanks

contrib/kubespray/quick-start/services-configuration.yaml.template
```diff
@@ -178,21 +178,25 @@ authentication:
 #       smtp-from: alert-sender@example.com
 #       smtp-auth-username: alert-sender@example.com
 #       smtp-auth-password: password-for-alert-sender
+#       customized-templates: # optional
+#       - your-customized-email-template
```
Binyang2014, line 182: Do we really need this option? Can we parse the folder to get the templates? I think you can modify the deploy python file to achieve this

suiguoxin: > Do we really need this option? Can we parse the folder to get the templates? I think you can modify the deploy python file to achieve this

Right, but I don't want to touch the python2 code...

suiguoxin: > Do we really need this option? Can we parse the folder to get the templates? I think you can modify the deploy python file to achieve this

How can we save the parsing results in `cluster_cfg` ?

Binyang2014: You can change https://github.com/microsoft/pai/blob/df25b985699b75a3813a2080ee6d437ca0db1bf2/src/alert-manager/config/alert_manager.py#L66-L69. Put the parsed result into `result` and return. Then you will get it from `cluster_cfg`

suiguoxin: > You can change
> 
> https://github.com/microsoft/pai/blob/df25b985699b75a3813a2080ee6d437ca0db1bf2/src/alert-manager/config/alert_manager.py#L66-L69
> 
> . Put the parsed result into `result` and return. Then you will get it from `cluster_cfg`

`alert_manager` is a class definition file, there is no guarantee on the path to pass

src/alert-manager/deploy/start.sh.template
```diff
@@ -19,6 +19,18 @@
...
+{% if 'email-admin' in cluster_cfg["alert-manager"]["actions-available"] %}
+kubectl create configmap alert-templates \
+{% for template in cluster_cfg["alert-manager"]["alert-handler"]["email-configs"]["templates"] -%}
+--from-file={{ template }}-html.ejs=alert-templates/{{ template }}/html.ejs \
+--from-file={{ template }}-subject.ejs=alert-templates/{{ template }}/subject.ejs \
+{% endfor -%}
+--dry-run -o yaml | kubectl apply --overwrite=true -f - || exit $?
+{% endif %}
+{% endif %}
+
```
Binyang2014, line 33: Please delete this configmap in stop.sh

Binyang2014 (approved):

src/alert-manager/deploy/start.sh.template
suiguoxin: > Please delete this configmap in stop.sh

Added, thanks

PR #5075, opened 2020-11-11
Support CPU jobs on GPU nodes in HiveD scheduler

### LINKED ISSUE #5058: Set cell indices in HiveD scheduler according to cell types
To support GPU cells and CPU cells on one single GPU node, HiveD scheduler needs to set cell indices in different env according to cell types:
```
if nvidia gpu:
  set NVIDIA_VISIBLE_DEVICES=id
else if amd gpu:
  set AMD_VISIBLE_DEVICES=id
else:
  set NVIDIA_VISIBLE_DEVICES=none
```

abuccts:
Support CPU jobs on GPU nodes in HiveD scheduler.

Closes #5058.


yqwang-ms (approved):
:shipit:

hzy46 (approved):

PR #5082, opened 2020-11-13
Support job transfer

shaiic-pai:


yiyione (approved):

debuggy (approved):

hzy46:
Test Case (the following items should be tested in both basic auth and AAD mode):

1. Set `webportal.enable-job-transfer = false` in `services-configuration.yaml`. Deploy the cluster, and confirm:
   - Submit a job. On the job detail page, the clone button should have no `Transfer` option.
   - On the user profile page, you cannot see the `Bounded Clusters` section.

2. Set `webportal.enable-job-transfer = true` in `services-configuration.yaml`. Deploy the cluster.
    1.  Confirm:
        - Submit a job. On the job detail page, the clone button should have a `Transfer` option.
        - On other user's job detail page, you cannot see the `Transfer` option.
        - Click the transfer button, it should warn you that `You should add some bounded clusters`.
        - On the user profile page, you can see the `Bounded Clusters` section.
   2. on the user profile page:
         - Add some bounded clusters
         - Try to delete some bounded clusters
   3. Test job transfer
        - Submit a job. On the job detail page, click `Transfer`.
        - Try to input wrong VC/storage/sku type, then click `confirm transfer`. The job should not be transferred.
        - Edit the job config to make it a valid config for the target cluster. Then click `Confirm transfer.`
        - Click `Stop it`
        - The job should be transferred, and you can see the URL on the job detail page. The previous job should be stopped.

PR #5091, opened 2020-11-18
Fix exit code in service-boot.sh

hzy46:


hzy46:
The previous pattern `<command> || exit $?` actually doesn't work.

Use `set -e` in the script instead.

suiguoxin (approved):

PR #5096, opened 2020-11-18
fix 404 error

AmberMsy:


src/webportal/src/app/job/job-view/fabric/job-detail/conn.js
```diff
@@ -183,7 +183,10 @@ export async function getContainerLogList(logListUrl) {
 export async function getContainerLog(tailLogUrls, fullLogUrls, logType) {
   const res = await fetch(tailLogUrls[logType]);
   if (!res.ok) {
-    throw new Error(res.status);
+    if (String(res.status) === '403') {
```
hzy46, line 186: Why convert it to string? Would `res.status === 403` work as expected?

AmberMsy: No, I have to add `String()` so it can work as expected. `res.status == '403'` work too but it can't pass the lint check.

hzy46 (approved):

PR #5129, opened 2020-11-26
[Marketplace]Use marketplace images from docker hub

yiyione:
After this change.
Admin can view the marketplace image tag in k8s dashboard, and don't have to build images by `paictl.py` before deploy marketplace.


debuggy:
Could we close this pr since it is not used currently?

PR #5166, opened 2020-12-11
Fix remove nodes doc issue

suiguoxin:
changes:
- remove node: no need to modify `hosts.yml`, as introduced in  [kubespray](https://github.com/kubernetes-sigs/kubespray/blob/release-2.11/docs/getting-started.md#remove-nodes)
- use `scale.yml`, `remove-node.yml` instead of `update-cluster.yml`
- `--limit` flag is no longer recommended in kubespray 2.11, we can use twice `-e` to specify the nodes to modify

Future refactor about using `layout.yaml` instead of `hosts.yaml` will be discussed separately.

### DIFF
--- a/docs/manual/cluster-admin/how-to-add-and-remove-nodes.md
+++ b/docs/manual/cluster-admin/how-to-add-and-remove-nodes.md
@@ -6,7 +6,19 @@ OpenPAI doesn't support changing master nodes, thus, only the solution of adding
 
 ### Preparation
 
-To add worker nodes, please check if the nodes meet [the requirements for worker machines](./installation-guide.md#installation-requirements).
+To add worker nodes, please check if the nodes meet the following requirements:
+
+  - Ubuntu 16.04 (18.04 should work, but not fully tested.)
+  - Assign each node a **static IP address**, and make sure nodes can communicate with each other. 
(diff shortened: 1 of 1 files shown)


docs/manual/cluster-admin/how-to-add-and-remove-nodes.md
```diff
@@ -6,19 +6,7 @@ OpenPAI doesn't support changing master nodes, thus, only the solution of adding
...
-    - You may use command `sudo docker run nvidia/cuda:10.0-base nvidia-smi` to check it. This command should output information of available GPUs if it is setup properly.
-    - Refer to [the installation guidance](./installation-faqs-and-troubleshooting.md#how-to-install-nvidia-container-runtime) if the it is not successfully set up.
-  - OpenPAI reserves memory and CPU for service running, so make sure there are enough resource to run machine learning jobs. Check hardware requirements for details.
-  - Dedicated servers for OpenPAI. OpenPAI manages all CPU, memory and GPU resources of servers. If there is any other workload, it may cause unknown problem due to insufficient resource.
+To add worker nodes, please check if the nodes meet [the requirements for worker machines](./installation-guide.md#installation-requirements).
```
hzy46, line 9: It is not the same.

Installation requires workers should have GPU, which is not applied when add/remove node.

I suggest leaving this change to the bug bash period.

hzy46 (approved):

PR #5184, opened 2020-12-21
layout schema check

suiguoxin:


contrib/kubespray/script/validate_layout_schema.py
```diff
@@ -0,0 +1,118 @@
...
+                    Optional('computing-device'): {
+                        'type': str,
+                        'model': str,
+                        'count': int,
+                    }
+                }
+            },
+            'machine-list': [
+                {
+                    'hostname': str,
```
hzy46, line 50: For hostname, please use this regex: `[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`

See https://github.com/kubernetes-sigs/kubespray/blob/release-2.11/roles/kubernetes/preinstall/tasks/0020-verify-settings.yml#L124

contrib/kubespray/script/validate_layout_schema.py
```diff
@@ -0,0 +1,118 @@
...
+        return False
+    if len(masters) > 1:
+        logger.error('More than one master machine specified.')
+        return False
+
+    # pai-master / pai-worker cannot be true at the same time
+    if 'pai-worker' in masters[0] and masters[0]['pai-worker'] == 'true':
+        logger.error("One machine can not be pai-master and pai-worker at the same time.")
+        return False
+
```
hzy46, line 93: We should also check there is at least one pai-worker

contrib/kubespray/script/validate_layout_schema.py
suiguoxin: > For hostname, please use this regex: `[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
> 
> See https://github.com/kubernetes-sigs/kubespray/blob/release-2.11/roles/kubernetes/preinstall/tasks/0020-verify-settings.yml#L124

Fixed, thx

contrib/kubespray/script/validate_layout_schema.py
suiguoxin: > We should also check there is at least one pai-worker.

Added

hzy46 (approved):

Issue #5149, opened 2020-12-03
Deploy: paictl add skip services options


Binyang2014:
Currently, when running paictl, we can not skip to deploy some services.
But some services are critical and can not be stopped, such as `storage-manager` and `dshuttle`. Seems we should try to skip this type of services.

suiguoxin:
Proposals:

#### Add an argument parser:
parser.add_argument("-n", "--service-list", default=None, help="Service list to manage")
**- parser.add_argument("-k", "--skip-service-list", default=None, help="Service list to skip")**

#### constraints:
- return error if the service name is included both args

PR #5193, opened 2020-12-23
add skip-service-list arg

suiguoxin:
#5149

### DIFF
--- a/deployment/serviceCmd.py
+++ b/deployment/serviceCmd.py
@@ -17,6 +17,7 @@
 
 
 import os
+import sys
 import readline
 import logging
 import logging.config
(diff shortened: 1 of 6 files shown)


hzy46 (approved):

deployment/serviceCmd.py
```diff
@@ -61,22 +63,25 @@ def add_arguments(parser):
     def process_args(self, args):
         if args.kube_config_path is not None:
             args.kube_config_path = os.path.expanduser(args.kube_config_path)
-        return args.service_list
+
```
Binyang2014, line 66: Try to use https://docs.python.org/2.7/library/argparse.html#mutual-exclusion?

suiguoxin: > Try to use https://docs.python.org/2.7/library/argparse.html#mutual-exclusion?

Fixed, thx

Binyang2014 (approved):

Issue #5140, opened 2020-12-02
add support to enflame DTU


beingj:
Track tasks to add support to enflame DTU
Copied from #5138
- [ ] deploy script: pre checks / installation items for each machine model
- [ ] exporter: additional exporters for machine models besides nvidia gpu / cpu machine.
- [ ] rest-server: When generating pod spec, rest-server should be aware of environmental variables used by other device in addition to NVIDIA_VISIBLE_DEVICES and PAI_AMD_VISIBLE_DEVICES.
- [ ] scheduler: health report for computing device. This is not mandatory since node-level health check is provided by k8s already.

hzy46:
Hi, for specific work items, please refer to the following:

### In Jan 2021 Release
  - [ ] add dtu device plugin in device plugin service
    - https://github.com/microsoft/pai/blob/86c73e32f8f2c8b2b651bd71a2855ff116c27c83/src/device-plugin/deploy/start.sh.template#L47-L49
    - https://github.com/microsoft/pai/blob/86c73e32f8f2c8b2b651bd71a2855ff116c27c83/src/device-plugin/deploy/delete.sh.template#L42-L43 
 - [ ] add dtu device check in installation script
   - Please write a [ansible role](https://docs.ansible.com/ansible/latest/user_guide/playbooks_reuse_roles.html) `enflame.com_dtu` and place it in https://github.com/microsoft/pai/tree/86c73e32f8f2c8b2b651bd71a2855ff116c27c83/contrib/kubespray/roles/requirement/computing-devices
   - Please reference it in https://github.com/microsoft/pai/blob/86c73e32f8f2c8b2b651bd71a2855ff116c27c83/contrib/kubespray/environment-check.yml#L55-L70
   - You can see the check for nvidia gpu for reference in https://github.com/microsoft/pai/tree/86c73e32f8f2c8b2b651bd71a2855ff116c27c83/contrib/kubespray/roles/requirement/computing-devices/nvidia.com_gpu

After Jan 2021 release, we expect openpai can support installation, scheduling and job submission with enflame.com/dtu.

### In Future Releases

- [ ] Metrics reporting: add exporter for dtu
- [ ] webportal

PR #5201, opened 2020-12-24
add dtu device plugin

beingj:
add Enflame DTU device plugin in device plugin service as requested by #5140


hzy46:
Please also add `enflame.com/dtu` here: https://github.com/microsoft/pai/blob/5fd85123b6840da3a53cb8408ec389e672e67db7/contrib/kubespray/environment-check.yml#L55-L70

hzy46 (approved):

PR #5229, opened 2021-01-07
remove irrelevant labels from nodes in cluster-configuration service

suiguoxin:
Issue: 
- Before `cluster-configration` service starts, the nodes may have already some labels. 
For example, master node will be added a label `pai-worker=true` after `quick-start-kubespray.sh` because this is set as a default kubelet start config.
These wrong put labels may leads to undisred behaviors. 
- `pai-worker` label is needed by job scheduler, the missing of this label may leads to job scheduling failure.

Solution:
- add labels according to config in `layout.yaml`, remove if not exist

Modif in this PR :
- start.sh: 
    - remove `pai-master`, `pai-worker` `pai-storage` label from nodes if not specified in `layout.yaml`
    - create namespace pai-storage only when it doesn't exist (avoid confusing err message in console)
    - remove redundant jinja2 empty lines


hzy46 (approved):

Binyang2014 (changes requested):
Please make sure not remove the label when updating the service

Binyang2014 (commented):
Please provide the test results after change the installation config

Binyang2014 (approved):

hzy46 (approved):

Binyang2014 (approved):

Issue #3387, opened 2019-08-16
Investigate GPU perf state add alert when GPU perf state is abnormal


Binyang2014:
Sometimes GPU perf state change to P8 and cause low performance for running job. Need to investigate why it happens.
If it's an abnormal issue, need to add alert.

Binyang2014:
Seem this issue again,
Fix by 
```bash
sudo nvidia-smi -pm ENABLED -i 3
sudo nvidia-smi -ac 2505,875 -i 3
```

Seems we can get this situation by nvdia-smi -q -d PERFORMANCE

PR #5238, opened 2021-01-10
add alert for GPU low perf state

Binyang2014:
GPU may drop into low perf state in some case. Which cause user workload run much slower. Add this alert. Fix: #3387


Binyang2014:
Fix: #3387

suiguoxin:
Can we add the solution to this alert to the troubleshooting doc ?

Binyang2014:
> Can we add the solution to this alert to the troubleshooting doc ?

Doc updated

suiguoxin (approved):

PR #5241, opened 2021-01-11
Fix installation check: K8S default scheduler only supports <= 1 type of computing device.

hzy46:


contrib/kubespray/script/validate_layout_schema.py
```diff
@@ -79,6 +85,11 @@ def check_layout(layout):
     if 'pai-worker' in masters[0] and masters[0]['pai-worker'] == 'true':
         logger.error("One machine can not be pai-master and pai-worker at the same time.")
         return False
+    # if cluster_config.enable_hived_scheduler is false, there should be <= 1 type of computing device
+    if 'enable_hived_scheduler' in cluster_config and cluster_config['enable_hived_scheduler'] is False:
```
suiguoxin, line 89: Will hived scheduler be enabled if `enable_hived_scheduler` is not set in `config.yaml` ?

contrib/kubespray/script/validate_layout_schema.py
```diff
@@ -79,6 +85,11 @@ def check_layout(layout):
...
+        if len(worker_computing_devices) > 1:
+            logger.error('K8S default scheduler only supports <= 1 type of computing device.')
```
suiguoxin, line 91: `logger.error('... as workers nodes')` ?
We should recommend the users to use hived in this case

contrib/kubespray/script/validate_layout_schema.py
hzy46: yes

contrib/kubespray/script/validate_layout_schema.py
hzy46: done

suiguoxin (approved):

Issue #5243, opened 2021-01-12
Can "View Full Log" display the log in a new tab?


ultmaster:
![image](https://user-images.githubusercontent.com/8463288/104271584-3ee71c00-54d6-11eb-8c7d-4b6dfad244dc.png)

It's inconvenient to have "view full log" directly download the log, especially when the downloaded log doesn't have file extensions for OS like Windows.

Binyang2014:
It's by design for download behavior. If you want to set default application to view log, we could add file extensions for the log file.

ultmaster:
@Binyang2014 Thanks for reply. I think that's true but again it's quite inconvenient. My download looks like:

![image](https://user-images.githubusercontent.com/8463288/104274987-f67f2c80-54dc-11eb-9350-c1a690024bc5.png)

When I click to open it, it always asks me to choose application:

![image](https://user-images.githubusercontent.com/8463288/104275013-039c1b80-54dd-11eb-88ee-b2c485f221d7.png)

I remember the design used to be opened in browser in previous versions. May I ask why that design has been abandoned?

Binyang2014:
The user log maybe quiet large. Render large file in browser may cause browser using much CPU/Mem and cause other pages slow down. Using download method will avoid this problem.
We can change download file name to `user-stdout.log` and `user-stderr.log`. Then you can choose vs code to open file with `.log` extension.

ultmaster:
Sounds reasonable to me. Thanks!

PR #5272, opened 2021-01-28
Add ".log" suffix for user-stdout/user-stderr/user-all log files

Binyang2014:
Add ".log" suffix for user-stdout/user-stderr/user-all log files
Fix: #5243


Gerhut:
Why not directly add `.log` to the actual url? Avoid breaking change?

Binyang2014:
> Why not directly add `.log` to the actual url? Avoid breaking change?

Because it's easier to implement and do not need to care about how log is stored.
The log url we returned is: `https://pai-master/log-manager/..../api/v1/logs/user-stdout?...`
The real log is stored at `/var/pai/log/.../user-stdout/current`
And we have some url such as  `https://pai-master/log-manager/..../api/v1/logs/user-stdout.1?...` for rotated log.
The real log is stored at `/var/pai/log/.../user-stdout/@....s`

If we change url, we need to extract the real log folder before try to get the content. Using this way to avoid string manipulate

Gerhut (approved):

PR #5273, opened 2021-01-28
bump DShuttle version

Binyang2014:


src/dshuttle-csi/deploy/dshuttle-csi-daemon.yaml.template
```diff
@@ -33,7 +33,7 @@ spec:
       initContainers:
       # This change aimed to avoid OS build file index for alluxio-fuse-fs, which will consume much disk space
       - name: change-updatedb-conf
-        image: dshuttle.azurecr.io/dshuttle/dshuttle-csi:25037dc
+        image: dshuttle.azurecr.io/dshuttle/dshuttle-csi:54cff8fc
```
abuccts, line 36: why change 7 digits to 8?
could you have a config setting for this image? so that only one place should be modified

PR #5281, opened 2021-02-02
send regular GPU utilization report with CronJob

suiguoxin:
- create a k8s CronJob which :
    - query `Prometheus` regularly for GPU utilization data
    - send cluster-usage alerts to `alert-manager`
    - the alerts will be handled by `alert-handler`

This feature was previously implemented by @Binyang2014 [here](https://github.com/microsoft/pai/blob/binyli/alert/src/alert-manager/deploy/alert-templates-notice/notice.tmpl), here we just adapt the code into `alert-manager`.

The alert mail:
image


src/alert-manager/deploy/alert-manager-cronjob.yaml.template
```diff
@@ -0,0 +1,50 @@
...
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+# DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+
+{% if cluster_cfg["alert-manager"]["use-pylon"] %}
+{% set external_url = cluster_cfg["pylon"]["uri"] %}
+{% else %}
+{% set external_url = cluster_cfg["alert-manager"]["url"] %}
+{% endif %}
```
Binyang2014, line 23: Why need these lines?

src/alert-manager/src/cluster-utilization/send_alert.py
```diff
@@ -0,0 +1,194 @@
...
+    job_infos = {}
+    job_usage = []
+    job_list = []
+    # get all jobs
+    headers = {'Authorization': "Bearer {}".format(TOKEN)}
+    resp = requests.get(rest_url, headers=headers)
+    if resp.ok:
+        job_list = resp.json()
+    else:
+        logging.warning("Failed to get job list")
```
Binyang2014, line 46: Throw exception here?

src/alert-manager/src/cluster-utilization/send_alert.py
```diff
@@ -0,0 +1,194 @@
...
+            logging.warning("request failed %s", resp.text)
+            continue
+        resp_json = resp.json()
+        if not resp_json["jobStatus"]["appLaunchedTime"]:
+            logging.warning("job not start, ignore it")
+            continue
+        job_info["start_time"] = datetime.fromtimestamp(
+            int(resp_json["jobStatus"]["appLaunchedTime"]) / 1000,
+            timezone.utc)
+        if not resp_json["jobStatus"]["appLaunchedTime"] or not resp_json[
```
Binyang2014, line 63: Do not need to check `resp_json["jobStatus"]["appLaunchedTime"]` here?

Binyang2014 (commented):
Can admin disable this report by comment the config?

src/alert-manager/src/cluster-utilization/send_alert.py
suiguoxin: > Do not need to check `resp_json["jobStatus"]["appLaunchedTime"]` here?

Removed, thx

src/alert-manager/src/cluster-utilization/send_alert.py
suiguoxin: > Throw exception here?

Fixed, thx

src/alert-manager/deploy/alert-manager-cronjob.yaml.template
suiguoxin: > Why need these lines?

Removed, thanks

suiguoxin:
> Can admin disable this report by comment the config?

The code is refactored. After this change, the report will only be enabled when the `schedule` field is set.

Binyang2014 (approved):

Issue #5270, opened 2021-01-28
generate an application token for every job


suiguoxin:
### Motivation
- enable runtime plugins to access rest-server

### Design
- Background: currently, all user tokens are saved in the same k8s secret in namespace 'pai-user-token', in `data` field
-  save `job-specific` tokens in another namespace named 'pai-job-token', **one secret for one user**
- do not return `job-specific` tokens in get tokens API by default
- verify tokens: check both normal tokens & job-specific tokens
- revoke job-specific tokens :
   - in database-controller, when jobs finished
   - in rest-server, `purge()` function, each time when `list()` / `verify()` is called

#### API change
- revoke tokens: 
  - revoke self: no constraint
  - else: not application_token

#### Schema of the job-specific secrets:
```yaml
apiVersion: v1
data:
  encode(frameworkName): token
kind: Secret
metadata:
  creationTimestamp: "2021-01-18T01:52:22Z"
  name: username
  namespace: **pai-job-token**
  resourceVersion: "10515250"
  selfLink: /api/v1/namespaces/pai-job-token/secrets/hash(username)
  uid: 72658f49-d620-45c5-ab07-e5fd0a14acb9
type: Opaque
```

### Work Items
- [x] create application token in rest-server, get a k8s secret definition #5282
- [x] create a token object in db, (add a column in table Framework named tokenSecretDef...), let dbc create the secret with the def #5282
- [x] revoke the token in dbc, (remove the token from DB) : realized in src/dbc #5282
- [x] filter/hide token in webportal/rest-server in user-profile page
    - save the secret in a different namespace, return only user-tokens by default #5292
- [x] revoke the job tokens after jobs finished
    - [x] in purge() function #5292

suiguoxin:
Change:

## Issues of the approach above:
- Save job-specific token in k8s secrets leads to more complicated purge() & verify() operation, which is of O(n) time complexity.

## New design
### sign 
sign a token with the following payload schema:
``` 
{
    username,
    application,
    jobSpecific, // indicate if the token if for a specific job only
    frameworkName, 
}
```
- no need to save the token in 

### verify
for job specific token verification:
- vefiry with jwt
- check if the job has completed
- no need to check k8s secret data

### purge
no need to purge job-specific tokens since they are not saved

PR #5282, opened 2021-02-02
generate an application token for every job

suiguoxin:
#5270 
- create application token in rest-server, get a k8s secret definition
- create a token object in db, (add a column in table Framework named tokenSecretDef...), let dbc create the secret with the def
- mount the token-secrets to initContainers & job containers
- revoke the token in dbc, (remove the token from DB) : realized in src/dbc

Test cases:
- submit a job with secrets
- check 'kubectl get secret | grep cred', you should see corresponding `configcred` and `tokencred`
- stop the job, the secrets should be deleted automatically


src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -442,6 +443,10 @@ const generateTaskRole = (
                   name: 'KUBE_APISERVER_ADDRESS',
                   value: apiserver.uri,
                 },
+                {
+                  name: 'REST_SERVER_URI',
+                  value: process.env.REST_SERVER_URI,
```
hzy46, line 448: Please don't use environmental variables here. Put it in `launcherConfig`

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -1053,19 +1090,24 @@ const put = async (frameworkName, config, rawConfig) => {
...
     : null;
 
-  // generate job config secret
+  // generate the job config secret definition
   const configSecretDef = config.secrets
     ? getConfigSecretDef(frameworkName, config.secrets)
     : null;
 
+  // create an application token
+  const token = await tokenModel.create(userName, true);
```
hzy46, line 1107: What if:

```
1. the write merger is down 
2. the token is created here
3. the job cannot be submitted to write merger
```

This token will never be deleted or revoked.

hzy46: Better to give each token a name or a label related to the job thus we can recycle them in the watchdog

src/rest-server/src/models/v2/job/k8s.js
suiguoxin: > Please don't use environmental variables directly here. Put it in `launcherConfig`

fixed, thx

src/rest-server/src/models/v2/job/k8s.js
suiguoxin: > Better to give each token a name or a label related to the job thus we can recycle them in the watchdog (maybe implemented in the future)

Added a TODO item here, will resolve this later

hzy46 (approved):

PR #5283, opened 2021-02-02
[Marketplace] Simplify marketplace setting

yiyione:
1. Admin can deploy marketplace in PAI quick start script (set enable_marketplace: true in config.yaml)
2. Simplify marketplace settings (only need to set `cluster.common.marketplace = "true"`
3. Add prerequisite to marketplace-webportal (require marketplace-restserver) and marketplace-restserver (require marketplace-db)
4. Will auto add marketplace to webportal-plugins if set `cluster.common.marketplace = "true"`

Related issue & test cases: https://github.com/microsoft/openpaimarketplace/issues/209#issue-799930634

### DIFF
--- a/contrib/kubespray/quick-start/services-configuration.yaml.template
+++ b/contrib/kubespray/quick-start/services-configuration.yaml.template
@@ -7,6 +7,7 @@ cluster:
     data-path: "/datastorage"
     qos-switch: "{{ env["cfg"]["qos-switch"] | default('false') }}"
     docker-data-root: "{{ env['cfg']['docker_data_root'] | default('/mnt/docker') }}"
+    marketplace: "true"
 
   # the docker registry to store docker images that contain system services like frameworklauncher, hadoop, etc.
   docker-registry:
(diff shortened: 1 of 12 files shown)


src/webportal/deploy/webportal.yaml.template
```diff
@@ -91,7 +91,24 @@ spec:
         - name: WEBPORTAL_PLUGINS
           # A raw JSON formatted value is required here.
           value: |
-            {{ cluster_cfg['webportal']['plugins'] }}
+            [
+{% if cluster_cfg['cluster']['common']['marketplace'] == "true" %}
```
hzy46, line 95: What if:

```
1. The admin deploys marketplace in a previous version of PAI
2. He upgrades PAI to `v1.6.0`
```

Will there be two duplicate marketplace entries here?

contrib/kubespray/quick-start/services-configuration.yaml.template
```diff
@@ -7,6 +7,7 @@ cluster:
     data-path: "/datastorage"
     qos-switch: "{{ env["cfg"]["qos-switch"] | default('false') }}"
     docker-data-root: "{{ env['cfg']['docker_data_root'] | default('/mnt/docker') }}"
+    marketplace: "true"
```
hzy46, line 10: Maybe `marketplace: "false"`? 

Or if we decide to turn on marketplace by default, better to notify all PAI devs.

src/webportal/deploy/webportal.yaml.template
yiyione: Yes, if the old webportal-plugin still in services-configurations.yaml, there will be two duplicate marketplace entries here.

yiyione: I will add an optional config to disable the auto-added marketplace-plugin in webportal.
The admin can avoid the duplicate marketplace entries by it.

contrib/kubespray/quick-start/services-configuration.yaml.template
yiyione: Done, set to "false".

hzy46 (approved):

debuggy (approved):

Issue #4889, opened 2020-09-09
make DBController tolerant to wrong framework request


hzy46:
- Currently, database controller assumes all framework requests it receives are valid. Thus it will try to sync every framework request to api server.
- However, it is hard to validate the request in the rest-server. There is chance that rest-server generates wrong framework requests. Then database controller will keep trying to sync the wrong request forever.
- We could make DBcontroller tolerant to wrong framework request.

yqwang-ms:
In previous design #4651,
We assume apiserver submission error as our platform issues, such as we generated invalid json object, so we always retry and expose the errors as job events to user.
![image](https://user-images.githubusercontent.com/32826762/92567522-658b1100-f2b0-11ea-8105-11e4bcf4e5b8.png)

Is that good enough?

hzy46:
> In previous design #4651,
> We assume apiserver submission error as our platform issues, such as we generated invalid json object, so we always retry and expose the errors as job events to user.
> ![image](https://user-images.githubusercontent.com/32826762/92567522-658b1100-f2b0-11ea-8105-11e4bcf4e5b8.png)
> 
> Is that good enough? As you cannot totally distingush between transient and permenant apiserver submission error.
> 
> BTW, any real example for the apiserver submission error?
> Curious how can we generate invalid framework object

If such error happens, I think the user may not able to stop the job. e.g. rest-server generates a wrong request object  -> the request object is stored in db ->  db controller tries to sync it to api server, but failed -> user stop the job, which will patch the `spec.executionType` field of the framework request object in db -> db controller tries to sync it to api server, but failed -> user will see his job is always in `Stopping` status.

I haven't found a real example.

yqwang-ms:
@hzy46  Valid point for the stop request cannot be satisfied, but we have to send stop to apiserver, because you do not know if the job really stopped (or not exist) in k8s. Considering this issue is most caused by our platform issues, under platform issues, everything may not work properly, so hang in `stopping` is acceptable, and we need to fix the platform issue ASAP/

hzy46:
@yqwang-ms 

If ApiServer non-404 error, can we **mock a Failed Framework event** to watcher? I think it is a clean solution for this problem.

hzy46:
It should be:
  - If ApiServer reports non-404 error, and it is caused by a invalid protocol, mock a Failed Framework event to watcher.
  - If ApiServer reports other non-404 error, add the event to db.

yqwang-ms:
[...]

Issue #4840, opened 2020-08-26
Failed to create Pod: couldn't get version/kind; json parse error: unexpected end of JSON input (\u003cempty\u003e)


yqwang-ms:
Job failed due to:
```
Exit Code: -200
Exit Type: USER_FAILURE
Exit Reason: 
Pod spec contains permanent error
Failed to create Pod: {"ErrStatus":{"metadata":{},"status":"Failure","message":"the object provided is unrecognized (must be of type Pod): couldn't get version/kind; json parse error: unexpected end of JSON input (\u003cempty\u003e)","reason":"BadRequest","code":400}}
```
However, the Pod body in http request is fine, so it seems like ApiServer or k8s client-go try to decode http body even if the underlay connection has been dropped.
So seems https://github.com/kubernetes/kubernetes/issues/51353 does not totally fix the issue.

Still need to confirm the root cause.
In the meanwhile, we can treat K8s BadRequest as unknown error instead of permanent (even if k8s code says `API calls that return BadRequest can never succeed.`), to mitigate this.

yqwang-ms:
Resolved by https://github.com/microsoft/pai/pull/4860
The pod decoding error will always be treated as transient error so forever retry.

PR #5284, opened 2021-02-03
make DBController tolerant to wrong framework request

hzy46:
to fix https://github.com/microsoft/pai/issues/4889

### DIFF
--- a/src/database-controller/src/common/framework.js
+++ b/src/database-controller/src/common/framework.js
@@ -24,10 +24,34 @@ const mockFrameworkStatus = () => {
 };
 
 const mockFailedFrameworkStatus = () => {
+  // Mock a failed status for framework.
+  // We only use it for frameworks that never start.
+  const dateStr = (new Date()).toISOString();
   return {
     state: 'Completed',
+    startTime: dateStr,
+    runTime: dateStr,
+    transitionTime: dateStr,
+    completionTime: dateStr,
     attemptStatus: {
-      completionStatus: null,
+      completionStatus: {
+        id: 0,
+        code: -1007,
+        diagnostics: 'Framework schema is submitted to database, but cannot be synchronized to API server due to unrecoverable error.',
+        phase: 'UnrecoverableSchemaSynchronizeError',
+        trigger: {
+          message: "Database controller cannot synchronize framework schema to API server.",
+          taskIndex: null,
+          taskRoleName: null,
+        },
+        type: {
+          attributes: ["Permanent"],
+          name: "Failed",
+        },
+        startTime: dateStr,
+        runTime: dateStr,
+        completionTime: dateStr,
+      },
       taskRoleStatuses: [],
     },
     retryPolicyStatus: {
@@ -260,6 +284,10 @@ class Snapshot {
     return this._snapshot.status.state;
   }
 
+  getTotalRetriedCount() {
+    return _.get(this._snapshot, 'status.retryPolicyStatus.totalRetriedCount', 0);
+  }
+
   getSnapshot() {
     return _.cloneDeep(this._snapshot);
   }
@@ -310,6 +338,13 @@ class Snapshot {
     // Manually set the framework's status to failed.
     // This is used when we cannot sync the framework request to api server,
     // and we're sure that the error is unrecoverable.
+
+    // For now, we only use it for frameworks that never start.
+    console.log(this.getTotalRetriedCount(), this.getState())
+    if (!this.getTotalRetriedCount() === 0 || this.getState() !== 'AttemptCreationPending') {
+      throw new Error('setFailed() only works for framework that never start!')
+    }
+
     this._snapshot.status = mockFailedFrameworkStatus();
   }
 }
--- a/src/database-controller/src/common/framework.js
+++ b/src/database-controller/src/common/framework.js
@@ -36,11 +36,11 @@ const mockFailedFrameworkStatus = () => {
     attemptStatus: {
       completionStatus: {
         id: 0,
-        code: -1007,
-        diagnostics: 'Framework schema is submitted to database, but cannot be synchronized to API server due to unrecoverable error.',
-        phase: 'UnrecoverableSchemaSynchronizeError',
+        code: -1100,
+        diagnostics: 'Job is submitted to database, but cannot be created in ApiServer due to permanent failures.',
+        phase: 'CreateFrameworkPermanentFailed',
         trigger: {
-          message: "Database controller cannot synchronize framework schema to API server.",
+          message: "Database controller cannot synchronize framework to APIServer.",
           taskIndex: null,
           taskRoleName: null,
         },
@@ -340,7 +340,6 @@ class Snapshot {
     // and we're sure that the error is unrecoverable.
 
     // For now, we only use it for frameworks that never start.
-    console.log(this.getTotalRetriedCount(), this.getState())
     if (!this.getTotalRetriedCount() === 0 || this.getState() !== 'AttemptCreationPending') {
       throw new Error('setFailed() only works for framework that never start!')
     }
--- a/src/database-controller/src/common/framework.js
+++ b/src/database-controller/src/common/framework.js
@@ -35,19 +35,14 @@ const mockFailedFrameworkStatus = () => {
     completionTime: dateStr,
     attemptStatus: {
       completionStatus: {
-        id: 0,
         code: -1100,
         diagnostics: 'Job is submitted to database, but cannot be created in ApiServer due to permanent failures.',
         phase: 'CreateFrameworkPermanentFailed',
-        trigger: {
-          message: "Database controller cannot synchronize framework to APIServer.",
-          taskIndex: null,
-          taskRoleName: null,
-        },
         type: {
           attributes: ["Permanent"],
           name: "Failed",
         },
+        id: 0,
         startTime: dateStr,
         runTime: dateStr,
         completionTime: dateStr,
@@ -340,6 +335,9 @@ class Snapshot {
     // and we're sure that the error is unrecoverable.
 
     // For now, we only use it for frameworks that never start.
+    // Also, requestSynced == false is guaranteed here.
+    // Because only call setFailed() from synchronizeHandler() in poller,
+    // and synchronizeHandler() is only called for requestSynced=false frameworks.
     if (!this.getTotalRetriedCount() === 0 || this.getState() !== 'AttemptCreationPending') {
       throw new Error('setFailed() only works for framework that never start!')
     }
--- a/src/database-controller/src/common/framework.js
+++ b/src/database-controller/src/common/framework.js
@@ -23,6 +23,21 @@ const mockFrameworkStatus = () => {
   };
 };
 
+const mockFailedFrameworkStatus = () => {
+  return {
+    state: 'Completed',
+    attemptStatus: {
+      completionStatus: null,
+      taskRoleStatuses: [],
+    },
+    retryPolicyStatus: {
+      retryDelaySec: null,
+      totalRetriedCount: 0,
+      accountableRetriedCount: 0,
+    },
+  };
+};
+
 const convertFrameworkState = (state, exitCode) => {
   switch (state) {
     case 'AttemptCreationPending':
@@ -289,6 +304,13 @@ class Snapshot {
       delete patchData.status;
     }
     this._snapshot = jsonmergepatch.apply(this._snapshot, patchData);
+  }
+
+  setFailed() {
+    // Manually set the framework's status to failed.
+    // This is used when we cannot sync the framework request to api server,
+    // and we're sure that the error is unrecoverable.
+    this._snapshot.status = mockFailedFrameworkStatus();
   }
 }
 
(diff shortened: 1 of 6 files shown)


src/database-controller/src/poller/utils.js
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+require('module-alias/register');
+const _ = require('lodash');
+const { v4: uuidv4 } = require('uuid');
+
+function isUnrecoverableResponse(response) {
+  // indentify if the response is recoverable
+  const statusCode = _.get(response, 'statusCode', 0);
+  const message = _.get(response, 'body.message', '');
```
yqwang-ms, line 10: If the error is connection reset or connection eof error, should return false

yqwang-ms: Such as 
 // Treat Permanent Framework Error as Permanent Error only if it must be
 // Permanent Error.
 if apiErrors.IsInvalid(apiErr) ||
  apiErrors.IsRequestEntityTooLargeError(apiErr) {
  // TODO: Also check net.IsConnectionRefused
  if net.IsConnectionReset(apiErr) || net.IsProbableEOF(apiErr) {
   // The ApiServer Permanent Error may be caused by Network Transient Error,
   // so treat it as Unknown Error for safety.
   return PodMatchResult{
    CodeInfo:    completionCodeInfoMap[CompletionCodePodCreationUnknownError],
    Diagnostics: diag,
   }
  } else {
   return PodMatchResult{
    CodeInfo:    completionCodeInfoMap[CompletionCodePodCreationPermanentError],
    Diagnostics: diag,
   }
  }
 }

---
In reply to: [570742486](https://github.com/microsoft/pai/pull/5284#discussion_r570742486) [](ancestors = 570742486)

src/database-controller/src/poller/utils.js
```diff
@@ -0,0 +1,66 @@
...
+}
+
+function generateClusterEventUpdate(snapshot, type, reason, message) {
+  const frameworkName = snapshot.getName();
+  const date = new Date();
+  const mockedObj = {
+    kind: 'Event',
+    apiVersion: 'v1',
+    metadata: {
+      name: `mocked-event-for-${frameworkName}`,
```
yqwang-ms, line 37: >mocked [](start = 13, length = 6)

db-controller-event-for-${frameworkName}

src/database-controller/src/common/framework.js
```diff
@@ -23,6 +23,21 @@ const mockFrameworkStatus = () => {
   };
 };
 
+const mockFailedFrameworkStatus = () => {
+  return {
```
yqwang-ms, line 27: We can only produce event and does not fail job in first stage.

Here the mocked failed state is not completed, such as miss the
attemptStatus's id, startTime, completionTime, completionStatus, etc.
status's startTime, completionTime, transitionTime, etc.

yqwang-ms, line 27: Make sure only mock job to fail when the first creation, i.e. not when we update it later. 
Otherwise, the mocked status is more complex (need status merge)

yqwang-ms: You can check other failed jobs to figure out all missing field here.

---
In reply to: [570746620](https://github.com/microsoft/pai/pull/5284#discussion_r570746620) [](ancestors = 570746620)

src/database-controller/src/poller/utils.js
hzy46: I did some investigation on this issue. There is no need to modify the code here.

First, k8s node client uses `request` to handle the HTTP request:

https://github.com/kubernetes-client/javascript/blob/61f8d3640485a177f1785601ba05a9884b6c34d8/src/gen/api/customObjectsApi.ts#L183-L193

For `request`, any network error will be raised directly. Thus there will be no valid `response`.

I also tested it by the following code:

```
const request = require('request');
request('https://download.pytorch.org/whl/cu101/torch-1.7.0%2Bcu101-cp36-cp36m-linux_x86_64.whl', function (error, response, body) {
  console.error('error:', error); // Print the error if one occurred
  console.error('response:', response)
 
});
```

If we cut the network connection during the request, the output will be:

```
error: { Error: read ECONNRESET
    at TCP.onread (net.js:622:25) errno: 'ECONNRESET', code: 'ECONNRESET', syscall: 'read' }
response: undefined
```

yqwang-ms: I mean apiserver may receive invalid json due to network issue, and return 422 to u

hzy46: This shouldn't happen in node.js because any network error will be raised directly.

yqwang-ms: See https://github.com/microsoft/pai/issues/4840
The nodejs k8s client may behavior good, if it will not try decode body when connection err

hzy46: You can find the logic here: https://github.com/kubernetes-client/javascript/blob/61f8d3640485a177f1785601ba05a9884b6c34d8/src/gen/api/customObjectsApi.ts#L412-L421

If there's any error, the node.js k8s client will call `reject(error);`, which means the error will be raised.

It starts to decode the body only when there is no error detected.

yqwang-ms: ok

yqwang-ms:
>       - PAI_FC

Add PAI_DB_CONTROLLER

---
Refers to: src/k8s-job-exit-spec/config/k8s-job-exit-spec.yaml:49 in 712fd44. [](commit_id = 712fd44405cc47c6cb3d8b7ddf90be3074772c79, deletion_comment = False)

src/k8s-job-exit-spec/config/k8s-job-exit-spec.yaml
```diff
@@ -482,6 +482,20 @@ spec:
     - reasonRegex: '(?i)^UnexpectedPredicateFailureType$'
       messageRegex: '(?ms).*'
 
+- code: -1007
+  phrase: UnrecoverableSchemaSynchronizeError
```
yqwang-ms, line 486: Adjust to below
###########################################################################
# Range: [-1099, -1000]
# Owner: PAI_FC
# Description: PAI_K8S issued failures
###########################################################################5291 



###########################################################################
# Range: [-1199, -1100]
# Owner: PAI_DB_CONTROLLER
# Description: PAI_K8S issued failures
###########################################################################5291

src/database-controller/src/common/framework.js
```diff
@@ -23,6 +23,45 @@ const mockFrameworkStatus = () => {
...
+      completionStatus: {
+        id: 0,
+        code: -1007,
+        diagnostics: 'Framework schema is submitted to database, but cannot be synchronized to API server due to unrecoverable error.',
+        phase: 'UnrecoverableSchemaSynchronizeError',
+        trigger: {
+          message: "Database controller cannot synchronize framework schema to API server.",
+          taskIndex: null,
+          taskRoleName: null,
+        },
```
yqwang-ms, line 46: Align with job spec

src/database-controller/src/common/framework.js
```diff
@@ -23,6 +23,45 @@ const mockFrameworkStatus = () => {
...
+  const dateStr = (new Date()).toISOString();
+  return {
+    state: 'Completed',
+    startTime: dateStr,
+    runTime: dateStr,
+    transitionTime: dateStr,
+    completionTime: dateStr,
+    attemptStatus: {
+      completionStatus: {
+        id: 0,
```
yqwang-ms, line 38: >0 [](start = 12, length = 1)

mock a md5 string like FC

src/database-controller/src/poller/index.js
```diff
@@ -95,6 +100,35 @@ function synchronizeHandler(snapshot, addOns, pollingTs) {
...
+        queue.add(async () => {
+          // For safety reason, we only consider the job that never starts.
+          // snapshot.setFailed() will raise an error if the framework is not in state AttemptCreationPending or has retry history.
+          snapshot.setFailed();
+          await postMockedEvent(snapshot, 'MODIFIED');
+          await databaseModel.FrameworkEvent.create(
+            generateClusterEventUpdate(
+              snapshot,
+              'Warning',
+              'UnrecoverableSynchronizeFailure',
```
yqwang-ms, line 119: >UnrecoverableSynchronizeFailure [](start = 15, length = 31)

Hard to understand, align to exit spec phase

src/database-controller/src/poller/index.js
```diff
@@ -95,6 +100,35 @@ function synchronizeHandler(snapshot, addOns, pollingTs) {
...
+              _.get(err, 'response.body.message', 'unknown'),
+            ),
+          );
+        }).catch(err => logger.error(
+           `An error happened when mock failed event for framework ${frameworkName} and pollingTs=${pollingTs}:`,
+           err,
+        ));
```
yqwang-ms, line 126: Remove catch here as it catch too much thing.
You can just catch in setFailed or just log warn in setFailed

src/database-controller/src/common/framework.js
```diff
@@ -290,6 +333,20 @@ class Snapshot {
...
+
+  setFailed() {
+    // Manually set the framework's status to failed.
+    // This is used when we cannot sync the framework request to api server,
+    // and we're sure that the error is unrecoverable.
+
+    // For now, we only use it for frameworks that never start.
+    console.log(this.getTotalRetriedCount(), this.getState())
+    if (!this.getTotalRetriedCount() === 0 || this.getState() !== 'AttemptCreationPending') {
+      throw new Error('setFailed() only works for framework that never start!')
```
yqwang-ms, line 345: and not synced to api server?

hzy46: `requestSyneced` is not managed by snapshot class.

For this change, we are sure that the framework's `requestSynced` field is false because we call `setFailed()` from `synchronizeHandler`. We only call `synchronizeHandler` for  `request=false` frameworks.

src/database-controller/src/poller/index.js
hzy46: done

src/k8s-job-exit-spec/config/k8s-job-exit-spec.yaml
hzy46: done. Use -1100 as the exit code.

src/database-controller/src/common/framework.js
hzy46: > > ```
> >   - PAI_FC
> > ```
> 
> Add PAI_DB_CONTROLLER
> 
> Refers to: src/k8s-job-exit-spec/config/k8s-job-exit-spec.yaml:49 in [712fd44](https://github.com/microsoft/pai/commit/712fd44405cc47c6cb3d8b7ddf90be3074772c79). [](commit_id = [712fd44](https://github.com/microsoft/pai/commit/712fd44405cc47c6cb3d8b7ddf90be3074772c79), deletion_comment = False)

done.

src/database-controller/src/common/framework.js
yqwang-ms: So add a comment here

src/k8s-job-exit-spec/config/k8s-job-exit-spec.yaml
```diff
@@ -482,6 +483,26 @@ spec:
     - reasonRegex: '(?i)^UnexpectedPredicateFailureType$'
       messageRegex: '(?ms).*'
 
+###########################################################################
+# Range: [-1199, -1100]
+# Owner: PAI_DB_CONTROLLER
+# Description: PAI_K8S issued failures
+###########################################################################
+
+- code: -1100
```
yqwang-ms, line 492: Remove blank line here?

src/database-controller/src/common/framework.js
```diff
@@ -23,6 +23,45 @@ const mockFrameworkStatus = () => {
...
+      completionStatus: {
+        id: 0,
+        code: -1100,
+        diagnostics: 'Job is submitted to database, but cannot be created in ApiServer due to permanent failures.',
+        phase: 'CreateFrameworkPermanentFailed',
+        trigger: {
+          message: "Database controller cannot synchronize framework to APIServer.",
+          taskIndex: null,
+          taskRoleName: null,
+        },
```
yqwang-ms, line 46: trigger is optional , only set when triggered by one specific task failure. So remove it here

src/database-controller/src/common/framework.js
hzy46: done. move id to `attemptStatus.id`.

src/database-controller/src/common/framework.js
hzy46: done

src/k8s-job-exit-spec/config/k8s-job-exit-spec.yaml
hzy46: done

src/database-controller/src/common/framework.js
hzy46: done

yqwang-ms (approved):
:shipit:

PR #5286, opened 2021-02-03
[Rest-Server] Add tokens check API for client

yiyione:
Add `GET /api/v2/tokens/check` API for OpenPAI client to verify token.
Currently only Marketplace will use it.
Just for internal use, didn't add to swagger.

image


debuggy (approved):

debuggy:
Is the response could also be decoded directly by jwt token?

PR #5291, opened 2021-02-06
Add save to template button in ui

debuggy:


src/webportal/src/app/job-submission/job-submission-page.jsx
```diff
@@ -356,6 +380,29 @@ export const JobSubmissionPage = ({
...
+  const onTemplateChange = useCallback((_, item) => {
+    if (item.key === 'No') {
+      return;
+    }
+    const jobConfig = JobProtocol.fromYaml(item.protocol);
+    const [
+      jobInfo,
+      taskRoles,
+      parameters,
+      ,
```
SwordFaith, line 392: This comma seems redundant.

yiyione (approved):

src/webportal/src/app/job-submission/job-submission-page.jsx
debuggy: It is used to ignore some returned values: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

PR #5292, opened 2021-02-07
save job specific token in sperate namespace

suiguoxin:
- save job specific token in sperate namespace
- revoke it in purge() function by check job state from DB

### DIFF
--- a/src/rest-server/src/models/token.js
+++ b/src/rest-server/src/models/token.js
@@ -17,6 +17,7 @@
 
 const jwt = require('jsonwebtoken');
 const uuid = require('uuid');
+const { Op } = require('sequelize');
 const { secret, tokenExpireTime } = require('@pai/config/token');
 const k8sSecret = require('@pai/models/kubernetes/k8s-secret');
 const k8sModel = require('@pai/models/kubernetes/kubernetes');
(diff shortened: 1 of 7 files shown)


src/rest-server/src/models/token.js
```diff
@@ -49,58 +65,83 @@ const sign = async (username, application, expiration) => {
...
   for (const [key, val] of Object.entries(data)) {
     try {
+      // expired tokens will be removed
       jwt.verify(val, secret);
-      result[key] = val;
     } catch (err) {
-      // pass;
+      continue;
+    }
+    if (jobSpecific) {
```
hzy46, line 77: If the user submitted a lot of jobs, e.g. > 1000, this will cause 1000+ rpc call to database, which will be a huge time delay. I suggest to collect all keys in `data` first, then validate them in db one time. You can use [this syntax](https://sequelize.org/master/manual/model-querying-basics.html#shorthand-syntax-for--code-op-in--code-) to select multiple frameworks' states.

src/rest-server/src/models/token.js
```diff
@@ -49,58 +65,83 @@ const sign = async (username, application, expiration) => {
...
+      // get framework from db
+      const frameworkName = deserialize(key);
+      const framework = await databaseModel.Framework.findOne({
+        attributes: ['state'],
+        where: { name: encodeName(frameworkName) },
+      });
+      // checkframework state, job-tokens of finished jobs will be removed
+      if (
+        !framework ||
+        ['SUCCEEDED', 'STOPPED', 'FAILED'].indexOf(framework.state) !== -1
```
hzy46, line 87: Suggest to use `framework.subState`.

`framework.subState === "Completed"` means the job is finished.

suiguoxin: > Suggest to use `framework.subState`.
> 
> `framework.subState === "Completed"` means the job is finished.

Fixed, thk

src/rest-server/src/models/token.js
suiguoxin: > If the user submitted a lot of jobs, e.g. > 1000, this will cause 1000+ rpc call to database, which will be a huge time delay. I suggest to collect all keys in `data` first, then validate them in db in one time. You can use [this syntax](https://sequelize.org/master/manual/model-querying-basics.html#shorthand-syntax-for--code-op-in--code-) to select multiple frameworks' states.

Fixed, thx

src/rest-server/src/models/token.js
```diff
@@ -67,30 +68,42 @@ const sign = async (username, application, expiration) => {
...
+        name: {
+          [Op.in]: frameworkNamesEncoded,
+        },
+        subState: {
+          [Op.ne]: 'Completed',
+        },
+      },
+    });
+    // job-tokens of finished jobs will be removed
+    data = Object.keys(data).reduce((filterd, key) => {
```
hzy46: typo filterd -> filtered

src/rest-server/src/models/token.js
```diff
@@ -67,30 +68,42 @@ const sign = async (username, application, expiration) => {
  */
 const purge = async (data, jobSpecific) => {
   const result = {};
+
+  if (jobSpecific) {
+    // get non-completed frameworks from db
+    const frameworkNamesEncoded = Object.keys(data)
+      .map(deserialize)
+      .map(encodeName);
```
hzy46, line 76: Can we avoid encodeName here? md5 encoding is cpu-intensive.

suiguoxin: > Can we avoid encodeName here? md5 encoding is cpu-intensive.

Fixed, thx

yiyione (approved):

hzy46 (approved):

PR #5294, opened 2021-02-08
[alert-manager] add gpu*hours info in cluster-utilization cronjob

suiguoxin:


src/alert-manager/src/cluster-utilization/send_alert.py
```diff
@@ -44,131 +58,154 @@ def get_job_usage_info(job_usage_result, rest_url):
...
+        user_infos[v["metric"]["username"]] = {
+            "username": v["metric"]["username"],
+            "usage": v["value"][1][:6] + "%", "resources_occupied": 0
+        }
     for job_name, job_info in job_infos.items():
         url = urllib.parse.urljoin(rest_url + "/", job_name)
         resp = requests.get(url, headers=headers)
         if not resp.ok:
             logging.warning("request failed %s", resp.text)
+            del job_infos[job_name]
```
Binyang2014, line 75: Please make sure this operation is safe. You delete element from `job_infos` while iterate it. I am not sure the behavior in this case.

src/alert-manager/src/cluster-utilization/send_alert.py
```diff
@@ -44,131 +58,154 @@ def get_job_usage_info(job_usage_result, rest_url):
...
         if not resp_json["jobStatus"]["appLaunchedTime"]:
             logging.warning("job not start, ignore it")
+            del job_infos[job_name]
             continue
-        job_info["start_time"] = datetime.fromtimestamp(
+        job_infos[job_name]["start_time"] = datetime.fromtimestamp(
             int(resp_json["jobStatus"]["appLaunchedTime"]) / 1000,
             timezone.utc)
+        if job_infos[job_name]["start_time"] < datetime.now(timezone.utc) - timedelta(days = 7):
+            job_infos[job_name]["start_time"] = datetime.now(timezone.utc) - timedelta(days = 7)
```
Binyang2014, line 88: Why add this line? Please do not change the value of `start_time`. You can use another key for this

src/alert-manager/src/cluster-utilization/send_alert.py
```diff
@@ -44,131 +58,154 @@ def get_job_usage_info(job_usage_result, rest_url):
...
             },
             "generatorURL": "alert/script"
-        }]
-        resp = requests.post(post_url, json=payload)
-        resp.raise_for_status()
-    logging.info("Finished sending alerts")
+        }
+        alerts.append(alert)
+    logging.info("Starting to send alerts...")
+    resp = requests.post(post_url, json=alerts)
```
Binyang2014, line 206: Please make sure this API is work

src/alert-manager/src/cluster-utilization/send_alert.py
suiguoxin: > Why add this line? Please do not change the value of `start_time`. You can use another key for this

Added another key `start_time_in_7d`.

src/alert-manager/src/cluster-utilization/send_alert.py
suiguoxin: > Please make sure this operation is safe. You delete element from `job_infos` while iterate it. I am not sure the behavior in this case.

Fixed & thanks for the reminder, there will be an issue in python3 if we delete elements while iterating.

src/alert-manager/src/cluster-utilization/send_alert.py
```diff
@@ -32,143 +31,183 @@ def wrapper(*args, **kwargs):
...
+        job_infos[job_name]["start_time"] = job_info["start_time"].strftime("%y-%m-%d %H:%M:%S")
+        job_infos[job_name]["resources_occupied"] = "{:.2f}".format(job_info["resources_occupied"])
+    for username, user_info in user_infos.items():
+        user_infos[username]["resources_occupied"] = "{:.2f}".format(user_info["resources_occupied"])
+
+    # sort usage info by resources occupied
+    job_usage = sorted(job_infos.values(), key=lambda x: float(x["resources_occupied"]), reverse=True)
+    user_usage = sorted(user_infos.values(), key=lambda x: float(x["resources_occupied"]), reverse=True)
+
+    return job_usage[:10], user_usage
```
suiguoxin, line 123: @Binyang2014 Here only the top 10 jobs (sorted by resources occupied) will be included in the alert since there can be thousands of jobs & there is an limitation in the post body size.

src/alert-manager/src/cluster-utilization/send_alert.py
suiguoxin: > Please make sure this API is work

Yep it works.

src/alert-manager/src/cluster-utilization/send_alert.py
```diff
@@ -32,143 +31,183 @@ def wrapper(*args, **kwargs):
...
+        if not matched_job:
+            logging.warning("Job %s not found.", job_name)
             continue
-        resp_json = resp.json()
-        # get job duration
-        if not resp_json["jobStatus"]["appLaunchedTime"]:
+        job_info = matched_job[0]
+
+        # ignore jobs not started
+        if not job_info["launchedTime"]:
```
Binyang2014, line 77: Please make sure this field is available for running jobs

Binyang2014 (approved):

Binyang2014:
The `Resources Occupied` seems not correct for running jobs. For job doesn't use gang allocation, if 3 tasks is running and 3 tasks is waiting, and each task using 1 GPU. The `Resources Occupied` will be `6GPU * duration` according to current code. But it is not correct.

src/alert-manager/src/cluster-utilization/send_alert.py
suiguoxin: > Please make sure this field is available for running jobs

will be fixed in https://github.com/microsoft/pai/pull/5307

suiguoxin:
> The `Resources Occupied` seems not correct for running jobs. For job doesn't use gang allocation, if 3 tasks is running and 3 tasks is waiting, and each task using 1 GPU. The `Resources Occupied` will be `6GPU * duration` according to current code. But it is not correct.

Fixed, get resources occupied from `Prometheus` instead.

Binyang2014 (approved):

PR #5302, opened 2021-02-11
fixed get wrong nvidia gpu card num.

lbin:
nvidia_gpu_count.stdout_lines[0] is null


hzy46:
this may be related to  https://github.com/microsoft/pai/issues/5306

lbin:
move to #5306

PR #5313, opened 2021-02-23
sign token with job info

suiguoxin:
#5270

### DIFF
--- a/src/rest-server/src/models/token.js
+++ b/src/rest-server/src/models/token.js
@@ -17,7 +17,6 @@
 
 const jwt = require('jsonwebtoken');
 const uuid = require('uuid');
-const { Op } = require('sequelize');
 const { secret, tokenExpireTime } = require('@pai/config/token');
 const k8sSecret = require('@pai/models/kubernetes/k8s-secret');
 const k8sModel = require('@pai/models/kubernetes/kubernetes');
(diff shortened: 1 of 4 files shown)


src/rest-server/src/models/token.js
```diff
@@ -28,21 +27,27 @@ const { encodeName } = require('@pai/models/v2/utils/name');
...
   k8sModel.createNamespace(userTokenNamespace);
-  k8sModel.createNamespace(jobTokenNamespace);
 }
 
-const sign = async (username, application, expiration) => {
+const sign = async (
+  username,
+  application,
+  expiration,
+  jobSpecific = 'false',
```
hzy46, line 40: while use string `'false'` instead of bool `false` here?

src/rest-server/src/models/token.js
```diff
@@ -136,25 +125,29 @@ const create = async (
       { encode: 'hex' },
     );
   } else {
-    const result = await purge(item, jobSpecific);
+    const result = purge(item, jobSpecific);
```
hzy46, line 128: remove `jobSpecific` here?

src/rest-server/src/models/token.js
```diff
@@ -177,18 +170,34 @@ const batchRevoke = async (username, filter, jobSpecific = false) => {
 
 const verify = async (token) => {
   const payload = jwt.verify(token, secret);
+  logger.debug(payload);
```
hzy46, line 173: Please remove this.

src/rest-server/src/models/token.js
suiguoxin: > while use string `'false'` instead of bool `false` here?

Typo, fixed

src/rest-server/src/models/token.js
suiguoxin: > remove `jobSpecific` here?

Fixed

src/rest-server/src/models/token.js
suiguoxin: > Please remove this.

Removed

hzy46 (approved):

PR #5343, opened 2021-03-03
Add an internal storage barrier to postgresql

### LINKED ISSUE #5341: Database may use unmounted host path
Currently we use internal storage to mount a file called `storage.ext4` to `/mnt/paiInternal` on the master node. Then postgresql uses `/mnt/paiInternal` as its data root.

Sometimes the path could be unmounted. If postgresql starts before internal-storage, it will use unmounted host path.

hzy46:
to fix #5341

### DIFF
--- a/src/internal-storage/build/internal-storage-barrier.k8s.dockerfile
+++ b/src/internal-storage/build/internal-storage-barrier.k8s.dockerfile
@@ -1,0 +1,24 @@
+# Copyright (c) Microsoft Corporation
+# All rights reserved.
+#
+# MIT License
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
+# documentation files (the "Software"), to deal in the Software without restriction, including without limitation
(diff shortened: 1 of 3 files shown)


Binyang2014 (commented):
Why not move internal storage into postgresql folder. If it only used by postgresql

hzy46:
> Why not move internal storage into postgresql folder. If it only used by postgresql

For current usage, there's no difference between putting it in `postgresql` or `internal-storage`.

But if there are other services relying on `internal-storage`, putting this in `internal-storage` can make it easier to reuse. It may happen in the future.

Binyang2014 (approved):
It's ok for me as a hot-fix. Please create another issue to move to to postgresql. Then we can make the code much more clean and leverage the k8s mechanism.

hzy46:
> It's ok for me as a hot-fix. Please create another issue to move to to postgresql. Then we can make the code much more clean and leverage the k8s mechanism.

I have record the new design to #5344 . It does have some advantages. However, considering there's only one service using internal storage now, it may be not worth the effort to implement the new mechanism. We can leave it to a future work.

PR #5371, opened 2021-03-10
Add nodes using paictl

Starmys:
Not ready to merge.


deployment/paiLibrary/paiOrchestration/temp_config.py
```diff
@@ -0,0 +1,68 @@
...
+            kube_config_path=self._kube_config_path
+        )
+        get_handler.run()
+    
+    def _generate_config_files(self):
+        self._logger.info("Generate config files: `hosts.yml` and `openpai.yml`")
+        linux_shell.execute_shell_raise(
+            shell_cmd="cd ./contrib/kubespray/ && python3 ./script/k8s_generator.py -l {} -c {} -o {}".format(
+                os.path.join(self._path, "layout.yaml"),
+                os.path.join(self._path, "services-configuration.yaml"),
```
suiguoxin, line 51: `-c` should be `config.yaml`. 
However, the problem is that `config.yaml` is not saved in k8s, please fix this in another PR.

deployment/paiLibrary/paiOrchestration/temp_kubespray.py
```diff
@@ -0,0 +1,63 @@
...
+        self._modify_files()
+    
+    def _clone_kubespray(self):
+        self._logger.info("Begin to clone Kubespray 2.11 from GitHub")
+        linux_shell.execute_shell_raise(
+            shell_cmd="git clone -b release-2.11 https://github.com/kubernetes-sigs/kubespray.git {}".format(self._path),
+            error_msg="Failed to clone Kubespray 2.11 from GitHub"
+        )
+    
+    def _modify_files(self):
```
suiguoxin, line 41: Please comment the reasons of modifying kubespray source somewhere

deployment/paiLibrary/paiOrchestration/temp_config.py
Starmys: Change to `-s` here?

deployment/paiLibrary/paiOrchestration/add_node.py
```diff
@@ -0,0 +1,52 @@
...
+
+import os
+import logging
+
+import temp_config
+import temp_kubespray
+from ..common import linux_shell
+
+
+class add_node:
```
suiguoxin, line 26: `class AddNode`. Please check the modified files with `pylint` even if we didn't set up pylint for the `deployment` folder in CI.

deployment/paiLibrary/paiOrchestration/temp_config.py
suiguoxin: > Change to `-s` here?

The problem is : Currently, when we call `paictl.py config push -p <> -m service` to update cluster configuration, only `layout.yaml` & `services-configuration.yaml` will be saved.  `config.yaml` should also be saved so that the admin users don't need to backup this file locally. 

I'm suggesting to fix this problem in another PR.

suiguoxin:
> * If there is no SSH key pair: need to set `user: <username>` and `password: <password>` in `services-configuration.yaml`
> * If there is SSH key pair:
>   
>   1. Mount `~/.ssh` folder into docker container
>   2. Set `user: <username>` in `services-configuration.yaml`

This is not included in this PR, right ?

Starmys:
> > * If there is no SSH key pair: need to set `user: <username>` and `password: <password>` in `services-configuration.yaml`
> > * If there is SSH key pair:
> >   
> >   1. Mount `~/.ssh` folder into docker container
> >   2. Set `user: <username>` in `services-configuration.yaml`
> 
> This is not included in this PR, right ?

Yes, would be solved along with `config.yaml`.

deployment/paiLibrary/paiOrchestration/temp_config.py
```diff
@@ -0,0 +1,68 @@
...
+                self._path
+            ),
+            error_msg="Failed to remove temporary config folder: {}, please remove it manually".format(self._path)
+        )
+    
+    def __del__(self):
+        self._logger.info("Remove temporary config folder")
+        linux_shell.execute_shell_raise(
+            shell_cmd="rm -r {}".format(self._path),
```
hzy46, line 60: Please do not use bare `rm` in the script.

You can refer to https://docs.python.org/3/library/tempfile.html for a cleaner solution to create a temporary directory.

deployment/paiLibrary/paiOrchestration/temp_kubespray.py
```diff
@@ -0,0 +1,63 @@
...
+        with open(os.path.join(self._path, "roles/remove-node/post-remove/tasks/main.yml"), "r") as f2:
+            data2 = yaml.load(f2, yaml.SafeLoader)
+        del data2[2]['run_once']
+        with open(os.path.join(self._path, "roles/remove-node/post-remove/tasks/main.yml"), "w") as f2:
+            yaml.dump(data2, f2, default_flow_style=False)
+    
+    def __del__(self):
+        self._logger.info("Remove temporary downloaded Kubespray folder")
+        linux_shell.execute_shell_raise(
+            shell_cmd="rm -r {}".format(self._path),
```
hzy46, line 58: Also, please use https://docs.python.org/3/library/tempfile.html instead

Issue #5347, opened 2021-03-04
support sort by completedTime in get the list jobs API


suiguoxin:
### Background
- [list Jobs API](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/microsoft/pai/master/src/rest-server/docs/swagger.yaml#operation/listJobs)
- sort by `completedTime` is not supported
- Related issue : #5348 

### Design 
sort by completed Time
- when return by DESC order, null will be at top
- when return by ASC order, null will be at bottom

### Implementation
- reference: https://www.designcise.com/web/tutorial/how-to-order-null-values-first-or-last-in-mysql

### Backlog:
Proposal 1: Add a query parameter
- `completedAfter=timestamp`, return the jobs completed after the timestamp.

Proposal 2: Add a query parameter,
`$filter=(startTime ge datetime'2014-01-01T00%3a00%3a00') and (completedTime le datetime'2014-04-30T00%3a00%3a00')`

`Odata-> Sequelize`: https://www.npmjs.com/package/odata-sequelize

Reference: https://docs.microsoft.com/en-us/azure/search/search-query-odata-filter

PR #5375, opened 2021-03-15
Support sorting by completionTime

suiguoxin:
#5347


hzy46 (approved):

suiguoxin:
Test cases:
- test get jobs list api [ordered by comletionTime (ASC & DESC](https://github.com/suiguoxin/pai/blob/master/src/rest-server/docs/swagger.yaml#L1158), the behavior should be as designed in #5347

PR #5387, opened 2021-03-22
Add custom ssl port support. resolves #5386

### LINKED ISSUE #5386: Request to support custom ssl port
I need openpai to support custom SSL port.
**Why is this needed**: 
If the administrator can customize the ssl port, it will be possible to deploy multiple websites on the same master node through an additional nginx service. But now the ssl port number is designated as 443, without modifying the source code, the administrator cannot customize the ssl port. The custom ssl port is very useful, especially when the master node has sufficient computing power to deploy multiple websites.
ssl port must be 443.
[              hostPort: 443](https://github.com/microsoft/pai/blob/accc5fda3d19df0ff7eba2411e9937d8ff6ed4b3/src/pylon/deploy/pylon.yaml.template#L88)
[basic-management-operations.md](https://github.com/microsoft/pai/edit/master/docs/manual/cluster-admin/basic-management-operations.md)

siaimes:
Signed-off-by: siaimes <34199488+siaimes@users.noreply.github.com>

### DIFF
--- a/deployment/clusterObjectModel/test/data/configuration-template-generate/services-configuration.yaml
+++ b/deployment/clusterObjectModel/test/data/configuration-template-generate/services-configuration.yaml
@@ -103,6 +103,7 @@ pylon:
   # port of pylon
   port: 80
   ssl:
+    enable: false
     port: 443
 
 hivedscheduler:
--- a/deployment/clusterObjectModel/test/data/configuration-template-generate/services-configuration.yaml
+++ b/deployment/clusterObjectModel/test/data/configuration-template-generate/services-configuration.yaml
@@ -102,6 +102,8 @@ prometheus:
 pylon:
   # port of pylon
   port: 80
+  ssl:
+    port: 443
 
 hivedscheduler:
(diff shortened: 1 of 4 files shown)


siaimes:
It seems that port 443 is still occupied, causing additional Nginx services to fail to respond to port 443 requests.

Binyang2014:
@siaimes  Is this PR still valid? I think you can simply change the `hostPort:` to your desire port number to achieve this

siaimes:
> @siaimes Is this PR still valid? I think you can simply change the `hostPort:` to your desire port number to achieve this

This PR has only been tested in v1.3.0, but there is a problem, and I have not adopted it myself.

I changed the 'hostPort' to '8443', and then started another service at '443'. The result of the experiment is that openpai can be migrated to '8443', but the service of '443' still cannot receive network traffic.

I also think we can simply change 'hostPort' to achieve this feature, but this doesn't seem to work in the v1.3.0 version, So for this PR, contributors may need to do more testing. Thank you.

PR #5390, opened 2021-03-24
Add "Solve Unmounted Database Problem" doc

hzy46:


docs/manual/cluster-admin/troubleshooting.md
```diff
@@ -149,3 +149,36 @@ Step 5. In the dev box container, start the internal storage service by `./paict
...
+5. Dump the database by `pg_dump -U root -h 127.0.0.1 -W  -t framework_events -t framework_history -t pod_events -t pods -t task_history -t tags -f exclude_frameworks_dump.sql -a openpai` and `pg_dump -U root -h 127.0.0.1 -W -t frameworks_tmp -f frameworks_tmp_dump.sql openpai`.
+6. In dev box container, run `./paictl.py service stop -n cluster-configuration postgresql`. Wait serveral minutes, run `./paictl.py service start -n cluster-configuration internal-storage postgresql`
+7. On master node, restore the data to `framework_tmp`: `psql  -U root -h 127.0.0.1 -W  -f frameworks_tmp_dump.sql openpai`
+8. Connect to the database `psql  -U root -h 127.0.0.1 -W openpai`. Then merge the tables by: `DELETE FROM frameworks WHERE name in ( SELECT name FROM frameworks_tmp );` and `INSERT INTO frameworks SELECT  * FROM frameworks_tmp;`.
```
Binyang2014, line 181: For this method, seems the data on host will replace the data on internal storage. Is it a safe way? If some status/events will be lost according to the method

hzy46: Events will not be lost. They are dumped to `exclude_frameworks_dump.sql` and then restored.

hzy46: Only frameworks with the same name will be overrided. These are usually running jobs thus the influence is minor.

Binyang2014 (approved):

PR #5391, opened 2021-03-24
Support Azure Storage as log server

Binyang2014:
Support Azure Storage as log server.

For full log RestServer will return log url which point to Azure Storage.
For tail-log, add an internal API to retrieve last N bytes log. (Since we can not get tail log though Azure without add http header)


src/rest-server/src/controllers/internal/tail-log.js
```diff
@@ -0,0 +1,59 @@
...
+const { ContainerClient } = require('@azure/storage-blob');
+
+const logger = require('@pai/config/logger');
+const createError = require('@pai/utils/error');
+
+const getTailLog = asyncHandler(async (req, res) => {
+  const tailLogSize = 16 * 1024 * 1024; // 16 KB
+  const logName = req.params.logName;
+  const queryString = req.url.substring(req.url.indexOf('?') + 1);
+  const account = process.env.LOG_AZURE_STORAGE_ACCOUNT;
```
hzy46, line 31: could you please put this config to launcher config?  https://github.com/microsoft/pai/blob/30f21b391db32ee0211bb024e20ae113f9317452/src/rest-server/src/config/launcher.js

src/rest-server/src/models/v2/job/log.js
```diff
@@ -104,6 +113,200 @@ const getLogListFromLogManager = async (
...
+  );
+  if (!taskStatus || !taskStatus.containerIp) {
+    logger.error(`Failed to find task to retrive log or task not started`);
+    throw NoTaskLogErr;
+  }
+
+  const podUid = taskStatus.containerId;
+
+  const account = process.env.LOG_AZURE_STORAGE_ACCOUNT;
+  const accountKey = process.env.LOG_AZURE_STORAGE_ACCOUNT_KEY;
```
hzy46, line 141: Also, I think it's better to put this into launcher config.

src/rest-server/src/routes/internal/tail-log.js
```diff
@@ -0,0 +1,27 @@
...
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+// module dependencies
+const express = require('express');
+const controller = require('@pai/controllers/internal/tail-log');
+
+const router = new express.Router();
+
+router.route('/:logName').get(controller.getTailLog);
```
hzy46, line 24: Should we check the PAI token when use this API?

Binyang2014: I think we don't need to check token here. The url will contain SAS token and Azure will use this to check the permission. And based on design, this API doesn't need Authentication header. Client can call the API directly to download log.

hzy46 (approved):

PR #5394, opened 2021-03-25
add config.yaml into push list

Starmys:


deployment/configCmd.py
```diff
@@ -176,13 +176,15 @@ def push_configuration(self, args):
         if args.kube_config_path != None:
             args.kube_config_path = os.path.expanduser(args.kube_config_path)
         push_list = [
+            "config.yaml",
```
hzy46, line 179: If `config.yaml` doesn't exist, what will happen?

Starmys: The same as `layou.yaml`, the logger will raise an IOError: `[Errno 2] No such file or directory: './tmp-config-1616644286/layout.yaml'`

hzy46: This will a breaking change. 

Please add the following text if `config.yaml` doesn't exist:


Cannot find config.yaml in your config folder. Before v1.7.0, this file is stored in ~/pai-deploy/cluster-cfg/config.yaml on the dev box machine. 
If you are upgrading from v1.7.0, please copy this file to the config folder.

hzy46 (approved):

Issue #5334, opened 2021-03-02
K8s API server's cert need renew each year


yiyione:
The k8s API server's cert will expire every year, and will cause OpenPAI cluster not available.
Certificate Management with kubeadm:
https://v1-16.docs.kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/#automatic-certificate-renewal

![image](https://user-images.githubusercontent.com/17357108/109600953-07e9ca00-7b59-11eb-9128-0455a305b4d0.png)

### Todo
- [ ] Document this requirement in the repo
- [ ] Add warning for the cert expire

Binyang2014:
refer: https://github.com/kubernetes/kubeadm/issues/581#issuecomment-471575078

yiyione:
Test case:
1. setup the alert-manager to enable the `email-admin` action (change the `admin-receiver` to a test address)
2. change the `schedule` and `alert-residual-days` in `alert-manager.cert-expiration-checker` from `services-configuration.yaml` to trigger the alert:
    ```yaml
    cert-expiration-checker:
      schedule: '* * * * *' # every minute
      alert-residual-days: 365 # always trigger
      cert-path: '/etc/kubernetes/ssl' # the k8s cert path in master node
    ```
3. use kubectl get pods to check whether the `cert-expiration-checker` cronjob be triggered
4. check the alert email or the logs from alert handler.

PR #5409, opened 2021-04-01
[Alert manager] k8s cert expiration checker

yiyione:
#5334


src/alert-manager/deploy/service.yaml
```diff
@@ -37,4 +38,4 @@ upgraded-script: upgraded.sh
 
 
 deploy-rules:
-  - in: pai-master
\ No newline at end of file
+  - in: pai-master
```
suiguoxin, line 41: The `deploy-rule` doesn't apply for CronJob, https://github.com/microsoft/pai/blob/master/deployment/paiLibrary/paiService/service_template_generate.py#L61
 we can add a nodeSelector in the CronJob definition.

src/alert-manager/config/alert-manager.yaml
```diff
@@ -31,3 +31,6 @@ cluster-utilization:
   configured: False
 use-pylon: False
 repeat-interval: '24h'
+cert-expiration-checker:
+  schedule: '0 0 * * *'
+  alert-residual-days: 30
```
suiguoxin, line 36: This file is loaded as the default config, this part should be configurable in `services-configuration.yaml`, refer to #5281

src/alert-manager/src/cert-expiration-checker/send_alert.py
```diff
@@ -0,0 +1,60 @@
...
+def send_alert(pai_url: str, certExpirationInfo: str):
+    trigger_time = str(datetime.now(timezone.utc).date())
+    post_url = pai_url.rstrip("/") + ALERT_PREFIX
+    alerts = []
+    alert = {
+        "labels": {
+            "alertname": "k8s cert expiration",
+            "severity": "warn",
+            "trigger_time": trigger_time,
+            "annotations": certExpirationInfo,
```
suiguoxin, line 34: Please provide a detailed summary in the `annotations` field like:
```yaml
annoations:
    summary: the k8s cert will be expired in {} days
``` 
The summary will be rendered in the email body

src/alert-manager/deploy/alert-manager-cert-expiration-check-cronjob.yaml.template
```diff
@@ -0,0 +1,43 @@
...
+    {%- else %}
+              value: "{{ cluster_cfg['pylon']['uri']}}"
+    {%- endif %}
+            - name: ALERT_RESIDUAL_DAYS
+              value: "{{ cluster_cfg["alert-manager"]["cert-expiration-checker"]["alert-residual-days"] }}"
+            volumeMounts:
+            - mountPath: /usr/local/bin/kubeadm
+              name: kubeadm
+            - mountPath: /etc/kubernetes/
+              name: kubenetes-config
```
Binyang2014, line 31: Can make the path configurable? We can't assume all env such as AKS use this path.

src/alert-manager/src/cert-expiration-checker/send_alert.py
```diff
@@ -0,0 +1,60 @@
...
+    alerts.append(alert)
+
+    logging.info("Sending alerts to alert-manager...")
+    resp = requests.post(post_url, json=alerts)
+    resp.raise_for_status()
+    logging.info("Alerts sent to alert-manager.")
+
+def main():
+    PAI_URI = os.environ.get("PAI_URI")
+    certExpirationInfo = os.popen('kubeadm alpha certs check-expiration --config="/etc/kubernetes/kubeadm-config.yaml"').read()
```
Binyang2014, line 47: The command seems depends on kubeadm version. Refer https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-certs/#check-certificate-expiration, the latest version don't need to add alpha in command.

One way to solve this is package kubeadm into docker image
Another way is check the cert manually. It's easy to use a lib to check cert expired time

src/alert-manager/deploy/alert-manager-cert-expiration-check-cronjob.yaml.template
yiyione: OK, I'll change it

src/alert-manager/config/alert-manager.yaml
yiyione: Done.

src/alert-manager/src/cert-expiration-checker/send_alert.py
yiyione: Done.

src/alert-manager/src/cert-expiration-checker/send_alert.py
yiyione: Done. use python OpenSSL.crypto.load_certificate(...) to read the crt file, and check the expired time.

suiguoxin:
Could you provide some documents to the end of this file : https://github.com/microsoft/pai/blob/master/docs/manual/cluster-admin/how-to-use-alert-system.md

yiyione:
> Could you provide some documents to the end of this file : https://github.com/microsoft/pai/blob/master/docs/manual/cluster-admin/how-to-use-alert-system.md

Added.

suiguoxin (approved):

siaimes:
> > Could you provide some documents to the end of this file : https://github.com/microsoft/pai/blob/master/docs/manual/cluster-admin/how-to-use-alert-system.md
> 
> Added.

> This feature will be enable by default, if the action `email-admin` is enabled.
> You can configure the `alert-manager`->`cert-expiration-checker` field in `services-configuration.yml`.
> `schedule`, `alert-residual-days` and `cert-path` are necessary fields for this feature, and we have default value for the fields.

Hi, @yiyione 

I have some doubts about this description. Is the warning message not sent to the upper right corner of the homepage? If the warning message is sent to the upper right corner of the homepage, can I use this feature without setting up a mail server?

Thanks.

siaimes:
[kubernetes/kubernetes/pull/52196](https://github.com/kubernetes/kubernetes/pull/52196)

In addition, I found a PR in the kubernetes repository that seems to fix the problem. Is this PR in this repository still necessary?

Binyang2014:
@siaimes Nice catch! We don't notice this feature. For this release, we will still use this PR. After we verified this feature, this PR can revert. Thanks for this notice!

yiyione:
> Hi, @yiyione
> 
> I have some doubts about this description. Is the warning message not sent to the upper right corner of the homepage? If the warning message is sent to the upper right corner of the homepage, can I use this feature without setting up a mail server?
> 
> Thanks.

It's a good question, and we are going to fix it.
After the PR #5440 merged, the admin will be able to view the alert on the `upper right corner of the homepage` without setting up the email.

Thanks.

PR #5421, opened 2021-04-12
Modify docs about config.yaml

Starmys:


deployment/confStorage/download.py
```diff
@@ -73,6 +73,11 @@ def download_cluster_configuration(self, local_path):
             self.logger.error("The configuration doesn't exists on your cluster. Please upload it first.")
             sys.exit(1)
 
+        if u'config.yaml' not in configuration_dict:
```
suiguoxin, line 76: typo ? `u'config.yaml'`

docs/manual/cluster-admin/how-to-add-and-remove-nodes.md
```diff
@@ -50,6 +50,8 @@ OpenPAI doesn't support changing master nodes, thus, only the solution of adding
       pai-worker: "true"
   ```
 
+- Make sure that you can access all nodes in the cluster using the settings in `<config-folder>/config.yaml`. If you use SSH key pairs to log in to nodes, please mount the folder `~/.ssh` on the dev box machine to `/root/.ssh` on the dev box docker container。
+
```
suiguoxin, line 54: Do we support ssh key pai log in during installation ? https://github.com/microsoft/pai/blob/master/docs/manual/cluster-admin/installation-guide.md#configyaml-example

deployment/confStorage/download.py
```diff
@@ -73,6 +73,11 @@ def download_cluster_configuration(self, local_path):
...
+            self.logger.warning("config.yaml doesn't exists on your cluster.")
+            self.logger.warning("Before v1.7.0, config.yaml is stored in ~/pai-deploy/cluster-cfg/config.yaml on the dev box machine.")
```
suiguoxin, line 78: Suggest adding a link for `config.yaml` schema reference, in case the admin user lose the backup

suiguoxin:
Can we add some words [here](https://github.com/microsoft/pai/blob/c31fa0c54e7b9f5050e7f904741a2112d9e90426/docs/manual/cluster-admin/how-to-add-and-remove-nodes.md#pull--modify-cluster-settings) to remind the users to check `config.yaml` before calling `./paictl.py add/remove` ?

docs/manual/cluster-admin/how-to-add-and-remove-nodes.md
Starmys: The user can leave the `password` blank and manually copy the SSH private key to nodes.

deployment/confStorage/download.py
Starmys: Keys in `configuration_dict` are unicode

suiguoxin (approved):

PR #5533, opened 2021-06-16
set cluster.advertise-address to avoid docker use non-private ip

Binyang2014:
If docker use ono-private ip. There will be error: `alertmanager: no private IP found`.
Set `cluster.advertise-address` to avoid this issue


suiguoxin (approved):

Binyang2014:
If docker use no-private ip. There will be error: alertmanager: no private IP found.
Set cluster.advertise-address to avoid this issue.
Refer: https://github.com/prometheus/alertmanager/issues/2284#issuecomment-640044282

PR #5541, opened 2021-06-24
add prometheus pushgateway service; adjust prometheus config

suiguoxin:
- add prometheus pushgateway service
- set `hornor_label` as true in Prometheus to keep `instance` label valuable

### DIFF
--- a/src/prometheus-pushgateway/deploy/prometheus-pushgateway.yaml.template
+++ b/src/prometheus-pushgateway/deploy/prometheus-pushgateway.yaml.template
@@ -50,8 +50,6 @@ spec:
         {%- if cluster_cfg['cluster']['common']['qos-switch'] == "true" %}
         resources:
           limits:
-            memory: "10Gi"
-          requests:
             memory: "256Mi"
         {%- endif %}
         args:
--- a/src/prometheus-pushgateway/deploy/prometheus-pushgateway.yaml.template
+++ b/src/prometheus-pushgateway/deploy/prometheus-pushgateway.yaml.template
@@ -40,9 +40,8 @@ spec:
       annotations:
         prometheus.io/scrape: "true"
         prometheus.io/path: "/prometheus-pushgateway/metrics"
-        prometheus.io/port: "{{ prometheus_pushgateway_port }}"
+        prometheus.io/port: "8080"
     spec:
-      hostNetwork: true
       containers:
       - name: prometheus-pushgateway
         image: {{ cluster_cfg["cluster"]["docker-registry"]["prefix"] }}prometheus-pushgateway:{{ cluster_cfg["cluster"]["docker-registry"]["tag"] }}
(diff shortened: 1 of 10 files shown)


src/prometheus-pushgateway/deploy/service.yaml
```diff
@@ -0,0 +1,37 @@
...
+# DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+cluster-type:
+  - yarn
+  - k8s
+
+prerequisite:
+  - cluster-configuration
+  - docker-cache
```
Binyang2014, line 24: Do we need this?

src/prometheus-pushgateway/deploy/prometheus-pushgateway.yaml.template
```diff
@@ -0,0 +1,72 @@
...
+    metadata:
+      name: prometheus-pushgateway
+      labels:
+        app: prometheus-pushgateway
+      annotations:
+        prometheus.io/scrape: "true"
+        prometheus.io/path: "/prometheus-pushgateway/metrics"
+        prometheus.io/port: "{{ prometheus_pushgateway_port }}"
+    spec:
+      hostNetwork: true
```
Binyang2014, line 45: Can we change hostNetwork to false?

src/prometheus-pushgateway/deploy/prometheus-pushgateway.yaml.template
```diff
@@ -0,0 +1,72 @@
...
+      containers:
+      - name: prometheus-pushgateway
+        image: {{ cluster_cfg["cluster"]["docker-registry"]["prefix"] }}prometheus-pushgateway:{{ cluster_cfg["cluster"]["docker-registry"]["tag"] }}
+        imagePullPolicy: Always
+        {%- if cluster_cfg['cluster']['common']['qos-switch'] == "true" %}
+        resources:
+          limits:
+            memory: "10Gi"
```
Binyang2014, line 53: Why need 10Gi memory...

src/prometheus-pushgateway/deploy/service.yaml
suiguoxin: > Do we need this?

Yes, why not ?

src/prometheus-pushgateway/deploy/prometheus-pushgateway.yaml.template
suiguoxin: > Why need 10Gi memory...

Changed to 256Mi limit

src/prometheus-pushgateway/deploy/prometheus-pushgateway.yaml.template
suiguoxin: > Can we change hostNetwork to false?

Not sure, why change it?

src/prometheus-pushgateway/deploy/service.yaml
Binyang2014: I think docker-cache is an optional services.

src/prometheus-pushgateway/deploy/prometheus-pushgateway.yaml.template
Binyang2014: using hostNetwork break the network isolation, If we don't need to access all network interface, set hostNetwork to false will make pod more security

suiguoxin: Changed

suiguoxin: We may remove this after pylon gets configured. Currently we'll not be able to access the Pushgateway portal with hostNetwork set as true.

Binyang2014 (approved):

PR #5547, opened 2021-06-29
Grafana config 

suiguoxin:


Binyang2014 (commented):
Can we combine the normal task/node with with pxc task/node view? Or you can create a new dashboard to show this

src/grafana/deploy/grafana-configuration/pai-jobview-dashboard.json.template
```diff
@@ -227,14 +227,14 @@
               "label": "GB",
               "logBase": 1,
               "max": null,
-              "min": null,
+              "min": 0,
```
Binyang2014, line 230: What's the difference for 0 and null?

suiguoxin: The yaxis may begin from negative values if set as null.

suiguoxin:
> Can we combine the normal task/node with with pxc task/node view? Or you can create a new dashboard to show this

Combined in the same dashboard, the queries without data will not be shown.

Binyang2014 (approved):

PR #5240, opened 2021-01-11
Eng doc update

hzy46:

PR #5567, opened 2021-07-18
Update basic-management-operations.md

### LINKED ISSUE #5465: Will the new version of the code affect the execution logic of the old version?
**Organization Name**:

**Short summary about the issue/question**: I encountered some version problems. Currently, the openpai I deployed is v1.6.0, but on the k8s dashboard, I see it is pulling the cert-expeiration-checker image, which is a feature introduced in v1.7.0.
image

**Brief what process you are following**: 

**How to reproduce it**:

**OpenPAI Environment**: ubuntu 16.04
- OpenPAI version: v1.6.0
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Hardware (e.g. core number, memory size, storage size, GPU type etc.):
- Others:

siaimes:
fix #5465

### DIFF
--- a/docs/manual/cluster-admin/basic-management-operations.md
+++ b/docs/manual/cluster-admin/basic-management-operations.md
@@ -79,6 +79,9 @@ Generally speaking, PAI services are daemon sets, deployments, or stateful sets 
 sudo docker run -itd \
         -e COLUMNS=$COLUMNS -e LINES=$LINES -e TERM=$TERM \
         -v /var/run/docker.sock:/var/run/docker.sock \
+        -v ${HOME}/pai-deploy/cluster-cfg:<config-folder>  \
+        -v ${HOME}/pai-deploy/kube:/root/.kube \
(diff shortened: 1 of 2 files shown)


siaimes:
According to [this script](https://github.com/microsoft/pai/blob/master/src/dev-box/build/start-script.sh), another solution is to set `OPENPAI_BRANCH_NAME` environment variable when starting the dev-box container, for example `-e OPENPAI_BRANCH_NAME=pai-1.8.y`. But, In my opinion, the better solution is to delete the script and directly check out the tag with the same version as the dev-box in the dev-box Dockerfile. Only in this way can the consistency of the versions of paictl.py and dev-box be guaranteed.

siaimes:
Conflict with this [PR](https://github.com/microsoft/pai/pull/5240).

siaimes:
PR #5567 conflicts with PR #5240, I will close it, a better solution needs further discussion.

PR #5616, opened 2021-09-15
remove log4j dependency from frameworklauncher

Binyang2014:
Remove log4j from `frameworklauncher`.
Framework launcher is not be used in pure k8s version and it uses log4j which has high security issue.
Remove this log4j dependency to fix security issue

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -184,23 +184,3 @@ jobs:
         yarn install --frozen-lockfiles
         yarn build
 
-  frameworklauncher:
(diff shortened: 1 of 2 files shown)


yqwang-ms:
Can we just remove log4j in pom?

yqwang-ms (changes requested):
Pls hold on, we can remove log4j or upgrade it

Binyang2014:
> Pls hold on, we can remove log4j or upgrade it

Remove the log4j dependency in pom file

yqwang-ms (approved):
thanks!

Issue #5654, opened 2021-11-11
Issues with alert manager.


siaimes:
**Organization Name**:

**Short summary about the issue/question**:
If the user uses a self-signed certificate, the alert-manager will be fail.
**Brief what process you are following**:

config file:

```
pylon:
    port: 80
    uri: "http://10.10.10.10:80"
    ssl:
      port: 8443
      crt_name: 10.10.10.10.crt
      crt_path: /cluster-configuration/ssl/10.10.10.10.crt
      key_name: 10.10.10.10.key
      key_path: /cluster-configuration/ssl/10.10.10.10.key
```

```
alert-manager:
#   port: 9093 # optional, do not modify this if you do not want to change the port alert-manager is listening on
  pai-bearer-token: '*********************'
  alert-handler: # alert-handler will only be enabled when this field is not empty
#     port: 9095 # optional, do not modify this if you do not want to change the port alert-handler is listening on
#     log-level: "info" # optional
    email-configs: # email-notification will only be enabled when this field is not empty
      admin-receiver: ********@163.com
      smtp-host: smtp.163.com
      smtp-port: 465
      smtp-from: ************@163.com
      smtp-auth-username: *************
      smtp-auth-password: *****************
  cluster-utilization: # cluster-utilization is a k8s CronJob which reports the GPU utilization of the cluster
    # for schedule syntax, refer to https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-schedule-syntax
    schedule: "* * * * *" # daily report at UTC 00:00
```

logs:

```
user@hostname:~$ sudo docker logs k8s_cluster-utilization_cluster-utilization-1636635840-s4lxj_default_559d12d2-a499-4327-9427-efadb0447901_3
2021-11-11 13:05:22,502 - INFO - send_alert.py:167 - Collecting cluster usage info...
2021-11-11 13:05:22,507 - DEBUG - connectionpool.py:943 - Starting new HTTPS connection (1): 10.10.10.10:8443
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/urllib3/connectionpool.py", line 677, in urlopen
    chunked=chunked,
  File "/usr/local/lib/python3.7/site-packages/urllib3/connectionpool.py", line 381, in _make_request
    self._validate_conn(conn)
  File "/usr/local/lib/python3.7/site-packages/urllib3/connectionpool.py", line 978, in _validate_conn
    conn.connect()
[...]

PR #5662, opened 2021-11-22
Use SSL port to send mail by default.

siaimes:
Use SSL port to send mail by default.

Signed-off-by: siaimes <34199488+siaimes@users.noreply.github.com>

### DIFF
--- a/src/alert-manager/src/alert-handler/controllers/mail.js
+++ b/src/alert-manager/src/alert-handler/controllers/mail.js
@@ -25,7 +25,7 @@ const path = require('path');
 const transporter = nodemailer.createTransport({
   host: process.env.EMAIL_CONFIGS_SMTP_HOST,
   port: parseInt(process.env.EMAIL_CONFIGS_SMTP_PORT),
-  secure: false,
+  secure: true,
   auth: {
     user: process.env.EMAIL_CONFIGS_SMTP_AUTH_USERNAME,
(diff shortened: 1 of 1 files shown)


siaimes:
https://github.com/microsoft/pai/issues/5654#issuecomment-973697005

https://github.com/microsoft/pai/blob/ea19af183e1a5e26ccd9fc2a410ffefd6ada9d52/src/alert-manager/src/alert-handler/controllers/mail.js#L28

PAI does not currently support the SSL protocol to send emails, this is not safe.

suiguoxin:
@siaimes  Thanks for the proposal, but it's not sufficient to simply set `secure` as true, it requires addition TLS set up & configuration

siaimes:
> @siaimes Thanks for the proposal, but it's not sufficient to simply set `secure` as true, it requires addition TLS set up & configuration

Thanks for your reply, I have tested it, just need to modify this line, then modify the port number in the configuration file to the SSL port (e.g. 465 for mail.163.com) of the mail server, it works.

siaimes:
[siaimes/alert-handler](https://github.com/siaimes/alert-handler/blob/main/Dockerfile)

```
    email-configs: # email-notification will only be enabled when this field is not empty
      admin-receiver: xxxx@163.com
      smtp-host: smtp.163.com
      smtp-port: 465
      smtp-from: xxxx@163.com
      smtp-auth-username: xxxx@163.com
      smtp-auth-password: xxxxxxxxxxxxxxxxx
```

Exactly, this image and config work for me.

PR #5674, opened 2021-12-15
fix env problem for adding nodes by paictl.py in dev-box docker

murez:
[CHANGE]: install `python 3.6` rather than `python 3.5`
[ADD]: `apt` install `sshpass` for ssh login new machines, `pip3` install `ansible==2.9.7`, `netaddr==0.7.19`, `passlib`.

### DIFF
--- a/src/dev-box/build/dev-box.common.dockerfile
+++ b/src/dev-box/build/dev-box.common.dockerfile
@@ -14,7 +14,7 @@ RUN apt-get -y update && \
       gawk \
       psmisc \
       python \
-      python3 \
+      python3.6 \
       python-yaml \
       python-jinja2 \
       python-urllib3 \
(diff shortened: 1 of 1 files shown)


src/dev-box/build/dev-box.common.dockerfile
```diff
@@ -14,7 +14,7 @@ RUN apt-get -y update && \
       gawk \
       psmisc \
       python \
-      python3 \
+      python3.6 \
```
Binyang2014, line 17: Why using python3.6 instead of default python version

murez: Because some pip packages are not surport or may cause some trouble when install ansible

murez: A better solution is change the whole docker base image from ubuntu16.04 to ubuntu18.04

PR #5656, opened 2021-11-15
Add release note for v1.8.1

yiyione:

PR #5676, opened 2021-12-27
[Cherry-pick] Add release note for v1.8.1 (#5656)

yiyione:
* add release note for v1.8.1

* update the version number in docs


suiguoxin:
We need to use `v1.8.1` here: https://github.com/microsoft/pai/blob/master/src/marketplace-restserver/deploy/marketplace-restserver.yaml.template#L21

yiyione:
> We need to use `v1.8.1` here: https://github.com/microsoft/pai/blob/master/src/marketplace-restserver/deploy/marketplace-restserver.yaml.template#L21

The marketplace doesn't have a v1.8.1 release, so we don't have a marketplace image with tag v1.8.1. (The marketplace image upload will be trigger by the release CI).
I think we can still use the v1.8.0 marketplace here.

suiguoxin (approved):

PR #5763, opened 2022-04-17
Fix job secret

### LINKED ISSUE #5609: Proposals to fix the container isolation issue and the possibility of multi-cluster DDP jobs.
Proposals to fix the container isolation issue and the possibility of multi-cluster DDP jobs.

1. Allow users to add a pair of asymmetric keys on the user profile page, similar to adding SSH Public Keys (Remind users: 1. This pair of keys is used for DDP jobs. If there is no DDP requirement, don't care about it; 2. This pair of keys should be generated separately instead of his own asymmetric key pair.).
2. Update code related to [these lines](https://github.com/microsoft/openpai-runtime/blob/4ab72398d404ec8b18d0eace9a27d954f46404f7/src/plugins/ssh/sshd.sh#L48-L62) to save this pair of asymmetric keys into `root/.ssh/authorized_keys` and `root/.ssh/id_rsa` of each job of the user respectively.

**Why is this needed**:

1. #5577 Q2 proposed a problem of isolation between containers, but after my detailed testing and viewing the source code, the problem was introduced by the [runtime](https://github.com/microsoft/openpai-runtime/blob/4ab72398d404ec8b18d0eace9a27d954f46404f7/src/plugins/ssh/sshd.sh#L48-L62).
2. If the user has multiple clusters, then the DDP job distributed in multiple clusters will be possible, and the user only needs to upload the same asymmetric key pair in each cluster.

A user can ssh into his own container first, and then can ssh into any user's container in the cluster without obstacles, which is quite dangerous.
webportal and runtime

siaimes:
Fix job secret, fix #5609 with https://github.com/microsoft/openpai-runtime/pull/53


siaimes:
## When adding or modifying users, automatically generate a job ssh key pair for each user if it does not exist.

image

## openpai-runtime will process this key pair.

image

## Functional tests are fine.

After updating rest-server and openpai-runtime, user A can no longer log in to user B's job containers, at the same time, the user's own containers can log in to each other.

## unit test

Help is needed to pass the unit test.

src/rest-server/src/models/v2/user.js
```diff
@@ -22,6 +22,24 @@ const groupModel = require('@pai/models/v2/group');
 
 const crudType = 'k8sSecret';
 const crudUser = crudUtil.getStorageObject(crudType);
+const keygen = require('ssh-keygen');
+const deasync = require('deasync');
+
+function getSSHKey() {
+  let ret;
+  const location = 'id_rsa';
+  keygen({ location: location }, function (err, result) {
```
Binyang2014, line 31: How about using native nodeJS library https://stackoverflow.com/questions/8520973/how-to-create-a-pair-private-public-keys-using-node-js-crypto

siaimes: https://stackoverflow.com/a/52775583

As you can see from the comment area of this answer, even if you use crypto, you still have to rely on the third-party package node-forge. I have also done relevant tests, and this solution is not feasible.

Binyang2014: We can rely on node-forge since it's a popular library. But ssh-keygen only has few stars, we don't know if the library still be maintained.

src/rest-server/src/models/v2/user.js
```diff
@@ -22,6 +22,24 @@ const groupModel = require('@pai/models/v2/group');
...
+    if (err) {
+      console.log(err);
+    }
+    ret = result;
+  });
+  // eslint-disable-next-line no-unmodified-loop-condition
```
Binyang2014, line 37: Remove this line. Please **DO NOT** disable eslint unless you have a good reason

src/rest-server/src/models/v2/job/k8s.js
```diff
@@ -557,12 +552,6 @@ const generateTaskRole = (
                 path: `/var/log/pai`,
               },
             },
-            {
-              name: 'job-ssh-secret-volume',
-              secret: {
-                secretName: 'job-ssh-secret',
-              },
-            },
```
Binyang2014, line 565: How dose the job container get the ssh secret after this change? We don't mount the secret as a file?

siaimes: It will in `/usr/local/pai/user-extension-secrets/userExtensionSecrets.yaml`, please see fix 53 of runtime which related to this.

https://github.com/microsoft/openpai-runtime/pull/53/files#diff-e21e1d55082d72593bd006834666b4a67bdff33b61d86a9346d2b4f1160141e7R56-R75

src/rest-server/src/models/v2/user.js
siaimes: 1. I have done relevant tests, and the solution provided by [https://stackoverflow.com/a/52775583](https://stackoverflow.com/a/52775583) using `crypto` and  `node-forge` is not feasible.
2. node v10 doesn't seem to support crypto.generateKeyPairSync.

src/rest-server/src/models/v2/user.js
```diff
@@ -33,10 +46,16 @@ const getAllUser = async () => {
 };
 
 const createUser = async (username, value) => {
+  if (!('jobSSH' in value.extension)) {
+    value.extension.jobSSH = generateSSHKeyPair();
```
Binyang2014, line 50: We don't need to pass parameter here?

Binyang2014 (approved):
Thanks for this change. LGTM
Please fix the CI then we can merge the change

src/rest-server/src/models/v2/user.js
siaimes: default 2048

raghavthind2005 (approved):

siaimes:
> Thanks for this change. LGTM Please fix the CI then we can merge the change

@Binyang2014 Thank you for your review, CI is fixed now.

siaimes:
After updating from the old version, the admin can delete useless secret by using the following command in the dev-box container:

```
kubectl delete secret job-ssh-secret
```

